PR #34, opened 2025-09-05
feat: add cursor support

ETNOL:
Pictures worth 1000 words

Example of it working in my Cursor IDE
image
Successful execution of the Cursor IDE option
image
Here's an example release with the new templates
https://github.com/ETNOL/spec-kit/releases/tag/v0.0.1

I don't use Cursor CLI and couldn't find docs on if it supports custom commands so its following Cursor's pattern for their IDE's commands. I'll manually test the CLI side once I get the opportunity.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -51,7 +51,9 @@ import readchar
 AI_CHOICES = {
     "copilot": "GitHub Copilot",
     "claude": "Claude Code",
-    "gemini": "Gemini CLI"
+    "gemini": "Gemini CLI",
+    "cursor": "Cursor CLI",
+    "cursor-ide": "Cursor IDE"
 }
 
 # ASCII Art Banner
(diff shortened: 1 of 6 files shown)


src/specify_cli/__init__.py
```diff
@@ -770,7 +779,7 @@ def init(
     with Live(tracker.render(), console=console, refresh_per_second=8, transient=True) as live:
         tracker.attach_refresh(lambda: live.update(tracker.render()))
         try:
-            download_and_extract_template(project_path, selected_ai, here, verbose=False, tracker=tracker)
+            download_and_extract_template(project_path, selected_ai, here, verbose=False, tracker=tracker, repo_owner=repo_owner)
```
ETNOL, line 782: Added an optional repo_owner param to allow me to test from my own release

localden:
@ETNOL thank you for your contribution! I refactored the CLI a bit and ensured that the Cursor integration is now supported.

PR #48, opened 2025-09-06
make ps1

bonsai:


localden:
There are a bunch of changes here - please explain the rationale for them and how the PS1 scripts will be used.

PR #88, opened 2025-09-08
Update README.md

sirnails:
explain that it has to be in vscode and in the copilot, not the spec kit cmd prompt


localden:
That section is not GitHub Copilot or VS Code-specific, so we will need to delegate it to another part of the README. I am also working on documentation for Spec Kit, so this will get folded there shortly. Thanks so much for the contribution!

PR #107, opened 2025-09-09
Fix: Use forked repo for template download

Mirza-Samad-Ahmed-Baig:
Problem: The code was hardcoded to download templates from the original repository, ignoring user forks.
  
Impact: This prevented users from creating projects with their own customized templates in their forked version.
  
Solution: The application now dynamically identifies the correct repository from the git configuration to download the templates.
  
Changes: I introduced a function to parse the git remote URL and used it to fetch templates from the user's forked repository.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -385,12 +385,40 @@ def init_git_repo(project_path: Path, quiet: bool = False) -> bool:
         os.chdir(original_cwd)
 
 
+def get_git_repo_info():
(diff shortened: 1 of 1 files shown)


localden:
@Mirza-Samad-Ahmed-Baig let's discuss this in a feature request first - I am not yet convinced that there is a need to expose forked repos here. The goal of the CLI is to be dead-simple to intregrate with, meaning that it relies on default spec templates.

Once you open an issue, happy to help review and discuss with the community there!

PR #128, opened 2025-09-10
Refactor HTTP client usage to utilize truststore for SSL context

ramusbucket:
When `specify` command is run behind a proxy (enterprise proxy like zscalar, or anti-virus web scan proxy, etc.), it hangs on `Fetch latest release` step, saying "contacting GitHub API". This PR fixes that issue with the use of local truststore with `httpx` and makes it work provided python is cert settings are configured according to the proxy provider's documentation. For example: https://help.zscaler.com/zia/adding-custom-certificate-application-specific-trust-store

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,7 @@ dependencies = [
     "httpx",
     "platformdirs",
     "readchar",
+    "truststore>=0.10.4",
 ]
 
 [project.scripts]
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -46,6 +46,11 @@ from typer.core import TyperGroup
 
 # For cross-platform keyboard input
 import readchar
+import ssl
+import truststore
+
+ssl_context = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+client = httpx.Client(verify=ssl_context)
 
 # Constants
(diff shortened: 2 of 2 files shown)


nomagicln:
Good jobs, but I think it would be better if there were an option to skip TLS verification.

localden:
This sounds like a reasonable change, but I am curious as @nomagicln called out - is it better to introduce a new package dependency vs. disable TLS verification altogether? Seems like the latter might be preferred.

ramusbucket:
Taken @localden  @nomagicln, disabling TLS verification altogether is one option. To provide more flexibility, I've implemented the `--skip-tls` cli option. As a secure alternative for users with enterprise proxy certificates or self-signed certificates, I've also retained a package dependency that enables the use of a custom truststore.

localden:
@ramusbucket I am more curious if _just disabling TLS verification is enough_? Seems like the threat vector of doing it is that _if you are getting MITM-d_, then you may download a bad package from someone that is injecting bad payloads. But even then, the CLI doesn't execute anything and just extracts the content so the risk is somewhat contained.

Also, looks like `pip` itself [is using](https://pip.pypa.io/en/latest/topics/https-certificates/#using-system-certificate-stores) `truststore`, so I am less concerned about this particular dependency.

I'll review this PR and we can merge it soon.

PR #130, opened 2025-09-10
Make Den the only Den and make him always have a flower

yarikoptic:
With this change we get

    ❯ git shortlog -sn
        39  Den Delimarsky 🌺
        ...

whenever on master we have

    ❯ git shortlog -sn
        20  Den Delimarsky
        19  Den Delimarsky 🌺
        ...

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Den Delimarsky 🌺 <53200638+localden@users.noreply.github.com>


localden:
Thank you @yarikoptic for this very important call-out 🤣 I should, indeed, always have a flower next to my name. That being said, let's not introduce this particular metadata piece in the code base for now.

PR #151, opened 2025-09-10
feat: add Windsurf AI support

raedkit:
## Description

This PR updates the release process to include a dedicated Windsurf package and modifies the CLI to use it directly, maintaining strict parity with the Copilot flow.

### Key Changes

1. **Release Process**:
   - Added a new `sdd-windsurf-package` build in both release.yml and manual-release.yml
   - The package includes `.windsurf/workflows/` with the necessary workflow files
   - Published as `spec-kit-template-windsurf-<version>.zip` in GitHub Releases

2. **CLI Updates**:
   - CLI directly downloads the Windsurf asset when `--ai windsurf` is used

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -51,7 +51,8 @@ import readchar
 AI_CHOICES = {
     "copilot": "GitHub Copilot",
     "claude": "Claude Code",
-    "gemini": "Gemini CLI"
+    "gemini": "Gemini CLI",
+    "windsurf": "Windsurf AI"
 }
 
 # ASCII Art Banner
@@ -635,10 +636,11 @@ def download_and_extract_template(project_path: Path, ai_assistant: str, is_curr
     return project_path
 
 
+
 @app.command()
 def init(
     project_name: str = typer.Argument(None, help="Name for your new project directory (optional if using --here)"),
-    ai_assistant: str = typer.Option(None, "--ai", help="AI assistant to use: claude, gemini, or copilot"),
+    ai_assistant: str = typer.Option(None, "--ai", help="AI assistant to use: claude, gemini, copilot, or windsurf"),
     ignore_agent_tools: bool = typer.Option(False, "--ignore-agent-tools", help="Skip checks for AI agent tools like Claude Code"),
     no_git: bool = typer.Option(False, "--no-git", help="Skip git repository initialization"),
     here: bool = typer.Option(False, "--here", help="Initialize project in the current directory instead of creating a new one"),
@@ -648,7 +650,7 @@ def init(
     
     This command will:
     1. Check that required tools are installed (git is optional)
-    2. Let you choose your AI assistant (Claude Code, Gemini CLI, or GitHub Copilot)
+    2. Let you choose your AI assistant (Claude Code, Gemini CLI, GitHub Copilot, or Windsurf AI)
     3. Download the appropriate template from GitHub
     4. Extract the template to a new project directory or current directory
     5. Initialize a fresh git repository (if not --no-git and no existing repo)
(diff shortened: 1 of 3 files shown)


src/specify_cli/__init__.py
```diff
@@ -659,6 +661,7 @@ def init(
         specify init my-project --ai claude
         specify init my-project --ai gemini
         specify init my-project --ai copilot --no-git
+        specify init my-project --ai windsurf
```
aashish-ak, line 664: Line 733 onwards: The elif block should have windsurf check as well.

localden:
@raedkit also just cross-checking, but do these commands actually work with Windsurf?

Their documentation is a bit ambiguous, stating that [slash commands are not supported](https://docs.windsurf.com/chat/overview#slash-commands), but I also see that you're packing yours in [workflows](https://docs.windsurf.com/windsurf/cascade/workflows), which seem to be working fine.

raedkit:
> @raedkit also just cross-checking, but do these commands actually work with Windsurf?
> 
> Their documentation is a bit ambiguous, stating that [slash commands are not supported](https://docs.windsurf.com/chat/overview#slash-commands), but I also see that you're packing yours in [workflows](https://docs.windsurf.com/windsurf/cascade/workflows), which seem to be working fine.

It's working fine, but the issue is that Windsurf have its own plan file. Which means that there's an ephemeral memory that I'm unable to modify which remind him each while to check its plan and not tasks.md. But it works ..
Need some extensive testing, if someone else want to test it and give me his feedback

src/specify_cli/__init__.py
raedkit: I'm sorry, but I don't get your point. Could you explain more precisely please ?

localden:
Tagging this as `needs-verification` to make sure that we're checking the changes before they merge. If this gets properly tested, zero issues with this going in.

adam-paterson:
@localden  Fellow Windsurf user here, I'm happy to verify if you'd like?

src/specify_cli/__init__.py
adam-paterson: The cli checks for the existence of tools before initialising them: https://github.com/github/spec-kit/pull/151/files#diff-8e3e8dfe8740d05f4d06dd457ea293ec88ac0ecc9bba6eb182f0f767db67ce9eR733-R748

You need to add the `check_tool` call to include Windsurf

```python
    # Check agent tools unless ignored
    if not ignore_agent_tools:
        agent_tool_missing = False
        if selected_ai == "claude":
          ...
        elif selected_ai == "windsurf":
            if not check_tool("windsurf", "Install from: https://windsurf.com/download"):
                console.print("[red]Error:[/red] Windsurf IDE is required for Windsurf projects")
                agent_tool_missing = True
          ...

```

raedkit: as stated in the README.md "The CLI will check if you have Claude Code or Gemini CLI installed." and this means it checks only the CLI commands not the IDEs like VSCode (used by Github Copilot). 
If we change this, we need to do the same thing for VSCode for Github Copilot.

localden:
@raedkit can you please incorporate the change suggestion from @adam-paterson? Also, please refresh your codebase with the latest `main` commits - looks like there are some conflicts that need to be resolved.

raedkit:
I see @localden has done all the necessary and integrated Windsurf. I will close this PR.
@localden are you OK with this ?

localden:
Yep, thanks @raedkit - I'm making sure to reference your contribution in the changelog.

topherfangio:
@localden @raedkit Are there any docs on how to use this new functionality?

raedkit:
> @localden @raedkit Are there any docs on how to use this new functionality?

You could check the video published by @localden on YouTube. It's the same logic across all IDEs and CLIs.
You could find the video on README.md

PR #349, opened 2025-09-18
feat(cli): add --force flag for specify init

adam-paterson:
## Summary
- add a --force option to the specify init command
- allow scripted environments to skip the interactive confirmation when --here and the directory is non-empty

## Reasoning
During the work on #159 and #339 the initialisation of spec-kit in the harness container prompts the user with a warning about existing template files. There was no way to bypass this in non-interactive environments.

## Testing
- uvx --from git+https://github.com/github/spec-kit.git specify init sandbox --ai opencode --script sh --ignore-agent-tools --no-git --force


adam-paterson:
@localden  Closing this due to duplication with 9319f04

adam-paterson:
yY??a?qqq

PR #418, opened 2025-09-21
feat/hack: Add Zed support

Josh-AS:
## Summary

Adds support for the Zed agent in the CLI and templates.  

Because Zed does **not support slash commands currently**, a workaround was implemented to allow similar functionality.

## Changes

- **Zed agent support**  
  - Command rules are compiled into a single `.rules` file, which is the format Zed accepts.  
  - Introduced prompting to define a `<slash_command>` template.  
  - Template rules are parsed into blocks and placed into the `.rules` file for Zed.  

## Notes / Limitations

- Attempting to use `/specify` (or any slash command) literally in Zed is still blocked.  
- Running the same commands **without the leading `/`** works in testing runs.  
- This approach is a hacky workaround to mimic slash command functionality in Zed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -102,6 +102,8 @@ jobs:
           - spec-kit-template-windsurf-ps-${{ steps.get_tag.outputs.new_version }}.zip
           - spec-kit-template-codex-sh-${{ steps.get_tag.outputs.new_version }}.zip
           - spec-kit-template-codex-ps-${{ steps.get_tag.outputs.new_version }}.zip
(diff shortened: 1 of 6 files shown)


Josh-AS:
Made draft as could do with some more CLI changes so if they select zed it onboards them correctly and doesn't tell them they can use /commands but instead tells them to run "specify A todo app"

Also `create-release-packages.sh` could be updated to bootstrap the .rules file with prompting for handling their execution and then `generate_commands` could be reverted back to normal, then .rules references wherever the release templates are stored.

Would be a lot cleaner this way, but bed time.

PR #473, opened 2025-09-22
Added 2 new commands that work well in Codex to replace /specify

timothyjoh:
I was digging into your Codex support this morning, and without the ability to use $ARGUMENTS when invoking a command in Codex, I create another 2-command step that works around that problem. I have tested this and it seems to work pretty well.

Steps: 
1. Invoke `/new` to clear the context
2. Invoke `/discuss`, Codex will ask a question like "Happy to help! To start, what’s the high-level goal for this new feature?" to which you can respond with a simple description. The agent will go back and forth with the user to clarify requirements one question at a time and then output a summary and prompt the user to continue with the next command
3. Invoke `/feature` which will take the above discussion and summary and execute the same process that `/specify` would have done. 
4. Continue on with `/plan` > `/tasks` > `/implement` as normal

> **Bonus** This process should also work fine with other agents, tho I did not test it all out yet. 

I have tested this by implementing 2 new features already this morning. Another added benefit that I can see is that maybe the discussion phase irons out a lot of the possible `NEEDS CLARIFICATION` segments during the planning phase.

I added some description to the Codex-specific warning text when initializing the project. But maybe some up-for-discussion thoughts:
1. Should we see if we can reduce this to a single step, call it `/specify-codex` and copy it into the project instead of the original `specify`? Could be cleaner than this, but I thought to keep them separate for the time being, in case OpenAI decides to implement command arguments in the near future, this might all be moot.
2. Anywhere else we should document this change, since it is Codex-specific, but also it could work well in others?
3. Other ideas?

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -1067,9 +1067,11 @@ def init(
     if selected_ai == "codex":
         warning_text = """[bold yellow]Important Note:[/bold yellow]
 
-Custom prompts do not yet support arguments in Codex. You may need to manually specify additional project instructions directly in prompt files located in [cyan].codex/prompts/[/cyan].
+Custom prompts do not yet support arguments in Codex. In the short-term (until Codex supports arguments) use this workaround pattern:
 
-For more information, see: [cyan]https://github.com/openai/codex/issues/2890[/cyan]"""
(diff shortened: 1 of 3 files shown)


ivangrynenko:
The way around this problem is quite simple. Update files in ~/.codex/prompts:
1. plan.md: `If the argument is missing, stop and ask the user to provide input.`
2. specify.md: Finish the first paragraph with `If no ``$ARGUMENTS`` was provided to you, stop and ask the user to provide the prompt of what to build or change.`
3. constitution.md: Add after the first paragraph this: `If no prompt was provided as part of this command, stop and ask the user to provide it. Use that input during the steps below.`

Codex will stop and ask you to provide the input, then apply it and use as if it has support for the arguments. 
@localden

localden:
Looks like this should now be working properly in Codex - thanks again for the contribution @timothyjoh!

Issue #516, opened 2025-09-23
Specs are being created in repo root and not it working directory


da1z:
I have monorepo with several projects in sub dirs. I open project dir with cursor (repo/project1).
spec-kit /specify command create specs in repo/specs instead of repo/project1/specs

localden:
Good suggestion - let me work through the requirements for it.

flavour:
#173

acgxv:
https://github.com/github/spec-kit/issues/581

bleow:
facing the same issue. the quickest and dirtiest fix is to change the `Get-FeatureDir` function to hardcode your working directory, i.e.
```
Join-Path $repoRoot "ADD/PATH_TO_WORKDIR_FROM_ROOT/HERE/specs/$Branch"
```


you could also comment out the try-catch block in `Get-RepoRoot` and just return the `$PSScriptRoot "../../.."` directly.

PR #526, opened 2025-09-23
Use workdir as root for everything instead of repo root

da1z:
opening to start discussion on the approach. let me know what you think

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,8 +377,7 @@ At this stage, your project folder contents should resemble the following:
 
 ```text
 ├── memory
-│	 ├── constitution.md
-│	 └── constitution_update_checklist.md
+│	 └── constitution.md
 ├── scripts
 │	 ├── check-prerequisites.sh
 │	 ├── common.sh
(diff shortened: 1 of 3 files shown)


da1z:
fix for https://github.com/github/spec-kit/issues/516

localden:
I see a potential flaw here - if we just use the working directory sourced from `pwd`, you're relying on the terminal being _in the right context_. So, if I am in VS Code, and happened to open the wrong folder in the terminal session, this can create the artifacts or look them up in the wrong location?

da1z:
> I see a potential flaw here - if we just use the working directory sourced from `pwd`, you're relying on the terminal being _in the right context_. So, if I am in VS Code, and happened to open the wrong folder in the terminal session, this can create the artifacts or look them up in the wrong location?

yes. if vs code opened in wrong working dir i would consider it as user error. and if it is wrong dir opened in vs code then none of / commands would work.

acgxv:
https://github.com/github/spec-kit/issues/581

mnriem:
@da1z Can you merge the conflicts so I can review the changes?

PR #527, opened 2025-09-23
feat: choose whether to remove the constitution.md or not (#507)

### LINKED ISSUE #507: on init don't overwrite .specify/memory/constitution.md if it exists
The only way to update, is to re-init ( or manually drag files ) it would be great if there was a check that .specify/memory/constitution.md exits, if so, keep it as is

My current workaround is just to git checkout .specify/memory/constitution.md, but it could cause some issue's if someone forgets to do  that

outp1:
Closes #507

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -32,6 +32,7 @@ import shlex
 import json
 from pathlib import Path
 from typing import Optional, Tuple
+from enum import Enum
 
 import typer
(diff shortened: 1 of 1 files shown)


outp1:
This is how the new feature looks like:
image

@localden, what do you think about it? Many users are struggling with the constitution.md file being overwritten. This approach could also potentially be used for other project files.

PR #562, opened 2025-09-24
Multi Agent Support

seanrobertwright:
With the realization that team members may not always use the same agent, the iteration implements this feature.  Users are now able to add multiple agents through the setup TUI, or via the command line.


localden:
@seanrobertwright thank you for the contribution! This would be a larger rewrite that also requires template updates that are not captured here. Let's first discuss the implementation in an issue and go from there.

seanrobertwright:
@localden I'll set something up to get this rolling.

Issue #540, opened 2025-09-24
The "Source Tree" section of plan-template.md leads to unpredictable/corrupt output (sometimes)


danwashusen:
I've noticed that when using Spec Kit within a brownfield project (e.g. `/specify Add feature XYZ to my project`) that implementation/test files end up in odd places. I traced it down to the `**Structure Decision**` in `plan-template.md`, which **sometimes** renders incorrectly without interpolating the actual project structure into the final `plan.md` file (see comments below for examples).

Asking GPT5 Codex High to analyse the issue yields:
```text
• I'm checking whether the instructions call for filling or customizing the Source Code section beyond
  using a fixed template. There isn't any explicit directive to interpolate or modify it; the template
  remains static per the .codex/prompts/plan.md instructions, especially Step 6, which guides artifact
  generation without mentioning changes to that section. I'll reference relevant lines and file paths to
  back this up in the response.

> No—the current instructions never tell the agent to interpolate or customize that block. The
  playbook just says to “Execute the implementation plan template … Let the template guide artifact
  generation” without describing any extra processing for the Source Code section beyond setting the
  plan.md keeps the generic Options 1/2/3 list, which is why it doesn’t reflect the repository’s actual
  layout.
```

airtonix:
seems like you need to be describing this in your `constitution.md`, since that is constantly referred to through out the prompts and templates. 

further more, in `/specify`, `/plan` and `/clarify`, you should probably be mentioning this in the initial prompt.

danwashusen:
> seems like you need to be describing this in your `constitution.md`, since that is constantly referred to through out the prompts and templates. 
> 
> further more, in `/specify`, `/plan` and `/clarify`, you should probably be mentioning this in the initial prompt.

No... `plan-template.md` does not consult the constitution (or anything else for that matter). It writes the template section to your `plan.md` in a basically hard coder manor...

I have the structure clearly described in my AGENTS.md (which is a more appropriate place than the constitution file). Relying on a constitutional exception to correct the file locations after the fact (because the `plan.md` is wrong) would make for an epic waste of context.

danwashusen:
Using Codex GPT5 High to run `/plan` the resulting `plan.md` file contains the following, the decision seems to indicate the templates version is close enough...  I'll try a "close enough isn't good enough" in my constitution... 🙄 

### Source Code (repository root)
```
# Option 2: Web application (frontend + backend)
backend/
├── src/
│   ├── models/
│   ├── services/
│   └── api/
[...]

PR #563, opened 2025-09-25
Refined wording in `plan-template.md` to ensure the Source Code section consistently works

danwashusen:
Summary

  - Brownfield runs of specify (e.g. /specify Add feature XYZ to my project) produce plan.md files whose “Source Code (repository root)” section stays the hard-coded scaffold from plan-template.md, so generated implementation/test files land in the wrong directories.
  - The template often just copies the generic Options 1/2/3 block into plan.md, wasting context and leaving agents to guess the true structure.
  - Multiple /plan runs with GPT5 Codex High/Medium show inconsistent behavior: some keep the placeholder Option 2 snippet, others revert to the full generic block, which confirms the template is never interpolated and is unreliable for real repo layouts.
  - Updating plan-template.md to include explicit instructions for interpolating the Source Tree into the fenced code block resolves the problem; after that change every /plan run renders the correct repo structure.

  PR Focus

  - Teach the planning template to populate the Source Tree section with the project’s actual directory layout instead of shipping the static Options list, ensuring Spec Kit generates artifacts in the intended locations.

See https://github.com/github/spec-kit/issues/540

### DIFF
--- a/templates/plan-template.md
+++ b/templates/plan-template.md
@@ -15,7 +15,7 @@ scripts:
 1. Load feature spec from Input path
    → If not found: ERROR "No feature spec at {path}"
 2. Fill Technical Context (scan for NEEDS CLARIFICATION)
-   → Detect Project Type from context (web=frontend+backend, mobile=app+api)
+   → Detect Project Type from file system structure or context (web=frontend+backend, mobile=app+api)
    → Set Structure Decision based on project type
 3. Fill the Constitution Check section based on the content of the constitution document.
 4. Evaluate Constitution Check section below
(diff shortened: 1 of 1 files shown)


danwashusen:
I've confirmed these changes still work on a "greenfield" project (where nothing exists yet) using GPT5 Codex Medium:

### Source Code (repository root)
```
web/
└── flappy-bird.html     # Single-file game artifact with inline CSS/JS

tests/
└── e2e/
    └── flappy-bird.spec.ts  # Playwright scenarios for controls, scoring, restart
```

**Structure Decision**: Treat project as a static web experience rooted at `/web/flappy-bird.html`, with browser-based Playwright tests under `/tests/e2e/`. No build tooling directories required because assets remain inline.

danwashusen:
As noted on [this comment](https://github.com/github/spec-kit/issues/540#issuecomment-3331155495) I've also tested on a reasonable complex "brownfield" project.

PR #608, opened 2025-09-26
Added --offline-template option to enable developers with corporate firewalls  to access spec-kit by using local template files instead of downloading from GitHub.

kumaresra:
Added --offline-template option to enable developers with corporate firewalls 
to access spec-kit by using local template files instead of downloading from GitHub.


localden:
Thanks for the contribution! At this time, we will not be adding this functionality - because Specify is a "scaffolding CLI," an "offline template" is basically "Download and extract the ZIP."

PR #658, opened 2025-09-29
Fix constitution.md path to relative path in plan template

newmind:
Update constitution reference from absolute path
to relative path  since templates will be
located within the .specify folder structure.

From
```
See `/memory/constitution.md`*
```
To
```
See `.specify/memory/constitution.md`*
```


localden:
Thank you for your contribution! Those paths should be dynamically substituted later at packaging time.

PR #722, opened 2025-10-03
Supporting multi ai agent editor

Guo-astro:
The scenario  is: planning with Claude Code Opus 4.1 and implementation with GPT-5 Codex. To enable the “/” command across tools, we need to initialize multiple AI providers so it works in both Claude Code and VS Code Copilot (Agent Mode). Copilot doesn’t expose certain models (e.g., Claude Code Opus 4.1).

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -215,6 +215,100 @@ def get_key():
 
 
 
+def select_multiple_with_arrows(options: dict, prompt_text: str = "Select options", default_keys: list = None) -> list:
+    """
+    Interactive multi-selection using arrow keys with Rich Live display.
+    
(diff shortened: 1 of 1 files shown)


localden:
Thanks for the contribution, @Guo-astro - we're re-working the multi-agent scenario as it will require some additional modifications to the scripts and templates. This is not just about the coding agent files.

Guo-astro:
> Thanks for the contribution, @Guo-astro - we're re-working the multi-agent scenario as it will require some additional modifications to the scripts and templates. This is not just about the coding agent files.

@localden  Thanks very much for the clearification , is there a link of issue or PR I can track into? Very happy to learn and comment

PR #766, opened 2025-10-07
Use the number prefix to find the right spec

thenets:
## Description

This PR updates the spec directory lookup mechanism to support multiple branches working on the same specification. Previously, branch names had to exactly match spec directory names, preventing multiple feature branches from sharing the same spec.

## Problem

The old implementation required an exact match between branch name and directory name:
- Branch `004-hashicorp-vault` → looked for `specs/004-hashicorp-vault/`
- Actual directory: `specs/004-vaults/`
- Result: **Failed** ❌

This prevented developers from creating multiple branches for the same specification (e.g., `004-fix-bug`, `004-add-feature`, `004-refactor`).

## Solution

Implemented prefix-based directory matching:
- Extracts numeric prefix from branch (e.g., `004` from `004-whatever`)
- Searches `specs/` for directories starting with that prefix
- Matches: `004-hashicorp-vault` → `specs/004-vaults/` ✅

## TODO

- [X] Add `bash` support
- [ ] Add `powershell` support

Note: I didn't create a powershell equivalent because I don't even know how to test it. I accept help with that, feel free to commit to my PR.

### DIFF
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -19,21 +19,21 @@ get_current_branch() {
         echo "$SPECIFY_FEATURE"
         return
     fi
-    
+
     # Then check git if available
     if git rev-parse --abbrev-ref HEAD >/dev/null 2>&1; then
         git rev-parse --abbrev-ref HEAD
         return
     fi
-    
+
     # For non-git repos, try to find the latest feature directory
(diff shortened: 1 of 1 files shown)


thenets:
Is anyone there?

PR #805, opened 2025-10-09
Feature :  Remote catalogue of constitutions

housekelly:
The intention here is to allow support for fetching pre-made constitutions from a remote github repository. It would be beneficial to allow a dev group or company to have well curated constitutions that contain all of the  steering guidelines, best practices, and company specifics to allow the specify-cli to read from a remote repo and allow the user to select one, instead of starting from scratch each time.

This feature adds a few new cli arguments to facilitate.

Screenshot in action.  
image

**File Modified**: `src/specify_cli/__init__.py`

**New Functions**:
- `fetch_remote_constitutions_list()` - Fetches list of constitutions from GitHub
- `fetch_remote_constitution_content()` - Downloads specific constitution content

**New Command**:
- `list-constitutions` - Browse available constitutions from a repository

**Enhanced Command**:
- `init` - Now supports 5 new options for remote constitutions:
  - `--constitution-repo`
  - `--constitution-name`
  - `--constitution-path`
  - `--constitution-branch`
  - `--constitution-interactive`

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -64,6 +64,94 @@ def _github_auth_headers(cli_token: str | None = None) -> dict:
     token = _github_token(cli_token)
     return {"Authorization": f"Bearer {token}"} if token else {}
 
+def fetch_remote_constitutions_list(repo_url: str, branch: str = "main", path: str = "", github_token: str = None) -> list[dict]:
+    """
+    Fetch a list of available constitutions from a remote GitHub repository.
+    
+    Args:
+        repo_url: GitHub repository URL (e.g., 'https://github.com/owner/repo' or 'owner/repo')
+        branch: Branch name to fetch from (default: 'main')
+        path: Path within the repository where constitutions are stored (default: root)
+        github_token: Optional GitHub token for private repositories
+        
+    Returns:
+        List of dicts with 'name', 'path', and 'download_url' keys
+    """
+    # Parse repo_url to extract owner and repo
+    repo_url = repo_url.strip()
+    if repo_url.startswith("https://github.com/"):
+        repo_url = repo_url.replace("https://github.com/", "")
+    repo_url = repo_url.rstrip("/")
+    
+    parts = repo_url.split("/")
+    if len(parts) < 2:
+        raise ValueError(f"Invalid repository URL: {repo_url}. Expected format: 'owner/repo' or 'https://github.com/owner/repo'")
+    
+    owner, repo = parts[0], parts[1]
+    
+    # Construct API URL to list directory contents
+    api_url = f"https://api.github.com/repos/{owner}/{repo}/contents/{path}"
+    if branch != "main":
+        api_url += f"?ref={branch}"
+    
+    try:
+        response = client.get(
+            api_url,
+            timeout=30,
+            follow_redirects=True,
+            headers=_github_auth_headers(github_token),
+        )
+        
+        if response.status_code == 404:
+            raise RuntimeError(f"Repository or path not found: {owner}/{repo}/{path}")
+        elif response.status_code != 200:
+            raise RuntimeError(f"GitHub API returned {response.status_code} for {api_url}")
+        
+        contents = response.json()
+        
+        # Filter for markdown files that look like constitutions
+        constitutions = []
+        for item in contents:
+            if item["type"] == "file" and item["name"].endswith(".md"):
+                constitutions.append({
+                    "name": item["name"].replace(".md", ""),
+                    "path": item["path"],
+                    "download_url": item["download_url"],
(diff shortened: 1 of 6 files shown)


housekelly:
@localden Curious on your thoughts here.  I feel it would be a positive feature for the project to support a remote catalogue of well crafted constitutions.  Please let me know if there is anything you feel needs changing.

dasiths:
@housekelly can we do this as part of https://github.com/github/spec-kit/issues/892

MadHuslista:
@housekelly  Nice one, kudos! I was questioning something like this while needing to switch between constitutions_c & constitution_py, in a Embedded C + Python codebase.  

How would you manage a sort of constitution tree or principles composition?  ( @localden  ) 

I'm seeing that there are couple of layers that arise from multiple constitutions: 
- Universals (principles used constantly for all constitutions) 
- Codebase specifics (for this repo use these ones, for C those other ones, fro Py. those others,  etc.. ) 
- Other layers.. 

So following up with your idea, I was thinking about setting constitutions or principles layers (not sure about the best atomicity) and enable a sort of "pick your poison" workflow in the command. 

I imagine something like: 
- Refresh availables => select const/principles => Trigger composition 

What do you think would be the best approach to the issue?

housekelly:
@dasiths happy to update it to support a folder which can include other notable files like your issue suggests. I feel like this functionality (remote repo support) belongs in the core list of cli functions.

As to @MadHuslista 's point.  I would prefer the tool not be very opinionated on the structure of the remote constitution for this very reason.  People will care about team practices, cloud practices, federal steering, etc.  If we support a remote repo i had envisioned writing some workflows that would combine those different layers into the final set of constitutions.  Let each group of devs worry about their own heirarchy

jedjohan:
Good stuff, exactly what our team starting spec-driven has been thinking about (we now have 2 repos frontend/backend) and they need to share a "domain knowledge" constitution. One follow-up idea: the new helpers in [__init__.py](vscode-file://vscode-app/c:/Users/jedjo/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (fetch_remote_constitutions_list, fetch_remote_constitution_content, and the CLI wiring) are tightly coupled to the GitHub REST API. 

If we can carve out a small provider abstraction there (parsing repo URLs, routing to provider-specific fetchers), the same CLI could support for example Azure DevOps repos too—most of the wiring is already done. That would let us reuse the new options in both the init and list-constitutions commands, and mirror the docs updates ([remote-constitutions.md](vscode-file://vscode-app/c:/Users/jedjo/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [README.md](vscode-file://vscode-app/c:/Users/jedjo/AppData/Local/Programs/Microsoft%20VS%20Code%20Insiders/resources/app/out/vs/code/electron-browser/workbench/workbench.html), CHANGELOG) for Azure DevOps usage. 

Long-term, having that abstraction in place would make it straightforward to plug in additional SCMs down the road (GitLab, Bitbucket, etc.), giving teams more flexibility without revisiting the CLI surfaces again

src/specify_cli/__init__.py
```diff
@@ -64,6 +64,94 @@ def _github_auth_headers(cli_token: str | None = None) -> dict:
...
+    repo_url = repo_url.rstrip("/")
+    
+    parts = repo_url.split("/")
+    if len(parts) < 2:
+        raise ValueError(f"Invalid repository URL: {repo_url}. Expected format: 'owner/repo' or 'https://github.com/owner/repo'")
+    
+    owner, repo = parts[0], parts[1]
+    
+    # Construct API URL to list directory contents
+    api_url = f"https://api.github.com/repos/{owner}/{repo}/contents/{path}"
```
RoyceLeonD, line 93: I know this is a GH project, but could you not hardcode to github. Instead only support fqdn with raw text or tar distributions (tar is less important). And Instead of it being a folder have it be a single file manifest. Similar to openapispec, ideally this model follows standard artifact + cdn distribution thus I get to use those same tools to publish/ manage/ version control. 

Example: we use Jfrog arifactory and would like to run our spec through CI/CD and then publish (thus we can security scan, we can have uptime healthchecks, retrieve clear usage, utilize our existing data-rentention policy), this is valuable for us since our developers' rbac is used in flight to hit a zero trust endpoint and only show specs related to their identify, we do a similar thing for packages. Not scope creeping. But rather would like to instead reduce the client side processing speckit is doing. Instead I think add a publish command/ action on github that treats specs like artifacts. 

This would flow nicely into other things you do. I personally don't like tis distribution model. Thoughts?

PR #825, opened 2025-10-10
Spec Kit Updates (0.0.19)

localden:


ben-edgar:
@localden I think I found a bug with this PR, I just tried installing spec kit for cursor and it looks like there's an issue since now the CLI tool is looking for `cursor-agent` in the release name instead of `cursor`. Did you intend to update the release name to use `cursor-agent` as well? I'm already working on opening up a PR to do that, but let me know if that's not desired!
Screenshot 2025-10-10 at 8 41 18 PM

PR #862, opened 2025-10-13
Update AI coding agents list in README

darkogj:
Added Kilo Code to the list of agents


localden:
Thank you for the contribution! I am actually going to clean up that agent list further to make sure we're not listing them fully in that section because it's already covered in the agent list.

PR #871, opened 2025-10-14
added offline mode for spec kit

kumaresra:
Add offline_template to enable developers with corporate firewall to access spec kit

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ Our research and experimentation focus on:
 ## 🔧 Prerequisites
 
 - **Linux/macOS/Windows**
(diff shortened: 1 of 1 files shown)


src/specify_cli/__init__.py
```diff
@@ -691,18 +732,16 @@ def download_and_extract_template(project_path: Path, ai_assistant: str, script_
...
-                tracker.complete("cleanup")
-            elif verbose:
-                console.print(f"Cleaned up: {zip_path.name}")
-    
+            tracker.add("cleanup", "Check temporary archive")
+            # Clean up downloaded ZIP file unless it's the specified offline template
+            if zip_path.exists():
+                if not offline_template or str(zip_path) != str(Path(offline_template).resolve()):
+                    zip_path.unlink()
+                if tracker:
```
lshgdut: this block should not be run where offline mode?

Issue #938, opened 2025-10-18
"specs" directory is not created in same level as ".specify" directory when .git is on a different directory


leobrines:
after executing `specify init .` and running the constitution script, the specify command creates the `specs/` directory in the ../ directory. This happens because I have the .git in ../.git path but it feels unexpected because it already has a .specify directory

Example:

First, I have this initial multiproject structure:

multiproject/.git
multiproject/project1/script1.sh
multiproject/project1/README.md
multiproject/project2/README.md

Then, I run the `specify init .` and the constitution script, all inside the project2 directory:

multiproject/.git
multiproject/project1/script1.sh
multiproject/project1/README.md
multiproject/project2/.specify (new directory)
multiproject/project2/.cursor(new directory)
multiproject/project2/README.md

Then, when I run the specify cursor command (previously project was open in cursor using `cursor project1` command), it creates the specs/ directory in an unexpected place:

multiproject/.git
multiproject/project1/script1.sh
multiproject/project1/README.md
multiproject/project2/.specify
multiproject/project2/.cursor
multiproject/project2/README.md
multiproject/specs/ (new directory)

I think because it is finding the .git directory so it thinks that is the correct place, but I think .specify should have priority over .git

PR #939, opened 2025-10-18
DRAFT: fix: prioritize specify directory

leobrines:
possible fix for issue https://github.com/github/spec-kit/issues/938

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -59,21 +59,22 @@ find_repo_root() {
     return 1
 }
 
-# Resolve repository root. Prefer git information when available, but fall back
-# to searching for repository markers so the workflow still functions in repositories that
-# were initialised with --no-git.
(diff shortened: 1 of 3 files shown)


scripts/bash/create-new-feature.sh
```diff
@@ -59,21 +59,22 @@ find_repo_root() {
...
+    if git rev-parse --show-toplevel >/dev/null 2>&1; then
+        HAS_GIT=true
+    else
+        HAS_GIT=false
     fi
-    HAS_GIT=false
+else
+    echo "Error: Could not determine repository root. Please run this script from within the repository." >&2
+    exit 1
 fi
```
jtama, line 78:
```suggestion
# Source common functions
source "$SCRIPT_DIR/common.sh"

# Get all paths and variables from common functions
eval $(get_feature_paths)
```

If you have already fixed the get_repo_root method, why not use it ?

PR #940, opened 2025-10-18
Update scripts to keep all spec-kit related files and folders in the .specify directory

seoljacob:
### Abstract
The specs dir was being made in the repo root dir. This PR makes updates to `scripts/bash/common.sh` and `scripts/bash/create-new-feature.sh` (along with their PowerShell counterparts) so that the specs dir is made in the .specify dir. This keeps all of the spec-kit related files and folders contained in one location.

### Testing Performed
Tested these changes locally, verifying that the workflow from /specify.consitution to /specify.implementation works as expected.

### DIFF
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -28,7 +28,7 @@ get_current_branch() {
 
     # For non-git repos, try to find the latest feature directory
     local repo_root=$(get_repo_root)
-    local specs_dir="$repo_root/specs"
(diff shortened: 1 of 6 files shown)


localden:
Thank you for the contribution, @seoljacob - at this time, we do not want to keep specs inside the `.specify` directory, as they are standalone code-related artifacts and not Spec Kit metadata.

PR #947, opened 2025-10-18
Add Chinese and Japanese translations of README

AndersHsueh:
add translations for readme.  (simpel-cn, and Japanese)


localden:
Thank you for the contribution! At this time, we will not be merging translations because they add maintenance burden that goes outside the scope of what the maintainers of this project have bandwidth for.

PR #999, opened 2025-10-22
feat(ci): upgraded github action versions and added dependabot

sombraSoft:
- Updates github actions versions
- Adds `dependabot.yml` to keep track of github actions versions


localden:
Thank you for the contribution, @sombraSoft. This change is not needed at this time.

Issue #407, opened 2025-09-20
[Feature Request] Configurable naming convention for /specify branch and specs/ folder


jmrtnz94:
Please add a way to configure the naming convention used when /specify creates a Git branch and the corresponding specs/<…> folder. Many teams need names that include ticket IDs or follow GitFlow-style prefixes.

Current behavior
/specify auto-creates a branch (e.g., 003-chat-system) and a matching directory under specs/…. The names are not user-configurable.

Requirement (behavioral)
Whatever approach we choose, the branch name and folder name should always stay in sync. A single computed value should be used to create both.

⸻

Approaches (seeking maintainer preference)

Kept in the repo’s style (simple CLI, optional config under .specify/):
	1.	One-off CLI flags (explicit, easy to trial)
	•	specify --branch-format "<template>"
	•	Folder path derived as specs/{branch} by default
	•	(Optional) --folder-format "<template>" so long as it includes {branch} (to guarantee sync)
	2.	Repo-wide config (opt-in)
	•	.specify/config.toml (or .specify/config.json)
	•	Example:

[naming]
branch_format = "feature/{ticket}-{kebab(summary)}"  # e.g., feature/PROJ-123-sync-webhooks
folder_format = "specs/{branch}"                     # keeps folder in lockstep with branch
# Optional validation:
validate_branch = "^(feature|bugfix|hotfix)/[A-Z]+-\\d+-[a-z0-9-]+$"

	•	Minimal token set (practical and easy to document):
	•	{seq} (zero-padded sequence like 003)
	•	{summary} / {kebab(summary)}
	•	{ticket} (via --ticket PROJ-123 or inferred by regex)
	•	{date} (YYYYMMDD)

	3.	Environment variables (handy in CI)
	•	SPECIFY_BRANCH_FORMAT, SPECIFY_FOLDER_FORMAT

Precedence (suggested): flag > env > config > default.

⸻

Back-compat & safety
	•	If no flags/env/config are provided, behavior remains exactly as today.
	•	Optional validate_branch lets teams enforce a regex and fail fast with a clear error.
	•	Sensible sanitization (lowercase, spaces → -, strip illegal chars).

⸻

Questions for maintainers
	•	Which interface would you prefer we implement first: flags, config, or both with precedence?
	•	TOML vs JSON for .specify/?
	•	Keep folder strictly derived as specs/{branch} (simplest sync), or allow a folder_format as long as it includes {branch}?
	•	Are the proposed tokens sufficient for v1?

[...]

PR #1009, opened 2025-10-22
Feature/configure branch name

ricardojmendez:
Contains several commits allowing the user to specify some values around branch and spec name:

- Branch prefix, allowing for feature/, bugfix/, or other patterns to match the user's / team's approach;
- An arbitrary spec number, to potentially tie it into the corresponding issue on Github, instead of just using a sequence.

It also includes an optional configuration file for default branch name, since I expect users will create one type of branch more often than others (eg. "feature").

This addresses some of the requests in #407.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -50,8 +50,6 @@
 		"kilocode.Kilo-Code",
 		// Roo Code
 		"RooVeterinaryInc.roo-cline",
-		// Amazon Developer Q
-		"AmazonWebServices.amazon-q-vscode",
 		// Claude Code
 		"anthropic.claude-code"
 	],
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -51,32 +51,33 @@ echo -e "\n🤖 Installing OpenCode CLI..."
 run_command "npm install -g opencode-ai@latest"
 echo "✅ Done"
 
-echo -e "\n🤖 Installing Amazon Q CLI..."
-# 👉🏾 https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-verify-download.html
+echo -e "\n🤖 Installing Kiro CLI..."
+# https://kiro.dev/docs/cli/
+KIRO_INSTALLER_URL="https://cli.kiro.dev/install"
+KIRO_INSTALLER_SHA256="7487a65cf310b7fb59b357c4b5e6e3f3259d383f4394ecedb39acf70f307cffb"
+KIRO_INSTALLER_PATH="$(mktemp)"
 
-run_command "curl --proto '=https' --tlsv1.2 -sSf 'https://desktop-release.q.us-east-1.amazonaws.com/latest/q-x86_64-linux.zip' -o 'q.zip'"
-run_command "curl --proto '=https' --tlsv1.2 -sSf 'https://desktop-release.q.us-east-1.amazonaws.com/latest/q-x86_64-linux.zip.sig' -o 'q.zip.sig'"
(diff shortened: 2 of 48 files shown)


ricardojmendez:
Any feedback on this PR, @localden?

Alexandre-Nourissier:
@ricardojmendez Thank you for this contribution, I am really looking forward to it. 

As it stand, it helps with the prefix of branches (feature/, bugfix/, etc...), but there is still the question of branch naming. It is less of a blocker, but some organizations have naming rules such as : `<WorkItem-Key>_name_of_the_branch`, where WorkItem-Key can be coming from Jira, Azure DevOps, etc...., and have an alphanumerical format including a shorthand for the project, such as "PROJECTKEY-1" (example: "feature/SHOP-39_add_coupons_to_cart"). It is also a way for work planning tools to automatically keep track of implementation work.  

So your approach is not addressing this aspect as it is. It seems that this would require a decent amount of configuration, (but can be simplified for the end user, since it is a convention).

That said, I am favorable with doing it step by step, and deliver the value of your PR as quick as possible, so we can consider going further at a later date.

Shehab-Muhammad:
Thanks so much for your effort @ricardojmendez 
However, I don't understand why it should be opinionated when it comes to branch name
branch names should be fully configurable
for example we are using clickup for tickets and we need to link our branchs to tickets on clickup
and it has its naming convention (includes ticket number, feature name and auther name)
I think the best way is to suggest the default existing way on `specify init` should ask user about branch naming convention if it's ok to use defult naming convention `001-user-auth` or enter custom naming convention or just ignore branchs and let user handle it manually
I think in the future it should git worktrees this's very important as well

ricardojmendez:
fwiw... On my tests, /specify seems to be perfectly fine with you creating the branch yourself and telling it to _Use the existing branch_.

You may need to update the branch validation code to avoid it grumbling, but I don't believe that it having an issue with the naming is a showstopper.

Shehab-Muhammad:
I think it should be supported, I don't have to change branch validation myself on each repo

mnriem:
@ricardojmendez As you said "Contains several commits allowing the user to specify some values around branch and spec name" can you break it up in multiple PRs. Specifically bug fixes should not tag along new features and vice versa

ricardojmendez:
@mnriem 

> "Contains several commits allowing the user to specify some values around branch and spec name"
> bug fixes should not tag along new features and vice versa

Not sure which part of that indicates that this includes bug fixes, or even features beyond the branch and naming (which go hand in hand).

mnriem:
I was referring to "An arbitrary spec number, to potentially tie it into the corresponding issue on Github, instead of just using a sequence." There were issues filed around that one. So I want to isolate that one from any other commit since it would address a bug (in this case we should validate if that one is no longer needed).

ricardojmendez:
> I was referring to "An arbitrary spec number, to potentially tie it into the corresponding issue on Github, instead of just using a sequence." There were issues filed around that one.

To be clear: this is not meant to address a bug on spec-kit, but so that a spec-kit user can say "this spec we are working on is for bugfix XYZ", which is internal to their project.

This is a self-contained PR specifically around branch naming. It does what it says on the tin.

If you think it's worth merging, I can figure out the current conflicts. If you don't, or still believe this addresses multiple unrelated issues, please do reject it.

mnriem:
OK, please. address the merge conflicts and the feedback Copilot gave where applicable. If not applicable please describe why it is not. Also make sure this is backwards compatible. Thanks a lot!

ricardojmendez:
@mnriem Fixed the couple conflicting files and pushed again.

gabrielhmsantos:
There is a similar PR ( #1511 ) open; I believe you could decide which one will implement this feature.

PR #1070, opened 2025-10-27
Refactor ESLint configuration checks in implement.md to address deprecation

### LINKED ISSUE #1069: The Implement command (re-)creates deprecated ESLint config files every run, causing issues
The [implement.md](https://github.com/github/spec-kit/blob/e6d6f3cdee99752baee578896797400a72430ec0/templates/commands/implement.md?plain=1#L70) command is re-creating a `.eslintignore` file every run. This file has been deprecated and causing warnings and config duplication: https://eslint.org/docs/latest/use/configure/ignore-deprecated

```markdown
   - Check if .eslintrc*or eslint.config.* exists → create/verify .eslintignore
```

danwashusen:
## Summary
- stop regenerating deprecated `.eslintignore` when a flat `eslint.config.*` is present
- keep `.eslintignore` guidance for legacy `.eslintrc*` configs only
- direct flat-config projects to rely on the config’s own `ignores` entries for required patterns

## Testing
- n/a (docs change)

Fixes #1069

### DIFF
--- a/templates/commands/implement.md
+++ b/templates/commands/implement.md
@@ -67,7 +67,8 @@ You **MUST** consider the user input before proceeding (if not empty).
      ```
 
    - Check if Dockerfile* exists or Docker in plan.md → create/verify .dockerignore
-   - Check if .eslintrc*or eslint.config.* exists → create/verify .eslintignore
+   - Check if .eslintrc* exists → create/verify .eslintignore
+   - Check if eslint.config.* exists → ensure the config's `ignores` entries cover required patterns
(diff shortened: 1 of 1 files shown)


danwashusen:
This feels a little odd to me because of the "If ignore file already exists" instructions, but it works and keeps the edit tight...

As noted in https://github.com/github/spec-kit/issues/1069, I feel this would be better as part of a tracked task created during planning. I'm happy to proceed down that path if you'd prefer @localden...

PR #1023, opened 2025-10-23
fix: check all git branches for feature numbering to prevent duplicates

Ashwinhegde19:
Fixes #975

## Problem
The `Get-NextBranchNumber` function was checking only branches with the same short name when determining the next feature number, causing duplicate feature numbers when multiple feature branches exist that haven't been merged to main.

## Solution
Updated the function to check ALL existing git branches matching the pattern `^\d{3}-` instead of only branches with the same short name.

## Changes
- Removed `$ShortName` parameter from `Get-NextBranchNumber` function
- Updated regex patterns to match ALL feature branches:
  - Remote: `refs/heads/(\d{3})-`
  - Local: `^\*?\s*(\d{3})-`
  - Specs: `^(\d{3})-`
- Updated function call to remove `-ShortName` parameter
- Now correctly finds highest feature number across ALL features

PR #1109, opened 2025-11-03
fix: check all git branches for feature numbering to prevent duplicates

### LINKED ISSUE #1101: Bash script creates duplicate feature numbers (filters by short name instead of checking all branches)
## Bug Description

The bash script `.specify/scripts/bash/create-new-feature.sh` generates duplicate feature numbers when multiple feature branches exist, similar to the PowerShell bug fixed in #1023.

## Root Cause

The `check_existing_branches` function (lines 84-112) filters branches by the specific `short_name`:

```bash
# Line 91: Only matches branches with the SAME short name
grep -E "refs/heads/[0-9]+-${short_name}$"
```

This causes it to check only branches matching the current feature name pattern, not ALL branches globally.

## Expected Behavior

Should scan all feature branches (like the non-git fallback in lines 201-207) and return the next global sequential number.

## Actual Behavior

With existing specs `001-build-an-account` and `002-oauth2-google-auth`, running the script with `--short-name multilingual-support` creates `001-multilingual-support` instead of `003-multilingual-support`.

## Fix

Apply the same fix from #1023: remove the `short_name` filter from `check_existing_branches` and match pattern `refs/heads/[0-9]+-` globally.

---

**Note**: PowerShell version already fixed in #1023, bash version needs same treatment.

Ashwinhegde19:
Fixes #1101

## Problem

The `check_existing_branches` function in `scripts/bash/create-new-feature.sh` was checking only branches with the same short name when determining the next feature number, causing duplicate feature numbers when multiple feature branches exist that haven't been merged to main.

This is the same bug that was fixed for PowerShell in #1023, but affecting the Bash script.

## Solution

Updated the function to check ALL existing git branches matching the pattern `[0-9]+-` instead of only branches with the same short name.

## Changes

- Updated grep patterns in `check_existing_branches` function
- **Remote branches (line 91)**: Changed from `grep -E "refs/heads/[0-9]+-${short_name}$"` to `grep -E "refs/heads/[0-9]+-"`
- **Local branches (line 94)**: Changed from `grep -E "^[* ]*[0-9]+-${short_name}$"` to `grep -E "^[* ]*[0-9]+-"`
- Now correctly finds highest feature number across ALL features
- Mirrors the fix from PowerShell script (#1023)

## Example

**Before (buggy behavior):**
- Existing branches: `001-build-an-account`, `002-oauth2-google-auth`
- Running script with `--short-name multilingual-support`
- Result: Creates `001-multilingual-support` ❌ (WRONG - duplicate!)

**After (correct behavior):**
- Existing branches: `001-build-an-account`, `002-oauth2-google-auth`
- Running script with `--short-name multilingual-support`
- Result: Creates `003-multilingual-support` ✅ (CORRECT - next sequential number)

## Related

This completes the fix started in #1023, resolving the duplicate numbering issue for both PowerShell and Bash users.

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -82,7 +82,6 @@ find_repo_root() {
 
 # Function to check existing branches (local and remote) and return next available number
 check_existing_branches() {
-    local short_name="$1"
     
     # Fetch all remotes to get latest branch info (suppress errors if no remotes)
     git fetch --all --prune 2>/dev/null || true
@@ -193,7 +192,7 @@ fi
 if [ -z "$BRANCH_NUMBER" ]; then
     if [ "$HAS_GIT" = true ]; then
         # Check existing branches on remotes
-        BRANCH_NUMBER=$(check_existing_branches "$BRANCH_SUFFIX")
+        BRANCH_NUMBER=$(check_existing_branches)
     else
         # Fall back to local directory check
         HIGHEST=0


maxmeyer3141:
HI there, i think this won't completly fix the problem with the branch numbering when using coding agents because the speckit.specify command prompt template instructs the agent to call the script with the Number parameter which bypasses this solution.
I suggest to update the git commands in https://github.com/github/spec-kit/blob/main/templates/commands/specify.md aswell.

Template (Source)
b. Find the highest feature number across all sources for the short-name:
Remote branches: git ls-remote --heads origin | grep -E 'refs/heads/[0-9]+-`<short-name>`$'
Local branches: git branch | grep -E '^[* ]*[0-9]+-`<short-name>`$'
Specs directories: Check for directories matching specs/[0-9]+-`<short-name>`

Template (fix)
b. Find the highest feature number across all sources for the short-name:
- Remote branches: git ls-remote --heads origin | grep -E 'refs/heads/[0-9]+-$'
- Local branches: git branch | grep -E '^[* ]*[0-9]+-$'
- Specs directories: Check for directories matching specs/[0-9]+-

Am i correct?

Ashwinhegde19:
> HI there, i think this won't completly fix the problem with the branch numbering when using coding agents because the speckit.specify command prompt template instructs the agent to call the script with the Number parameter which bypasses this solution. I suggest to update the git commands in https://github.com/github/spec-kit/blob/main/templates/commands/specify.md aswell.
> 
> Template (Source) b. Find the highest feature number across all sources for the short-name: Remote branches: git ls-remote --heads origin | grep -E 'refs/heads/[0-9]+-`<short-name>`$' Local branches: git branch | grep -E '^[* ]*[0-9]+-`<short-name>`$' Specs directories: Check for directories matching specs/[0-9]+-`<short-name>`
> 
> Template (fix) b. Find the highest feature number across all sources for the short-name:
> 
> * Remote branches: git ls-remote --heads origin | grep -E 'refs/heads/[0-9]+-$'
> * Local branches: git branch | grep -E '^[* ]*[0-9]+-$'
> * Specs directories: Check for directories matching specs/[0-9]+-
> 
> Am i correct?

@maxmeyer3141 You're right! I've just added commits a005e81 and b0dc835 that update both the Bash script and the `specify.md` template to check all branches instead of filtering by short-name. Thanks for catching that! 🙌

localden:
Thanks for the contribution - latest version should include a number of fixes in this domain.

PR #1113, opened 2025-11-04
feat: Add /speckit.innovate command for parallel brainstorming

copyboy:
- Add new /speckit.innovate command template with dynamic multi-role system
- Support three brainstorming modes (Quick/Standard/Deep)
- Implement cross-domain inspiration system
- Add structured challenge-rebuttal debates between agents
- Include session persistence and resumability features
- Add user impact analysis from multiple stakeholder perspectives
- Create comprehensive agent-roles.yaml configuration with 15+ role categories
- Add brainstorms directory structure template and documentation
- Update README.md to document the new command
- Update CHANGELOG.md with complete feature description

This command facilitates creative ideation using parallel subagent discussions, helping teams brainstorm innovative solutions through diverse expert perspectives.


offendingcommit:
I think this has some merit, but I would invoke bash or powershell commands like the other speckit commands do.

copyboy:
> I think this has some merit, but I would invoke bash or powershell commands like the other speckit commands do.

I didn't understand your question because I provided a command /speckit.innovate to trigger a brainstorming session.

offendingcommit:
So the other commands include bash or powershell scripts with load things into context and I was only suggesting that you could include some of the same. 😄

localden:
@copyboy can you share more about how you tested this and what the output is across different scenarios? I think I get the idea, but would love to see how this works in practical terms.

PR #1196, opened 2025-11-15
 preserve-constitution flag

roee-hersh:


offendingcommit:
Seems like an easy thing. Although, I'd prefer this to be the default and maybe leverage git to restore it. Using --force to overwrite it.

Additionally, I'd also like to see something similar with the templates like plan-template etc since sometimes as constitutions get updated so do some of the preflight checks in the templates make sense to keep a specific format.

PR #1198, opened 2025-11-15
Fix branch numbering to increment globally across all features

danieldekay:
The `/specify` command was always creating branches starting at `001`, ignoring existing feature branches with different short-names.

## Problem

Given existing branches `001-user-auth`, `002-payment-flow`, `003-dashboard`, creating a new feature would incorrectly produce `001-admin-panel` instead of `004-admin-panel`.

## Root Cause

The `check_existing_branches()` (bash) and `Get-NextBranchNumber()` (PowerShell) functions filtered by exact short-name match:

```bash
# Before: Only finds branches matching the specific short-name
grep -E "refs/heads/[0-9]+-${short_name}$"

# After: Finds all feature branches
grep -E "refs/heads/[0-9]+-"
```

This meant each unique short-name started its numbering from 001, rather than continuing the global sequence.

### Bash (`scripts/bash/create-new-feature.sh`)
- Changed remote, local, and specs directory searches to match pattern `[0-9]+-*` instead of `[0-9]+-${short_name}`
- Added octal-to-decimal conversion (`10#$num`) to handle leading zeros correctly

### PowerShell (`scripts/powershell/create-new-feature.ps1`)
- Applied equivalent regex changes to match all feature branches
- Maintains parity with bash implementation

### Documentation (`templates/commands/specify.md`)
- Updated instructions to clarify branch numbers increment globally across all features

## Behavior

Branch numbers now increment sequentially regardless of short-name:
- `001-user-auth` → `002-payment-flow` → `003-dashboard` → `004-admin-panel`

<details>

<summary>Original prompt</summary>

> Fix the branch numbering issue with /specify command. Currently, when creating a new branch with /specify, it doesn't correctly find the right branch number and always starts at 001. The command should properly detect existing branches and increment the branch number correctly based on the highest existing branch number.

</details>

*This pull request was created as a result of the following prompt from Copilot chat.*
> Fix the branch numbering issue with /specify command. Currently, when creating a new branch with /specify, it doesn't correctly find the right branch number and always starts at 001. The command should properly detect existing branches and increment the branch number correctly based on the highest existing branch number.

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -134,21 +134,23 @@ check_existing_branches() {
     # Fetch all remotes to get latest branch info (suppress errors if no remotes)
     git fetch --all --prune 2>/dev/null || true
     
-    # Find all branches matching the pattern using git ls-remote (more reliable)
-    local remote_branches=$(git ls-remote --heads origin 2>/dev/null | grep -E "refs/heads/[0-9]+-${short_name}$" | sed 's/.*\/\([0-9]*\)-.*/\1/' | sort -n)
+    # Find all feature branches using git ls-remote (check all branches, not just matching short-name)
+    local remote_branches=$(git ls-remote --heads origin 2>/dev/null | grep -E "refs/heads/[0-9]+-" | sed 's/.*\/\([0-9]*\)-.*/\1/' | sort -n)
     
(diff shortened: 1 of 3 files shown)


localden:
Thank you for the contribution! This should already be addressed in the latest version.

PR #1236, opened 2025-11-23
feat: add --template-path option to specify init

cometzero:
## Summary

This PR adds a new `--template-path` option to the `specify init` command, allowing users to initialize projects from local template files instead of downloading from GitHub.

## Problem

Previously, `specify init` always downloaded templates from GitHub, which required manual copying of template ZIP files when implementing or testing new agents. This created friction in the development workflow.

## Solution

Add `--template-path` option to `specify init` that accepts both file and directory paths:
- **File path**: Directly uses the specified ZIP file
- **Directory path**: Searches for template files matching the pattern `spec-kit-template-{ai}-{script}-*.zip`

## Usage

```bash
# Using a directory containing template files
specify init my-project --ai claude --script sh --template-path .genreleases

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -736,36 +736,66 @@ def download_template_from_github(ai_assistant: str, download_dir: Path, *, scri
     }
     return zip_path, metadata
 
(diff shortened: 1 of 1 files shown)


localden:
Thank you for your contribution. At this time, we have no plans on adding local template support because the templates themselves are just ZIP files - one can extract them in whatever folder they need.

PR #1237, opened 2025-11-23
fix: use global maximum for branch numbering to prevent collisions

Mearman:
The check_existing_branches (bash) and Get-NextBranchNumber (PowerShell) functions were only looking for branches/specs matching the SAME short name when determining the next feature number. This caused collisions where multiple features could be assigned the same number if they had different short names.

For example, if feature 023-ci-optimization existed, creating a new feature with a different short name would incorrectly use 001 instead of 024.

This fix changes both functions to:
1. Use get_highest_from_branches() / Get-HighestNumberFromBranches to find the highest number across ALL branches globally
2. Use get_highest_from_specs() / Get-HighestNumberFromSpecs to find the highest number across ALL spec directories globally
3. Return the maximum of both + 1

The helper functions already existed but were not being used. This fix properly utilizes them to ensure features are numbered sequentially regardless of their short names.

---

## Additional Fix: Octal Interpretation Bug

**Bug**: When `--number 027` was passed to the bash script, `printf '%03d'` interpreted it as octal, converting 027 (octal) to 23 (decimal).

**Root Cause**: The bash `printf %03d` command treats strings with leading zeros as octal notation.

**Fix**: Added `$((10#$BRANCH_NUMBER))` to force base-10 interpretation before formatting.

**Example**:
- **Before**: `--number 027` → octal 027 → decimal 23 → feature 023
- **After**: `--number 027` → decimal 27 → feature 027

**Note**: The PowerShell version does not have this bug because `[int]` type conversion in PowerShell does not treat leading zeros as octal.

---

**Issues Fixed**:
1. Branch number collisions when creating features with different names
2. Incorrect feature numbers when using `--number` with leading zeros

**Impact**: Ensures features are numbered correctly and sequentially

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -246,7 +246,7 @@ if [ -z "$BRANCH_NUMBER" ]; then
     fi
 fi
 
-# Force base-10 interpretation to prevent octal conversion (027 → 23)
+# Force base-10 interpretation to prevent octal conversion (e.g., 010 → 8 in octal, but should be 10 in decimal)
 FEATURE_NUM=$(printf "%03d" "$((10#$BRANCH_NUMBER))")
 BRANCH_NAME="${FEATURE_NUM}-${BRANCH_SUFFIX}"
 


Mearman:
@localden I've been running into this bug quite often

Issue #1332, opened 2025-12-12
10th spec given same name as first spec


robshapiro:
My first spec was `001-phase1-quick-wins`. I just created my 10th feature, which was created as spec `001-handyman-request`. 

You can see that the 10th feature has the same number as the first, `001`. 

Feels like it should never create a spec w/ the same "ID"? In the image below you'll see two created w/ 001

<img width="828" height="816" alt="Image" src="https://github.com/user-attachments/assets/7ad93475-2ae1-4cc5-9c02-64c4929cae6d" />

kassi:
This happened to me once on the 3rtd, once on the 5th feature I wanted to implement.
At some point the script under the hood doesn't really recognize correctly what's already there.

I looked into `create-new-feature.sh` and found a couple of functions around determining existing dirs and branches for the numbering. Then I looked into the specify function and realized that it actually describes to the agent to determine the next number using some other commands before it actually runs `create-new-feature.sh`.
My output at this point says:

I’m going to (a) create the next numbered feature branch + spec scaffold, (b) inspect the submodule CSV metadata files to match their actual columns/order, then (c) write the spec and the separate requirements checklist file.
```
git fetch --all --prune
```
```
git ls-remote --heads origin | rg "refs/heads/[0-9]+-my-featurename$" || true
zsh: command not found: rg
```
```
git branch --list | rg "^[* ]*[0-9]+-my-featurename$" || true
zsh: command not found: rg
```
Quick adjustment: your shell doesn’t have rg, so I’ll compute the next feature number via a small python filter over git output (no grep/rg needed), then run the feature-creation script once.

before it actually lists some python code do determine the number.

This is somehow contradictory.

BretJohnson:
I believe the issue here is that the LLM instructions in https://github.com/github/spec-kit/blob/main/templates/commands/specify.md for determining the index are ambiguous - or arguably just wrong. The steps below literally say to (1) determine the short name and (2) determine the next sequential index number FOR THAT PARTICULAR SHORT NAME.  That is, make the index number not a global counter but a counter specific to a particular short name, which means that most of the time you'll get a "001" since having duplicated short names is rare.

The prompt instructions weren't always so explicit about that, but that changed with https://github.com/github/spec-kit/pull/1019, merged in October. So then I believe that some LLMs do that. While other LLMs interpret the instructions, especially instructions before that PR, as creating a global unique counter and doing that.

[...]

PR #1340, opened 2025-12-15
fix: validate prefix uniqueness when --number option is used

ggeutzzang:
## Summary

When `--number` is explicitly provided to `create-new-feature.sh`, this PR adds validation to ensure the specified prefix number is not already in use by another spec directory.

## Problem

Even after PR #1237 fixed the auto-detection logic, users can still create collisions by explicitly using the `--number` option:

```bash
# Existing spec
specs/001-feature-a/

# User runs with explicit --number
./create-new-feature.sh --number 1 --short-name "feature-b" "Description"

# Result: COLLISION!
specs/001-feature-b/  # Same prefix as 001-feature-a!
```

## Solution

Added validation after `FEATURE_NUM` is calculated:
- Checks all existing spec directories for the same prefix
- Allows retry of the same feature (same suffix)
- Blocks different features from using duplicate prefixes
- Provides helpful error message guiding users to auto-detection

## Test

```bash
$ ./create-new-feature.sh --number 1 --short-name "conflict" "Test"
Error: Prefix 001 is already used by '001-existing-feature'
Hint: Remove --number option to auto-detect the next available number
```

## Related

Related to #1332 - This addresses a remaining edge case not covered by #1237

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -248,6 +248,24 @@ fi
 
 # Force base-10 interpretation to prevent octal conversion (e.g., 010 → 8 in octal, but should be 10 in decimal)
 FEATURE_NUM=$(printf "%03d" "$((10#$BRANCH_NUMBER))")
+
+# Validate that the prefix number is not already in use by another spec
+# This prevents collisions when --number option is explicitly provided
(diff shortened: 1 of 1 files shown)


ggeutzzang:
@copilot review

Issue #685, opened 2025-09-30
support for fish


jamesainslie:
I, and others, use fish. Which is not POSIX compliant, and there are some necessary changes for the underlying logic to work. I have implemented a [PR 683](https://github.com/github/spec-kit/pull/683) 

## Specific Bash/Zsh Incompatibilities with Fish Shell

Based on the bash scripts in Spec Kit, here are the critical areas that would **fail catastrophically** in Fish Shell:

1. Variable Assignment Syntax
2. Command Substitution Syntax
3. Special Variables (`BASH_SOURCE`, `$$`, etc.)
4. Conditional Test Syntax
5. Array Syntax
6. Regex Matching with `BASH_REMATCH`
7. Error Handling (`set -e`, `set -u`, `set -o pipefail`)
8. Function Definition and Local Variables
9. `eval` with Output Capture
10. Here Documents and String Output
11. Arithmetic Expansion
12. Source/Sourcing Scripts
13. Argument Parsing

## Summary: Why Fish Shell Support is Non-Trivial

These aren't minor syntax differences—they're **fundamental incompatibilities** requiring complete rewrites:

| Area | Bash/Zsh | Fish | Complexity |
|------|----------|------|------------|
| Variable assignment | `VAR=value` | `set VAR value` | High |
| Arrays | `arr=() arr+=(val)` | `set arr; set -a arr val` | Very High |
| Conditionals | `[[ ]]` | `test` or `string match` | High |
| Regex | `=~` with `BASH_REMATCH` | `string match -r` | Very High |
| Special vars | `$$`, `BASH_SOURCE` | `$fish_pid`, `status filename` | Medium |
| Functions | `func() { }` | `function func; end` | Medium |
| Error handling | `set -e -u -o pipefail` | Different model | High |

davidvartanian:
Fish support would be amazing, I also use Fish and I guess many others who haven't left feedback yet also do! 🙏

PR #1358, opened 2025-12-20
feat(shell): initial Fish support

Francesco146:
Since it's a big change, i'll open multiple PR to simplify the review process.

Linked to the old: #685


localden:

PR #1360, opened 2025-12-20
feat(shell): add fish support to `common`

Francesco146:


localden:

PR #1363, opened 2025-12-20
feat(shell): add fish support to `update-agent-context`

Francesco146:
this is the last PR needed. i split them up because there are so many changes, and it would have resulted in a PR of about 1000 LoC, which would not have been pleasant to review :D


localden:

PR #1610, opened 2026-02-15
Implement polyglot scripts to avoid os specific commands in commands

nacho4d:


mnriem:
@nacho4d Can you address the feedback Copilot added?

PR #1632, opened 2026-02-19
Feat/ai skills

mnriem:


DovieW:
why would I want the agent changing modes by itself?

PR #1679, opened 2026-02-24
Speckit 2level specs

andrewvlgafc:
hello.  second try )
extended version of /speckit.specify 

usage /speckit.specify library: FB client
this make new feature not in /specs folder but in /specs/library 
i think it is good for grouping feaches


mnriem (changes requested):
Can you do this as an extension? See https://github.com/github/spec-kit/tree/main/extensions

andrewvlgafc:
> Can you do this as an extension? See https://github.com/github/spec-kit/tree/main/extensions

I think that on this link is a kit to create new commands? I only inprove existent. is'nt it?

mnriem (changes requested):
Can you please do it as an extension so the community can try it out. I would like the community weigh in and that is the best way to see.

PR #1706, opened 2026-02-26
chore: Update outdated GitHub Actions versions

mnriem:
This PR updates outdated GitHub Action versions.

- Updated `actions/checkout` from `v4` to `v6` in `.github/workflows/release.yml`
- Updated `actions/checkout` from `v4` to `v6` in `.github/workflows/docs.yml`
- Updated `actions/checkout` from `v4` to `v6` in `.github/workflows/lint.yml`

This is a reopened version of #1508 which appeared to be blocked.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


mnriem:
Also closes https://github.com/github/spec-kit/pull/1508

PR #1810, opened 2026-03-11
fix: clean up command templates (specify, analyze)

pierluigilenoci:
## Summary

- **specify.md**: Fix self-referential step number — step 6c says "proceed to step 6" but should say "proceed to step 7" (Report completion)
- **specify.md**: Remove empty `## General Guidelines` heading that has no content before `## Quick Guidelines`
- **analyze.md**: Deduplicate `{ARGS}` — the `## Context` section at the bottom repeats the same arguments already present in the `## User Input` section at the top

## Test plan

- [ ] Verify `specify.md` step 6c correctly references step 7
- [ ] Verify no empty headings remain in `specify.md`
- [ ] Verify `analyze.md` still passes `{ARGS}` context correctly via the `## User Input` section

### DIFF
--- a/templates/commands/analyze.md
+++ b/templates/commands/analyze.md
@@ -182,4 +182,6 @@ Ask the user: "Would you like me to suggest concrete remediation edits for the t
 - **Use examples over exhaustive rules** (cite specific instances, not generic patterns)
 - **Report zero issues gracefully** (emit success report with coverage statistics)
 
-Context for analysis: {ARGS}
+## Context
+
+{ARGS}


pierluigilenoci:
cc @localden @sigent as top contributors to these files — would appreciate your review.

templates/commands/specify.md
```diff
@@ -129,7 +129,7 @@ Given that feature description, do this:
 
    c. **Handle Validation Results**:
 
-      - **If all items pass**: Mark checklist complete and proceed to step 6
+      - **If all items pass**: Mark checklist complete and proceed to step 7
```
dhilipkumars, line 132: Good catch

templates/commands/analyze.md
```diff
@@ -182,6 +182,4 @@ Ask the user: "Would you like me to suggest concrete remediation edits for the t
 - **Use examples over exhaustive rules** (cite specific instances, not generic patterns)
 - **Report zero issues gracefully** (emit success report with coverage statistics)
 
-## Context
```
dhilipkumars, line 185: i would still leave the `## Context` to keep the hierarchy and importance when LLM interprets this

pierluigilenoci:
@dhilipkumars @mnriem I've addressed the feedback — restored the `## Context` heading in `analyze.md` as suggested. Could you please take another look? Thanks!

templates/commands/analyze.md
pierluigilenoci: Fixed

pierluigilenoci: Agreed — restored the `## Context` heading in 1f16e27. The markdown hierarchy is important for LLM parsing.

templates/commands/specify.md
pierluigilenoci: Thanks! Step reference was indeed pointing to the wrong step.

PR #1828, opened 2026-03-13
Add specify doctor command for project health diagnostics

KhawarHabibKhan:
**Extension Submission**

**Extension Name:** Project Health Check
**Extension ID:** doctor
**Version:** 1.0.0
**Author:** KhawarHabibKhan
**Repository:** https://github.com/KhawarHabibKhan/spec-kit-doctor

Closes #1821

**Description**

A diagnostic extension that scans a Spec Kit project and reports health issues across 6 areas: project structure, AI agent configuration, feature specifications, scripts, extensions, and git status.

Initially submitted as a core CLI command. Based on feedback from contributors suggesting this fits better as an extension, reverted the core approach and rebuilt as a standalone extension following the extension publishing guide.

Checklist

  -   Valid extension.yml manifest
  -   README.md with installation and usage docs
  -   LICENSE file included
  -   GitHub release created (v1.0.0)
  -   All commands working
  -   Added to extensions/catalog.community.json
  -   Added to extensions/README.md Available Extensions table
  -   Bash and PowerShell script parity

PR #1829, opened 2026-03-13
fix: auto-correct prefix conflicts in create-new-feature scripts

daniel-graham:
## Problem

Passing an explicit feature number can reuse a prefix already owned by another directory under `specs/`. Downstream prefix resolution then sees multiple matching feature directories and fails.

## Solution

Treat an explicit `--number` / `-Number` as a preferred value:

- If no spec directory uses that formatted prefix, keep the requested number.
- If a matching spec directory exists, choose the next unused prefix above `max(spec prefixes)`, skipping prefixes also owned by timestamp directories, and warn on stderr.
- Match directories only; similarly named files do not trigger correction.
- Preserve exact-directory reuse when `--allow-existing-branch` / `-AllowExistingBranch` is supplied, including canonical truncated names and directories with same-prefix siblings.

The guardrail intentionally checks local spec directories only. It does not fetch or scan git branches.

## Changes

- `scripts/bash/create-new-feature.sh`: track explicit numbers, detect spec-prefix conflicts, and update help text.
- `scripts/powershell/create-new-feature.ps1`: mirror Bash behavior with directory-only matching.
- `scripts/python/create_new_feature.py`: keep the current Python script variant behaviorally aligned, including unreadable-directory handling.
- `tests/test_timestamp_branches.py`: cover correction, 4+ digit continuation, matching files, dry-run behavior, and exact-directory reuse.
- `tests/test_create_new_feature_python_parity.py`: verify Bash, PowerShell, and Python parity, including timestamp-prefix collision skipping.

Current `main` behavior remains intact for 64-bit feature numbers, timestamp prefixes, automatic numbering, dry-run output, and allow-existing reuse.

## Validation

- `131 passed, 1 skipped` in the two affected test modules.
- `5,255 passed, 6 skipped` in the full pytest suite.
- Ruff checks passed for changed Python files.
- Bash ShellCheck passed after excluding three pre-existing warnings outside the changed lines.
- PowerShell parser validation passed.
- `git diff --check` passed.

## Agent disclosure

This revision was authored autonomously on behalf of @daniel-graham by Codex (model: GPT-5).

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -8,6 +8,7 @@ ALLOW_EXISTING=false
 SHORT_NAME=""
 BRANCH_NUMBER=""
 USE_TIMESTAMP=false
+NUMBER_EXPLICIT=false
 ARGS=()
 i=1
 while [ $i -le $# ]; do
@@ -48,6 +49,9 @@ while [ $i -le $# ]; do
                 exit 1
             fi
             BRANCH_NUMBER="$next_arg"
+            if [ -n "$BRANCH_NUMBER" ]; then
+                NUMBER_EXPLICIT=true
+            fi
             ;;
         --timestamp)
             USE_TIMESTAMP=true
@@ -60,7 +64,7 @@ while [ $i -le $# ]; do
             echo "  --dry-run           Compute feature name and paths without creating directories or files"
             echo "  --allow-existing-branch  Reuse an existing feature directory if it already exists"
             echo "  --short-name <name> Provide a custom short name (2-4 words) for the feature"
-            echo "  --number N          Specify branch number manually (overrides auto-detection)"
+            echo "  --number N          Prefer a feature number (auto-corrected if its specs prefix exists)"
             echo "  --timestamp         Use timestamp prefix (YYYYMMDD-HHMMSS) instead of sequential numbering"
             echo "  --help, -h          Show this help message"
             echo ""
@@ -91,6 +95,7 @@ if [ -z "$FEATURE_DESCRIPTION" ]; then
 fi
 
 MAX_FEATURE_NUMBER=9223372036854775807
+MAX_BRANCH_LENGTH=244
 
 is_feature_number_in_range() {
     local value="$1"
@@ -128,10 +133,38 @@ get_highest_from_specs() {
     echo "$highest"
 }
 
+# Return success when a spec directory owns the given numeric prefix.
+spec_prefix_exists() {
+    local specs_dir="$1"
+    local feature_num="$2"
+
+    for spec_path in "$specs_dir/${feature_num}-"*; do
+        [ -d "$spec_path" ] && return 0
+    done
+    return 1
+}
+
 # Function to clean and format a branch name
 clean_branch_name() {
     local name="$1"
     echo "$name" | tr '[:upper:]' '[:lower:]' | sed 's/[^a-z0-9]/-/g' | sed 's/-\+/-/g' | sed 's/^-//' | sed 's/-$//'
+}
+
+# Fit a feature prefix and suffix within GitHub's branch-name limit.
+fit_branch_name() {
+    local feature_num="$1"
+    local branch_suffix="$2"
+    local branch_name="${feature_num}-${branch_suffix}"
+
+    if [ ${#branch_name} -gt $MAX_BRANCH_LENGTH ]; then
+        local prefix_length=$(( ${#feature_num} + 1 ))
+        local max_suffix_length=$((MAX_BRANCH_LENGTH - prefix_length))
+        local truncated_suffix
+        truncated_suffix=$(printf '%s' "$branch_suffix" | cut -c "1-$max_suffix_length" | sed 's/-$//')
+        branch_name="${feature_num}-${truncated_suffix}"
+    fi
+
+    printf '%s' "$branch_name"
 }
 
 # Quote a value for POSIX shell reuse, byte-identical to Python's shlex.quote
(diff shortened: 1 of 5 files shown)


mnriem (changes requested):
Thanks for tackling this — the problem is real and the guardrail idea is the right direction. Before we can merge, there are a few things we need to address: the --number flag's contract changes from a hard override to a suggestion, which is a silent breaking change for any automation relying on deterministic output; the recalculation path skips git fetch so the corrected number could still collide with unfetched remote branches; the check only covers spec directories but not existing branches, leaving an inconsistent experience; the help text still says "overrides auto-detection"; and the PowerShell script doesn't get the same guardrail, breaking cross-platform parity. Could you scope the guardrail to only fire when --number was explicitly passed, add a fetch before recalculating, extend the collision check to branches, update the help text, and port it to the PowerShell side as well?

daniel-graham:
## Changes made in response to review

Thanks for the thorough review @mnriem! All feedback addressed across 3 follow-up commits:

### @mnriem review (requested changes)

| Feedback | Resolution |
|----------|------------|
| `--number` contract changes silently from override to suggestion | Guardrail now **only fires when `--number` was explicitly passed** (`NUMBER_EXPLICIT` flag in Bash, `$PSBoundParameters.ContainsKey` in PowerShell) |
| Recalculation skips `git fetch` | Guardrail now calls `git fetch --all --prune` before checking branches, and delegates recalculation to `check_existing_branches` / `Get-NextBranchNumber` which also fetch |
| Collision check only covers spec dirs, not branches | Trigger now checks **both** `specs/NNN-*` dirs and `git branch -a` for matching prefixes |
| Help text still says "overrides auto-detection" | Updated to "Preferred branch number (auto-corrected if prefix already exists in specs or branches)" |
| PowerShell missing parity | Full guardrail ported to `create-new-feature.ps1` with identical logic |

### @copilot review (suggestions)

| Feedback | Resolution |
|----------|------------|
| PR title only mentions `.sh` | Updated to reference both scripts |
| Use `$PSBoundParameters.ContainsKey` instead of sentinel | Adopted — idiomatic PowerShell explicit-parameter detection |
| `git branch -a` without fetch (PS) | Added `git fetch --all --prune` before branch check |
| `git branch -a` without fetch (Bash) | Same fix applied |

### Also fixed
- Removed unrelated `templates/commands/specify.md` changes that had leaked into the PR

All changes validated on macOS with both Bash and PowerShell (`pwsh`):
- Explicit `--number` with conflict → warns + auto-corrects
- No `--number` (auto-detect) → guardrail does not fire

mnriem (changes requested):
Please pull in latest from main and address Copilot feedback. If not applicable, please explain why.

daniel-graham:
## Follow-up: rebased on main + addressed remaining feedback

### What changed

1. **Rebased on latest `main`** — picks up PR #1828 ("Add specify doctor command")
2. **Fixed `compgen` to only match directories** — previously `compgen -G "specs/NNN-*"` would also match files (e.g., `specs/001-notes.md`), causing false positives. Now iterates matches and only triggers on `-d` (directories)
3. **Confirmed `templates/commands/specify.md` is clean** — no diff vs upstream/main. The bot's earlier comment was from a stale review before the revert was pushed

### Final PR diff

Only 2 files changed (82 additions, 2 deletions):
- `scripts/bash/create-new-feature.sh`
- `scripts/powershell/create-new-feature.ps1`

No other files are modified.

mnriem (changes requested):
Please address the additional Copilot comments. Thank for staying with it!

daniel-graham:
## Latest update (rebased + squashed to 1 commit)

Addressed the remaining Copilot feedback and cleaned up from internal review:

1. **Bash: `--number` range validation** — now rejects values outside `[1,999]` with a clear error, since downstream tooling expects a 3-digit prefix
2. **PowerShell: `[ValidateRange(1, 999)]`** — added to the `-Number` parameter for parity with Bash
3. **Double-fetch eliminated** — Bash inlines the max-finding using already-fetched local branch data; PowerShell defers the single fetch to `Get-NextBranchNumber` (only called when a conflict is found)
4. **Stale comments cleaned up** — removed references to `check_existing_branches` in the inlined code path

### Note for @mnriem on range validation

We added `[ValidateRange(1, 999)]` to the PowerShell `-Number` parameter (and equivalent validation in Bash) to prevent malformed prefixes from values outside the 3-digit range. This means `-Number 0` now errors instead of silently triggering auto-detection — users should simply omit `-Number` for auto-detect behavior.

If there's an intentional reason to allow values outside `[1, 999]` (e.g., future support for 4-digit prefixes), let us know and we'll adjust the bounds accordingly.

### Final diff stat

```
 scripts/bash/create-new-feature.sh    | 62 ++++++++++-
 scripts/powershell/create-new-feature.ps1 | 44 +++++++-
 2 files changed, 104 insertions(+), 2 deletions(-)
```

mnriem (changes requested):
I would prefer the range to allow for 5 digits so it takes a longer time to run out. Now that we are using AI we probably will go through number more quickly so 3 digits seems a bit small. And please address Copilot feedback. if not applicable please explain why.

daniel-graham:
> I would prefer the range to allow for 5 digits so it takes a longer time to run out. Now that we are using AI we probably will go through number more quickly so 3 digits seems a bit small. And please address Copilot feedback. if not applicable please explain why.

Agreed, I imagine we may want to either add some padding 00NNN to existing directories, or leave that up to the user, and have the script simply progress naturally to 5 digits from 999 -> 01000?

mnriem:
Probably should keep the prepending for the current scheme and once it overflows no more prepending as that will keep it backwards compatible

daniel-graham:
Rebased this PR onto current `main` and rebuilt the guardrail around current behavior.

The revised implementation checks spec directories only and contains no git fetch or branch-scanning path, so the earlier fetch-order and stale-remote comments no longer apply. Current core auto-numbering also derives from spec directories. The earlier 3-digit/5-digit range discussion is superseded by current `main`, which supports prefixes through signed 64-bit maximum values and formats them with a minimum width of three digits.

Bash, PowerShell, and the current Python variant now share these contracts:

- explicit number stays unchanged when its spec prefix is unused;
- existing spec-directory prefix advances to `max(spec prefixes) + 1`;
- matching files are ignored;
- dry-run computes the corrected result without writes;
- exact-directory reuse remains available through the allow-existing option;
- timestamp and automatic-number paths remain unchanged.

Validation on commit `05bf2bff971ed1df074e05d9eb2eb197532e0b9f`:

- affected modules: 127 passed, 1 skipped;
- full suite: 5,251 passed, 6 skipped;
- Ruff, Bash ShellCheck for changed behavior, PowerShell parser, and `git diff --check` passed;
- independent final diff review found no actionable issues.

Fresh GitHub workflow runs were created but currently report `action_required` before executing jobs, consistent with the approval gate for fork-origin pull requests rather than a test failure.

Posted on behalf of @daniel-graham by Codex (model: GPT-5, autonomous).

daniel-graham:
Local review loop complete for `12756f5`.

Fixed five cross-implementation defects:

- Canonical branch-name truncation now happens before exact allow-existing comparison.
- An exact allow-existing target wins even when a sibling directory shares its numeric prefix.
- Python matches Bash and PowerShell when the specs directory cannot be listed.
- Corrected explicit numbers advance past prefixes already owned by timestamp directories, addressing Copilot review `4763816309`.
- Empty `--number` remains omitted in Bash, matching PowerShell and Python even when automatic numbering lands on a timestamp-owned prefix, addressing Copilot review `4764870736`.

Validation:

- Affected modules: `131 passed, 1 skipped`.
- Full suite: `5,255 passed, 6 skipped`.
- Ruff, Python compilation, Bash syntax and ShellCheck, PowerShell parser, and `git diff --check`: passed.
- Independent post-fix review and direct Bash/PowerShell/Python probes: clean.

The four workflow runs are awaiting maintainer approval and have not executed the code.

Posted on behalf of @daniel-graham by Codex (model: GPT-5, autonomous).

mnriem:

mnriem:

mnriem (approved):

mnriem:
Thank you!

PR #1836, opened 2026-03-13
fix: update-agent-context timestamp regex to match template output

### LINKED ISSUE #1823: [Bug] PowerShell update-agent-context script does not refresh "Last updated" in template-generated agent files
## Summary

The PowerShell `update-agent-context.ps1` script appears intended to refresh the `Last updated` date in generated agent context files, but it does not update files created from the current `agent-file-template.md`.

## Reproduction

1. Create or initialize an agent file from `.specify/templates/agent-file-template.md`.
2. Observe that the template writes the date line in this format:

```md
Auto-generated from all feature plans. Last updated: 2026-03-11
```

3. Run the PowerShell updater:

```powershell
.specify/scripts/powershell/update-agent-context.ps1 -AgentType cursor-agent
```

4. Observe that other sections such as `Active Technologies` and `Recent Changes` are updated, but the `Last updated` line remains unchanged.

## Expected behavior

The updater should refresh the date in existing agent files when they are updated.

## Actual behavior

The date stays stale for template-generated files.

## Root cause

There appears to be a format mismatch between the template and the update logic:

- Template emits:

```md
Auto-generated from all feature plans. Last updated: [DATE]
```

- PowerShell updater only rewrites lines matching this pattern:

```powershell
if ($line -match '\*\*Last updated\*\*: .*\d{4}-\d{2}-\d{2}') {
    $output.Add(($line -replace '\d{4}-\d{2}-\d{2}',$Date.ToString('yyyy-MM-dd')))
}
```

Because the existing file line does not use the `**Last updated**:` format, it never matches and is never updated.

## Suggested fix

Either:

1. Update the regex in `update-agent-context.ps1` so it also matches the template-generated line format, or
2. Change the template to emit the same format the updater expects.

Option 1 seems safer because it preserves the current generated file style and updates existing files correctly.

## Impact

This creates misleading metadata in generated agent context files: users can see that the body content has been refreshed while the `Last updated` header remains stale.

### LINKED ISSUE #1834: update-agent-context scripts fail to refresh 'Last updated' date from template
Hi,

I ran into an issue where the `update-agent-context` scripts (both bash and PowerShell) don't refresh the "Last updated" timestamp in agent files that were generated from the template.

The template outputs plain text:
```
Last updated: 2025-03-01
```

But both scripts look for bold Markdown markers:
```
**Last updated**: 2025-03-01
```

So the regex never matches and the date is never refreshed on subsequent runs.

This seems to be the same issue as #1823.

**Files:**
- `scripts/bash/update-agent-context.sh` (line ~479)
- `scripts/powershell/update-agent-context.ps1` (line ~334)

mango766:
Updates the "Last updated" timestamp regex in both `update-agent-context.sh` and `update-agent-context.ps1` to match the plain text format used by the template, in addition to the bold Markdown format.

The template outputs `Last updated: [DATE]` (plain text) but the scripts were only looking for `**Last updated**: [DATE]` (bold Markdown), so the timestamp was never refreshed on subsequent runs.

Fixes #1823
Closes #1834

### DIFF
--- a/scripts/bash/update-agent-context.sh
+++ b/scripts/bash/update-agent-context.sh
@@ -476,7 +476,7 @@ update_existing_agent_file() {
         fi
         
         # Update timestamp
-        if [[ "$line" =~ \*\*Last\ updated\*\*:.*[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9] ]]; then
+        if [[ "$line" =~ (\*\*)?Last\ updated(\*\*)?:.*[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9] ]]; then
             echo "$line" | sed "s/[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]/$current_date/" >> "$temp_file"
         else
             echo "$line" >> "$temp_file"
--- a/scripts/powershell/update-agent-context.ps1
+++ b/scripts/powershell/update-agent-context.ps1
@@ -331,7 +331,7 @@ function Update-ExistingAgentFile {
             if ($existingChanges -lt 2) { $output.Add($line); $existingChanges++ }
             continue
         }
-        if ($line -match '\*\*Last updated\*\*: .*\d{4}-\d{2}-\d{2}') {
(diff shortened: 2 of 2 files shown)


mnriem:
@mango766 Thank you!

PR #1949, opened 2026-03-24
Fix/feedback bito

ivo-toby:
## Testing

## AI Disclosure

- [ ] I **did** use AI assistance (describe below)

### DIFF
--- a/hooks/scripts/claude-block-force.sh
+++ b/hooks/scripts/claude-block-force.sh
@@ -30,9 +30,9 @@ DANGEROUS_PATTERNS=(
 )
 
 # Check if command matches any dangerous pattern at a command boundary
-# Matches at start of string or after command separators (&&, ||, ;)
+# Matches at start of string or after command separators (&&, ||, ;, |)
 for pattern in "${DANGEROUS_PATTERNS[@]}"; do
-    if [[ "$COMMAND" =~ (^|&&|\|\||;)[[:space:]]*$pattern ]]; then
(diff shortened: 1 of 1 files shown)


ivo-toby:
Wrong repo.. Nothing to see here

mnriem:
@ivo-toby What is minispec?

ivo-toby:
@mnriem I forked SpecKit and created a version that addresses the issue of engineers losing the mental models of their codebases. I love SpecKit for personal projects, but for (work) production code, I don't feel confident enough just to have an agent write everything, with me ending up reviewing hundreds of lines of code. After 150 lines, you're skimming; after 250, you're rubberstamping. I need to be engaged with my codebase. So that's what MiniSpec tries to do. And for me this works quite well. I keep having a decent understanding of what I push to production, and it's way more fun compared to "only" reviewing code.

mnriem:
@ivo-toby Just curious to see if the recent additions to Spec Kit as a Kit aka extensions, presets and switching agents would have not required you to fork? E.g if the workflow is too heavy your own extension can define its own workflow without even using the core commands. Presets can overlay domain specific or organization specific requirements

Thoughts?

ivo-toby:
@mnriem looks like something I could (should) have used! Extensions will also open up a lot of possibilities for internal tooling tied to specs. Very nice. Will certainly look into moving the ideology of MiniSpec into an extension

mnriem:
My personal favorite and yes I created it is the Pirate Speak preset. It is listed as one of our community walkthroughs at https://github.com/github/spec-kit?tab=readme-ov-file#-community-walkthroughs Love to read the Ship's manifest!

PR #1969, opened 2026-03-25
fix(scripts): honor PowerShell agent and script filters

### LINKED ISSUE #1946: [Bug]: PowerShell offline scaffolding ignores -Agents / -Scripts, causing --offline to fail on Windows
### Bug Description

--offline scaffolding fails on Windows because the PowerShell release script drops the -Agents and -Scripts arguments internally, so no bundled package is produced.

This shows up as:

Error: --offline was specified but scaffolding from bundled assets failed.
Detail: release script produced no output
The root cause is in .github/workflows/scripts/create-release-packages.ps1: the helper function Normalize-List uses param([string]$Input), which conflicts with PowerShell’s automatic $input variable. Because PowerShell variable names are case-insensitive, the passed value is effectively lost and the filtered build lists become empty.

### Steps to Reproduce

On Windows with pwsh installed, run:
specify init my-project --ai opencode --script ps --offline
Observe the offline init failure:
Error: --offline was specified but scaffolding from bundled assets failed.
Detail: release script produced no output
To reproduce the lower-level script bug directly, run:
pwsh -File .github/workflows/scripts/create-release-packages.ps1 -Version v0.0.0 -Agents opencode -Scripts ps
Observe that the script prints:
Building release packages for v0.0.0
Agents:
Scripts:

Archives in .genreleases:
No spec-kit-template-opencode-ps-v0.0.0.zip is created.

### Expected Behavior

--offline should scaffold successfully on Windows when pwsh is installed and the selected agent/script pair is supported.

The PowerShell release script should honor:

-Agents opencode -Scripts ps
and generate the expected offline package output.

### Actual Behavior

Offline scaffolding fails with:

Error: --offline was specified but scaffolding from bundled assets failed.
Detail: release script produced no output
The underlying PowerShell script exits successfully but ignores the filtered values, prints empty Agents: and Scripts: lines, and produces no output package.

### Specify CLI Version

0.4.0

### AI Agent

opencode

### Operating System

Windows 11

### Python Version

3.12.11

### Error Logs

```shell
Error: --offline was specified but scaffolding from bundled assets failed.
Common causes: missing bash/pwsh, script permission errors, or incomplete wheel.
Remove --offline to attempt a GitHub download instead.
Detail: release script produced no output
```

### Additional Context

Root cause appears to be a PowerShell script bug in .github/workflows/scripts/create-release-packages.ps1.

Problematic code:

function Normalize-List {
    param([string]$Input)

    if ([string]::IsNullOrEmpty($Input)) {
        return @()
    }

    $items = $Input -split '[,\s]+' | Where-Object { $_ } | Select-Object -Unique
    return $items
}
Because $Input conflicts with PowerShell’s automatic $input variable, the function can return an empty result even when called with values like opencode or ps.

Suggested fix:

Rename the function parameter from $Input to $Value or similar
Update call sites from Normalize-List -Input ... to Normalize-List -Value ...
Suggested regression test:

Run create-release-packages.ps1 with -Agents opencode -Scripts ps
Assert stdout contains:
Agents: opencode
Scripts: ps
Assert spec-kit-template-opencode-ps-v0.0.0.zip is created

chenry-me:
Fixes #1946 

## Description

Fixes a Windows offline scaffolding bug in `create-release-packages.ps1`.

The PowerShell helper `Normalize-List` used `$Input` as its parameter name, which conflicts with PowerShell's automatic `$input` variable. This caused `-Agents` and `-Scripts` filtering to be ignored, so `specify init --offline` could fail with `release script produced no output`.

This change renames the parameter and updates the call sites so the selected agent/script filters are honored correctly.

## Testing

Manual verification:
- Reproduced the failure with `specify init ... --offline --ai opencode --script ps` before the fix
- Verified that offline scaffolding succeeds after the fix

Note:
- I did not mark the full pytest suite as passing in my Windows environment because there are unrelated bash/path/tooling failures, including missing `zip` for the bash release-packaging path and tests that pass Windows absolute paths directly to bash.

## AI Disclosure

I used OpenCode (GPT-5.4) to help analyze the code, identify the root cause, and draft parts of the fix description (to help me communicate more clearly in English). I manually reproduced the bug, reviewed the relevant script behavior, and verified the final fix locally.

### DIFF
--- a/.github/workflows/scripts/create-release-packages.ps1
+++ b/.github/workflows/scripts/create-release-packages.ps1
@@ -498,13 +498,13 @@ $AllAgents = @('claude', 'gemini', 'copilot', 'cursor-agent', 'qwen', 'opencode'
 $AllScripts = @('sh', 'ps')
 
 function Normalize-List {
-    param([string]$Input)
+    param([string]$Value)
 
-    if ([string]::IsNullOrEmpty($Input)) {
+    if ([string]::IsNullOrEmpty($Value)) {
         return @()
     }
 
-    $items = $Input -split '[,\s]+' | Where-Object { $_ } | Select-Object -Unique
+    $items = $Value -split '[,\s]+' | Where-Object { $_ } | Select-Object -Unique
     return $items
 }
 
@@ -527,7 +527,7 @@ function Validate-Subset {
 
 # Determine agent list
 if (-not [string]::IsNullOrEmpty($Agents)) {
-    $AgentList = Normalize-List -Input $Agents
+    $AgentList = Normalize-List -Value $Agents
     if (-not (Validate-Subset -Type 'agent' -Allowed $AllAgents -Items $AgentList)) {
         exit 1
     }
@@ -537,7 +537,7 @@ if (-not [string]::IsNullOrEmpty($Agents)) {
 
 # Determine script list
 if (-not [string]::IsNullOrEmpty($Scripts)) {
-    $ScriptList = Normalize-List -Input $Scripts
+    $ScriptList = Normalize-List -Value $Scripts
     if (-not (Validate-Subset -Type 'script' -Allowed $AllScripts -Items $ScriptList)) {
         exit 1
     }


mnriem:
Thank you!

PR #1994, opened 2026-03-27
fix: prevent extension command shadowing

### LINKED ISSUE #1963: Extension commands can shadow core spec-kit commands
## Problem

An extension can register commands (or aliases) whose output file names collide with core spec-kit commands (`analyze`, `clarify`, `checklist`, `constitution`, `implement`, `plan`, `specify`, `tasks`, `taskstoissues`). When this happens, the extension command **overwrites** the core command file (e.g. `.claude/commands/speckit.analyze.md`), effectively hijacking it.

### How it happens

1. **Via aliases** — command aliases are not validated at all. An extension could declare an alias like `speckit.analyze`, which would produce the file `speckit.analyze.md` — directly overwriting the core `analyze` command.
2. **Via extension ID** — while extension command names must follow `speckit.{ext}.{cmd}` (3 parts), there's no check preventing an extension from using an ID that shadows a core command namespace. An extension with ID `analyze` would register commands under `speckit.analyze.*`, creating potential confusion.
3. **Cross-extension conflicts** — there's also no check for command name collisions between two different installed extensions.

### Relevant code

- **Manifest validation** (`src/specify_cli/extensions.py:150-160`): validates command name pattern `speckit.{ext}.{cmd}` but does not check against core command names or existing extensions
- **Alias registration** (`src/specify_cli/agents.py:479-491`): aliases are registered without any name format validation
- **Output file naming** (`src/specify_cli/agents.py:395-404`): `_compute_output_name` derives the file name directly from the command name — no conflict detection

## Proposed solution

Add validation during `extension add` (in `install_from_directory` / `install_from_zip`) that:

1. **Rejects command names and aliases that conflict with core commands** — maintain a list of core command names and check against it during manifest validation
2. **Rejects command names that conflict with already-installed extensions** — check existing registered commands in the extension registry
3. **Validates alias format** — aliases should follow the same `speckit.{ext}.{cmd}` pattern as command names

### UX options on conflict

- **Fail with error** (recommended for security) — clearly explain which names conflict and suggest renaming
- **Warn and prompt** — ask the user if they want to proceed (with `--force` flag to bypass)

## Impact

- **Security**: a malicious extension could silently replace core spec-kit commands with arbitrary instructions
- **Reliability**: users may not realize a core command has been overwritten until they get unexpected behavior

afurm:
## Description

Fixes #1963.

This PR hardens extension installation so extensions cannot silently shadow core Spec Kit commands or collide with commands from other installed extensions.

Specifically, it:
- adds install-time validation for extension command and alias names
- rejects extension IDs and command namespaces that overlap with core `speckit.*` commands
- rejects malformed short aliases that do not follow `speckit.{extension}.{command}`
- rejects command/alias collisions with already-installed extensions
- updates extension tests and shipped extension docs/examples to use namespaced aliases

This is needed to close a security and reliability gap where an extension could overwrite core command files or another extension's command files during installation.

## Testing

- Ran `uv run specify --help` successfully and confirmed `extension` commands are present
- Ran the full test suite successfully with `uv sync --extra test && uv run pytest` (`876 passed`)
- Also ran `uv run --extra test pytest tests/test_extensions.py` while developing the fix

- [x] Ran existing tests with `uv sync --extra test && uv run pytest`
- [ ] Tested with a sample project (not applicable for this change)

## AI Disclosure

Used Codex to inspect the issue, implement the validation changes, update tests/docs, and run verification commands.

### DIFF
--- a/src/specify_cli/extensions.py
+++ b/src/specify_cli/extensions.py
@@ -192,7 +192,7 @@ class ExtensionManifest:
                 raise ValidationError("Command missing 'name' or 'file'")
 
             # Validate command name format
-            if not re.match(r'^speckit\.[a-z0-9-]+\.[a-z0-9-]+$', cmd["name"]):
+            if EXTENSION_COMMAND_NAME_PATTERN.match(cmd["name"]) is None:
                 raise ValidationError(
                     f"Invalid command name '{cmd['name']}': "
                     "must follow pattern 'speckit.{extension}.{command}'"


afurm:
> Please address Copilot feedback. If not applicable, please explain why

Addressed in `3571f75`.

I fixed all 3 Copilot items:
- Enforced `extension.id` namespace ownership.
- Replaced the hard-coded core command list with discovery from bundled templates, plus a parity test.
- Updated the docs/RFC wording to match the implemented validation and migration behavior.

Validation:
- `uv run --with '.[test]' pytest tests/test_extensions.py -q`
- `148 passed`

mnriem:
@afurm This does not prevent presets from overriding commands?

afurm:
> @afurm This does not prevent presets from overriding commands?

Correct. This change does not prevent presets from overriding commands.

The new validation only runs in the **extension install path** (`ExtensionManager.install_from_directory()` / `install_from_zip()` in `src/specify_cli/extensions.py`). Presets still use the separate preset registration flow in `src/specify_cli/presets.py`, where command overrides are applied by priority and restored on removal.

So the intent here is narrower: prevent **extensions** from claiming core-command namespaces or another extension’s namespace. It does not change the existing preset override mechanism.

mnriem:
Thank you!

mbachorik:
Thank you!!!

PR #1929, opened 2026-03-21
feat(cli): add `specify extension init` scaffolding command

mvanhorn:
## Description

Add a `specify extension init <name>` subcommand that scaffolds a new extension directory from the built-in template (`extensions/template/`), substituting user-provided metadata into all placeholder values.

The extension ecosystem has 23+ community extensions published to the catalog, each manually created by copying `extensions/template/` and hand-editing 7 files. Every other stage of the extension lifecycle has CLI support (search, add, remove, enable, disable, update, set-priority, catalog management) - except creation. This closes that gap.

**What it does:**
- Copies `extensions/template/` to `spec-kit-<name>/` in the current or specified directory
- Substitutes placeholders: extension ID, display name, author, description, repository URL, date
- Validates extension ID format (lowercase, hyphen-separated, `^[a-z][a-z0-9]*(-[a-z0-9]+)*$`)
- Replaces `EXAMPLE-README.md` as the new `README.md`
- Optionally initializes a git repo (`--no-git` to skip)
- Prompts interactively when CLI args are omitted
- Prints a next-steps panel matching the style of `specify init`

**Usage:**
```bash
specify extension init my-linter
specify extension init my-linter --author "Jane Doe" --description "Lint spec files"
specify extension init my-linter --output ~/projects --no-git
```

This command does NOT require a spec-kit project - you can scaffold an extension anywhere, then test it with `specify extension add --dev /path/to/extension`.

The template source is resolved from the installed package first (works in wheel/air-gapped), falling back to the source tree layout.

## Testing

- [x] Ran existing tests with `pytest` - all 810 existing tests pass
- [x] Added 23 new test cases in `tests/test_extension_init.py`:
  - Extension ID validation (10 tests): valid names, uppercase rejection, special chars, leading numbers, empty strings
  - Title case conversion (3 tests)
  - Template discovery (2 tests)
  - CLI integration (8 tests): full scaffold, placeholder substitution, invalid name rejection, existing directory rejection, --no-git, README replacement, next-steps output, date substitution
- [x] Tested locally: `specify extension init my-test --no-git` produces a working extension directory with all placeholders replaced

## AI Disclosure

- [x] I **did** use AI assistance (describe below)

This contribution was developed with AI assistance (Claude Code). The implementation follows existing patterns from the `specify init` command and `extension add` subcommand.

PR #2000, opened 2026-03-27
feat(extensions): scaffold config templates on extension add/enable

mvanhorn:
## Description

Rework of #1929 per @mnriem's feedback. Instead of a standalone `specify extension init` command, config scaffolding now runs automatically as part of the extension lifecycle:

- `extension add`: after installing an extension, config templates from `provides.config` are deployed to `.specify/`
- `extension enable`: when re-enabling a disabled extension, missing config templates are deployed

Existing user-customized config files are never overwritten. Extensions without a `provides.config` section are unaffected.

Changes:
- `ExtensionManifest.config` property reads `provides.config` from the manifest
- `ExtensionManager.scaffold_config()` copies config templates to the project, skipping existing files
- `extension_add` replaces the "Configuration may be required" warning with automatic deployment
- `extension_enable` deploys config on re-enable

## Video Demo

Config scaffolding demo

The demo shows: config template exists in extension dir, no project config yet, `scaffold_config` deploys it, re-running preserves existing config (idempotent).

## Testing

- [x] Ran existing tests with `uv sync --extra test && uv run python -m pytest` - all 890 tests pass
- [x] Added 4 new tests covering scaffold_config: deploy, preserve existing, no config section, missing template
- [x] Tested with a sample project (see demo above)

## AI Disclosure

This contribution was developed with AI assistance (Claude Code + Codex CLI).

Closes #1929


mnriem:
On a side note you moved us into the two thousands! ;)

mvanhorn:
Ha, PR #2000\! Didn't even notice. Lucky number.

mnriem (changes requested):
Please address Copilot feedback. If not applicable, please explain why.

mvanhorn:
Addressed all Copilot feedback in 86606ef:

1. **Path traversal protection** - `scaffold_config()` now rejects template/target names with `..` segments or absolute paths
2. **File type check** - uses `is_file()` instead of `exists()` to reject directory templates
3. **Manifest validation** - `config` property normalizes malformed entries (non-list or non-dict values return `[]`)
4. **Richer messaging** - returns `(deployed, skipped)` tuple so the CLI distinguishes "already exists" from "template missing"
5. **Security tests** - added test coverage for path traversal rejection, directory template rejection, and malformed manifest handling

mvanhorn:
Rebased and addressed the review. Since #3014 split `extensions.py` into the `extensions/` package while this sat, this is a port rather than a rebase: the manager and manifest changes now land in `extensions/__init__.py` and the CLI wiring in `extensions/_commands.py`. Same behaviour, new home.

On the eight comments:

- **Consistent return** -- `scaffold_config()` returns `(deployed, skipped_existing, failed)` on every path, including the missing-manifest case, which previously returned a bare `[]` and would have raised on unpack.
- **Misleading message** -- the new `failed` list separates "already exists (preserved)" from "template missing/invalid", and both `extension_add` and `extension_enable` report them distinctly.
- **Path containment** -- template paths must resolve inside the extension dir and targets inside `.specify/`, checked with `resolve()` + `relative_to()` rather than a `..` substring test.
- **File type** -- symlinks and non-regular files are rejected before any copy.
- **Manifest shape** -- `ExtensionManifest.config` returns `[]` unless `provides.config` is a list of dicts, and `scaffold_config()` surfaces a malformed section as a failure instead of iterating it.
- **enable crash** -- `extension_enable` catches manifest errors and warns instead of tracebacking.
- **Tests** -- the malformed-manifest test now uses a valid manifest with a non-list `provides.config`, and there are new cases for symlink templates, path traversal, absolute paths, directory templates, and the missing-manifest tuple shape.

`pytest` is 5282 passed / 172 skipped on the full suite.

mnriem:

mvanhorn:
1db0301 addresses these, and rebases onto main since the branch had drifted 75 commits.

The config location was the real one. `ConfigManager._get_project_config()` reads `.specify/extensions/<id>/<id>-config.yml`, so writing to the `.specify/` root put the file somewhere the extension never looks. Deployment moved there, and two existing tests that asserted the old path are updated rather than worked around.

Containment is now checked component by component before `.specify` becomes the root, matching the shared_infra safe-write path. Resolving it first and trusting the result meant a symlinked component let every target pass `relative_to` and `copy2` write outside the project.

`mkdir` moved inside the `OSError` handler, so a nested target whose parent cannot be created is recorded as failed instead of raising after `extension add` already installed the extension.

The "Configuration may be required" warning is conditional now. It ran unconditionally right after the scaffolding output, so it contradicted a successful scaffold and fired for extensions with no `provides.config` at all.

New tests cover the corrected location, a symlinked config root, and an uncreatable nested target. Suite is 6019 passing.

The earlier comments on `src/specify_cli/extensions.py` predate the split into the package; the return-type, path-traversal, symlink-template and manifest-shape points they raised are all handled in the current `extensions/__init__.py`.

mnriem:

mvanhorn:
Addressed the outstanding Copilot finding.

You were right that scaffolding accepted targets removal wouldn't preserve. `remove(keep_config=True)` rmtree's every subdirectory and keeps only top-level `-config.yml` / `-config.local.yml`, and the backup path globs that same pattern, so a nested target was silently deleted by `extension add --force` and replaced with the template default.

I took the first of the two options and constrained scaffold targets to the preserved convention, rather than widening remove, backup, rescue and restore to walk manifest targets recursively. Narrowing one entry point is a smaller surface than teaching four removal paths a new contract, and it fails loudly at scaffold time instead of losing data at update time.

The `nested/test-config.yml` test was only using nesting to reach the mkdir handler, and since `config_dir` is the extension directory itself that path is unreachable for top-level targets, so I replaced it with tests for the constraint and for the `.local.yml` form.

443 tests in `tests/test_extensions.py` pass.

For what it's worth, the other findings in this thread were already fixed in 1db0301 on the 29th: the 3-tuple return on the missing-manifest path, deploying to `.specify/extensions/<id>/` where the config contract actually reads, the symlink containment guard, and moving mkdir inside the `OSError` handler.

mnriem (approved):

mnriem:
Thank you!

mvanhorn:
Thanks @mnriem for merging the extension config template scaffolding. Happy it rounds out the extension flow.

Issue #318, opened 2025-09-17
documentation: CLI reference is misleading


stevenlafl:
Didn't know how to categorize this, however I am coming into it not having looked at any YouTube videos (though I know they exist) based on recommendation from Reddit. **Please understand this is more about the nature of the documentation itself - not to be construed as my being unable to figure it out (I demonstrate the obvious later) nor that I am currently having issues.**

## Problem (User Journey)

I am a new technical user who prefers to install CLI tools first.

In the Getting Started:

```
uvx --from git+https://github.com/github/spec-kit.git specify init <PROJECT_NAME>
```

This I understand. It's completely unambiguous. I skip the section as I'm not ready to begin a new project.

Under **CLI Reference** section it is used like a standalone command-line tool. How do I install that? I go above and examine the Getting Started section.

Now I see the problem. I cannot literally type `specify` as stated without also using uvx, strictly following the instructions. The instructions seem to be more geared toward those using a permanent CLI installation, though no section clearly states how to install it that way.

## Suggestion

I see that there is a GIF in the documentation where it shows `uv run specify` but that actual command text is not shown nor used in the documentation. This is not screen-reader friendly and not actually stated in the text.

I believe we should add something like:

```
Install the CLI with:
   uv tool install --from git+https://github.com/github/spec-kit.git specify-cli

Upgrade the CLI with:
  uv tool upgrade specify-cli
```

Or at the very least properly state the `uv run specify` bit present in the GIF prior to the command listing to users know how to execute them without using the longform `uvx --from ...`

This would put the instructions more in line with the majority of other existing projects.

foertel:
> Under CLI Reference section it is used like a standalone command-line tool. How do I install that?

Just use the `uvx --from` syntax from the Getting Started Section for every call. It could be a little bit clearer in the docs, but I think it's fine. An installable version will follow later, for now only the `uvx` syntax is supported.

But atm you only run this command once for init, so ... 🤷

stevenlafl:
> But atm you only run this command once for init, so ... 🤷 

Yeah. Looking at only the docs on the homepage it's really unclear how you even get an agent to use the workflow. How these "slash commands" are even possible. I generally don't like to run things without understanding it, and normally these kinds of workflows/commands are built into CLI tools like Claude Code itself nowhere near modifiable by the user.

[...]

PR #2001, opened 2026-03-27
Clarify CLI reference usage in README

ArthurFay:
## Summary
This pull request clarifies the CLI reference in the README by explaining that the examples assume the persistent installation method was used, which makes the `specify` command available directly in PATH.

It also adds a note above the examples section explaining that users who did not install the CLI persistently should use the `uvx --from ... specify ...` form instead.

## Issue
Addresses confusion described in issue #318.

## Files Changed
- README.md

## Impact
These documentation updates help reduce confusion for new users and make the CLI usage instructions easier to follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ Community projects that extend, visualize, or build on Spec Kit:
 ## 🔧 Specify CLI Reference
 
 The `specify` command supports the following options:
-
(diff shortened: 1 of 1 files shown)


mnriem (changes requested):

PR #1757, opened 2026-03-04
fix: use global branch numbering instead of per-short-name detection

fsilvaortiz:
## Summary
- The `specify.md` prompt instructed the AI to search for existing branches filtered by exact short-name, so every new feature started at `001` since no branches matched the new short-name
- The underlying `create-new-feature.sh` already has correct global numbering logic via `check_existing_branches()` that searches ALL branches and spec directories
- Simplified step 2 to tell the AI to NOT pass `--number`, letting the script auto-detect the next globally available number

Closes #1744
Closes #1468

## Root cause
The prompt at step 2b told the AI to grep for branches matching the exact short-name:
```
git ls-remote --heads origin | grep -E 'refs/heads/[0-9]+-<short-name>$'
```
And step 2c said: "If no existing branches/directories found **with this short-name**, start with number 1". Since each new feature has a unique short-name, the search always returned empty and numbering reset to 001.

## Test plan
- [x] Verified the script correctly assigns `001`, `002`, `003` to 3 different features when `--number` is NOT passed
- [x] Reproduced the bug: passing `--number 1` manually causes duplicate `001` prefixes (confirming the AI override was the problem)
- [x] Full test suite passes (120 tests)
- [x] Verified the change is a net removal of 24 lines of flawed logic, replaced with 4 lines of clear instructions

🤖 Generated with [Claude Code](https://claude.com/code)

Issue #1744, opened 2026-03-03
[Bug]: All new specs start with 001


RunOrVeith:
### Bug Description

Every time I start building a new feature, the resulting branch starts with the prefix 001. It does not increment.
There are errors from subsequent scripts complaining about duplicate prefixes.
See my current branches: 

<img width="549" height="44" alt="Image" src="https://github.com/user-attachments/assets/374a8145-965e-4e30-96ed-55086b855721" />

### Steps to Reproduce

1. run `/speckit.specify` 
2. a new branch with prefix 001 is created

<img width="331" height="79" alt="Image" src="https://github.com/user-attachments/assets/921e39b9-a2b8-42f9-be83-980ffaa259c8" />

### Expected Behavior

Every time I create a new spec, the counter increments

### Actual Behavior

The counter stays at 001

### Specify CLI Version

0.1.13

### AI Agent

Claude Code

### Operating System

macOs 

### Python Version

3.14

### Error Logs

```shell

```

fsilvaortiz:
Submitted a fix in PR #1757.

**Root cause**: The `specify.md` prompt told the AI to search for branches matching the exact short-name (`grep -E 'refs/heads/[0-9]+-<short-name>$'`). Since each new feature has a unique short-name, the search always returned empty, and the prompt said "start with number 1" — causing every feature to get `001`.

**Fix**: The underlying `create-new-feature.sh` already has correct global numbering via `check_existing_branches()`. The fix removes the AI's flawed per-short-name detection and tells it to NOT pass `--number`, letting the script handle it.

🤖 Generated with [Claude Code](https://claude.com/code)

fsilvaortiz:
Submitted a fix in PR #1757.

**Root cause**: The `specify.md` prompt instructed the AI to search for branches filtered by the exact short-name (`grep -E 'refs/heads/[0-9]+-<short-name>$'`), so every new feature started at 001 since no branches matched the new short-name. The underlying `create-new-feature.sh` already has correct global numbering logic via `check_existing_branches()`.

**Fix**: Removed the AI's flawed per-short-name number detection and simplified the prompt to NOT pass `--number`, letting the script auto-detect the next globally available number across all branches and spec directories.

🤖 Generated with [Claude Code](https://claude.com/code)

echarrod:
We're using Copilot in VSCode + Copilot CLI, but still (intermittently) get this error, even on latest version:

Version 0.3.0:

[...]

PR #2036, opened 2026-03-31
fix(scripts): validate --number flag against existing branches/specs

fsilvaortiz:
## Summary

Fixes the intermittent branch numbering reset reported in #1744 (see [comment](https://github.com/github/spec-kit/issues/1744#issuecomment-4073538688) by @echarrod).

PR #1757 fixed the `specify.md` prompt to tell the AI not to pass `--number`, but AI assistants **intermittently ignore** that instruction, causing new features to get `001` instead of the next available number.

This PR adds **script-level validation** as defense in depth:

- When `--number` is passed, the script checks specs directories and git branches (after fetch) for collisions
- If the number is already in use, it warns on stderr and auto-detects the next available number
- Skips validation when `--allow-existing-branch` is set (intentional reuse)
- Applied symmetrically to both Bash and PowerShell scripts

## Test plan

- [x] All 942 existing tests pass
- [x] Updated `test_without_flag_still_errors` → `test_without_flag_auto_detects_on_collision` to reflect new behavior
- [x] Verified collision detection against specs directories
- [x] Verified collision detection against git branches (local + remote)
- [x] Verified `--allow-existing-branch` bypasses validation correctly
- [x] Bash syntax validated

🤖 Generated with [Claude Code](https://claude.com/code)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -287,9 +287,8 @@ else
             done
         fi
 
-        # Check git branches for collision (fetch first to catch remote-only branches)
+        # Check git branches for collision
         if [ "$NUMBER_IN_USE" = false ] && [ "$HAS_GIT" = true ]; then
-            git fetch --all --prune 2>/dev/null || true
             branches=$(git branch -a 2>/dev/null || echo "")
             if [ -n "$branches" ]; then
                 while IFS= read -r branch; do
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -251,7 +251,7 @@ if [ "$USE_TIMESTAMP" = true ]; then
 else
     # Validate --number input when provided
     if [ -n "$BRANCH_NUMBER" ]; then
-        if ! echo "$BRANCH_NUMBER" | grep -q '^[0-9]\+$' || [ "$((10#$BRANCH_NUMBER))" -lt 1 ]; then
+        if ! echo "$BRANCH_NUMBER" | grep -Eq '^[0-9]+$' || [ "$((10#$BRANCH_NUMBER))" -lt 1 ]; then
             >&2 echo "Error: --number requires a positive integer, got '$BRANCH_NUMBER'"
             exit 1
         fi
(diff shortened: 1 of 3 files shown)


mnriem (changes requested):

fsilvaortiz:
Addressed Copilot suggestion in e602275. Added a static guard test (`test_powershell_has_number_collision_validation`) that verifies the PS script contains the collision detection logic — matching the existing pattern used for PowerShell tests in this file, since `pwsh` availability is not guaranteed in CI.

fsilvaortiz:
Re: Copilot comment about `--number 0` — already addressed in 59153f7. The validation was moved before the `if/elif` block and now rejects 0 explicitly (`-lt 1` check). Test coverage added in 6cf7bd1 (`test_rejects_zero`).

mnriem (changes requested):
One more small thing according to Copilot

mnriem:
@fsilvaortiz So more Copilot feedback to address

fsilvaortiz:
Addressed the double-fetch comments in 135163b (bash) and a8e710e (PowerShell): removed `git fetch --all --prune` from the collision check block. If a collision is detected, `check_existing_branches`/`Get-NextBranchNumber` already fetch internally during auto-detect fallback. The collision check uses `git branch -a` with whatever refs are currently available.

fsilvaortiz:
Re: double fetch optimization suggestions (comments on bash line 308 and PS line 260):

Acknowledged — the collision-then-auto-detect path does fetch twice. This is intentional and accepted as a trade-off:

1. **Correctness first**: The collision check needs fresh remote refs to avoid false negatives. The auto-detect fallback (`check_existing_branches`/`Get-NextBranchNumber`) fetches internally and we cannot skip that without modifying existing function signatures.

2. **Rare path**: The double fetch only happens when `--number` is passed AND collides — the exact scenario this PR defends against. In normal operation (no `--number` flag), the existing single-fetch path is unchanged.

3. **Scope**: Introducing a `compute_next_branch_number_from_cache()` function or adding a `-SkipFetch` parameter to `Get-NextBranchNumber` would expand this PR beyond its fix scope and modify stable interfaces. If the maintainer feels this optimization is valuable, it can be a follow-up PR.

The first comment (PS missing fetch) was already addressed in 29302c2.

echarrod (approved):

mnriem (changes requested):
Please resolve conflicts

PR #2051, opened 2026-04-01
Install Claude Code as native skills and align preset/integration flows

### LINKED ISSUE #2031: [Bug]: .claude/commands/ no longer recognized in Claude Code v2.1.88 — needs migration to .claude/skills/
### Bug Description

After upgrading to Claude Code v2.1.88, all speckit slash commands (/speckit.analyze, /speckit.plan, etc.) installed via 
  specify init in .claude/commands/ return "Unknown skill" when invoked. The /skills dialog does not list them either.     
  
  Claude Code has unified the old "commands" and "skills" systems — .claude/commands/ is no longer scanned for slash       
  commands. Only .claude/skills/<name>/SKILL.md format is recognized.

### Steps to Reproduce

1. Install spec-kit and run specify init (creates files in .claude/commands/)
  2. Upgrade Claude Code to v2.1.88                                                                                        
  3. Type /speckit.analyze in the prompt
  4. See: "Unknown skill: speckit.analyze"   

### Expected Behavior

Slash commands should be available as /speckit.analyze, /speckit.plan, etc.  

### Actual Behavior

All speckit commands return "Unknown skill" — Claude Code no longer reads .claude/commands/.  

### Specify CLI Version

spec-kit (latest via specify init)  

### AI Agent

Claude Code

### Operating System

macOS (Darwin 25.3.0)

### Python Version

Python 3.14.2

### Error Logs

```shell

```

### Additional Context

 Related to #291 but a different root cause — this is not a path/cwd issue. The .claude/commands/ mechanism itself has
  been deprecated in favor of .claude/skills/. The specify init command needs to be updated to output skills-format files  
  (.claude/skills/<name>/SKILL.md with proper frontmatter including name, description, user-invocable: true).

afurm:
Fixes #2031.

This PR moves Claude Code support from generated slash-command files in `.claude/commands` to native skills in `.claude/skills`, and aligns the surrounding init, hook, and preset flows with that model.

What changed:
- Added a dedicated Claude integration that writes skills to `.claude/skills/<skill>/SKILL.md`
- Updated Claude init behavior so `specify init --ai claude` installs skills by default
- Updated CLI messaging and usage output to show Claude skill invocations like `/speckit-plan`
- Updated hook rendering so Claude skills are displayed with `/speckit-*` invocation syntax
- Updated README examples and docs to reflect Claude’s skills-first workflow
- Added focused integration tests for Claude skill generation and init behavior

Follow-up fixes included in this branch:
- Interactive Claude selection now uses the integration path too, so selecting Claude from the prompt produces the same integration metadata, scripts, and skill output as `--ai claude`
- Preset skill propagation now creates missing Claude skill directories for new preset-provided commands, instead of only updating already-existing skills

Why this is needed:
- Claude now expects spec-kit to be installed as native skills rather than legacy command files
- Without these fixes, Claude projects behaved differently depending on how init was invoked, and fresh Claude skills projects could silently miss brand-new preset commands added by presets

Reviewer guide:
- Claude integration implementation: `src/specify_cli/integrations/claude/__init__.py`
- Claude init/autopromotion flow: `src/specify_cli/__init__.py`
- Preset skill propagation fix: `src/specify_cli/presets.py`
- Hook invocation rendering: `src/specify_cli/extensions.py`
- Main tests: `tests/integrations/test_integration_claude.py`, `tests/test_presets.py`, `tests/test_extensions.py`

## Testing

Targeted tests run:
- `uv run pytest tests/integrations/test_integration_claude.py tests/test_presets.py -q`
- `uv run pytest tests/integrations/test_cli.py -q`
- `uv run pytest tests/test_extensions.py -k HookInvocationRendering -q`

## AI Disclosure

Used Codex to review the branch, identify regressions in Claude preset/init flows, implement the fixes, and add/update targeted tests.

### DIFF
--- a/tests/integrations/test_cli.py
+++ b/tests/integrations/test_cli.py
@@ -76,6 +76,33 @@ class TestInitIntegrationFlag:
         assert "--integration copilot" in result.output
         assert (project / ".github" / "agents" / "speckit.plan.agent.md").exists()
 
+    def test_ai_claude_here_preserves_preexisting_commands(self, tmp_path):
+        from typer.testing import CliRunner
+        from specify_cli import app
+
+        project = tmp_path / "claude-here-existing"
+        project.mkdir()
+        commands_dir = project / ".claude" / "commands"
+        commands_dir.mkdir(parents=True)
+        command_file = commands_dir / "speckit.specify.md"
+        command_file.write_text("# preexisting command\n", encoding="utf-8")
+
+        old_cwd = os.getcwd()
+        try:
+            os.chdir(project)
+            runner = CliRunner()
+            result = runner.invoke(app, [
+                "init", "--here", "--force", "--ai", "claude", "--ai-skills", "--script", "sh", "--no-git", "--ignore-agent-tools",
+            ], catch_exceptions=False)
+        finally:
+            os.chdir(old_cwd)
+
+        assert result.exit_code == 0, result.output
+        assert "--integration claude" in result.output
+        assert command_file.exists()
+        assert command_file.read_text(encoding="utf-8") == "# preexisting command\n"
+        assert (project / ".claude" / "skills" / "speckit-plan" / "SKILL.md").exists()
+
     def test_shared_infra_skips_existing_files(self, tmp_path):
         """Pre-existing shared files are not overwritten by _install_shared_infra."""
         from typer.testing import CliRunner
--- a/tests/test_ai_skills.py
+++ b/tests/test_ai_skills.py
@@ -756,39 +756,6 @@ class TestLegacyDownloadPath:
 
         assert not (tmp_path / "evil.txt").exists()
 
-    def test_here_mode_commands_preserved(self, tmp_path, monkeypatch):
-        """For --here on existing repos, commands must NOT be removed."""
-        from typer.testing import CliRunner
-
-        runner = CliRunner()
-        # Create a mock existing project with commands already present
-        target = tmp_path / "existing"
-        target.mkdir()
-        agent_folder = AGENT_CONFIG["claude"]["folder"]
-        cmds_dir = target / agent_folder.rstrip("/") / "commands"
-        cmds_dir.mkdir(parents=True)
-        (cmds_dir / "speckit.specify.md").write_text("# spec")
-
-        # --here uses CWD, so chdir into the target
-        monkeypatch.chdir(target)
-
-        def fake_download(project_path, *args, **kwargs):
-            pass  # commands already exist, no need to re-create
-
-        with patch("specify_cli.integrations.get_integration", return_value=None), \
-             patch("specify_cli.download_and_extract_template", side_effect=fake_download), \
-             patch("specify_cli.ensure_executable_scripts"), \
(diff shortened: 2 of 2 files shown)


mnriem (changes requested):
Can you please make sure that any Claude specific tests are in test_integration_claude.py. This will make it easier for anyone to see what Claude specific tests there are. Thanks!

afurm:
> Can you please make sure that any Claude specific tests are in test_integration_claude.py. This will make it easier for anyone to see what Claude specific tests there are. Thanks!

Done. I moved the Claude-specific tests into `tests/integrations/test_integration_claude.py` and pushed the cleanup in `ce96610`.

I also reran the affected test files after the move to make sure this was only organizational and didn’t change behavior.

afurm:
> Please address Copilot feedback

Done

afurm:
Addressed one suggestion directly and reviewed the other in detail.

What I fixed:
- Switched `ClaudeIntegration` to use `skills_dest()` now that it subclasses `SkillsIntegration`.
- Fixed a related native-skill preset cleanup issue I found while validating the review.
- Added regression tests for the native-skill preset flow.

What I did not change:
- I did not apply the `create_missing_skills == "/SKILL.md"` change as written. Native skill agents already create new preset skills through `_register_commands()`, so flipping that condition directly would be incorrect for the current code paths.

afurm:
Addressed. I removed the test from the legacy-path suite and replaced it with a CLI-level test that covers the current Claude integration behavior instead. It now verifies that `specify init --here --ai claude` preserves pre-existing `.claude/commands` while installing skills, which felt more appropriate given that the legacy path is expected to be removed soon.

mnriem:
Thank you!

mariozig:
@afurm I noticed we have a mix of references to skills using both `speckit-` and `speckit.`. Some use dash instead of dot. 

Is this intentional? ...is the project switching to dash?  

I noticed after taking this code i now see both:

Screenshot 2026-04-02 at 10 23 02 AM

Thank you!

afurm:
> @afurm I noticed we have a mix of references to skills using both `speckit-` and `speckit.`. Some use dash instead of dot.
> 
> Is this intentional? ...is the project switching to dash?
> 
> I noticed after taking this code i now see both:
> 
> Screenshot 2026-04-02 at 10 23 02 AM
> Thank you!

For Claude, the dashed form is intentional. The PR switched Claude to skills in `.claude/skills`, so the user-facing invocation is now `/speckit-*`.

The dotted form still exists in the codebase as the canonical internal command ID (`speckit.plan`, `speckit.specify`, etc.) and is still used by other agents, extensions, presets, and metadata, so seeing both styles in source is expected.

What you are likely seeing locally is both legacy `.claude/commands` entries and new `.claude/skills` entries present at the same time after the upgrade. That would explain why Claude shows both. So I don’t think this points to a bug in the merged naming change itself, but it may indicate a follow-up migration/cleanup gap for older Claude installs.

mnriem:
@afurm Is correct

mariozig:
Got it! Thank you both. 💪

PR #2099, opened 2026-04-05
feat: add --allow-model-invocation flag for agent-orchestrated workflows

### LINKED ISSUE #2098: [Feature]: add `--allow-model-invocation` flag for agent-orchestrated workflows
### Problem Statement

Since v0.4.5 (#2051), `specify init --ai claude` installs speckit commands as native skills under `.claude/skills/` with `disable-model-invocation: true` hardcoded in `build_skill_frontmatter()` (`agents.py`). This prevents Claude from invoking speckit skills via the Skill tool — only the user can trigger them manually via `/speckit-specify`, `/speckit-plan`, etc.

This is a reasonable default for most users (prevents Claude from auto-triggering side-effect-heavy workflow steps). However, it breaks agent-orchestrated speckit workflows where Claude is instructed (via rules/steering files) to invoke speckit skills programmatically as part of a multi-step pipeline.

Currently this only affects `--ai claude` (the flag is inside an `if agent_name == "claude"` guard in `build_skill_frontmatter()`), but other skill-based agents may benefit from the same override if they support model-initiated invocation.

Note: #2081 refactored the injection into `_inject_frontmatter_flag()` but kept the hardcoded `True` behavior.

### Proposed Solution

Add an `--allow-model-invocation` flag to `specify init` that sets `disable-model-invocation: false` (or omits the key entirely) in generated SKILL.md frontmatter.

Persist the choice in `init-options.json` so it survives re-init and is respected by `specify extension add`:

```json
{
  "ai_skills": true,
  "allow_model_invocation": true
}
```

**Implementation sketch:**

1. **`specify init`**: accept `--allow-model-invocation` flag, persist to `init-options.json`
2. **`build_skill_frontmatter()` in `agents.py`**: read the option from `init-options.json` (via `load_init_options()`) and conditionally set the flag
3. **`specify extension add`**: already calls `build_skill_frontmatter()` via `CommandRegistrar`, so it would inherit the behavior automatically if the method reads `init-options.json`

**Affected code:**

- `build_skill_frontmatter()` in `src/specify_cli/agents.py` — where the flag is hardcoded
- `_inject_frontmatter_flag()` added in #2081 — the shared helper that injects the flag post-setup
- `init()` in `src/specify_cli/__init__.py` — to accept and persist the new option
- `load_init_options()` — already used by `resolve_skill_placeholders()`, same pattern

### Alternatives Considered

**Manual sed after every init/extension install:**

```bash
sed -i '' 's/disable-model-invocation: true/disable-model-invocation: false/' .claude/skills/speckit-*/SKILL.md
```

This works but must be re-applied after every `specify init --force` or `specify extension add` that regenerates skills. Not sustainable across many projects.

**Keeping legacy `.claude/commands/` alongside skills:** Duplicates files and they drift over time.

### Component

### AI Agent

Claude Code

### Use Cases

1. Users who build structured orchestration flows in their Claude Code steering (e.g., rules that map speckit steps 1–17 to specific skills and have Claude invoke them sequentially) cannot use the native skills path — the Skill tool cannot see or invoke skills with `disable-model-invocation: true`.
2. Teams that want Claude to drive the speckit workflow autonomously (with user approval gates at key steps) while keeping the native skills format.
3. Extension authors who want their extensions to be invocable by the model as part of automated pipelines.

### Acceptance Criteria

- [ ] `specify init --ai claude --allow-model-invocation` generates SKILL.md files with `disable-model-invocation: false`
- [ ] The flag is persisted in `init-options.json` as `allow_model_invocation: true`
- [ ] `specify init --force` respects the persisted option
- [ ] `specify extension add` respects the persisted option when generating skills
- [ ] Default behavior (no flag) remains unchanged (`disable-model-invocation: true`)
- [ ] Documentation is updated

### Additional Context

> This issue was authored with assistance from Claude Code.

srobroek:
## Summary
- Add `--allow-model-invocation` flag to `specify init`
- Persist choice in `.specify/init-options.json`
- `build_skill_frontmatter()` reads the option and conditionally sets `disable-model-invocation`
- Extension and preset installs automatically respect the persisted option

## Implementation Details

The flag defaults to `false` (skills disabled for model invocation) to maintain backward compatibility. When a user runs:

```bash
specify init --integration claude --allow-model-invocation
```

The generated SKILL.md files will have `disable-model-invocation: false`, allowing the AI model to invoke speckit skills programmatically for agent-orchestrated workflows.

## Changes
- **CLI**: Added `--allow-model-invocation` option to `specify init` command
- **Persistence**: Flag stored in `.specify/init-options.json` for consistent behavior
- **Skill Generation**: `CommandRegistrar.build_skill_frontmatter()` now accepts optional `project_root` parameter and reads init-options to determine the flag value
- **Integration Setup**: `ClaudeIntegration.setup()` conditionally injects `disable-model-invocation` based on the persisted option
- **Extensions & Presets**: All skill generation paths (init, extension install, preset install) respect the setting
- **Tests**: Added comprehensive test coverage for all scenarios

## Testing

All existing tests pass, plus 4 new tests covering:
1. Default behavior (disable-model-invocation: true)
2. Flag enabled behavior (disable-model-invocation: false)
3. `build_skill_frontmatter()` respects init-options.json
4. Preset installs respect the setting

```bash
pytest tests/integrations/test_integration_claude.py::TestAllowModelInvocation -xvs
# 4 passed
```

Fixes github/spec-kit#2098

**AI Disclosure (per CONTRIBUTING.md)**: Claude Code was used for code generation. Changes were reviewed and tested by the submitter.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -843,6 +843,7 @@ def init(
     branch_numbering: str = typer.Option(None, "--branch-numbering", help="Branch numbering strategy: 'sequential' (001, 002, ...) or 'timestamp' (YYYYMMDD-HHMMSS)"),
     integration: str = typer.Option(None, "--integration", help="Use the new integration system (e.g. --integration copilot). Mutually exclusive with --ai."),
     integration_options: str = typer.Option(None, "--integration-options", help='Options for the integration (e.g. --integration-options="--commands-dir .myagent/cmds")'),
+    allow_model_invocation: bool = typer.Option(False, "--allow-model-invocation", help="Allow AI model to invoke speckit skills programmatically (sets disable-model-invocation: false)"),
 ):
     """
     Initialize a new Specify project.
@@ -1125,6 +1126,27 @@ def init(
             if ai_skills:
                 integration_parsed_options["skills"] = True
 
+            # Persist the CLI options BEFORE calling setup() so that integrations
+            # can read them during skill generation (e.g., allow_model_invocation).
+            init_opts = {
+                "ai": selected_ai,
+                "integration": resolved_integration.key,
+                "branch_numbering": branch_numbering or "sequential",
+                "here": here,
+                "preset": preset,
+                "script": selected_script,
+                "speckit_version": get_speckit_version(),
+            }
+            # Ensure ai_skills is set for SkillsIntegration so downstream
+            # tools (extensions, presets) emit SKILL.md overrides correctly.
+            from .integrations.base import SkillsIntegration as _SkillsPersist
+            if isinstance(resolved_integration, _SkillsPersist):
+                init_opts["ai_skills"] = True
+            # Persist allow_model_invocation flag if specified
+            if allow_model_invocation:
+                init_opts["allow_model_invocation"] = True
+            save_init_options(project_path, init_opts)
+
             resolved_integration.setup(
                 project_path, manifest,
                 parsed_options=integration_parsed_options or None,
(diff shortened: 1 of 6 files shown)


mnriem:
In order to not bleed in integration specific options into the CLI we have added a singular --integration-options flag that should be used that then subsequently shunts it into the integration. For your case it should look like --integration-optons="--allow-model-invocation". Also note that the integration is solely responsible for this additional flag and we should probably make sure CommandRegistry renders the skill altogether (we might not be there yet, but that is where we will end up).

mnriem (changes requested):
Please address comments above

kennedy-whytech:
support for the option to --allow-model-invocation, so there's more potential for integration.

mnriem:
You must use --integration-options="--allow-model-invocation" and not parse them in the top level init.py. Each integration is responsible for its own options.

srobroek:
apologies completely missed the GH notifications, working on updating this now.

srobroek:
As it turns out 0.8.4 already sets disable-model-invocation to false by default, so this is now redundant.

Issue #2101, opened 2026-04-06
fix: Extension SKILL.md files contain unresolvable relative paths to extension subdirectories


mbachorik:
## Summary

When an extension is installed, its commands are registered as SKILL.md files
(for agents like Kimi and Codex). Extension command bodies commonly reference
files using paths relative to the extension root — e.g.:

```
Read agents/control/commander.md for instructions.
Load knowledge-base/agent-scores.yaml for calibration.
Use templates/kill-report.md as output format.
```

After installation the extension files live at
`.specify/extensions/<id>/agents/...`, `.specify/extensions/<id>/knowledge-base/...`,
etc. But the generated SKILL.md files emit these bare relative paths verbatim.
The AI agent resolves them relative to the workspace root, where they don't
exist, so the referenced files are never found.

## Reproduction

1. Install an extension that has subdirectories (`agents/`, `knowledge-base/`,
   `templates/`, `scripts/`) referenced from its command bodies (e.g. `echelon`).
2. Inspect the generated SKILL.md:
   ```
   cat .kimi/skills/speckit-echelon-run/SKILL.md | grep "agents/"
   ```
   Result: `Read agents/control/commander.md for instructions.`
   
   Expected: `Read .specify/extensions/echelon/agents/control/commander.md for instructions.`

## Root cause

`CommandRegistrar.render_skill_command()` renders the command body without
rewriting extension-relative paths to their installed locations. The method
has no knowledge of which source directory the body came from, so it cannot
perform the substitution.

## Fix

Add `CommandRegistrar._rewrite_extension_paths(text, extension_id, extension_dir)`
that discovers the subdirectories that exist inside the installed extension
and rewrites matching `subdir/...` references to `.specify/extensions/<id>/subdir/...`.
Call it from `render_skill_command()` when `source_dir` is provided, and thread
`source_dir` through from `register_commands()`.

PR forthcoming against `main`.

## Affected agents

Any agent using the `/SKILL.md` extension format — currently `kimi` and `codex`.

marcelsafin:
Since #2102 was closed in April and not reopened, I would like to pick this up. Plan:

1. Failing test: install an extension whose command body references extension-relative paths (`agents/...`, `templates/...`), assert the generated SKILL.md references resolve from the skill file location.
2. Fix in the shared skill-rendering path: rewrite extension-relative references to the installed extension directory (project-relative), so it covers extension add, install, upgrade, and switch.
[...]

PR #2102, opened 2026-04-06
fix(extensions): rewrite extension-relative paths in generated SKILL.md files

mbachorik:
Fixes github/spec-kit#2101

## What

When an extension command body references files relative to the extension root
(e.g. `agents/control/commander.md`, `knowledge-base/scores.yaml`), those
paths are now correctly rewritten to `.specify/extensions/<id>/agents/...` in
the generated SKILL.md files.

## Why

Extension commands are registered as SKILL.md files for skill-format agents
(currently `kimi` and `codex`). After installation the extension files live
at `.specify/extensions/<id>/...`, but `render_skill_command()` had no path
rewriting — it wrote the bare relative paths verbatim. The AI agent then tries
to read `agents/control/commander.md` from the workspace root, which doesn't
exist, causing the skill's state machine to fail immediately.

## How

- Add `CommandRegistrar._rewrite_extension_paths(text, extension_id, extension_dir)`:
  discovers which subdirectories exist inside the installed extension and
  rewrites matching `subdir/…` references to `.specify/extensions/<id>/subdir/…`
  using conservative regex anchoring (path boundary characters only).
- Call it from `render_skill_command()` when `source_dir` is provided.
- Pass `source_dir` from `register_commands()` to `render_skill_command()`.

## Test plan

- [x] `test_skill_registration_rewrites_extension_relative_paths` — new test
  verifying `agents/`, `templates/`, `scripts/`, and `knowledge-base/` paths
  are rewritten while non-extension dirs (e.g. `memory/`) are left alone.
- [x] All 228 existing extension + ai-skills tests pass.

### DIFF
--- a/tests/test_integration_foobar_extension.py
+++ b/tests/test_integration_foobar_extension.py
@@ -1,0 +1,276 @@
+"""
+Integration tests: install a real extension into a temp project and verify
+that generated SKILL.md files have correct .specify/extensions/<id>/… paths
+instead of bare extension-relative references.
+
+Set the SPECKIT_TEST_EXT_DIR environment variable to the path of a local
+extension checkout before running. Tests are skipped automatically when
+the variable is not set or the directory does not exist.
+
+Example:
+    SPECKIT_TEST_EXT_DIR=~/work/my-extension pytest tests/test_integration_foobar_extension.py
+"""
+
+import json
+import os
+import re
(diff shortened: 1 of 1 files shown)


mbachorik:
Closing — will reopen against main from a clean branch after the active PR merges.

PR #2152, opened 2026-04-10
fix(specify): correct self-referencing step number in validation flow

ismaelJimenez:
## Description

Fix self-referencing step number in the `specify` command template's validation flow.

Step 7c ("If all items pass") incorrectly said "proceed to step 7", which points back to the current step (Specification Quality Validation), creating an infinite loop. Changed to "proceed to step 8" (Report completion).

## Testing

- [x] Verified the step numbering is correct by reviewing the full outline (steps 1–9)

## AI Disclosure

- [x] I **did not** use AI assistance for this contribution

### DIFF
--- a/templates/commands/specify.md
+++ b/templates/commands/specify.md
@@ -183,7 +183,7 @@ Given that feature description, do this:
 
    c. **Handle Validation Results**:
 
-      - **If all items pass**: Mark checklist complete and proceed to step 7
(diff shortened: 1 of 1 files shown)


ismaelJimenez:
@mnriem This is an small fix for a typo we have on the specify template

mnriem:
Thank you!

Issue #2165, opened 2026-04-10
Add notice that git extension will no longer be enabled by default at 0.10.0


mnriem:
### Summary

The `git` extension is currently bundled and enabled by default during `specify init`. Starting with v1.0.0, it will still be bundled but no longer enabled by default — users will need to explicitly opt in. We should add messaging now so users are aware of the upcoming change.

### Proposed behavior

During `specify init`, display a notice when the git extension is auto-enabled, e.g.:

> ⚠️ The git extension is currently enabled by default, but starting with v0.10.0 it will require explicit opt-in via `specify init --extension git` or post-init with `specify extension add git`.

### Scope

- Notice should appear **only during `specify init`**
- No changes to the git extension itself or its bundled status
- The git extension remains bundled (shipped with spec-kit) — only the default-enabled behavior changes at 0.10.0

### Acceptance criteria

- [ ] `specify init` shows a notice when the git extension is auto-enabled
- [ ] Notice is visible but non-blocking (yellow/warning styling)
- [ ] Message clearly states the timeline (v0.10.0) and how to opt in

aaronrsun:
https://github.com/github/spec-kit/pull/2197

PR #2197, opened 2026-04-12
Added warning message about git extension no longer being enabled by default

aaronrsun:
Addresses: #2165
Spec doc: #2196


mnriem (changes requested):
Would it be possible that the extension itself vends the message during its setup?

aaronrsun:
> Would it be possible that the extension itself vends the message during its setup?

@mnriem Addressed your comment as well as all of Copilot's. Please take a look again.

src/specify_cli/__init__.py
```diff
@@ -1277,6 +1277,8 @@ def init(
 
             ensure_constitution_from_template(project_path, tracker=tracker)
 
+            _git_ext_freshly_installed = False
```
mnriem, line 1280: Can this live in extensions.py somehow? I want extensions to be handled by just one point in the code. If not possible, please explain why.

mnriem (changes requested):

Issue #2168, opened 2026-04-10
Make git extension opt-in and remove `--no-git` at 0.10.0


mnriem:
### Summary

At 0.10.0, finalize the git extension default change and `--no-git` removal:

1. **Git extension becomes opt-in** — no longer auto-enabled during `specify init`. Users must pass `--extension git` or add it post-init with `specify extension add git`.
2. **Remove `--no-git` flag** — no longer needed since the git extension won't be enabled by default.

This is the follow-through on the deprecation notices added earlier.

### Depends on

- #2165 (deprecation notice for git extension default)
- #2166 (`--extension` flag on `specify init`)
- #2167 (deprecation notice for `--no-git`)

### Acceptance criteria

- [ ] `specify init` no longer auto-enables the git extension
- [ ] Git extension remains bundled (shipped with spec-kit) and installable via `--extension git`
- [ ] `--no-git` flag is removed
- [ ] Documentation updated to reflect the change
- [ ] CHANGELOG entry noting the breaking change

Issue #2167, opened 2026-04-10
Deprecate `--no-git` flag on `specify init` (removal at 0.10.0)


mnriem:
### Summary

The `--no-git` flag on `specify init` skips git repository initialization and git extension installation. With the introduction of `--extension` (#2166) and the git extension no longer being enabled by default at 0.10.0, `--no-git` becomes redundant and should be deprecated now and removed at 0.10.0.

### Proposed behavior

- Add a deprecation warning when `--no-git` is used: *"⚠️ `--no-git` is deprecated and will be removed in v0.10.0. The git extension will no longer be enabled by default — use `--extension git` to opt in."*
- `--no-git` continues to function until 0.10.0
- Removal happens in #TBD (0.10.0 issue)

### Depends on

- #2166 (`--extension` flag)

### Acceptance criteria

- [ ] Using `--no-git` emits a visible deprecation warning
- [ ] `--no-git` still functions as before (non-breaking)
- [ ] Remove `--no-git` at 0.10.0

Issue #2047, opened 2026-04-01
[Feature]: Configurable Branching Strategy Without Command Forking


PeterDaumQS:
### Problem Statement

Spec Kit currently hardcodes feature branch naming and feature directory resolution around two built-in formats:

- sequential: `001-feature-name`
- timestamp: `YYYYMMDD-HHMMSS-feature-name`

This makes adoption difficult for teams that use ticket-based or organization-specific naming such as:

- `US12321`
- `PDC-12312`
- `feature/US12312`
- `PDC-12312-bug`
- `GH-123-improvement-this-is-a-cool-feature`

The current customization story pushes users toward one of two undesirable options:

- rewriting core command files through presets
- directly modifying core command files or scripts

This proposal introduces a first-class, configuration-driven branching strategy that lets teams adapt Spec Kit to their branch conventions without forking commands.
It also preserves a no-git workflow where feature resolution works without branches.

## Problem

The current implementation couples three concerns too tightly:

1. branch creation strategy
2. feature identity
3. spec directory lookup

Today, the workflow assumes that:

- `/speckit.specify` will always generate a short slug
- branch naming is always sequential or timestamp-based
- follow-up commands can recover feature context by parsing a numeric or timestamp prefix from the current branch name

That assumption is embedded in both command instructions and bash scripts, which means the branching model is not practically extensible.

### Hardcoded creation flow

The specify command instructs the agent to generate a short branch slug and only supports sequential or timestamp prefixing.

### Hardcoded validation

Feature branch validation only accepts numeric or timestamp-prefixed names.

### Hardcoded feature directory resolution

Follow-up commands infer the feature directory from a parsed branch prefix, which breaks for company-style branch names.

### Poor customization seam

Presets can override commands, but that requires replacing the entire command prompt rather than configuring only the branching policy. Extensions can add behavior, but they are not the right abstraction for redefining core feature identity semantics.

### Proposed Solution

Introduce a first-class branching configuration file, for example:

- `.specify/branching.yml`

This file should define how Spec Kit:

- validates branch names
- extracts a stable feature identifier from a branch name
- derives the spec directory name
- generates a new branch during `/speckit.specify`
- resolves feature identity when no git repository is present

The key design change is:

- do not use a central registry as the default source of truth
- do not require the spec directory to equal the raw branch name
- instead derive a stable `feature_id` from the branch name using configured rules

This keeps the model deterministic, avoids merge hotspots, and still supports parallel work.

[...]

Issue #1921, opened 2026-03-20
[Feature]: Make branch creation configurable


ramonduraes:
### Problem Statement

Hi all,

Thanks for the work on this toolkit, it’s been really useful to structure spec-driven workflows.

I’d like to suggest making the automatic branch creation optional or configurable.

Right now, creating a new branch for each spec is the default behavior. This works well for teams using a branch-per-initiative model, but it can add friction for teams already using trunk-based development or custom branching strategies.

## Proposal

It would be helpful to support something like:

- On first run, the `specify` command could ask whether the user wants to use automatic branch creation or not
- Persist this choice in a config file, for example `.spec-kit.json`
- A setting such as `"branchStrategy": "auto" | "none" | "manual"`
- Optional CLI flags like `--no-branch` to override behavior per execution

## Why this matters

This would allow teams to adopt the spec structure without being tied to a specific Git workflow, making the toolkit more flexible and easier to integrate into existing engineering practices.

Happy to help or contribute if this direction makes sense.

### Proposed Solution

It would be great to make the automatic branch creation optional and configurable.

On the first run, the `specify` command could prompt the user to choose whether to use automatic branch creation or not. This would allow teams to align the tool with their existing Git workflow from the start.

This preference could then be persisted in a config file, for example `.spec-kit.json`, with a setting like:

- `"branchStrategy": "auto" | "none" | "manual"`

Additionally, CLI flags such as `--no-branch` could be used to override the behavior per execution when needed.

This approach keeps the current behavior for teams that rely on it, while allowing more flexibility for teams using trunk-based development or custom branching strategies.

### Component

Specify CLI (initialization, commands)

### AI Agent (if applicable)

None

mnriem:
Does the --no-git do the trick for you?

See https://github.com/github/spec-kit?tab=readme-ov-file#specify-init-arguments--options

bmcnaboe:
First, thank you, Spec Kit team, for a very helpful tool!

@mnriem, as far as I can tell, the --no-git option only affects the project's initialization (i.e., whether or not to create a new repo during the initial setup). This is completely different from whether or not to create a new branch for each spec, assuming the project is using git.

I support @ramonduraes's request. I have a branching flow that uses a different naming convention than Spec Kit, and checking out to the new branch feels a bit presumptuous and unhelpful when working on multiple specs at once. As a side note, although it shouldn't really matter, I'm using Worktrunk to manage worktrees and associated branches.

[...]

PR #2209, opened 2026-04-13
feat: add modifies_hooks support for extension hook overrides

dango85:
## Summary

- Extensions can now declare a `modifies_hooks` section in `extension.yml` to **disable** or **set_optional** hooks from other installed extensions during `specify extension add`
- Solves the conflict where extensions like [spec-kit-worktree-parallel](https://github.com/dango85/spec-kit-worktree-parallel) need to disable the git extension's `before_specify -> speckit.git.feature` hook (which forces branch switching, breaking the worktree-first workflow)

## Related Issues

This PR addresses a common pain point across many open issues — the inability for extensions to coordinate hook behavior, especially around git branch creation:

| Issue | Title | Impact |
|-------|-------|--------|
| #654 | Rethink the branch thing | 142 upvotes — widespread frustration with forced branching; `modifies_hooks` lets extension authors provide alternatives |
| #61 | Spawn worktree when creating new branch | 36+ upvotes — worktree extensions need to disable branch-first hooks |
| #1476 | Native git worktree support for parallel agents | Parallel agents need stable checkout; this PR is the extension-level answer |
| #1680 | Support using existing branch (without auto branch creation) | Extensions can now disable `git.feature` to stay on the current branch |
| #1921 | Make branch creation configurable | `modifies_hooks` provides the mechanism for conditional branch behavior |
| #2047 | Configurable branching strategy without command forking | Extensions declare `modifies_hooks` instead of forking the `specify` command |
| #2168 | Make git extension opt-in at 1.0.0 | `modifies_hooks` provides a granular alternative: keep git for commits, disable branch creation |
| #2167 | Deprecate `--no-git` flag | Eliminates the need for `--no-git` as a blunt instrument — extensions surgically disable specific hooks |
| #343 | Add hook system for workflow automation | Extends the hook system with inter-extension coordination |
| #1850 | Preset system: hook system for post-install automation | `modifies_hooks` is the install-time hook automation mechanism |
| #1016 | Allow branching to be optional in git repos | `modifies_hooks: action: disable` on the branch hook is exactly this |
| #377 | Multi agent in one project | Parallel agents need worktree isolation, which needs the branch hook disabled |
| #537 | Remove all git functionality | Extreme version of the same frustration — `modifies_hooks` lets people disable git hooks selectively |

## Safety Model

| Guardrail | Behavior |
|-----------|----------|
| **Explicit consent** | Interactive Y/N prompt before applying any modifications |
| **CI / non-TTY** | Defaults to NO (fail-closed) |
| **Granular targeting** | Requires `hook` + `extension` + `command` triple |
| **Limited actions** | Only `disable` and `set_optional` (no removal, no arbitrary mutation) |
| **Mandatory reason** | `reason` field required and shown in consent prompt |
| **Community warning** | Extra warning when modifying non-bundled extension hooks |
| **Reversible** | Original state restored on `specify extension remove` |
| **Audit trail** | Applied modifications stored in registry `_modified_hooks` |

## Schema

```yaml
modifies_hooks:
  - hook: before_specify
    extension: git
    command: speckit.git.feature
    action: disable
    reason: "Worktree-parallel keeps primary checkout stable; branch creation happens via git worktree add -b"
```

## Consent UX

```
Extension 'worktrees' requests the following hook modifications:

  Hook                   Target Extension   Command                      Action       Reason
  -----------------------------------------------------------------------------------------------------------------------
  before_specify         git                speckit.git.feature          disable      Worktree-parallel keeps primary...

Apply these modifications? [Y/n]:
```

## Test plan

- [x] 9 new tests added covering: consent accepted, consent declined, non-TTY, uninstall restore, validation (missing fields, bad action), target not found, already-disabled no-op
- [x] All 165 existing extension tests pass (zero regressions)

### DIFF
--- a/src/specify_cli/extensions.py
+++ b/src/specify_cli/extensions.py
@@ -9,6 +9,7 @@ without bloating the core framework.
 import json
 import hashlib
 import os
+import sys
 import tempfile
 import zipfile
 import shutil
@@ -217,6 +218,32 @@ class ExtensionManifest:
                         f"Hook '{hook_name}' missing required 'command' field"
                     )
 
+        # Validate modifies_hooks (if present)
+        modifies_hooks = self.data.get("modifies_hooks")
+        if modifies_hooks is not None:
+            if not isinstance(modifies_hooks, list):
+                raise ValidationError(
+                    "Invalid modifies_hooks: expected a list"
+                )
+            _VALID_MODIFY_ACTIONS = {"disable", "set_optional"}
+            _REQUIRED_MODIFY_FIELDS = {"hook", "extension", "command", "action", "reason"}
+            for idx, entry in enumerate(modifies_hooks):
+                if not isinstance(entry, dict):
+                    raise ValidationError(
+                        f"Invalid modifies_hooks[{idx}]: expected a mapping"
+                    )
+                missing = _REQUIRED_MODIFY_FIELDS - set(entry.keys())
+                if missing:
+                    raise ValidationError(
+                        f"modifies_hooks[{idx}] missing required fields: {', '.join(sorted(missing))}"
+                    )
+                action = entry["action"]
+                if action not in _VALID_MODIFY_ACTIONS:
+                    raise ValidationError(
+                        f"modifies_hooks[{idx}] invalid action '{action}': "
+                        f"must be one of {', '.join(sorted(_VALID_MODIFY_ACTIONS))}"
+                    )
+
         # Validate commands (if present)
         for cmd in commands:
             if "name" not in cmd or "file" not in cmd:
@@ -263,6 +290,15 @@ class ExtensionManifest:
     def hooks(self) -> Dict[str, Any]:
         """Get hook definitions."""
         return self.data.get("hooks", {})
+
+    @property
+    def modifies_hooks(self) -> List[Dict[str, Any]]:
+        """Get hook modification declarations.
+
+        Each entry targets a specific hook from another extension and
+        requests a limited action (disable / set_optional) with a reason.
+        """
+        return self.data.get("modifies_hooks", [])
 
     def get_hash(self) -> str:
         """Calculate SHA256 hash of manifest file."""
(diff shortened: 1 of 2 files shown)


mnriem:
Are you aware that you can disable an extension?

dango85:
Thanks for the feedback @mnriem!

Yes — though `specify extension disable git` turns off **all** git hooks (including `speckit.git.commit`, `speckit.git.initialize`, etc.). For the worktree-parallel case, we only need to skip the `before_specify → speckit.git.feature` hook while keeping everything else working.

Today there's no per-hook granularity at the CLI level, so the only option is manually editing `extensions.yml`. `modifies_hooks` lets extension authors declare that intent so users don't have to discover and fix the conflict themselves.

Happy to adjust the approach if there's a preferred direction — e.g., a `specify extension disable-hook git before_specify` command could work just as well.

mnriem:
I am very hesitant to allow one extension to change the configuration of another. It creates a whole sleuth of other potentital problems with dependencies.

I think the best approach here is to use a preset that overrides just the commands (either core command or extension command) you need to behave differently. 

That way you are in control of what happens. As otherwise any change to the Git extension could very easily break you?

Note at 1.0.0 time the Git extension becomes an opt-in so you cannot rely on the fact it will be auto enabled then

dango85:
Thanks for the clear guidance — I agree.

I won’t merge or pursue cross-extension configuration (one extension changing another’s hooks/config on install). It’s too easy to break when the Git extension changes, and it doesn’t fit Git becoming opt-in at 1.0.0.

For parallel worktrees, the fix is explicit user control: e.g. set enabled: false on the Git extension’s before_specify → speckit.git.feature entry in .specify/extensions.yml when you need a stable primary checkout, and use git worktree add -b so feature branches don’t require git checkout on the shared root.

I’ll treat any stronger “do X before specify” workflow as a preset that overrides only the commands that need to differ.

I’m closing this PR; follow-up is documentation (and a separate preset repo/zip if useful), not more hook-override machinery in core.

PR #2258, opened 2026-04-17
fix: suppress CRLF warnings in auto-commit.ps1

### LINKED ISSUE #2253: [Bug]:  The auto-commit.ps1 script fails when git emits CRLF warnings to stderr
### Bug Description

From ClaudeCode:  The auto-commit.ps1 script fails when git emits CRLF warnings to stderr (it treats any stderr output as a fatal error). This is a known issue on Windows repos without core.autocrlf=false. 

```
 Bash(powershell.exe -NoProfile -File "I:/test/.specify/extensions/git/scripts/powershell/auto-commit.ps1" after_plan)
  ⎿  Error: Exit code 1
     git : warning: in the working copy of '.specify/memory/constitution.md', LF will be replaced by CRLF the next time Git
     touches it
     At I:\test\.specify\extensions\git\scripts\powershell\auto-commit.ps1:120 char:13
     + $diffHead = git diff --quiet HEAD 2>$null; $d1 = $LASTEXITCODE
     +             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         + CategoryInfo          : NotSpecified: (warning: in the... Git touches it:String) [], RemoteException
         + FullyQualifiedErrorId : NativeCommandError
```

### Steps to Reproduce

in a speck project, run `/speckit-git-commit`

### Expected Behavior

Good commit with no errors and handled warnings. 

### Actual Behavior

commit fails...see the powershell error above. 

### Specify CLI Version

0.7.1

### AI Agent

Claude Code

### Operating System

windows 11

### Python Version

3.13.12

### Error Logs

```shell

```

### Additional Context

_No response_

SARAMALI15792:
## Description

Fixes #2253

The `auto-commit.ps1` script fails on Windows when git emits CRLF line-ending warnings to stderr. With `$ErrorActionPreference = 'Stop'`, PowerShell converts any stderr output from native commands into terminating `ErrorRecord` exceptions — crashing the script before the `2>$null` redirection can suppress them.

### Root Cause

PowerShell raises a `NativeCommandError` when it sees *any* stderr output from a native command under `$ErrorActionPreference = 'Stop'`, terminating the script before stderr redirection takes effect. Lines using `2>$null` are ineffective because the error is raised before the redirection is processed.

### Solution

Temporarily set `$ErrorActionPreference = 'Continue'` around all native git calls that may emit stderr warnings (repo detection, change detection, and staging/commit). This prevents stderr from becoming a terminating error while still allowing `2>&1` capture for diagnostics on real failures. Each block uses `try/finally` to guarantee the preference is restored.

### Changes

- **`extensions/git/scripts/powershell/auto-commit.ps1`** — Three git call sites wrapped with `$ErrorActionPreference = 'Continue'` + `try/finally`:
  1. `git rev-parse` (repo detection)
  2. `git diff` / `git ls-files` (change detection — the original crash site)
  3. `git add` / `git commit` (staging and commit)
- **`tests/extensions/git/test_git_extension.py`** — New `TestAutoCommitPowerShellCRLF` class with 5 tests using `core.autocrlf=true` + LF-ending tracked files. On Windows runners the CRLF warnings fire and validate the fix end-to-end.

### Why This Is Safe

- Only affects `$ErrorActionPreference` scope; restored via `finally` blocks
- Uses `'Continue'` (not `'SilentlyContinue'`) so error output is still captured in `$out` for diagnostics
- Exit code semantics unchanged (`$LASTEXITCODE` is unaffected by `ErrorActionPreference`)
- Bash version is unaffected (stderr redirection works correctly there)

## Testing

- All 15 PowerShell tests pass locally (macOS with pwsh)
- New CRLF tests pass trivially on non-Windows (no warnings emitted); on the Windows runner `core.autocrlf=true` triggers actual CRLF warnings to exercise the fix
- Existing `TestAutoCommitPowerShell` tests confirm no regressions

### DIFF
--- a/extensions/git/scripts/powershell/auto-commit.ps1
+++ b/extensions/git/scripts/powershell/auto-commit.ps1
@@ -36,10 +36,14 @@ if (-not (Get-Command git -ErrorAction SilentlyContinue)) {
     exit 0
 }
 
-try {
-    git rev-parse --is-inside-work-tree 2>$null | Out-Null
-    if ($LASTEXITCODE -ne 0) { throw "not a repo" }
-} catch {
+# Temporarily relax ErrorActionPreference so git stderr warnings
+# (e.g. CRLF notices on Windows) do not become terminating errors.
+$savedEAP = $ErrorActionPreference
+$ErrorActionPreference = 'SilentlyContinue'
+git rev-parse --is-inside-work-tree 2>$null | Out-Null
+$isRepo = $LASTEXITCODE -eq 0
+$ErrorActionPreference = $savedEAP
+if (-not $isRepo) {
     Write-Warning "[specify] Warning: Not a Git repository; skipped auto-commit"
     exit 0
 }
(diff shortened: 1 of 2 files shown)


mnriem (approved):

mnriem:
Thank you!

Issue #2185, opened 2026-04-12
[Bug]: specify init fails with "No matching release asset found for claude"


MUHAMMEDHAFEEZ:
### Bug Description

The ```specify init --here --ai claude``` command fails during initialization because it cannot find the required template asset for Claude.

What went wrong?

When running the initialization command with ```--ai claude```, the CLI attempts to fetch a release asset matching the pattern ```spec-kit-template-claude-ps```, but no assets are found, causing the setup to stop.

### Steps to Reproduce

Navigate to any directory

Run: ```specify init --here --ai claude```
Confirm prompt with `y`

Observe the error:  ```No matching release asset found for claude (expected pattern: spec-kit-template-claude-ps)```

### Expected Behavior

The CLI should:

Either download the correct template for claude
Or fallback to a default template
Or provide a clear message about unsupported AI options

### Actual Behavior

The CLI fails with:

```No matching release asset found for claude
(no assets)```

and does not proceed with initialization.

### Specify CLI Version

0.4.2

### AI Agent

Claude Code

### Operating System

Windows 10 (Build 10.0.26200, AMD64)

### Python Version

Python 3.13.0

### Error Logs

```shell
No matching release asset found for claude (expected pattern: spec-kit-template-claude-ps)
Available Assets:
(no assets)
```

### Additional Context

The issue might be related to missing release assets or incorrect naming conventions for Claude templates.
Possibly a regression or incomplete support for Claude in the current version.

MUHAMMEDHAFEEZ:
I found the issue.

The problem happens when using:

`specify init --here --ai claude
`
because it tries to load a Claude template asset that doesn’t exist, which causes the error.

 **Solution**

Use this instead:

`specify init . --integration claude
`
This works correctly and completes the setup without needing any missing assets.

**This seems related to the discussion in issue #2170. **

we need to add it to the README.md and all the docs 

@mnriem

mnriem:
It is not a bug. Since 0.4.5 templates are NOT separately downloadable as we needed to be able to make sure folks were not filing breaking changes against the templates while using an old version of the Specify CLI. Please upgrade your Specify CLI as per the install instructions

sajjad-salam:
<img width="779" height="120" alt="Image" src="https://github.com/user-attachments/assets/b8e9aecd-2982-4ccf-ac92-21ca585d8b07" />
there is no --integration

mnriem:
Which version is this?

sajjad-salam:
[...]

PR #2265, opened 2026-04-17
fix: fall back to older releases when latest has no template assets

Dimsum-art:
## Problem

GitHub releases v0.4.5–v0.6.2 ship with zero template assets, causing `specify init` to hard-fail with exit code 1. The last release with template zips was v0.4.4. Tracked in #2185.

## Fix

Four-step download strategy in `download_template_from_github()`:

1. **Cache-first** — return a cached copy if < 7 days old (or always if `SPECIFY_OFFLINE=1`)
2. **Release fallback** — try `/releases/latest`; if no matching assets, walk up to 10 older releases via `/releases?per_page=10`
3. **Stale-cache rescue** — if all releases have no assets but a cached zip exists, return it with a warning instead of hard-failing
4. **Hard fail only** — raise `Exit(1)` only when no release AND no cache can supply the asset, with a clear message listing versions checked

Three cache helpers added before the function:
- `_get_cache_dir()` — returns `~/.cache/specify/templates/` (created on demand)
- `_find_cached_template(pattern)` — returns newest cached zip by mtime (deterministic)
- `_cache_template(zip_path, metadata)` — copies zip + writes `cache-meta.json` (best-effort, never raises)

`SPECIFY_OFFLINE=1` env var skips all network calls and uses the cache directly.

## Changes

- `src/specify_cli/__init__.py` — cache helpers + rewritten download function
- `tests/test_download_fallback.py` — 5 new tests covering all fallback paths
- `tests/conftest.py` — ensures tests import from `src/` not the installed package

## Tests

```
5 passed in 0.10s
✓ test_falls_back_to_older_release_when_latest_has_no_assets
✓ test_uses_stale_cache_when_all_releases_have_no_assets
✓ test_offline_mode_uses_cache_without_network
✓ test_offline_mode_without_cache_raises_exit
✓ test_successful_download_is_cached
```

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -30,6 +30,7 @@ import tempfile
 import shutil
 import shlex
 import json
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


mnriem:
Sorry, but we removed the zip bundles since 0.4.5 (see its release notes at https://github.com/github/spec-kit/releases/tag/v0.4.5) and as such we cannot take this contribution in as it is expected that users upgrade the CLI as per install instructions going forward

PR #2271, opened 2026-04-18
feat: implement multi-agent support and 'use' command

### LINKED ISSUE #2257: [Feature]: Support multiple installed integrations in the same project
### Problem Statement

`specify integration` now supports `install`, `uninstall`, and `switch`, but it still appears to enforce a single installed integration per project.

I'm frustrated when a repository needs to support multiple agents used by different people on the same team, because the second integration cannot be installed alongside the first one.

Repro with `specify 0.6.1`:

```bash
specify init --here --ai claude --offline --force
specify integration install codex
```

Observed behavior:

```text
Integration 'claude' is already installed.
Run specify integration uninstall first, or use specify integration switch codex.
```

Also:

- `specify integration list` shows only one installed integration
- `.specify/integration.json` still models a single current integration

This means `#1924` / `#2065` solved integration management and switching, but not simultaneous multi-install in one project.

### Proposed Solution

Allow multiple integrations to be installed in the same repository, while keeping one integration as the default/active one.

Concretely, I would like:

- multiple installed integrations in one project
- one `default_integration`
- explicit installed integration state, for example `installed_integrations`
- `specify integration install <agent>` to work without requiring uninstall/switch first when another integration is already present
- `specify integration use <agent>` (or equivalent) to change the default integration without uninstalling others

Conceptually, something like:

```json
{
  "default_integration": "claude",
  "installed_integrations": ["claude", "codex"]
}
```

### Alternatives Considered

- `specify integration switch <agent>` works for one person changing tools, but it does not solve shared multi-agent repos
- uninstalling and reinstalling integrations works as a workaround, but it is awkward and not a good fit for teams using different agents in the same project
- re-running `specify init` with another agent can leave artifacts on disk, but the supported CLI flow still behaves as single-install

### Component

### AI Agent (if applicable)

Claude Code

### Use Cases

1. A team shares one repository, but some developers use Claude and others use Codex.
2. A project wants to keep both agent-specific command/skill setups available without forcing users to switch the whole repo state.
3. A repo is maintained over time by different contributors who do not all use the same AI agent, but still want to share the same `.specify/` workflow.

### Acceptance Criteria

- [ ] It is possible to install more than one integration in the same project
- [ ] One integration can be marked as the default/current integration
- [ ] Installing a second integration does not require uninstalling the first one
- [ ] The project metadata explicitly represents multiple installed integrations
- [ ] `specify integration list` shows all installed integrations clearly
- [ ] Documentation explains the difference between installed integrations and the default integration

### Additional Context

Related issues:

- #377
- #1065
- #1924
- #2065

This request is not about replacing `switch`. `switch` is useful. The gap is support for repositories where multiple integrations need to coexist at the same time.
````

UmmeHabiba1312:
This PR introduces the ability to manage multiple AI integrations within a single project.
- Updated `save_init_options` to maintain a list of `installed_integrations`.
- Modified `init` to allow adding new agents to an existing project without conflict.
- Added `list-integrations` command to view all installed agents.
- Added `use` command to switch between active integrations.

Fixes #2257


mnriem:
Explain why this is needed? This will muddy the waters for any coding agent as now it has potentially N number of agents / commands and skills in the repository that it sees?

UmmeHabiba1312:
Thanks for the feedback @mnriem . This is primarily for Team Portability (#2257) allowing developers to use different tools (Claude, Cursor, etc.) on the same repo without constant re-installs.

To avoid 'muddying the waters,' we use Logical Isolation:

Directory Silos: Agents look for their own folders (e.g., .claude/ vs .cursor/) and don't interfere with each other similar to having both .vscode and .idea folders.

State Switch: The default_integration metadata acts as a source of truth to identify the active agent.

If needed, we can enhance the use command to automatically update .gitignore or .cursorignore to hide inactive agents from the active one's context.

Does this isolation strategy work for you?

mnriem:
You are saying they do not interfere, but you cannot guarantee that as LLMs will do what they do. So in order for me to accept this PR I request you to do the following:

1. Only allow installing more than one integration using the --force. If attempted without it indicate that installing multiple integration is not recommended
2. Merge list-integrations with list command
3. Update docs/reference/integrations.md to reflect the change. Also indicating it is NOT recommended

mnriem:
Superseded by #2389

PR #2273, opened 2026-04-19
fix(scripts): persist .specify/feature.json during feature creation

grafvonb:
## Description

The current `/speckit.specify` workflow expects `.specify/feature.json` to exist, but the shipped `create-new-feature` scripts do not persist it.

That creates a mismatch between documented behavior and actual script behavior:
- the command flow and downstream tooling expect `feature_directory` metadata
- feature creation creates the branch and spec directory
- but no `.specify/feature.json` file is written

As a result, later steps may rely on metadata that the standard feature-creation flow never produces.

This PR fixes that by persisting `.specify/feature.json` during feature creation in both the Bash and PowerShell scripts. It keeps the change intentionally small and aligns the shipped scripts with the workflow’s existing expectations.

## Testing

Additional verification:
- `pytest tests/test_timestamp_branches.py`

Results in this environment:
- `44 passed, 6 skipped`

The skipped tests are PowerShell runtime tests because `pwsh` is not available in this environment.

## AI Disclosure

Used Codex to inspect the existing `/speckit.specify` flow, implement the minimal script changes, and add/update regression tests.

### DIFF
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -372,4 +372,3 @@ except Exception:
     # Callers running under set -e should use: TEMPLATE=$(resolve_template ...) || true
     return 1
 }
-
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -324,6 +324,7 @@ fi
 
 FEATURE_DIR="$SPECS_DIR/$BRANCH_NAME"
 SPEC_FILE="$FEATURE_DIR/spec.md"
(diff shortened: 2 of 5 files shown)


mnriem (changes requested):
@grafvonb Can you please review your changes in light of the Git extension having moved to its own extension? I see the use of BRANCH_NAME? That is suspect as that is a Git specific thing.

grafvonb:
@mnriem you are right, but the problem exists also in the git extension. I created a new pr for that #2297.

PR #2276, opened 2026-04-20
fix(integrations): migrate Antigravity (agy) layout to .agents/ and deprecate --skills

baveku:
## Changes
- **Directory Migration:** Switched the default integration storage folder from `.agent/` to `.agents/`.
- **Auto-Migration:** Added a seamless `shutil.move` migration step in `AgyIntegration.setup()` to auto-rename `.agent/` to `.agents/` if the new directory doesn't exist yet, preventing breakage for legacy projects.
- **Backward Compatibility:** Retained `--skills` as a deprecated, no-op flag instead of removing it completely to ensure existing automation scripts do not break during option parsing.
- **Version Warning:** Added a setup warning indicating that the new `.agents/` layout requires Antigravity `v1.20.5` or newer.

## Testing

- [x] Verified the directory change in `agy/integrations/__init__.py`

## AI Disclosure

### DIFF
--- a/src/specify_cli/integrations/agy/__init__.py
+++ b/src/specify_cli/integrations/agy/__init__.py
@@ -1,13 +1,20 @@
 """Antigravity (agy) integration — skills-based agent.
 
-Antigravity uses ``.agent/skills/speckit-<name>/SKILL.md`` layout.
+Antigravity uses ``.agents/skills/speckit-<name>/SKILL.md`` layout.
 Explicit command support was deprecated in version 1.20.5;
-``--skills`` defaults to ``True``.
+``--skills`` defaults to ``True``. The ``.agents/`` path replaces ``.agent/`` starting in v1.19.5.
 """
 
 from __future__ import annotations
 
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
 from ..base import IntegrationOption, SkillsIntegration
+
+if TYPE_CHECKING:
+    from ..manifest import IntegrationManifest
+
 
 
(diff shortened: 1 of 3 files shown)


mnriem (changes requested):
Please address Copilot feedback. Also note that if this requires a newer version of agy installed you should warn users about this when the integration gets installed. Also please include a test when changing behavior

mnriem:
Please address Copilot support. BTW what is the reason for mentioning --skills? Can we get rid of that? Thoughts?

baveku:
> Please address Copilot support. BTW what is the reason for mentioning --skills? Can we get rid of that? Thoughts?

Good catch! Since explicit command support was deprecated in v1.20.5, the `.agents/skills/` layout is now the only supported behavior for the Antigravity integration. The `--skills` flag was essentially a redundant legacy option with no actual branching logic. 
I've completely removed it from `AgyIntegration`, cleaned up the outdated docstrings, and updated the test suite to reflect its removal.

mnriem:
Please address Test & Lint errors

mnriem:

mnriem:
Please change PR title / description to match intent of the PR as per Copilot feedback

baveku:
> Please change PR title / description to match intent of the PR as per Copilot feedback

I've updated the PR title and description to accurately reflect the changes. It now explicitly details the auto-migration step we added, the retention of the --skills flag for backward compatibility, and the new version warning. Thanks for the feedback!

mnriem (approved):

mnriem:
Thank you!

PR #2296, opened 2026-04-21
fix(agents): block directory traversal in command write paths (#2229)

### LINKED ISSUE #2229: security: alias path traversal in CommandRegistrar.register_commands() for flat agents
## Summary

Free-form aliases (permitted since #2125 / 4deb90f) can contain path separators that escape the agent commands directory when registered by flat-file agents (claude, gemini, copilot, etc.).

## Root cause

\`CommandRegistrar._compute_output_name()\` returns the alias name unchanged for non-SKILL.md agents:

\`\`\`python
if agent_config["extension"] != "/SKILL.md":
    return cmd_name  # alias returned as-is
\`\`\`

A manifest alias like \`../foo\` produces:

\`\`\`
commands_dir / "../foo.md"  →  escapes commands_dir
\`\`\`

\`alias_file.parent.mkdir(parents=True, exist_ok=True)\` will even create the parent directories.

## Not affected

SKILL.md agents (codex, kimi) are safe — \`_compute_output_name\` replaces \`.\` with \`-\` and prepends \`speckit-\`, neutralising traversal sequences.

## Suggested fix

Validate alias strings against a safe identifier pattern before writing, or resolve \`alias_file\` and assert it stays within \`commands_dir\` before writing.

## Discovery

Identified during review of PR #2027.

chordpli:
## Summary

Closes #2229.

Extends the alias containment guard from `b67b285` (Goose PR #2015) to the two remaining write paths in `CommandRegistrar` that derive filenames from free-form command or alias names:

- **Primary command write** in `register_commands()` — previously unguarded; a primary command name like `/absolute/evil` could escape `commands_dir` and attempt to create parent directories outside it.
- **`write_copilot_prompt()`** — companion `.prompt.md` writes under `.github/prompts/` previously had no containment check.

Both are now protected by a shared `_ensure_inside()` helper.

## Design decision

Per maintainer guidance on #2229, this uses a **lexical** containment check (`os.path.normpath` + `Path.is_relative_to`) rather than a resolved-path check. Path traversal via `..` or absolute paths is blocked at all three write sites, while intentionally symlinked sub-directories under an agent's commands directory (for example, `.claude/skills/shared -> /team/shared-skills`) remain supported to preserve backwards compatibility for existing extension setups.

This also aligns the existing alias guard from `b67b285` with the same lexical policy so all three write sites behave consistently.

## Tests

Added 22 regression cases in `tests/test_registrar_path_traversal.py`:

| Area | Cases | Notes |
|------|-------|-------|
| Primary command name (gemini, copilot) | 8 | new coverage |
| Alias name (gemini, copilot) | 8 | regression for `b67b285` |
| `write_copilot_prompt` direct | 4 | public `@staticmethod`, new coverage |
| Safe baseline | 1 | positive regression |
| Symlinked sub-dir preserved | 1 | backward-compatibility contract; skips when symlink creation is not permitted |

Payloads: `../pwned`, `../../etc/passwd`, `subdir/../../escape`, `/absolute/evil`.

Validated locally with:
- `uv run pytest -q tests/test_registrar_path_traversal.py`
- `uv run pytest -q tests/test_extensions.py -k 'copilot_companion_prompt_created or copilot_aliases_get_companion_prompts or register_commands_for_copilot'`
- `uvx ruff check src/specify_cli/agents.py tests/test_registrar_path_traversal.py`

## Out of scope

- Pre-planted symlink attacks (an attacker would already need write access under `commands_dir`) — not addressed by the lexical containment policy chosen for backwards compatibility.
- Primary command names coming from trusted `ExtensionManager.install_from_directory()` flows are still validated upstream by `EXTENSION_COMMAND_NAME_PATTERN`; this PR adds defense in depth for direct callers of the public `register_commands()` API.

### DIFF
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -6,6 +6,7 @@ Used by both the extension system and the preset system to write
 command files into agent-specific directories in the correct format.
 """
 
+import os
 from pathlib import Path
 from typing import Dict, List, Any
 
@@ -399,6 +400,28 @@ class CommandRegistrar:
 
         return f"speckit-{short_name}"
 
+    @staticmethod
+    def _ensure_inside(candidate: Path, base: Path) -> None:
+        """Validate that a write target stays within the expected base directory.
+
+        Uses lexical normalization so traversal via ``..`` or absolute paths is
+        rejected while intentionally symlinked sub-directories remain
+        supported.
+
+        Args:
+            candidate: Path that will be written.
+            base: Directory the write must remain within.
+
+        Raises:
+            ValueError: If the normalized candidate path escapes ``base``.
+        """
+        normalized = Path(os.path.normpath(candidate))
+        base_normalized = Path(os.path.normpath(base))
+        if not normalized.is_relative_to(base_normalized):
+            raise ValueError(
+                f"Output path {candidate!r} escapes directory {base!r}"
+            )
+
     def register_commands(
         self,
         agent_name: str,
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

Issue #1982, opened 2026-03-26
[Bug]: 0.4.2 removed "extension" command from specify


ChristianPresley:
### Bug Description

extension add fails with no command found:

```
specify extension add verify-tasks
Usage: specify [OPTIONS] COMMAND [ARGS]...
Try 'specify --help' for help.                   
Error │ No such command 'extension'. Did you mean 'version'
``` 

Help does not list "extension" in version 0.4.2

```
specify --help
GitHub Spec Kit - Spec-Driven Development Toolkit                                                                       
 Usage: specify [OPTIONS] COMMAND [ARGS]...                                                                                                                                                    
 Setup tool for Specify spec-driven development projects                                                                                                                                       
 Options 
│ --help          Show this message and exit.                                                                                                                                                 
 Commands 
│ init     Initialize a new Specify project from the latest template.
│ check    Check that all required tools are installed.
│ version  Display version and system information.
```

### Steps to Reproduce

1. Run `specify extension add <extension>`
2. Run `specify --help`
3. Run `specify extension --help`

### Expected Behavior

Extensions are installed or the command is available.

### Actual Behavior

Command not found.

### Specify CLI Version

0.4.2

### AI Agent

Not applicable

### Operating System

Ubuntu 24.04

### Python Version

Python 3.12.3

### Error Logs

```shell

```

mnriem:
I am unable to reproduce the issue

specify extension add verify-tasks
Error: 'verify-tasks' is available in the 'community' catalog but installation is not allowed from that catalog.

To enable installation, add 'verify-tasks' to an approved catalog (install_allowed: true) in .specify/extension-catalogs.yml.

mbachorik:
i have just reproduced:

michalbachorik@Michals-MacBook-Pro xxxx % uv tool install specify-cli --force
Resolved 20 packages in 338ms
Prepared 2 packages in 142ms
Installed 20 packages in 21ms
 + annotated-doc==0.0.4
 + anyio==4.13.0
 + certifi==2026.2.25
 + click==8.3.1
 + h11==0.16.0
 + httpcore==1.0.9
 + httpx==0.28.1
 + idna==3.11
 + markdown-it-py==4.0.0
 + mdurl==0.1.2
 + platformdirs==4.9.4
 + pygments==2.19.2
 + readchar==4.2.1
 + rich==14.3.3
 + shellingham==1.5.4
 + socksio==1.0.0
 + specify-cli==1.0.0
 + truststore==0.10.4
[...]

PR #2411, opened 2026-04-29
fix: include --from git+... in upgrade hint to avoid PyPI squat package

Quratulain-bilal:
## Summary

The upgrade hint surfaced by spec-kit'''s compatibility errors was missing `--from git+https://github.com/github/spec-kit.git`, so users who followed it landed on an unrelated PyPI package called `specify-cli` (no author, no project URLs) which ships a stub CLI that lacks `extension`, `preset`, and most spec-kit commands. This is the root cause behind #1982 ("0.4.2 removed extension command from specify").

Three call sites were affected:

- `src/specify_cli/extensions.py:1111` — `CompatibilityError` on incompatible extension
- `src/specify_cli/presets.py:579` — `PresetCompatibilityError` on incompatible preset
- `extensions/EXTENSION-DEVELOPMENT-GUIDE.md:672` — troubleshooting doc

`extensions.py` already defines `REINSTALL_COMMAND` (the canonical, full command). The fix reuses that constant in `extensions.py` and imports it from `presets.py` so the three sites can never drift again. The doc fix also adds a one-line note explaining the PyPI name collision so the same advice does not get re-stripped during a future copyedit.

## Why a PR (per maintainer request)

In #1982 @mnriem asked for a PR to make this clear:

> No idea so we will have to make sure our docs make it clear that one CANNOT use PyPi. Want to take on a PR for that?

The original reporter said they would bundle it with another fix but it has been ~30 days with no PR, so I picked it up.

## Test plan

- [x] `pytest tests/test_extensions.py tests/test_presets.py` — 434 passed
- [x] `python -c "from specify_cli.extensions import REINSTALL_COMMAND; print(REINSTALL_COMMAND)"` — prints the full GitHub-sourced command
- [x] `python -c "from specify_cli.presets import REINSTALL_COMMAND; print(REINSTALL_COMMAND)"` — same value, re-exported
- [ ] Manual: trigger an extension version mismatch and verify the error message includes `--from git+https://github.com/github/spec-kit.git`

Refs #1982

### DIFF
--- a/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
+++ b/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
@@ -669,7 +669,7 @@ hooks:
 
 **Error**: `Extension requires spec-kit >=0.2.0`
 
-- **Fix**: Update spec-kit with `uv tool install specify-cli --force`
+- **Fix**: Update spec-kit with `uv tool install specify-cli --force --from git+https://github.com/github/spec-kit.git`. The bare `specify-cli` package on PyPI is a different, unrelated project — installing it without `--from git+...` will give you a stub CLI that does not include `extension`, `preset`, or other spec-kit commands.
 
 **Error**: `Command file not found`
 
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #2428, opened 2026-05-01
chore(integrations): clean up docs and project guard

PascalThuet:
Closes #2427.

## What changed

- Removed AGENTS.md guidance for update-agent-context scripts that are no longer in the tree.
- Moved the existing spec-kit project guard helper before the integration commands and reused it for list/install/uninstall/switch/upgrade.
- Added a small regression test for those integration commands outside a spec-kit project.

## Validation

- uvx ruff check AGENTS.md src/specify_cli/__init__.py tests/integrations/test_cli.py
- uv run pytest tests/integrations/test_cli.py tests/integrations/test_integration_catalog.py -q
- uv run pytest -q

PR #2436, opened 2026-05-02
fix(workflows): require project for catalog list

PascalThuet:
Follow-up to #2428.

## What changed

- Make `specify workflow catalog list` use the shared spec-kit project guard.
- Extend the project guard regression test to cover that subcommand.

## Validation

- `uvx ruff check src/specify_cli/__init__.py tests/integrations/test_cli.py`
- `uv run pytest tests/integrations/test_cli.py -q`
- `uv run pytest tests/test_workflows.py -q`


mnriem (approved):

mnriem:
Thank you!

PR #2443, opened 2026-05-02
[Security hardening] Require explicit opt-in for shell workflow steps

### LINKED ISSUE #2440: [Security hardening] Require explicit opt-in for workflow shell steps
## Summary

Workflow `shell` steps currently execute local shell commands from workflow YAML. This is a powerful and useful capability, but catalog-installed or downloaded workflows should make that execution boundary explicit.

## Why

A workflow definition can contain arbitrary shell commands. Users should have a clear prompt or policy gate before a workflow with shell execution runs, especially when the workflow came from a catalog, URL, or third-party source.

## Proposed direction

- Add a workflow-level permission such as `requires.permissions.shell: true`.
- Reject or pause before running shell steps unless the workflow declares the capability and the user opts in.
- Surface the exact command or a summarized command list before execution.
- Keep local/development workflows ergonomic, but make downloaded/catalog workflows visibly executable.

## Acceptance criteria

- Workflows with `type: shell` require an explicit declaration or opt-in.
- Existing bundled workflows without shell steps continue to run unchanged.
- Tests cover shell steps with and without the permission declaration.
- Documentation explains that shell workflows execute local code with user privileges.

PascalThuet:
## Summary
- Require workflows with shell steps to declare requires.permissions.shell: true.
- Detect shell steps recursively across conditional, switch, loop, and fan-out nesting.
- Block direct execution of shell workflows without the permission and document the new opt-in.

## Security context
Shell workflow steps execute local commands via the shell. This makes that capability an explicit workflow permission instead of an implicit behavior.

Closes #2440

## Validation
- git diff --check
- uvx ruff check src/specify_cli/workflows/engine.py tests/test_workflows.py
- uv run pytest tests/test_workflows.py -q

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -193,6 +193,14 @@ steps:
       args: "{{ inputs.spec }}"
 ```
 
+Workflows that include any `type: shell` step must also declare:
+
+```yaml
+requires:
+  permissions:
+    shell: true
+```
+
 This produces the following execution flow:
 
 ```mermaid
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -47,8 +47,8 @@ class WorkflowDefinition:
         if not isinstance(self.default_options, dict):
             self.default_options = {}
 
-        # Requirements (declared but not yet enforced at runtime;
-        # enforcement is a planned enhancement)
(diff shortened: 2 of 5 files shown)


mnriem (changes requested):
See associated issue for my comments

PR #2481, opened 2026-05-07
fix: move branch validation after --paths-only exit

### LINKED ISSUE #2477: [Bug]: check-prerequisites.sh --paths-only still runs branch validation, contradicting --help
## Bug Description

`scripts/bash/check-prerequisites.sh --paths-only` is documented as "no validation" / "no prerequisite validation", but the script runs `check_feature_branch` **before** the `--paths-only` early exit. The result: when the current branch does not match the feature-branch naming convention, `--paths-only` exits non-zero with a branch-format error, even though the caller asked for paths only and explicitly opted out of validation.

Source: https://github.com/github/spec-kit/blob/main/scripts/bash/check-prerequisites.sh#L81-L113

```bash
# Get feature paths and validate branch
_paths_output=$(get_feature_paths) || { ... ; exit 1; }
eval "$_paths_output"
unset _paths_output
check_feature_branch "$CURRENT_BRANCH" "$HAS_GIT" || exit 1   # ← runs even with --paths-only

# If paths-only mode, output paths and exit (support JSON + paths-only combined)
if \$PATHS_ONLY; then
    ...
    exit 0
fi
```

The header documentation (line 14) says: \`--paths-only        Only output path variables (no validation)\` and \`--help\` (line 54) says \`(no prerequisite validation)\`. Branch validation is the most user-visible form of validation, so this contradicts the documented behavior.

## Steps to Reproduce

1. Initialize a Spec Kit project: \`specify init test-paths-only --integration claude\`
2. Switch to a non-feature branch: \`git checkout -b chore/some-cleanup\`
3. Run paths-only: \`./.specify/scripts/bash/check-prerequisites.sh --paths-only\`

## Expected Behavior

\`--paths-only\` succeeds and emits \`REPO_ROOT\`, \`BRANCH\`, \`FEATURE_DIR\`, etc., regardless of branch name (the caller asked for paths, not validation).

## Actual Behavior

Exit 1 with branch-validation error (the message produced by \`check_feature_branch\` in \`common.sh\`), no path variables emitted.

## Specify CLI Version

0.7.4.dev0

## AI Agent

Claude Code

## Suggested Fix

Move \`check_feature_branch\` after the \`if \$PATHS_ONLY; then ... exit 0; fi\` block so \`--paths-only\` truly skips validation. Same source ordering already exists in \`scripts/powershell/check-prerequisites.ps1\` — please verify both are aligned.

## Context

Surfaced while bulk-installing Spec Kit across 20 internal projects. \`/speckit-tasks\` and similar commands invoke \`check-prerequisites.sh --paths-only --json\` from non-feature chore branches and unexpectedly fail with branch-format errors despite the documented opt-out.

mnriem:
Superseded by #2672 which landed the same fix (paths-only skips branch validation). Closing.

### DIFF
--- a/scripts/bash/check-prerequisites.sh
+++ b/scripts/bash/check-prerequisites.sh
@@ -78,11 +78,10 @@ done
 SCRIPT_DIR="$(CDPATH="" cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/common.sh"
 
-# Get feature paths and validate branch
+# Get feature paths
 _paths_output=$(get_feature_paths) || { echo "ERROR: Failed to resolve feature paths" >&2; exit 1; }
 eval "$_paths_output"
 unset _paths_output
-check_feature_branch "$CURRENT_BRANCH" "$HAS_GIT" || exit 1
 
 # If paths-only mode, output paths and exit (support JSON + paths-only combined)
 if $PATHS_ONLY; then
@@ -111,6 +110,9 @@ if $PATHS_ONLY; then
     fi
     exit 0
 fi
+
+# Validate branch name
+check_feature_branch "$CURRENT_BRANCH" "$HAS_GIT" || exit 1
 
 # Validate required directories and files
 if [[ ! -d "$FEATURE_DIR" ]]; then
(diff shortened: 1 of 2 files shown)


mnriem:
@copilot apply changes based on the comments in [this thread](https://github.com/github/spec-kit/pull/2481#pullrequestreview-4246124586)

PR #2482, opened 2026-05-07
fix(kiro-cli): replace literal $ARGUMENTS with prose fallback

### LINKED ISSUE #1926: [Bug]: $ARGUMENTS placeholder not substituted in Kiro CLI — file-based prompts don't support arguments
## Summary

`$ARGUMENTS` is never substituted when SpecKit prompts are invoked via the **Kiro CLI** (`kiro-cli chat`). The literal string `$ARGUMENTS` is passed to the AI model, causing the agent to either error out or prompt the user to re-enter their input.

## Environment

- **AI Tool**: Kiro CLI (`kiro-cli chat`)
- **SpecKit version**: latest (github/spec-kit)
- **OS**: macOS

## Steps to Reproduce

1. Install SpecKit into a project (`.specify/` directory present)
2. Open `kiro-cli chat` in the project root
3. Invoke `/speckit.specify Add employee onboarding flow`
4. Observe the prompt fires but `$ARGUMENTS` is passed literally to the model

## Expected Behaviour

The text typed after `/speckit.specify` (i.e. `Add employee onboarding flow`) should be substituted for `$ARGUMENTS` in the prompt body before the model receives it.

## Actual Behaviour

The model receives the literal string `$ARGUMENTS` and responds:

> The `$ARGUMENTS` placeholder wasn't replaced — no feature description was provided.

## Root Cause

Kiro CLI's file-based prompt system **does not support argument substitution**. Per the [official Kiro documentation](https://kiro.dev/docs/cli/chat/manage-prompts/):

> "File-based prompts (local and global) do not support arguments."

Arguments are only supported for MCP server prompts (`@server-name/prompt-name <arg>`). Local prompts stored in `.kiro/prompts/` (or `.specify/` via symlink/copy) receive no argument injection mechanism.

## Impact

All SpecKit commands that rely on `$ARGUMENTS` substitution are broken on Kiro CLI:
- `/speckit.specify`
- Any other command that uses `$ARGUMENTS` in its prompt body

## Suggested Fix

One or more of the following:

1. **Document the limitation** — add a note to the SpecKit README / agent compatibility matrix that Kiro CLI does not support `$ARGUMENTS` in file-based prompts.
2. **Workaround guidance** — instruct Kiro users to type their feature description as a follow-up message immediately after invoking the prompt (the prompt instructions already hint at this: *"Assume you always have it available in this conversation even if `$ARGUMENTS` appears literally"*), but this is not obvious to users.
3. **MCP server prompt** — expose SpecKit commands as MCP prompts, which Kiro *does* support with arguments (`@speckit/specify "Add employee onboarding"`).
4. **Prompt self-detection** — add a preamble to affected prompts that detects the literal `$ARGUMENTS` string and instructs the model to ask the user for input gracefully rather than erroring.

## References

- [Kiro CLI — Manage Prompts](https://kiro.dev/docs/cli/chat/manage-prompts/#passing-arguments-to-mcp-prompts)

eldar702:
## Summary

Replace the literal `"$ARGUMENTS"` value in `KiroCliIntegration.registrar_config["args"]` with a prose fallback string so that rendered Kiro CLI prompt files no longer ship a broken shell-style placeholder.

Fixes #1926

## Problem

When SpecKit prompts are invoked via Kiro CLI (`kiro-cli chat`), the literal string `$ARGUMENTS` was passed to the AI model unchanged. Every parameterized SpecKit slash-command under Kiro CLI was effectively unusable — the agent either errored out or asked the user to repeat their input.

The integration code feeds `registrar_config["args"]` straight into `process_template()` (`base.py:717-719`), which then substitutes both `{ARGS}` and `$ARGUMENTS` tokens in command bodies with that value. Setting it to `"$ARGUMENTS"` made the substitution a no-op, so the literal token remained in every rendered prompt.

## Solution

Kiro CLI file-based prompts do **not** natively support any argument-substitution syntax — `$ARGUMENTS`, `{{userMessage}}`, and `$INPUT` are all delivered to the model verbatim (see [kirodotdev/Kiro#4141](https://github.com/kirodotdev/Kiro/issues/4141) and [kiro.dev/docs/cli/chat/manage-prompts](https://kiro.dev/docs/cli/chat/manage-prompts/)). Switching to a different shell-style token would just ship a different broken literal.

Instead, set `args` to a prose fallback that gets substituted into the rendered prompt and tells the model to take its argument from the user's next conversational message:

```python
_KIRO_ARG_FALLBACK = "(the user will provide the argument in this conversation)"
```

This is the approach used by other integrations whose target CLI lacks native argument injection.

## Test plan

- [x] New regression test `test_rendered_prompts_do_not_contain_raw_arguments` — calls `setup()`, walks every rendered `.kiro/prompts/*.md`, asserts `$ARGUMENTS` is **not** present. Fails on `main`, passes after the fix.
- [x] New regression test `test_rendered_prompts_contain_kiro_arg_placeholder` — asserts the new prose value **is** substituted in (proves substitution fired, not just that the placeholder was removed).
- [x] Inherited `test_registrar_config` from `MarkdownIntegrationTests` overridden in `TestKiroCliIntegration` because the base assertion `args == "$ARGUMENTS"` no longer holds for kiro-cli.
- [x] `pytest` full suite: **2788 passed, 34 skipped** — zero regressions.

## Notes

This change was AI-assisted. The fix was selected after a 3-agent solution-design debate in which two alternative approaches (`{{userMessage}}` and `$INPUT` placeholder swaps) were rejected after web research confirmed Kiro file-based prompts do not substitute any token; the prose-fallback was the only approach that actually solves the bug. The deliberation also produced the regression-test pair that proves both directions — `$ARGUMENTS` must be absent **and** the new value must be present — so we can't accidentally regress by changing the fallback to an empty string.


mnriem (changes requested):
This PR should also update `docs/reference/integrations.md` to document the argument limitation in the Kiro CLI notes column. Currently it only says `Alias: --integration kiro`. It should note that Kiro CLI file-based prompts [do not support argument substitution](https://kiro.dev/docs/cli/chat/manage-prompts/) and that `$ARGUMENTS` placeholders are replaced with a prose fallback at install time.

This is consistent with how other integrations document their quirks in that table (e.g., Goose noting YAML recipe format, Pi noting MCP limitations, skills-based integrations noting their layout).

eldar702:
Thanks for both reviews! Pushed `835de7d` addressing both findings.

### 1. Maintainer (mnriem) — docs

Updated the Kiro CLI row in `docs/reference/integrations.md` to lead with the limitation, following the Pi row precedent (limitation first, alias preserved at end):

> Kiro CLI does not substitute `$ARGUMENTS` in file-based prompts, so Spec Kit ships a prose fallback at render time (see [Manage prompts](https://kiro.dev/docs/cli/chat/manage-prompts/) and issue [#1926](https://github.com/github/spec-kit/issues/1926)). Alias: `--integration kiro`

### 2. Copilot — test (lines 27 + 56)

The previous `test_registrar_config` would have silently passed if `$ARGUMENTS` were swapped for `$INPUT`, `{{userMessage}}`, `<args>`, or any other unsubstituted token. Replaced with a **dual-layer regression guard**:

- **Layer 1 — `test_registrar_config_args_is_exact_prose_fallback`**: pins `args` to the imported `_KIRO_ARG_FALLBACK` constant. Wording drift now requires a deliberate paired commit (production constant + test).
- **Layer 2 — `test_registrar_config_args_does_not_look_like_a_placeholder_token`**: independent regression guard built on a 7-pattern regex set covering Bash (`$X`, `${X}`, `${X:-default}`), Mustache/Handlebars/Jinja (`{{X}}`, `{{{X}}}`), Liquid/Jinja control (`{% %}`), Python str.format / .NET (`{0}`, `{var}`), angle-bracket (`<X>`), and Windows (`%X%`). Anchored to the full string so legitimate prose mentioning a placeholder ("the `{{magic}}` of placeholders") is not flagged.

Also fixed the line-56 tautology — `test_rendered_prompts_contain_kiro_arg_placeholder` now imports `_KIRO_ARG_FALLBACK` directly instead of reading it back from `registrar_config["args"]`, so it stays independent of the field it should be guarding.

### Why two tests instead of one?

Different regression classes deserve different diagnostic messages:
- Layer 1 fails with `expected 'X', got 'Y'` — pinpoints accidental wording drift.
- Layer 2 fails with `value matches a known placeholder-token shape` — pinpoints the class of regression Copilot called out.

Either alone leaves a gap. Together they're ~25 extra lines.

### Validation

- `pytest tests/integrations/test_integration_kiro_cli.py -v` → 25/25 passed
- `pytest tests/` → 2790 passed, 34 skipped — no regressions
- Adversarial regex matrix: 8/8 prose accept (incl. the actual `_KIRO_ARG_FALLBACK` value), 15/15 placeholder shapes reject (`$ARGUMENTS`, `$INPUT`, `$args`, `${ARGS}`, `${ARGS:-default}`, `{{userMessage}}`, `{{ x | y }}`, `{{{var}}}`, `{% if x %}{{ x }}{% endif %}`, `<args>`, `%USERNAME%`, `{0}`, `{name}`, `{0:>5}`, `{name!r}`).

mnriem (approved):

mnriem:
Thank you!

PR #2502, opened 2026-05-08
fix(workflows): auto-detect project integration instead of hardcoding "copilot"

### LINKED ISSUE #2406: [Bug]: Workflow engine hardcodes "copilot" default, ignoring project's initialized integration (e.g., opencode)
### Bug Description

When running a workflow without explicitly passing the integration via `--input integration=<name>`, the workflow engine defaults to `copilot` (defined in `workflow.yml`). This overrides the project's actual initialized integration stored in `.specify/init-options.json` and `.specify/integration.json`. 

If the `copilot` CLI happens to be installed on the system, the workflow incorrectly dispatches the command to Copilot instead of the configured project AI (e.g., `opencode`). This results in a confusing `No such agent: speckit.specify` error, because the project only contains agent files for the initialized AI, not Copilot.

### Steps to Reproduce

1. Initialize a project exclusively with Opencode: `specify init --ai opencode`
2. Ensure the `copilot` CLI is coincidentally installed on your system (`/usr/bin/copilot`).
3. Run a workflow without an explicit integration input: 
   `specify workflow run speckit --input spec="..."`
4. The workflow fails at the first step (`speckit.specify`).

### Expected Behavior

The workflow engine should auto-detect the project's integration from `.specify/integration.json` (or `.specify/init-options.json`) and use it as the default. If I initialized the project with `opencode`, the workflow should dispatch to `opencode`.

### Actual Behavior

The run fails with: `"No such agent: speckit.specify"`.

**The Failure Path:**
1. The `workflow.yml` (lines 22-25) sets the default input for integration to `"copilot"`.
2. The `CommandStep` resolves the integration input, choosing `"copilot"` over the project default.
3. `CopilotIntegration.dispatch_command()` attempts to run: `copilot -p "<spec>" --agent speckit.specify`.
4. Because `/usr/bin/copilot` exists, `shutil.which("copilot")` succeeds and dispatch proceeds.
5. Copilot fails because the commands live at `.opencode/command/speckit.*.md`, **not** `.github/agents/speckit.*.agent.md`.

### Specify CLI Version

0.7.4.dev0

### AI Agent

Claude Code

### Operating System

Ubuntu 24.04.4 LTS

### Python Version

3.12.3

### Error Logs

```shell

```

### Diagnostic Information & Open Questions
If we fix the workflow to correctly dispatch to `opencode` (or pass `--input integration=opencode` manually), we might hit a secondary bug. 

Currently, `OpencodeIntegration` delegates to `MarkdownIntegration.build_exec_args()`, which generates the CLI command: `["opencode", "-p", prompt]`. 
It needs to be verified if the `opencode` CLI actually supports the `-p` shorthand in non-interactive mode. Standard `opencode` usage typically expects `opencode run <message>` or `opencode --prompt <message>`. 
*(Diagnostic commands to verify this internally: `opencode --prompt "test"` vs `opencode run "test"`).*

### Suggested Fixes & Quality Improvements
1. **Dynamic Default Resolution:** The workflow YAML schema should support `integration: auto`, telling the engine to read `.specify/integration.json` instead of hardcoding `"copilot"`.
2. **Workflow Pre-flight Check:** When `specify workflow run` is invoked, it should warn or abort if the workflow's resolved integration mismatches the project's initialization state.
3. **Opencode Integration Test:** Add a contract test specifically for `OpencodeIntegration.build_exec_args()` to ensure the generated CLI flags (like `-p`) are actually supported by the `opencode` CLI in headless mode.
4. **Better CLI Error Bubbling:** When `CommandStep._try_dispatch()` fails, surface the actual `stderr` from the CLI tool rather than dropping it.

markuswondrak:
## Description

Fixes the hardcoded `"copilot"` default in `workflow.yml` that caused the workflow engine to dispatch to Copilot even when the project was initialized with a different integration (e.g. opencode).

Introduces an `"auto"` sentinel for the `integration` input: when `auto` is the resolved value, the engine reads `.specify/integration.json` to determine the active integration at runtime.

**Changes:**
- `integration_state.py` — new `read_integration_state()` function with typed error classes (`IntegrationStateError`, `IntegrationStateSchemaError`)
- `__init__.py` — `_read_integration_json()` now delegates to `read_integration_state()` instead of inlining the logic
- `workflows/engine.py` — `_resolve_default()` / `_resolve_workflow_integration()` resolve the `"auto"` sentinel at runtime
- `workflow.yml` — default changed from `"copilot"` to `"auto"`; added `opencode` to `requires.integrations.any`
- Tests extended for state reader and engine auto-detection

Fixes #2406

## Testing

## AI Disclosure

GitHub Copilot assisted with code generation and test scaffolding.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -61,13 +61,14 @@ from .integration_runtime import (
 )
 from .integration_state import (
     INTEGRATION_JSON,
-    INTEGRATION_STATE_SCHEMA,
+    IntegrationStateError as _IntegrationStateError,
+    IntegrationStateSchemaError as _IntegrationStateSchemaError,
     dedupe_integration_keys as _dedupe_integration_keys,
     default_integration_key as _default_integration_key,
     installed_integration_keys as _installed_integration_keys,
     integration_setting as _integration_setting,
     integration_settings as _integration_settings,
-    normalize_integration_state as _normalize_integration_state,
+    read_integration_state as _read_integration_state,
     write_integration_json as _write_integration_json_file,
 )
 from .shared_infra import (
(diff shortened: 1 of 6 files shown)


mnriem (changes requested):

PR #2543, opened 2026-05-13
refactor: extract _assets.py and _utils.py from __init__.py (PR-2/8)

darion-yaphet:
## Summary

- Extract bundle path resolution and version lookup into `_assets.py` (stdlib only, zero internal imports)
- Extract system utilities (subprocess, tool detection, file operations) into `_utils.py` (imports only from `._console`)
- Re-export all moved symbols from `__init__.py` for full backward compatibility
- Update `test_check_tool.py` to patch both `specify_cli` and `specify_cli._utils` namespaces since constants are now defined in `_utils`

## Files Changed

| File | Change |
|------|--------|
| `src/specify_cli/_assets.py` | New — 115 lines, bundle path & version helpers |
| `src/specify_cli/_utils.py` | New — 281 lines, subprocess & tool detection utilities |
| `src/specify_cli/__init__.py` | Shrunk by ~380 lines, re-exports preserved |
| `tests/test_utils_assets_imports.py` | New — import regression guard |
| `tests/test_check_tool.py` | Patch both namespaces for moved constants |

## Test plan

- [ ] `pytest tests/test_utils_assets_imports.py` — all symbols importable from `specify_cli`
- [ ] `pytest tests/test_check_tool.py` — patch paths correct after move
- [ ] `pytest` (full suite) — no regressions

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -33,8 +33,6 @@ import zipfile
 import tempfile
 import shutil
 import json
-import json5
-import stat
 import shlex
 import urllib.error
 import urllib.request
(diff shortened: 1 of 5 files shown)


mnriem (approved):

mnriem:
Thank you!

PR #2571, opened 2026-05-15
Fix paths-only validation in check-prerequisites

### LINKED ISSUE #2477: [Bug]: check-prerequisites.sh --paths-only still runs branch validation, contradicting --help

prantikmedhi:
Fixes #2477.

- Move branch validation after the paths-only early exit in bash and PowerShell
- Add a regression test for paths-only on a non-feature branch

Validated with:
- bash scripts/bash/check-prerequisites.sh --json --paths-only
- bash scripts/bash/check-prerequisites.sh --json (still fails on main as expected)

### DIFF
--- a/scripts/bash/check-prerequisites.sh
+++ b/scripts/bash/check-prerequisites.sh
@@ -78,11 +78,10 @@ done
 SCRIPT_DIR="$(CDPATH="" cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/common.sh"
 
-# Get feature paths and validate branch
+# Get feature paths
 _paths_output=$(get_feature_paths) || { echo "ERROR: Failed to resolve feature paths" >&2; exit 1; }
 eval "$_paths_output"
 unset _paths_output
-check_feature_branch "$CURRENT_BRANCH" "$HAS_GIT" || exit 1
 
 # If paths-only mode, output paths and exit (support JSON + paths-only combined)
 if $PATHS_ONLY; then
@@ -111,6 +110,8 @@ if $PATHS_ONLY; then
     fi
     exit 0
 fi
+
+check_feature_branch "$CURRENT_BRANCH" "$HAS_GIT" || exit 1
 
 # Validate required directories and files
 if [[ ! -d "$FEATURE_DIR" ]]; then
--- a/scripts/powershell/check-prerequisites.ps1
+++ b/scripts/powershell/check-prerequisites.ps1
@@ -56,12 +56,8 @@ EXAMPLES:
 # Source common functions
 . "$PSScriptRoot/common.ps1"
 
-# Get feature paths and validate branch
+# Get feature paths
 $paths = Get-FeaturePathsEnv
-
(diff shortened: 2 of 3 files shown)


mnriem:

prantikmedhi:
Addressed the Copilot feedback in the regression test fixture:

- initialize the temp repo directly on `main` and create an initial empty commit so branch detection is stable
- strip `SPECIFY_FEATURE` and `SPECIFY_FEATURE_DIRECTORY` from the Bash/PowerShell test subprocess environments

Validation: `uv sync --extra test && uv run python -m pytest tests/test_check_prerequisites_paths_only.py` (1 passed, 1 skipped because `pwsh` is not available locally).

mnriem (changes requested):

PR #2574, opened 2026-05-15
feat: add self-check tip to check output

### LINKED ISSUE #2573: Add self-check tip to specify check
Follow-up to #2498, #2519, #2548, and #2549.

## Problem

`specify check` verifies local tools, but users may also need to know whether their installed `specify` CLI is outdated. This matters because some integration features depend on the CLI version. If a user has an old local CLI, a feature can look unsupported even though it exists in newer releases.

Users can run `specify self check` today, but `specify check` is the command they are likely to use when diagnosing their setup.

## Revised scope

Based on maintainer feedback, keep `specify check` as an offline environment scan. Do not add `specify check --latest` or any network lookup to this command.

Instead, add a discoverability tip to `specify check` output:

```text
Tip: Run 'specify self check' to verify you have the latest CLI version.
```

This keeps version lookup responsibility in `specify self check` while making it easier to discover from the normal setup check.

## Tests

Add tests for:

- `specify check` prints the `specify self check` tip;
- `specify check` does not call the latest-release lookup.

Suggested command:

```bash
pytest tests/test_check_tool.py tests/test_upgrade.py -v
```

PascalThuet:
## Summary

Closes #2573.

Keeps `specify check` as an offline environment scan and adds a discoverability tip for the existing `specify self check` version lookup.

## Changes

- Add a `specify self check` tip to `specify check` output.
- Keep `specify check` network-free; no `--latest` option is added.
- Document the offline/version-check split in the core command reference and upgrade troubleshooting.
- Add tests that assert the tip is shown and that `specify check` does not call the latest-release lookup.

## Validation

Local:

- `uv run pytest tests/test_check_tool.py tests/test_upgrade.py -q`
- `uvx ruff check src/ tests/test_check_tool.py tests/test_upgrade.py`
- `git diff --check origin/main`
- `uv run specify check --help`

CI:

- GitHub Actions will run the full PR validation suite.

## AI Disclosure

Used AI assistance to inspect the existing check/self-check flow, adjust the PR based on review feedback, add tests/docs, and run local validation. The changes were reviewed before submission.

### DIFF
--- a/docs/reference/core.md
+++ b/docs/reference/core.md
@@ -69,6 +69,8 @@ specify check
 
 Checks that required tools are available on your system: `git` and any CLI-based AI coding agents. IDE-based agents are skipped since they don't require a CLI tool.
 
+This command stays offline. If a command behaves like an older Spec Kit version or an expected CLI feature is missing, run `specify self check` to check whether your local CLI is behind the latest release.
+
 ## Version Information
 
 ```bash
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -388,6 +388,14 @@ Only Spec Kit infrastructure files:
 
 ### "CLI upgrade doesn't seem to work"
 
+If a command behaves like an older Spec Kit version, first check for local CLI drift:
+
+```bash
+specify self check
+```
+
+`specify check` is an offline environment scan; `specify self check` is the CLI version lookup.
+
 Verify the installation:
 
 ```bash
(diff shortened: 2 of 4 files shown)


mnriem:
Thanks for the work here! We left a suggestion on #2573 — `specify check` is an offline environment scan and `specify self check` already owns the version-lookup logic. Rather than adding `--latest` with network code, we think a simpler tip line in `specify check` output would solve the discoverability problem without blurring the two commands' responsibilities:

```
Tip: Run 'specify self check' to verify you have the latest CLI version.
```

Would you be open to adjusting the PR in that direction?

PascalThuet:
Updated in `e886a15` based on the suggestion: removed the `--latest` approach, kept `specify check` network-free, added the `specify self check` tip, and updated docs/tests. Local validation passes: targeted pytest, ruff, and diff whitespace check.

mnriem:
Thank you!

PR #2582, opened 2026-05-15
fix: sanitize extension name in download path to prevent path traversal

mnriem:
## Summary

Fixes the path traversal vulnerability

The `extension` argument in `specify extension add <name> --from <url>` was interpolated directly into the temporary ZIP download path:

```python
zip_path = download_dir / f"{extension}-url-download.zip"
```

Absolute paths or `../` segments in the extension argument could escape `.specify/extensions/.cache/downloads/`, causing the CLI to write downloaded bytes to an attacker-chosen local path and then delete that path in cleanup.

## Fix

1. **`Path(extension).name`** strips all directory components (handles `../` traversal and absolute paths)
2. **Residual separator replacement** (`/` and `\` → `_`) as a belt-and-suspenders measure
3. **`resolve().relative_to()` containment guard** rejects anything that still escapes — consistent with the existing Zip Slip guard pattern in `extensions.py`

The fix is applied before any I/O, so both the write and delete vectors are eliminated.

## Tests

New test file `tests/test_extension_add_path_traversal.py`:
- 4 parametrized traversal payloads (`../pwned`, `../../etc/passwd`, `subdir/../../escape`, `/tmp/evil`)
- 1 clean-name regression test (normal names are unaffected)

All existing path traversal and extension tests continue to pass.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -3646,19 +3646,29 @@ def extension_add(
                 # (consistent with shared_infra._ensure_safe_shared_directory):
                 # `mkdir(parents=True)` would silently traverse a symlinked ancestor,
                 # so each component is checked and created individually instead.
+                # Each existing component is also re-resolved and required to land
+                # under project_root, which catches non-symlink directory aliases
+                # (e.g. Windows junctions / mount points) that resolve outside.
                 import uuid as _uuid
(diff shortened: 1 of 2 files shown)


mnriem:
@copilot apply changes based on the comments in [this thread](https://github.com/github/spec-kit/pull/2582#pullrequestreview-4300610733)

PR #2589, opened 2026-05-15
fix(catalogs): reject boolean priority in extension and preset catalo…

Quratulain-bilal:
## What

  The extension- and preset-catalog config readers silently accept a boolean `priority` value. Because `bool` is a
  subclass of `int` in Python, a config like:

  ```yaml
  catalogs:
    - name: mine
      url: https://example.com/catalog.json
      priority: yes      # PyYAML parses this to True
  ```

  is coerced via `int(True) == 1` and quietly treated as a valid priority of 1 — silently reordering the catalog stack
  instead of raising the same `Invalid priority` error that a typo like `priority: "fast"` already raises.

  The sibling integration-catalog reader in `src/specify_cli/catalogs.py` already guards this case (see `catalogs.py:137`,
   plus the existing `test_load_catalog_config_rejects_boolean_priority` in
  `tests/integrations/test_integration_catalog.py`). The extension and preset readers had drifted from that precedent.

  ## Why this matters

  - `priority: yes` / `priority: no` / `priority: true` are easy YAML typos — silently accepting them changes catalog
  resolution order without any warning, which is a frustrating bug to diagnose because the config "looks valid" and no
  error fires.
  - The three catalog readers (`catalogs.py`, `extensions.py`, `presets.py`) validate the same shape and should fail in
  the same way. Today, the same malformed input raises in one and is silently accepted in two.

  ## The change

  - `src/specify_cli/extensions.py`  add `isinstance(raw_priority, bool)` guard before `int(...)` in
  `ExtensionCatalog._load_catalog_config`, raising `ValidationError` with the same message shape used for other invalid
  priorities.
  - `src/specify_cli/presets.py`  same guard in `PresetCatalog._load_catalog_config`, raising `PresetValidationError`.
  - Both call sites mirror the existing `catalogs.py:137` pattern exactly, with a short comment explaining the
  `bool`-is-`int` subtlety so the rationale is visible to the next reader.

  ## Tests

  - `tests/test_extensions.py` — `test_load_catalog_config_rejects_boolean_priority`
  - `tests/test_presets.py` — `test_load_catalog_config_rejects_boolean_priority`

  Both follow the template of the existing integration-catalog regression test, so the three catalog validators now have
  parallel coverage.

  ```
  $ python -m pytest tests/test_extensions.py::TestCatalogStack tests/test_presets.py::TestPresetCatalog -q
  38 passed in 2.44s
  $ python -m ruff check src/
  All checks passed!
  ```

  ## Scope

  Intentionally small: no behaviour change for any priority that was previously accepted (real integers, numeric strings,
  missing values all behave exactly as before). Only `priority: true` / `priority: false`  which were almost certainly
  never intended as priority 1 / priority 0  now raise a clear error.

### DIFF
--- a/src/specify_cli/extensions.py
+++ b/src/specify_cli/extensions.py
@@ -1768,12 +1768,24 @@ class ExtensionCatalog:
                 skipped_entries.append(idx)
                 continue
             self._validate_catalog_url(url)
+            raw_priority = item.get("priority", idx + 1)
+            # Reject bools explicitly: ``bool`` is a subclass of ``int`` so
+            # ``int(True)`` silently returns 1, which would let a YAML
+            # ``priority: true`` slip through as a valid priority of 1. The
+            # sibling integration-catalog reader in ``catalogs.py`` already
+            # guards this; mirror the check here so the three catalog
+            # validators stay consistent.
+            if isinstance(raw_priority, bool):
+                raise ValidationError(
+                    f"Invalid priority for catalog '{item.get('name', idx + 1)}': "
+                    f"expected integer, got {raw_priority!r}"
+                )
             try:
-                priority = int(item.get("priority", idx + 1))
+                priority = int(raw_priority)
             except (TypeError, ValueError):
                 raise ValidationError(
                     f"Invalid priority for catalog '{item.get('name', idx + 1)}': "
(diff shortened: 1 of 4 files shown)


mnriem (approved):

mnriem:
Thank you!

PR #2615, opened 2026-05-18
refactor: create commands/ package and move init handler (PR-4/8)

darion-yaphet:
Description

  Part 4 of 8 in the __init__.py module split refactor.

  Creates the commands/ subpackage and moves the init command handler (~670 lines) out of __init__.py into commands/init.py, using a register(app) pattern to avoid
  circular imports.

  New modules:
  - _agent_config.py — extracts AGENT_CONFIG, AI_ASSISTANT_HELP, AI_ASSISTANT_ALIASES, SCRIPT_TYPE_CHOICES, DEFAULT_INIT_INTEGRATION as a base-layer module (no internal
   imports), breaking the circular dependency that would otherwise occur when commands/init.py imports from __init__.py
  - commands/__init__.py — package marker
  - commands/init.py — init command via register(app) pattern; dependencies still in __init__.py are resolved via lazy imports inside the command body
  - commands/{integration,preset,extension,workflow}.py — placeholder stubs for future PRs

  Backward compatibility: All re-exported symbols (AGENT_CONFIG, AI_ASSISTANT_HELP, SCRIPT_TYPE_CHOICES, select_with_arrows, _locate_bundled_workflow) remain importable
   from specify_cli.

  __init__.py reduction: ~780 lines removed.

  Testing

  - Tested locally with uv run specify --help
  - Ran existing tests with uv sync && uv run pytest
  - Tested with a sample project (if applicable)

  uv run pytest: 2910 passed, 35 skipped. Fixed one test whose mock.patch target moved from specify_cli._stdin_is_interactive →
  specify_cli.commands.init._stdin_is_interactive.

  AI Disclosure

  - I did use AI assistance (describe below)

  Code structure and implementation generated with Claude Code; reviewed and validated manually.


mnriem:
Note for each of the subcommands I would probably like to have the commands co-located with their actual subdirectory where the rest of the functionality lives. E.g integrations in `integrations` directory.

darion-yaphet:
I will fix the issue of moving to the subdirectory later.

mnriem:
So not sure if we should merge this as you are going to have to move each of the sub commands into their corresponding directory anyway? Thoughts?

darion-yaphet:
> So not sure if we should merge this as you are going to have to move each of the sub commands into their corresponding directory anyway? Thoughts?

I think I didn't explain it clearly enough.
In this PR, I will move the commands to a subdirectory. But I really don't have time today. I will make the changes later.

darion-yaphet:
PR-4 lays the architectural foundation. It introduces the commands/ package and moves the init command (~670 lines) from __init__.py into commands/init.py. As part of
   this, agent configuration constants (AGENT_CONFIG, AI_ASSISTANT_HELP, etc.) are extracted into a dedicated _agent_config.py module to eliminate the circular import
  that would otherwise arise in later PRs. This PR also establishes the migration pattern that all subsequent PRs follow: each command module exposes a register(app)
  function called by __init__.py at startup, and any shared helpers still living in __init__.py are accessed via lazy imports inside command function bodies to keep the
   dependency graph acyclic.

  PRs 5 through 8 are straightforward applications of that same pattern, relocating the remaining four command groups one PR at a time:

  - PR-5 moves the 11 integration subcommands (list, install, use, uninstall, switch, upgrade, search, info, and the three catalog subcommands) into
  integrations/_commands.py (~1,200 lines)
  - PR-6 moves the 12 preset subcommands into presets/_commands.py (~730 lines)
  - PR-7 moves the 10 extension subcommands into extensions/_commands.py (~1,250 lines)
  - PR-8 moves the 9 workflow subcommands into workflows/_commands.py (~625 lines)

  Once all four PRs land, __init__.py is reduced from ~4,500 lines to a ~500-line registration layer — little more than imports and register() calls.

  A few things I'd love your input on:

  - Co-location convention — we're placing _commands.py alongside the domain code rather than keeping everything under a central commands/ directory. Does this feel
  intuitive when navigating the codebase, or would you prefer a different layout?
  - The register(app) pattern — each command module wires itself into the Typer app via a single register() call. Any concerns about discoverability or testability with
   this approach?
  - PR granularity — we're splitting this into one PR per command group so each diff stays reviewable. Would you rather see fewer, larger PRs, or is this cadence
  working for you?
  - Anything I might have missed — if you spot edge cases, naming issues, or patterns that conflict with other parts of the codebase, please flag them early.

  Happy to walk through any part of this in more detail.

mnriem:
Thank you and yes please go ahead!

PR #2631, opened 2026-05-19
fix(cursor-agent): enable headless CLI dispatch end-to-end (-p --trust --approve-mcps --force + Windows .cmd shim resolution)

### LINKED ISSUE #2629: [Feature]: Enable CLI dispatch for cursor-agent integration so 'specify workflow run' can drive Cursor headlessly
### Summary

`specify workflow run` should be able to drive the `cursor-agent` CLI headlessly — the same way it already drives `claude`, `codex`, `gemini`, etc. — so that the full speckit workflow (`specify` → `plan` → `tasks` → `implement`) can run end-to-end against Cursor from the command line, in addition to the existing in-IDE skill flow.

Today this fails: the `cursor-agent` integration is configured with `requires_cli: False` and does not override `build_exec_args`, so the workflow dispatcher returns `None` and surfaces a misleading `"CLI not found or not installed"` error even when the CLI is installed and working.

A focused PR implementing this is open as **#2631** (https://github.com/github/spec-kit/pull/2631).

### Problem

When running:

```powershell
specify workflow run speckit `
  --input "spec=some feature description" `
  --input "integration=cursor-agent" `
  --input "scope=full"
```

The first step fails with:

```
Cannot dispatch command 'speckit.specify': integration 'cursor-agent' CLI not found or not installed.
Install the CLI tool or check 'specify integration list'.
```

But:

1. `cursor-agent` CLI is installed and on PATH (`cursor-agent --version` → `2026.05.16-0338208`, executable at `C:\Users\eman\AppData\Local\cursor-agent\cursor-agent.cmd`).
2. The CLI fully supports non-interactive / headless execution — `cursor-agent --help` shows `-p / --print` ("Has access to all tools, including write and shell"), `--output-format text|json|stream-json`, `--model`, `--workspace`, and `--trust` (mandatory in headless contexts).
3. `cursor-agent -p --trust --output-format text "Say only the word OK"` returns `OK` with exit code 0 in ~23 s on this machine.

### Root cause

`src/specify_cli/integrations/cursor_agent/__init__.py` declares:

```python
config = {
    ...,
    "install_url": None,
    "requires_cli": False,
}
```

and does not override `build_exec_args`. In `src/specify_cli/workflows/steps/command/__init__.py`, the dispatcher's `_try_dispatch` checks `impl.build_exec_args("test") is None` (line ~130) and returns `None` for cursor-agent, which the caller surfaces as the generic `"CLI not found or not installed"` message — even though the actual CLI is healthy.

In `src/specify_cli/integrations/base.py`, `SkillsIntegration.build_exec_args` only emits args when `config["requires_cli"]` is truthy, so flipping the flag alone is not quite enough — `cursor-agent` requires `--trust` for headless execution (without it the CLI exits non-zero with a "Workspace Trust Required" prompt), which the default `[<key>, -p, <prompt>]` template does not provide.

### Expected behavior

`specify workflow run` should dispatch cursor-agent the same way it dispatches `claude`:

```
[cursor-agent, -p, --trust, /speckit-<name> <args>, --output-format json]
```

or, for streamed text output:

```
[cursor-agent, -p, --trust, /speckit-<name> <args>]
```

That argv shape is verified working against the current cursor-agent CLI release (`2026.05.16-0338208`).

### Proposed fix (single-file, ~30 LOC)

In `src/specify_cli/integrations/cursor_agent/__init__.py`:

1. Set `config["requires_cli"] = True`.
2. Point `config["install_url"]` at `https://docs.cursor.com/en/cli/overview`.
3. Override `build_exec_args` to inject `--trust` after `-p`:

   ```python
   def build_exec_args(self, prompt, *, model=None, output_json=True):
       if not self.config or not self.config.get("requires_cli"):
           return None
       args = [self.key, "-p", "--trust", prompt]
       if model:
           args.extend(["--model", model])
       if output_json:
           args.extend(["--output-format", "json"])
       return args
   ```

This is a strict addition: the existing in-IDE skill flow (via `.cursor/skills/speckit-*/SKILL.md`) continues to work unchanged, and the 34 existing `test_integration_cursor_agent.py` tests still pass.

A PR with the change plus 6 new tests that pin the argv shape, `requires_cli`, `install_url`, and the hyphenated `/speckit-<name>` skill invocation form is in flight.

### Environment

- OS: Windows 10 (10.0.26200), PowerShell 7
- spec-kit (specify CLI): `0.8.7` and `0.8.11` (both reproduce; `specify self upgrade` did not fix)
- cursor-agent: `2026.05.16-0338208`
- Workflow: bundled `speckit` (Full SDD Cycle v1.0.0)
- Python: 3.12.13 (under the `specify-cli` `uv tool` venv)

### Why this matters

Cursor is one of the most prominent agent options users select with `specify init --integration cursor-agent` (auto-promoted from `--ai cursor-agent`). Today, *every* such user who tries `specify workflow run` hits a dead end with an error message that tells them to "install the CLI" — which is wrong (the CLI is already installed) and doesn't hint at the underlying integration-level limitation.

Enabling CLI dispatch closes the loop, makes the bundled `speckit` workflow usable end-to-end with Cursor from the terminal (mirroring the claude/codex/gemini story), and removes the misleading error.

### Background context

PR #2156 (`feat(cursor-agent): migrate commands to skills under .cursor/skills`) intentionally moved cursor-agent commands into `.cursor/skills/` as IDE skills, which is the right call for in-IDE usage. This change does not undo that — the SKILL.md install path is preserved exactly. It only adds a CLI dispatch path on top, so the same integration can be driven from inside the IDE **or** from `specify workflow run`, depending on what the user wants.

One-TheOnly:
﻿## Summary

Enable `specify workflow run` to dispatch the `cursor-agent` CLI headlessly end-to-end on Windows, matching the way it already dispatches `claude` / `codex` / `gemini`. Existing in-IDE skill flow (via `.cursor/skills/speckit-*/SKILL.md`) is preserved exactly — this PR only adds the CLI dispatch path on top.

Fixes #2629.

**Integration config (final):** `requires_cli` remains **`False`** so `specify init --ai cursor-agent` works without the CLI installed (IDE/skills flow unchanged). Workflow dispatch is enabled by overriding `build_exec_args()` — the same pattern as `CopilotIntegration` — with CLI presence enforced at dispatch time via `shutil.which`, not as a hard `specify init` precheck.

## Problem

Before this PR, `specify workflow run speckit --input integration=cursor-agent ...` fails immediately with:

```
Cannot dispatch command 'speckit.specify': integration 'cursor-agent' CLI not found or not installed.
Install the CLI tool or check 'specify integration list'.
```

even when `cursor-agent` is installed and healthy (`cursor-agent --version` returns `2026.05.16-0338208`).

Three separate gaps each independently block the headless path; this PR fixes all three.

#### 1) `1c55988 fix(cursor-agent): enable CLI dispatch via -p --trust headless mode`

`CursorAgentIntegration` had `requires_cli: False` and did not override `build_exec_args`, so `SkillsIntegration.build_exec_args` returned `None`, the dispatcher in `workflows/steps/command/__init__.py::_try_dispatch` returned `None`, and the misleading CLI-missing error was surfaced. The CLI is actually fully headless-capable:

- `-p / --print` — print mode, "Has access to all tools, including write and shell"
- `--output-format text | json | stream-json`
- `--model <name>`
- `--trust` — "Trust the current workspace without prompting (only works with --print/headless mode)"

Without `--trust`, the CLI exits non-zero with `⚠ Workspace Trust Required`.

This commit (superseded on `requires_cli` by `ff07cc0` below — kept here for history):

- `config["install_url"]`: `None` → `https://docs.cursor.com/en/cli/overview`
- First override of `build_exec_args` to emit `[self.key, "-p", "--trust", prompt, …]` (later commits add `--approve-mcps --force`; `ff07cc0` sets `requires_cli` back to `False` for IDE-first init).

#### 2) `454d3d0 fix(integrations): resolve .cmd/.bat shims before subprocess.run`

After commit (1), the dispatcher still failed on Windows with `FileNotFoundError [WinError 2]`. Root cause:

- `shutil.which("cursor-agent")` returns `cursor-agent.CMD` (honoring `PATHEXT`).
- `subprocess.run(["cursor-agent", ...])` calls `CreateProcess`, which does **not** consult `PATHEXT`, so the bare command name fails to launch.

The same bug affects every `.cmd`/`.bat`-wrapped CLI on Windows (verified for `codex` as well; only `.exe` CLIs like `claude` were unaffected).

Fix in `IntegrationBase.dispatch_command`: resolve `exec_args[0]` via `shutil.which` before `subprocess.run`. No-op on POSIX (no `PATHEXT` to consult); transparent fix on Windows.

#### 3) `595bcab fix(cursor-agent): inject --approve-mcps --force for headless MCP/tool access`

After commits (1)+(2), `subprocess.run` succeeds and cursor-agent starts, but the first MCP call (e.g. to read a DingTalk PRD) fails with `The tool call was rejected. I can't retrieve the document name without permission to call the tool.` Two more flags are needed:

- `--approve-mcps` — without it, every MCP server stays `not loaded (needs approval)` and tool calls are silently dropped.
- `--force` — without it, the agent halts on the first tool-call approval prompt; the workflow returns a misleading non-zero exit. With it, cursor-agent matches the implicit "trusted environment" semantics that `claude -p` and `codex --exec` already have by default — the right semantics for an unattended `specify workflow run`.

Final argv after all three commits:

```python
args = [self.key, "-p", "--trust", "--approve-mcps", "--force", prompt]
if model: args.extend(["--model", model])
if output_json: args.extend(["--output-format", "json"])
```

#### 4) `ff07cc0 refactor(cursor-agent): express dispatch support via build_exec_args() instead of requires_cli`

Addresses Copilot review: keep `requires_cli: False` so `specify init` does not hard-fail for IDE-only users. Remove the `requires_cli` guard inside `build_exec_args()`; dispatch support is signalled by overriding `build_exec_args()` (mirrors `CopilotIntegration`). Headless argv shape unchanged: `-p --trust --approve-mcps --force`.

#### 5) `422b250 test(cursor-agent): use urlparse hostname check and cover dispatch without requires_cli`

Addresses CodeQL: replace `"cursor.com" in url` substring check with `urlparse(...).hostname` comparison; add regression test that dispatch works with `requires_cli=False`.

## End-to-end verification (real DingTalk PRD, just a URL as input)

Repro on Windows 10 + cursor-agent `2026.05.16-0338208`:

```powershell
cd D:\Code\MBM\MBM-Backend     # real Maven multi-module Java project
specify workflow run speckit `
  --input "spec=https://alidocs.dingtalk.com/i/nodes/<dingtalk-prd-node>" `
  --input "integration=cursor-agent" `
  --input "scope=full"
```

Two independent runs (`df8c4b9b`, `388d3c15`) both reach the `review-spec` gate:

```
specify step:
  dispatched=true, exit_code=0, status=completed, duration ~156s / ~147s

agent behaviour (autonomously, from a URL-only spec input):
  - calls dingtalk-doc MCP get_document_content to fetch the PRD
  - reads project constitution + existing specs/ layout
  - decides specs/ subpath itself (e.g. specs/tooling/mold-assembly/202605/<feature>/)
  - produces spec.md (31.5 KB) + checklists/requirements.md
  - updates .specify/feature.json and specs/menu-dictionary.md
  - marks 3 [NEEDS CLARIFICATION] items for Gate 1 review

next step (review-spec gate):
  (gate started; status would be `paused` on a normal completion)
```

(Both runs above were eventually cancelled by `Ctrl+C` while the gate step was just starting, which surfaced an `[Errno 22] Invalid argument` from the interrupted `subprocess.run`. That was an artifact of my cancellation, **not** a gate-step bug — `review-spec` is a `gate` step and its designed terminal status is `paused`. Independent of this PR either way.)

## Backward compatibility

- In-IDE skill flow (`.cursor/skills/speckit-*/SKILL.md`, `--skills` default) — **unchanged**.
- `multi_install_safe`, `context_file`, `registrar_config`, `options()` — unchanged.
- POSIX behaviour — `shutil.which` lookup is a no-op for already-absolute exec_args[0], and the `--approve-mcps --force` flags exist on Linux/macOS cursor-agent identically.
- `catalog.json`, `workflow.yml`, templates — untouched.
- Other integrations (`claude`, `codex`, `gemini`) — their `build_exec_args` paths are not touched; the only shared change is `base.py::dispatch_command` adding the `shutil.which` shim resolution, which only takes effect when `shutil.which` returns a *different* path than the input (i.e. `.cmd`/`.bat` shim case on Windows).

## Testing

```
$ uv run python -m pytest tests/integrations/test_integration_cursor_agent.py -v
============================= 43 passed in 3.72s ==============================

$ uv run python -m pytest -q
========== 12 failed, 2867 passed, 109 skipped in 372.85s (0:06:12) ===========
```

- 44 cursor-agent integration tests pass (CI green on this branch).
- New tests pin:
  - `requires_cli is False` (IDE-first init) and `install_url` is set; dispatch via `build_exec_args()` override
  - `install_url` hostname validated with `urlparse` (not substring match)
  - Exact argv layout `[cursor-agent, -p, --trust, --approve-mcps, --force, …]` (default + text-output + with-model variants)
  - All four mandatory headless flags are always present together
  - Hyphenated skill invocation (`/speckit-plan` not `/speckit.plan`)
  - `.cmd` shim resolution: subprocess receives full `cursor-agent.CMD` path, not bare `cursor-agent`
  - Passthrough when `shutil.which` returns `None`
- Existing `tests/test_workflows.py::TestCommandStep::test_dispatch_with_mock_cli` and `test_dispatch_failure_returns_failed_status` updated to patch `shutil.which` at the `base.py` layer (was previously only patching the dispatcher layer, which made the test environment-sensitive — passed on CI with no `claude` on PATH, failed locally with `claude` installed).
- Full repo: **2867 passed, 0 regressions from this PR.** The 12 remaining pre-existing failures are unrelated Windows `symlink` privilege issues (`WinError 1314`) requiring elevated runner — not introduced or affected by this PR.

## Out of scope (deliberately)

- The misleading `"CLI not found or not installed. Install the CLI tool."` error message itself (it conflates 5 distinct dispatch failure modes into one). Worth a separate PR to split per-failure-mode messages — happy to do that as a follow-up if maintainers want.
- Adding `cursor-agent` to bundled `speckit` workflow's `requires.integrations.any` list (currently `["copilot", "claude", "gemini"]`). Not strictly required — `specify workflow run` does not enforce that list strictly today — but worth a follow-up so `specify workflow info speckit` is consistent with the actual support matrix.
- `docs/reference/integrations.md` page for cursor-agent (none exists today; happy to add if wanted).
- Adding a `speckit.git.feature` step to `speckit` workflow.yml (current workflow doesn't auto-create feature branches; that's by design but worth surfacing as an opt-in for teams that want branch-per-feature).

### DIFF
--- a/src/specify_cli/integrations/base.py
+++ b/src/specify_cli/integrations/base.py
@@ -202,6 +202,16 @@ class IntegrationBase(ABC):
             )
             raise NotImplementedError(msg)
 
+        # Windows: ``subprocess.run`` calls ``CreateProcess`` which does not
+        # consult ``PATHEXT``, so a bare command name like ``cursor-agent``
+        # that resolves to ``cursor-agent.cmd`` fails with ``WinError 2``.
+        # Resolve via ``shutil.which`` (which does honor ``PATHEXT``) so
+        # ``.cmd``/``.bat`` shims work transparently.  On POSIX this is a
+        # no-op for absolute paths and a harmless lookup otherwise.
+        resolved = shutil.which(exec_args[0])
+        if resolved:
+            exec_args = [resolved, *exec_args[1:]]
+
         cwd = str(project_root) if project_root else None
 
         if stream:
--- a/src/specify_cli/integrations/cursor_agent/__init__.py
+++ b/src/specify_cli/integrations/cursor_agent/__init__.py
@@ -41,14 +41,30 @@ class CursorAgentIntegration(SkillsIntegration):
     ) -> list[str] | None:
         """Build CLI arguments for non-interactive ``cursor-agent`` execution.
 
-        Uses ``-p`` (print/headless mode, which gives access to all
-        tools including write and shell) plus ``--trust`` (bypass the
-        Workspace Trust prompt — mandatory for headless execution; the
-        CLI exits non-zero without it).
+        Mandatory headless flags:
+
+        * ``-p`` — print/headless mode (access to all tools)
+        * ``--trust`` — bypass Workspace Trust prompt (CLI exits non-zero
+          otherwise)
+        * ``--approve-mcps`` — auto-approve MCP server loading (otherwise
+          MCP servers stay ``not loaded (needs approval)`` and tool calls
+          to them are silently dropped)
+        * ``--force`` — auto-approve tool invocations (shell/write/MCP),
+          matching the implicit "trusted environment" semantics that other
+          integrations (``claude -p``, ``codex --exec``) get by default
+
+        Together these are the minimum set required to make
+        ``specify workflow run speckit --input integration=cursor-agent``
+        behave the same way as it does for ``claude`` / ``codex``.
+        Verified locally: with ``--approve-mcps --force`` the agent can
+        call any configured MCP server (e.g. ``dingtalk-doc``) and write
+        files during ``/speckit-*`` skill execution; without them the run
+        either drops tool calls or exits non-zero on the first approval
+        prompt.
         """
         if not self.config or not self.config.get("requires_cli"):
             return None
-        args = [self.key, "-p", "--trust", prompt]
+        args = [self.key, "-p", "--trust", "--approve-mcps", "--force", prompt]
         if model:
             args.extend(["--model", model])
         if output_json:
--- a/src/specify_cli/integrations/cursor_agent/__init__.py
+++ b/src/specify_cli/integrations/cursor_agent/__init__.py
@@ -3,9 +3,11 @@
 Cursor Agent uses the ``.cursor/skills/speckit-<name>/SKILL.md`` layout.
 Commands are deprecated; ``--skills`` defaults to ``True``.
 
-CLI dispatch via ``cursor-agent -p --trust <prompt>`` is supported so
-``specify workflow run`` can drive cursor-agent headlessly, in addition
-to the existing in-IDE skill flow.
+The IDE/skills flow is the primary path and works without the
+``cursor-agent`` CLI being installed (``requires_cli=False``).  Workflow
+dispatch via ``cursor-agent -p --trust <prompt>`` is offered as an
+opt-in capability — the presence of ``build_exec_args()`` is what
+indicates dispatch support, mirroring ``CopilotIntegration``.
 """
 
 from __future__ import annotations
@@ -20,7 +22,12 @@ class CursorAgentIntegration(SkillsIntegration):
         "folder": ".cursor/",
         "commands_subdir": "skills",
         "install_url": "https://docs.cursor.com/en/cli/overview",
-        "requires_cli": True,
+        # IDE-first integration: ``specify init --ai cursor-agent`` must
+        # work without the ``cursor-agent`` CLI installed (the IDE flow
+        # uses skills directly).  Workflow dispatch additionally requires
+        # the CLI on PATH, but that's enforced at dispatch time via
+        # ``shutil.which`` rather than as a hard ``specify init`` precheck.
+        "requires_cli": False,
     }
     registrar_config = {
         "dir": ".cursor/skills",
@@ -40,6 +47,13 @@ class CursorAgentIntegration(SkillsIntegration):
         output_json: bool = True,
     ) -> list[str] | None:
         """Build CLI arguments for non-interactive ``cursor-agent`` execution.
+
+        Always returns argv (no ``requires_cli`` guard) so workflow
+        dispatch is supported even though the integration's ``config``
+        sets ``requires_cli=False`` to keep the IDE-only flow unblocked.
+        This mirrors ``CopilotIntegration``: dispatch support is signalled
+        by overriding ``build_exec_args()``, not by the ``requires_cli``
+        flag (which is reserved for the ``specify init`` precheck).
 
         Mandatory headless flags:
 
@@ -62,8 +76,6 @@ class CursorAgentIntegration(SkillsIntegration):
         either drops tool calls or exits non-zero on the first approval
         prompt.
         """
-        if not self.config or not self.config.get("requires_cli"):
-            return None
         args = [self.key, "-p", "--trust", "--approve-mcps", "--force", prompt]
         if model:
             args.extend(["--model", model])
--- a/src/specify_cli/integrations/cursor_agent/__init__.py
+++ b/src/specify_cli/integrations/cursor_agent/__init__.py
@@ -5,9 +5,9 @@ Commands are deprecated; ``--skills`` defaults to ``True``.
 
 The IDE/skills flow is the primary path and works without the
 ``cursor-agent`` CLI being installed (``requires_cli=False``).  Workflow
-dispatch via ``cursor-agent -p --trust <prompt>`` is offered as an
-opt-in capability — the presence of ``build_exec_args()`` is what
-indicates dispatch support, mirroring ``CopilotIntegration``.
+dispatch via ``cursor-agent -p --trust --approve-mcps --force <prompt>``
+is offered as an opt-in capability — the presence of ``build_exec_args()``
+is what indicates dispatch support, mirroring ``CopilotIntegration``.
 """
 
 from __future__ import annotations
--- a/src/specify_cli/integrations/cursor_agent/__init__.py
+++ b/src/specify_cli/integrations/cursor_agent/__init__.py
@@ -2,6 +2,10 @@
 
 Cursor Agent uses the ``.cursor/skills/speckit-<name>/SKILL.md`` layout.
 Commands are deprecated; ``--skills`` defaults to ``True``.
+
+CLI dispatch via ``cursor-agent -p --trust <prompt>`` is supported so
+``specify workflow run`` can drive cursor-agent headlessly, in addition
+to the existing in-IDE skill flow.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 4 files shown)


mnriem (changes requested):

mnriem:
Please address code scanning resulsts

One-TheOnly:
Thanks for the review! I've addressed both points in ff07cc0 + 422b250:

- **Copilot (requires_cli)**: reverted `requires_cli` to `False` so `specify init` no longer hard-fails for IDE-only users. Dispatch support is now expressed by overriding `build_exec_args()` (same shape as `CopilotIntegration`), and the `requires_cli` guard in `build_exec_args()` is removed.
- **CodeQL (URL substring)**: replaced the `"cursor.com" in url` substring check with an exact `urlparse(...).hostname` comparison, plus a regression test.

`pytest tests/integrations/test_integration_cursor_agent.py` is green (44 passed) and `ruff check src/` is clean. Could you re-run the workflows when you get a chance?

mnriem:

One-TheOnly:
This autofix suggestion was applied.

mnriem:
Can you please update the PR description as per Copilot feedback

One-TheOnly:
﻿@mnriem Updated the PR description to match the current implementation:

- **`requires_cli` stays `False`** (IDE-first `specify init`; dispatch via `build_exec_args()` override, same as `CopilotIntegration`).
- Documented review commits **`ff07cc0`**, **`422b250`**, and **`6e3819c`** (docstring flags).
- Corrected the Testing section (no longer claims `requires_cli is True`).

CI on `6e3819c` is all green. Ready for another look when you have time.

mnriem (approved):

PR #2647, opened 2026-05-20
fix: refresh speckit_version in init-options on integration update

PascalThuet:
## Summary

- Refresh `.specify/init-options.json` `speckit_version` whenever active integration metadata is updated.
- Refresh the same version field for multi-agent install/upgrade paths without changing the default integration fields.
- Add regression tests for default and non-default integration update paths.

## Why

`init-options.json` stores active integration metadata such as `integration`, `ai`, `context_file`, `script`, and `ai_skills`, but `speckit_version` stayed pinned to the version that originally initialized the project.

After upgrading Spec Kit, that made `init-options.json` diverge from `.specify/integration.json` and the integration manifests, which already record the current Spec Kit version.

For multi-agent projects, non-default integration installs/upgrades also update `.specify/integration.json`, so this keeps the global init-options version aligned while preserving the active `ai`, `integration`, and `context_file` values.

If the intended meaning of this field is historical instead, it should probably be renamed or documented separately, for example as `initialized_speckit_version`.

## Validation

- `uv run pytest tests/integrations/test_integration_subcommand.py -v`

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -1332,6 +1332,15 @@ def _write_integration_json(
     )
 
 
+def _refresh_init_options_speckit_version(project_root: Path) -> None:
+    """Refresh only the Spec Kit version recorded in init-options.json."""
+    opts = load_init_options(project_root)
+    if not isinstance(opts, dict) or not opts:
+        return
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #2649, opened 2026-05-20
fix: render script command hints with active agent separator

PascalThuet:
## Summary

- Render user-facing script command hints with the active integration's `invoke_separator`.
- Update bash and PowerShell prerequisite guards in `setup-tasks` and `check-prerequisites`.
- Add regression tests for hyphenated agent command hints.

## Why

Shared scripts were hardcoding dotted slash commands such as `/speckit.plan`, `/speckit.specify`, and `/speckit.tasks`.

That is correct for internal command IDs and dot-based integrations, but it is confusing for agents whose installed commands use `invoke_separator: "-"`, where the resolvable slash command is `/speckit-plan`.

## Implementation

The shared script helpers now read `.specify/integration.json`, resolve the default integration's stored `invoke_separator`, and format the displayed slash command accordingly. If state is missing or malformed, they fall back to the dotted form.

## Validation

- `git diff --check`
- `bash -n scripts/bash/common.sh`
- `bash -n scripts/bash/setup-tasks.sh`
- `bash -n scripts/bash/check-prerequisites.sh`
- `uv run pytest tests/test_setup_tasks.py -v`
- `uv run pytest tests/test_timestamp_branches.py -v`

PowerShell was not available in the local environment, so the PowerShell regression tests were collected but skipped.


mnriem (changes requested):
Can you remove the unnecessary whitespace changes at the end of the files where it shows up?

PascalThuet:
done :)

mnriem (changes requested):

mnriem:

mnriem (approved):

mnriem:
Thank you!

PR #2662, opened 2026-05-21
fix: while/do-while loop condition reads stale iteration-0 step output

### LINKED ISSUE #2592: [Bug]: `while`/`do-while` loop condition reads stale iteration-0 step output
### Bug Description

`workflows/engine.py` namespaces nested step IDs per iteration in
`while` / `do-while` loops:

```python
# engine.py around line 678
ns_copy["id"] = f"{step_id}:{ns_copy['id']}:{_loop_iter + 1}"
```

But the loop's `condition` (resolved via
`expressions.evaluate_condition` →
`_resolve_dot_path` → `_build_namespace`) reads step output by the
**unprefixed** step ID. So a condition like
`{{ steps.my-gate.output.choice }} != 'approve'` reads the
**iteration-0** value on every subsequent iteration, because
`steps.my-gate` in the namespace points at the original
un-namespaced step output and is never updated to track
`steps.my-gate:my-loop:1`, `:2`, etc.

Net: gate-driven loop termination based on a gate's verdict
**never fires**. The loop runs to `max_iterations`.

### Steps to Reproduce

```yaml
schema_version: "1.0"
workflow:
  id: while-loop-bug-repro
  name: "While-loop namespacing bug reproducer"
  version: "0.0.1"
  workflow_status: test
requires:
  speckit_version: ">=0.8.7"
  integrations:
    any: ["claude"]
inputs: {}
steps:
  - id: my-loop
    type: while
    condition: "{{ steps.my-gate.output.choice }} != 'approve'"
    max_iterations: 5
    body:
      - id: my-gate
        type: gate
        prompt: "Approve, improve, rebuild, or abort?"
        choices: ["approve", "improve", "rebuild", "abort"]
```

```bash
specify workflow run while-loop-bug-repro
# choose 'approve' on the first prompt
```

### Expected Behavior

Loop terminates after iteration 1; condition evaluates `false`.

### Actual Behavior

Loop continues for all `max_iterations` (= 5), prompting again on
every iteration. Condition evaluator on iteration 1+ reads the
iteration-0 stale `output.choice`.

### Specify CLI Version

`engine.py` byte-identical (sha256 `543dbb3676…`) on every recent
0.8.x release; bug present in current `main` (`0.8.12.dev0`).

### AI Agent

Claude Code

### Operating System

Any (pure-Python engine code; not platform-specific).

### Python Version

Python 3.11+

### Error Logs

Not an error — silent wrong behaviour. The loop runs to
`max_iterations` and exits normally, ignoring the gate verdict.

### Additional Context

**Scope**: the bug is in the workflow engine's condition resolver
(`expressions.evaluate_condition`), not in any integration. The
reproducer happens to use Claude as the integration the gate step
runs through, but the bug manifests identically with any agent —
the dropdown above just reflects the reproducer environment.

`while` and `do-while` are documented step kinds (engine.py:97).
If they don't work for the canonical use case (gate-driven
termination), they're effectively dead. Workflow authors today must
manually unroll review loops into nested `switch`-after-`gate`
chains, capped at 2 iterations.

#### Possible fix shapes

- **A.** Inside a loop body, expose the most recent iteration's
  nested-step output under an alias like
  `{{ steps.my-gate.latest.output.choice }}`.
- **B.** Auto-resolve unprefixed step refs in loop conditions to
  the latest namespaced version. Backward-compatible — existing
  condition text starts working.
- **C.** Expose iteration counter as `{{ context.iteration }}` and
  let condition authors reference iteration-suffixed IDs manually.

B is least friction. Want shape agreement before opening a PR
(engine internals).

**AI disclosure:** drafted with Claude Opus including reproducer +
analysis. Human-reviewed.

mnriem:
## Summary

Fixes #2592 — `while`/`do-while` loop conditions always read the iteration-0 step output, ignoring updates from subsequent iterations.

## Root cause

The engine namespaces nested step IDs per iteration (e.g., `my-loop:my-gate:1`) to preserve per-iteration history. However, the condition resolver (`evaluate_condition` → `_resolve_dot_path`) looks up step output by the original unprefixed key (`steps.my-gate`), which is only written during iteration 0 and never updated afterward.

## Fix

Keep namespaced step IDs for execution (preserving unique log entries, state keys, and `on_step_start` callback IDs per iteration), but execute each body step individually and immediately alias its result back to the unprefixed key after it completes successfully. This ensures:

- `evaluate_condition()` sees the latest iteration's output
- Inter-step references within multi-step loop bodies work correctly (step B sees step A's output from the current iteration, not a stale previous one)
- Namespaced history entries are preserved for debugging (`parent:child:1`, `parent:child:2`, etc.)
- Pause/fail is checked after each step — aliasing only occurs for successfully completed steps
- Fully backward-compatible — same namespaced key format, same log IDs, no changes to condition syntax or workflow YAML

## Changes

- **`src/specify_cli/workflows/engine.py`** — Execute loop body steps one at a time with namespaced IDs, aliasing each result back to the unprefixed key immediately after successful completion
- **`tests/test_workflows.py`** — 5 new engine-level regression tests in `TestWorkflowEngine`:
  - `test_while_loop_condition_reads_latest_iteration` (positive: loop stops early)
  - `test_do_while_loop_condition_reads_latest_iteration` (positive: loop stops early)
  - `test_while_loop_runs_to_max_when_condition_stays_true` (negative: loop exhausts max_iterations)
  - `test_do_while_loop_runs_to_max_when_condition_stays_true` (negative: loop exhausts max_iterations)
  - `test_while_loop_multi_step_body_inter_step_refs` (multi-step body: step B sees step A's current-iteration output)

## Testing

All 2998 existing + new tests pass (2998 passed, 1 skipped, 0 failures).

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -674,15 +674,24 @@ class WorkflowEngine:
                             break
                         # Namespace nested step IDs per iteration
                         iter_steps = []
+                        original_ids = {}
                         for ns in result.next_steps:
                             ns_copy = dict(ns)
                             if "id" in ns_copy:
-                                ns_copy["id"] = f"{step_id}:{ns_copy['id']}:{_loop_iter + 1}"
+                                orig = ns_copy["id"]
+                                ns_copy["id"] = f"{step_id}:{orig}:{_loop_iter + 1}"
+                                original_ids[ns_copy["id"]] = orig
                             iter_steps.append(ns_copy)
                         self._execute_steps(
                             iter_steps, context, state, registry,
                             step_offset=-1,
                         )
+                        # Copy namespaced results back to unprefixed
+                        # keys so loop conditions see updated values.
+                        for namespaced, orig in original_ids.items():
+                            if namespaced in context.steps:
+                                context.steps[orig] = context.steps[namespaced]
+                                state.step_results[orig] = context.steps[namespaced]
                         if state.status in (
                             RunStatus.PAUSED,
                             RunStatus.FAILED,
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -672,30 +672,30 @@ class WorkflowEngine:
                     for _loop_iter in range(max_iters - 1):
                         if not evaluate_condition(condition, context):
                             break
-                        # Snapshot current results under namespaced
-                        # keys for per-iteration history before they
-                        # are overwritten by the next iteration.
+                        # Namespace nested step IDs per iteration
+                        # so logs and state keys are unique.
+                        # Execute one step at a time and alias each
+                        # result back to the unprefixed key so that
+                        # later steps in the same body and the loop
+                        # condition see the latest values.
                         for ns in result.next_steps:
-                            orig = ns.get("id")
-                            if orig and orig in context.steps:
-                                ns_key = f"{step_id}:{orig}:{_loop_iter}"
-                                context.steps[ns_key] = context.steps[orig]
-                                state.step_results[ns_key] = context.steps[orig]
-                        # Execute body with original step IDs so
-                        # results land at the unprefixed keys.  Both
-                        # inter-step references within the body and
-                        # the loop condition naturally see the latest
-                        # values without a copy-back.
-                        self._execute_steps(
-                            result.next_steps, context, state, registry,
-                            step_offset=-1,
-                        )
-                        if state.status in (
-                            RunStatus.PAUSED,
-                            RunStatus.FAILED,
-                            RunStatus.ABORTED,
-                        ):
-                            return
+                            ns_copy = dict(ns)
+                            orig = ns_copy.get("id")
+                            if orig:
+                                ns_copy["id"] = f"{step_id}:{orig}:{_loop_iter + 1}"
+                            self._execute_steps(
+                                [ns_copy], context, state, registry,
+                                step_offset=-1,
+                            )
+                            if orig and ns_copy["id"] in context.steps:
+                                context.steps[orig] = context.steps[ns_copy["id"]]
+                                state.step_results[orig] = context.steps[ns_copy["id"]]
+                            if state.status in (
+                                RunStatus.PAUSED,
+                                RunStatus.FAILED,
+                                RunStatus.ABORTED,
+                            ):
+                                return
 
             # Fan-out: execute nested step template per item with unique IDs
             if step_type == "fan-out":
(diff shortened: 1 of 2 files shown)


doquanghuy (commented):
Thanks @mnriem! I checked out the branch locally and ran the new regression tests — all 4 pass. The fix logic is sound: per-step aliasing of the namespaced ID back to the unprefixed key cleanly handles the condition-resolver gap I reported in #2592.

A few observations for consideration, none are blockers:

### 1. Test placement (worth fixing)

The 4 new tests are inserted into the `TestWorkflowValidation` class (at lines 1892, 1951, 1996, 2047 — `TestWorkflowValidation` spans lines 1208 → ~2099 in this PR's view), but they exercise engine *execution* behaviour, not validation. Future maintainers grepping for loop tests in `TestWhileStep` (line 910) or `TestDoWhileStep` (line 965) won't find them, and the file's validation-vs-execution boundary gets muddied.

A small move to those existing classes (or a new `TestLoopIterationOutput` adjacent to them) keeps the file organisation clean.

### 2. Coverage of a claim in the PR description

The PR description states:
> Inter-step references within multi-step loop bodies work correctly (step B sees step A's output from the current iteration, not a stale previous one)

The logic supports this — the per-step `for ns in result.next_steps` aliases after each step before the next runs — but no test exercises a multi-step body. A small `steps: [stepA, stepB]` test where `stepB` reads `{{ steps.stepA.output.stdout }}` would lock that contract explicitly.

### 3. The canonical bug-report scenario goes untested

My issue's reproducer used a `gate` step inside the loop (pause-resume across iterations). The PR's tests use only shell steps, so the pause-then-resume path through the new aliasing logic isn't exercised. Not a regression risk (resume re-runs the loop from scratch, so any aliased PAUSED-state value gets overwritten on the next iter-0), but the canonical use case from the bug report goes uncovered. A monkey-patched gate test would lock that no surprises lurk in the pause-then-alias ordering.

### 4. Minor: PR description vs. code

PR description:
> Pause/fail is checked after each step — no aliasing of partial results

But the code does alias *before* checking status:
```python
self._execute_steps([ns_copy], context, state, registry, step_offset=-1)
if orig and ns_copy["id"] in context.steps:
    context.steps[orig] = context.steps[ns_copy["id"]]
    state.step_results[orig] = context.steps[ns_copy["id"]]
if state.status in (RunStatus.PAUSED, RunStatus.FAILED, RunStatus.ABORTED):
    return
```

This is fine in practice — `_execute_steps` writes `context.steps[step_id]` at engine.py:606 *before* the pause check at :619, so the alias picks up a valid (if PAUSED-state) result, and resume re-runs the loop from scratch anyway. But the description implies a stricter guarantee than the code. Either tighten the wording or guard the alias with a status check (`if state.status not in (PAUSED, FAILED, ABORTED):`) — both reach the same end state.

Otherwise, the fix is exactly what #2592 asked for. Thanks for jumping on it!

**AI disclosure:** drafted with Claude Opus, human-reviewed.

PR #2670, opened 2026-05-21
docs: fix broken pipx homepage URLs to point to pipx.pypa.io

jpatel3:
This PR updates outdated pipx homepage URLs (which returned 404 at the old pypa.github.io location) to point to the official, active domain at https://pipx.pypa.io/.


mnriem:
Thank you!

PR #2675, opened 2026-05-22
Fix support section TOC anchor and add missing emoji

Selfeer:
- Added missing 💬 emoji to the Support section in the README
- Fixed the Support entry in the Table of Contents to correctly link to the section anchor generated by GitHub markdown rendering

P.S. This bothered me more than it should have, so I had to fix it 😄


mnriem:
Thank you!

PR #2280, opened 2026-04-20
fix(powershell): ensure UTF-8 templates are written without BOM

Nimraakram22:
Description
This PR fixes an issue where PowerShell 5.1 (default on Windows) writes files like CLAUDE.md with a UTF-8 Byte Order Mark (BOM). This invisible prefix causes parsing errors in modern tools like Claude Code and other Unix-style CLI utilities.

Changes
Stripped BOM from existing templates: Found and removed the BOM from all .md files in the templates/ and presets/ directories.

Hardened PowerShell logic: Replaced Copy-Item with a .NET based write method ([System.IO.File]::WriteAllText) in setup-plan.ps1 and create-new-feature.ps1 to force "UTF-8 No BOM" output, regardless of the source template's encoding.

Verification
Verified using the file command in MINGW64 to ensure output files now show as UTF-8 Unicode text instead of UTF-8 Unicode (with BOM) text.

PR #2686, opened 2026-05-23
fix(cli): pin UTF-8 encoding on init-options and .extensionignore I/O

Quratulain-bilal:
## What

  `Path.read_text()` / `Path.write_text()` default to the system locale codec, which is cp1252 / gb2312 / cp932 on Windows.
   Two user-facing file paths in spec-kit were calling them without an explicit `encoding=` argument:

  - **`src/specify_cli/__init__.py:400, 412`** — `save_init_options` / `load_init_options` for
  `.specify/init-options.json`. A peer machine with a different default locale (or a UTF-8 Unix CI runner reading a file
  written on a cp1252 Windows host) cannot decode the file, raising `UnicodeDecodeError`. `UnicodeDecodeError` is a
  subclass of `ValueError` — not `OSError` / `json.JSONDecodeError` — so the existing fall-back `except` tuple in
  `load_init_options` also misses it and the error propagates raw to the CLI.

  - **`src/specify_cli/extensions.py:764`** — `.extensionignore` pattern reader. The very next line already normalises
  backslashes *"so Windows-authored files work"*, proving the codebase expects Windows authors to write this file.
  Multibyte UTF-8 patterns (Chinese filenames, accented directory names) silently mojibake when the host locale is not
  UTF-8, so the patterns fail to match and unintended files are shipped with the extension.

  ## Reproducer

  ```python
  # On a non-UTF-8 Windows host (e.g. cp1252):
  from specify_cli import save_init_options, load_init_options
  save_init_options(Path("."), {"project_name": "café"})
  # File written as cp1252.

  # On a UTF-8 host (or another Windows host with a different default codec):
  load_init_options(Path("."))
  # UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9...
  # — propagates raw because UnicodeDecodeError is a ValueError, not an OSError.
  ```

  ## Why this matters

  - The sibling integration-catalog reader at `src/specify_cli/integrations/catalog.py:150,156,193,202,374` already pins
  `encoding="utf-8"` everywhere. PR #2280 (`f684305`) fixed the symmetric PowerShell-template BOM bug. The two paths in
  this PR were the remaining drifted ones.
  - `init-options.json` is meant to be a portable record of how a project was scaffolded — a peer cloning the repo on a
  different OS / locale must be able to read it. Today they can't if the original author's project name (or any future
  field) contains non-ASCII.
  - `.extensionignore` already explicitly supports Windows authors (see line 766). UTF-8 patterns are part of that same
  contract.

  ## The change

  - `src/specify_cli/__init__.py` — pin `encoding="utf-8"` on both `write_text` and `read_text`; extend the existing
  `except` tuple in `load_init_options` to include `UnicodeDecodeError` so a peer file written in a non-UTF-8 codec falls
  back to `{}` per the existing contract instead of crashing.
  - `src/specify_cli/extensions.py` — pin `encoding="utf-8"` on the `.extensionignore` reader.

  ## Tests

  - `tests/test_presets.py::TestInitOptions` — parametrized non-ASCII round-trip (CJK / Latin-1 / Greek / emoji) plus a
  `0xe9`-byte corrupted-file fallback test.
  - `tests/test_extensions.py::TestExtensionIgnore` — Japanese (`ドキュメント/`) and Latin-1 (`café/`) ignore patterns
  correctly exclude their directories during install.

  ```
  $ python -m pytest tests/test_presets.py::TestInitOptions tests/test_extensions.py::TestExtensionIgnore -q
  22 passed in 4.56s
  $ python -m ruff check src/
  All checks passed!
  ```

  ## Scope

  Intentionally narrow: no behaviour change for ASCII content (UTF-8 is a superset). Only non-ASCII content that previously
   round-tripped accidentally (when host locale happened to be UTF-8) or silently mojibaked (when it wasn't) now
  round-trips reliably across all hosts.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -397,14 +397,27 @@ def save_init_options(project_path: Path, options: dict[str, Any]) -> None:
     """
     dest = project_path / INIT_OPTIONS_FILE
     dest.parent.mkdir(parents=True, exist_ok=True)
-    # Pin UTF-8 explicitly: ``Path.write_text`` defaults to the system
-    # locale codec, which is cp1252 / gb2312 / cp932 on Windows. A
-    # locale-encoded write succeeds locally but produces a file a peer
-    # machine (different locale) or Unix CI cannot decode. The sibling
-    # integration-catalog writer in ``integrations/catalog.py`` already
-    # pins ``encoding="utf-8"`` for the same reason.
+    # Write JSON as real UTF-8 instead of ``\uXXXX`` escape sequences
+    # (``ensure_ascii=False``) and pin the file encoding to match.
+    #
+    # The default ``json.dumps`` output is ASCII-only — any non-ASCII
+    # character is encoded as a ``\uXXXX`` escape — so without the
+    # ``ensure_ascii=False`` flip below the encoding pin alone would be
+    # a no-op for any payload we plausibly write today. We pair the two
+    # so the on-disk bytes match a human's expectation of "this file is
+    # UTF-8" (greppable, readable in editors that don't decode JSON
+    # escapes, friendly to peers running ``cat`` or ``Get-Content``) and
+    # so the encoding pin is a real contract instead of a future hedge.
+    #
+    # ``Path.write_text`` without ``encoding=`` falls back to the system
+    # locale codec (cp1252 / gb2312 / cp932 on Windows), which would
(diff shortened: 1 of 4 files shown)


mnriem:

mnriem (approved):

mnriem:
Thank you!

PR #2765, opened 2026-05-29
fix(copilot): resolve active spec template

### LINKED ISSUE #2761: Copilot speckit.specify does not honor preset spec-template override
# Copilot /speckit.specify does not honor preset spec-template override

## Summary

When using the Copilot integration, a preset that overrides only `spec-template` appears to install and resolve correctly, but `/speckit.specify` still uses the core/default template because the generated Copilot agent instructions hardcode `.specify/templates/spec-template.md` instead of resolving the effective template through the preset resolution stack.

This makes template-only preset overrides unreliable for Copilot, even though the preset documentation says templates and commands are resolved independently and command overrides are optional.

## Environment

- OS: Windows
- Spec Kit CLI: `specify 0.8.18.dev0`
- Integration: Copilot / VS Code
- Project initialized with `.github/agents/*.agent.md` and `.github/prompts/*.prompt.md`

## What The Documentation Implies

The presets documentation says:

- Presets can provide template files and command files.
- Each file is looked up independently, so different files can come from different layers.
- Command overrides are optional.
- If only different template sections are needed, a template override should be sufficient.
- `specify preset resolve <name>` shows the winning file.

## Reproduction

1. Install a preset that provides `templates/spec-template.md` but does not override `speckit.specify`.
2. Verify the template resolves from the preset:

```powershell
specify preset resolve spec-template
```

Observed resolver output:

```text
spec-template:
C:\git\harness\crm-project\.specify\presets\harness-presets\templates\spec-template.md
(top layer from: harness-presets v0.1.0)
```

3. Run `/speckit.specify` from Copilot in VS Code.
4. Inspect the generated `spec.md`.

## Expected Behavior

`/speckit.specify` should use the effective `spec-template` selected by the preset resolution stack. In this case, it should use:

```text
.specify/presets/harness-presets/templates/spec-template.md
```

A preset that only overrides `spec-template` should affect the generated spec without requiring a command override.

## Actual Behavior

The generated spec follows the core/default template instead of the preset template.

The generated Copilot agent file contains hardcoded instructions like:

```text
Copy `.specify/templates/spec-template.md` to `SPECIFY_FEATURE_DIRECTORY/spec.md` as the starting point
Load `.specify/templates/spec-template.md` to understand required sections.
```

Because the agent reads the core path directly, it bypasses the preset resolver even though `specify preset resolve spec-template` reports the preset template as the winner.

## Why This Matters

This creates a mismatch between preset documentation and Copilot runtime behavior:

- `specify preset resolve spec-template` says the preset template wins.
- The Copilot `/speckit.specify` agent still reads the core template path.
- Users are forced to override `speckit.specify` as a workaround, even when they only want to change the spec template shape.

## Suggested Fix

Update the Copilot `speckit.specify` command/agent instructions so they resolve the effective template instead of hardcoding `.specify/templates/spec-template.md`.

For example, the generated agent instructions could say:

```text
Resolve the active `spec-template` using the Spec Kit preset resolution stack, equivalent to `specify preset resolve spec-template`, then read that resolved file as the starting template.
Do not directly read `.specify/templates/spec-template.md` unless the resolver returns that path.
```

Alternatively, the command-generation layer could materialize the winning template path into the generated agent instructions.

## Additional Note

After adding a `speckit.specify` command override, Copilot loads the preset command correctly. However, that should not be required for a template-only override according to the preset documentation.

WOLIKIMCHENG:
## Description

Fixes #2761.

This updates the Copilot `speckit.specify` command instructions so generated agents resolve the active `spec-template` through the Spec Kit preset/template resolution stack instead of hardcoding the core template path. This lets project overrides, presets, and extensions supply the effective spec template for new specs.

## Testing

- [x] Tested locally with `uv run --no-sync specify --help`
- [x] Ran existing tests with `uv run --no-sync pytest -q` (`3104 passed, 40 skipped`)
- [x] Ran focused Copilot integration tests with `uv run --no-sync pytest tests/integrations/test_integration_copilot.py -q` (`47 passed`)
- [x] Ran `uvx ruff check src/`
- [x] Ran `python -m compileall -q src tests/integrations/test_integration_copilot.py`
- [x] Ran `git diff --check -- templates/commands/specify.md tests/integrations/test_integration_copilot.py`
- [ ] Tested with a sample project (not applicable; covered by Copilot integration generation test)

### DIFF
--- a/templates/commands/specify.md
+++ b/templates/commands/specify.md
@@ -91,7 +91,8 @@ Given that feature description, do this:
 
    **Create the directory and spec file**:
    - `mkdir -p SPECIFY_FEATURE_DIRECTORY`
-   - Copy `templates/spec-template.md` to `SPECIFY_FEATURE_DIRECTORY/spec.md` as the starting point
+   - Resolve the active `spec-template` through the Spec Kit preset/template resolution stack (equivalent to `specify preset resolve spec-template`)
+   - Copy the resolved `spec-template` file to `SPECIFY_FEATURE_DIRECTORY/spec.md` as the starting point
    - Set `SPEC_FILE` to `SPECIFY_FEATURE_DIRECTORY/spec.md`
    - Persist the resolved path to `.specify/feature.json`:
      ```json
@@ -107,7 +108,7 @@ Given that feature description, do this:
    - The spec directory name and the git branch name are independent — they may be the same but that is the user's choice
    - The spec directory and file are always created by this command, never by the hook
 
-4. Load `templates/spec-template.md` to understand required sections.
+4. Load the resolved active `spec-template` file to understand required sections.
 
 5. Follow this execution flow:
     1. Parse user description from arguments
--- a/tests/integrations/test_integration_copilot.py
+++ b/tests/integrations/test_integration_copilot.py
@@ -147,6 +147,21 @@ class TestCopilotIntegration:
             assert "__SPECKIT_COMMAND_" not in content, f"{agent_file.name} has unprocessed __SPECKIT_COMMAND_*__"
             assert "\nscripts:\n" not in content
 
+    def test_specify_agent_resolves_active_spec_template(self, tmp_path):
+        """Generated specify agent must not hardcode the core spec template."""
+        from specify_cli.integrations.copilot import CopilotIntegration
+        copilot = CopilotIntegration()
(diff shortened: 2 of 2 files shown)


mnriem:
Thank you!

PR #2792, opened 2026-06-01
fix: resolve GitHub release asset API URL for private repo extension downloads

lselvar:
## Summary

- Fixes `specify extension add` when the catalog `download_url` points at a GitHub release URL for a private or SSO-protected repository
- For private/SSO repos, browser release URLs (`https://github.com/<owner>/<repo>/releases/download/<tag>/<asset>.zip`) redirect to an HTML/SSO page instead of the ZIP asset, causing installation to fail with `BadZipFile: File is not a zip file`

## Fix — two download URL shapes are now handled

**Browser release URL** (commit 1):
- Detects GitHub browser release asset URLs during download
- Resolves the matching asset via the GitHub REST API: `GET /repos/<owner>/<repo>/releases/tags/<tag>`
- Downloads the matched asset URL with `Accept: application/octet-stream`
- Falls back to the original download URL if the API call fails

**Direct REST asset URL** (commit 2):
- When `download_url` is already a GitHub REST release asset URL (`https://api.github.com/repos/<owner>/<repo>/releases/assets/<id>`), skips the metadata lookup and downloads directly with `Accept: application/octet-stream`
- That header is required because GitHub returns release asset metadata JSON by default; `Accept: application/octet-stream` returns the ZIP payload

**Other URLs**: existing downloader behavior is unchanged.

Auth is preserved end-to-end through the existing `specify_cli.authentication.http.open_url` infrastructure.

## Test plan

- [ ] Run focused test suite to confirm new and existing behavior:
  ```bash
  UV_NATIVE_TLS=true SSL_CERT_FILE=/opt/homebrew/etc/openssl@3/cert.pem UV_DEFAULT_INDEX=https://pypi.org/simple PYTHONPATH=src uv run --extra test pytest tests/test_extensions.py -k 'make_request or fetch_single_catalog_sends_auth_header or download_extension_sends_auth_header or download_extension_accepts_direct_github_rest_asset_url'
  ```
  Expected: 13 passed, 190 deselected
- [ ] Verify `specify extension add` against an extension hosted on a private GitHub repo using a browser release URL — asset should be resolved via the API and downloaded correctly
- [ ] Verify `specify extension add` against a catalog using a direct REST asset URL (`api.github.com/repos/.../releases/assets/<id>`) — should download directly with auth and `Accept: application/octet-stream`
- [ ] Verify `specify extension add` still works for public GitHub release URLs and non-GitHub URLs (fallback paths not triggered)

> **AI disclosure**: This PR was developed with Claude Code assistance.

### DIFF
--- a/src/specify_cli/extensions.py
+++ b/src/specify_cli/extensions.py
@@ -1728,13 +1728,51 @@ class ExtensionCatalog(CatalogStackBase):
         from specify_cli.authentication.http import build_request
         return build_request(url)
 
-    def _open_url(self, url: str, timeout: int = 10):
+    def _open_url(
+        self,
+        url: str,
+        timeout: int = 10,
+        extra_headers: Optional[Dict[str, str]] = None,
+    ):
         """Open a URL with provider-based auth, trying each configured provider.
 
         Delegates to :func:`specify_cli.authentication.http.open_url`.
         """
         from specify_cli.authentication.http import open_url
-        return open_url(url, timeout)
+        return open_url(url, timeout, extra_headers=extra_headers)
+
+    def _resolve_github_release_asset_api_url(
+        self,
+        download_url: str,
+        timeout: int = 60,
+    ) -> Optional[str]:
+        """Resolve a GitHub browser release asset URL to its API asset URL."""
+        import urllib.error
+        from urllib.parse import unquote, urlparse
+
+        parsed = urlparse(download_url)
+        if parsed.hostname != "github.com":
+            return None
+
+        parts = [unquote(part) for part in parsed.path.strip("/").split("/")]
+        if len(parts) < 6 or parts[2:4] != ["releases", "download"]:
+            return None
+
+        owner, repo, tag = parts[0], parts[1], parts[4]
+        asset_name = "/".join(parts[5:])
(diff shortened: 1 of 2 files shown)


mnriem:
Just curious, can it refer to the REST URL directly?

lselvar:
> Just curious, can it refer to the REST URL directly?

Yes, it can refer to the REST asset URL directly, but the downloader still needs to send `Accept: application/octet-stream`; otherwise GitHub returns the asset metadata JSON rather than the ZIP, which leads to the same `BadZipFile` failure.

I used the browser release URL as the input because catalogs commonly get/store GitHub’s `browser_download_url`, which is easier to author and stable by tag + asset name. The REST asset URL requires the numeric asset id.

That said, I think it makes sense to support both. Have updated the PR so if `download_url` is already a GitHub REST asset URL like:

`[https://api.github.com/repos/<owner>/<repo>/releases/assets/<asset_id>`](https://api.github.com/repos/%3Cowner%3E/%3Crepo%3E/releases/assets/%3Casset_id%3E%60)

we skip the lookup and just download it with `Accept: application/octet-stream`. The browser URL path would continue to resolve to that REST URL internally.

mnriem:
Thank you!

PR #2803, opened 2026-06-02
fix: recover active skills registration for extensions

### LINKED ISSUE #2769: [Bug]: Misleading empty `registered_commands` and silent per-agent skip during `specify extension add`
## Correction note

This report was originally filed with the title "silently skips command registration when the target agent directory does not exist", and described the extension command as silently unavailable after install. That framing was based on reading the source only. After reproducing on spec-kit 0.8.16, the original framing is incorrect for skills-mode agents such as Claude: the command does register and is available. The accurate and smaller issue is about a misleading registry field and a silent per-agent skip. The corrected details are below.

## Bug Description

During `specify extension add`, the `.specify/extensions/.registry` field `registered_commands` can read `{}` (empty) even when the extension command was actually registered and is available to the agent. In addition, `register_commands_for_all_agents` skips an agent whose output directory does not exist without printing any warning. The result is that an empty `registered_commands` is not a reliable signal that registration failed, which is confusing when inspecting the registry or the install output.

## Steps to Reproduce (verified on 0.8.16)

Normal case:

1. `specify init . --ai claude --no-git`
2. `specify extension add <extension> --dev`

Result:

```
registered_commands: {'claude': ['speckit.archive.run']}
registered_skills:   []
.claude/skills/speckit-archive-run/SKILL.md      # created, command available
```

This works as expected.

Case where the agent skills directory is absent at install time (for example, the project was initialized for a different agent, or `.claude/skills/` was removed):

1. `specify init . --ai claude --no-git`
2. `rm -rf .claude/skills`
3. `specify extension add <extension> --dev`

Result:

```
registered_commands: {}                          # claude skipped silently
registered_skills:   ['speckit-archive-run']     # skills fallback created it
.claude/skills/speckit-archive-run/SKILL.md      # still created, still available
```

The installer prints "Extension installed successfully" and "1 agent skill auto-registered". The command is available either way.

## Expected Behavior

1. When `register_commands_for_all_agents` skips an agent because its output directory does not exist, the CLI should print a warning rather than skip silently.
2. The registry and the post-install summary should reflect the real registration state across both `registered_commands` and `registered_skills`, so an empty `registered_commands` is not mistaken for a failure.

## Actual Behavior

1. The per-agent skip is silent. In `src/specify_cli/agents.py`, `register_commands_for_all_agents` only registers when `agent_dir.exists()`:

   ```python
   for agent_name, agent_config in self.AGENT_CONFIGS.items():
       detect_dir_str = agent_config.get("detect_dir")
       if detect_dir_str:
           detect_path = project_root / detect_dir_str
           if not detect_path.exists():
               continue
       agent_dir = self._resolve_agent_dir(agent_name, agent_config, project_root)
       if agent_dir.exists():          # silent skip when absent
           ...
   ```

2. In `src/specify_cli/extensions.py`, `install_from_directory` stores whatever `register_commands_for_all_agents` returns, including `{}`, so the registry can show `registered_commands: {}` even when the skills path (`_register_extension_skills`) registered the command and created the SKILL.md.

3. In `src/specify_cli/__init__.py`, the success path lists `manifest.commands` and reports `registered_skills`, but never reflects an empty or partial `registered_commands`, so the registry field is a misleading signal.

## Impact

Low. This is mainly an observability and UX issue. For skills-mode agents such as Claude, the command is still registered and available (it appears as `/speckit-archive-run` under `.claude/skills/`). A genuinely unavailable command would require a non-skills agent whose command directory is absent at install time, with no skills fallback, which is a narrow case.

## Suggested Fix

- Print a warning when an agent is skipped during registration because its directory is absent.
- Make the registry and install summary reflect actual registrations across both `registered_commands` and `registered_skills`, so an empty `registered_commands` is not read as a failure.

## Environment

- Specify CLI Version: 0.8.16 (behavior also present on `main`, 0.8.19.dev0)
- AI Agent: Claude Code
- Operating System: macOS (Darwin Kernel 25.4.0, arm64)
- Python Version: 3.14.5

## Additional Context

This was originally surfaced through two community extension reports:

- https://github.com/stn1slv/spec-kit-archive/issues/2
- https://github.com/stn1slv/spec-kit-reconcile/issues/2

On 0.8.16, those extensions install and register correctly for Claude. The command appears as a skill under `.claude/skills/speckit-<name>-run/SKILL.md`. The original reporters appear to have inspected `registered_commands` and `.claude/commands/`, which are not where skills-mode registration lands on current versions.

minbang930:
## Description

Fixes #2769.

Extension command registration now resolves the initialized active skills-backed agent's skills directory before writing command artifacts. If that active skills directory is missing, extension install can recreate it safely and register the command artifact instead of leaving registry state out of sync with generated files.

The recovery path is limited to extension command registration. Preset registration keeps the existing directory-detection behavior, and shared skills directories are still guarded so unrelated agents are not marked as registered, including when the active shared-dir registration fails after recreating the directory.

## Testing

Additional details:

- `UV_CACHE_DIR=/tmp/uv-cache uv sync --extra test && UV_CACHE_DIR=/tmp/uv-cache uv run pytest -q` — `3312 passed, 40 skipped, 19 warnings in 74.72s`
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_extension_skills.py -q` — `40 passed in 0.38s`
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest -q` — `3313 passed, 40 skipped, 19 warnings in 75.10s`
- `UV_CACHE_DIR=/tmp/uv-cache uv run specify --help` — command completed and printed CLI help
- Sample project flow: `specify init <tmp>/sample --ai claude --no-git --ignore-agent-tools`, removed `.claude/skills`, then ran `specify extension add selftest`; registry recorded the Claude command and `.claude/skills/speckit-selftest-extension/SKILL.md` was created.
- `git diff --check HEAD~1..HEAD` — clean

## AI Disclosure

I used Codex to help reproduce the issue, draft the regression tests, and propose the implementation. I reviewed the diffs, checked the active/shared skills directory edge cases, ran the tests above, and refined the change before preparing this PR.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -86,6 +86,12 @@ from ._agent_config import (
     DEFAULT_INIT_INTEGRATION as DEFAULT_INIT_INTEGRATION,
     SCRIPT_TYPE_CHOICES as SCRIPT_TYPE_CHOICES,
 )
+from ._init_options import (
+    INIT_OPTIONS_FILE as INIT_OPTIONS_FILE,
+    is_ai_skills_enabled as _is_ai_skills_enabled,
+    load_init_options as load_init_options,
+    save_init_options as save_init_options,
+)
 
 app = typer.Typer(
     name="specify",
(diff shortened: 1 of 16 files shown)


minbang930:
Addressed the Copilot review feedback in b84bd53494ad7f059cbd23abddaebfa9145114ae:

- Set `active_created_skills_dir` immediately after the missing active skills-dir recovery path resolves the directory, so shared-dir agents are not treated as detected even if active registration later raises `OSError`.
- Added regression coverage for the shared `.agents/skills` case where active `agy` registration fails and `codex` must not register.
- Added explicit `encoding="utf-8"` to the test `read_text()` call.
- Re-ran `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_extension_skills.py -q` — `40 passed in 0.38s`.
- Re-ran `UV_CACHE_DIR=/tmp/uv-cache uv run pytest -q` — `3313 passed, 40 skipped, 19 warnings in 75.10s`.

The comment about `ExtensionManager.install_from_directory()` not using the new flag appears to be covered already. That path calls the extension wrapper `CommandRegistrar.register_commands_for_all_agents(...)`, which delegates with `create_missing_active_skills_dir=True`.

mnriem (changes requested):

minbang930:
Addressed the latest Copilot feedback in 673fc935: missing active skills-dir recovery now preserves Hermes’ configured global `~/.hermes/skills` output instead of redirecting registration to the project-local marker.

Local verification: targeted extension tests passed, and the full suite passed: 3374 passed, 40 skipped, 19 warnings.

minbang930:
Addressed the latest Copilot feedback in 24ae633: `detect_dir` markers now require `Path.is_dir()` so regular files do not count as integrations.
Local verification: targeted extension tests passed, and the full suite passed: 3375 passed, 40 skipped.

mnriem:
Please make sure checklist is after plan

minbang930:
Addressed the latest review feedback in 8a3f907: `checklist` now stays after `plan`, and unusable skills dirs skip registration instead of aborting install.

Local verification: targeted tests passed, and the full suite passed: 3378 passed, 40 skipped, 19 warnings.

mnriem (approved):

PR #2813, opened 2026-06-02
fix(workflows): validate run_id in RunState.load before touching the …

Quratulain-bilal:
## What

  `RunState.load(run_id, project_root)` interpolates the caller-supplied `run_id` directly into a filesystem path before
  validating it:

  ```python
  # src/specify_cli/workflows/engine.py:317
  @classmethod
  def load(cls, run_id: str, project_root: Path) -> RunState:
      """Load a run state from disk."""
      runs_dir = project_root / ".specify" / "workflows" / "runs" / run_id
      state_path = runs_dir / "state.json"
      if not state_path.exists():
          ...
      with open(state_path, encoding="utf-8") as f:
          state_data = json.load(f)
      state = cls(run_id=state_data["run_id"], ...)  # <-- validation only here
  ```

  `RunState.__init__` already validates `run_id` against `r'^[a-zA-Z0-9][a-zA-Z0-9_-]*$'`, but only on the stored
  `state_data["run_id"]` *after* the file lookup has already happened. The `run_id` argument passed to `load` is reachable
  from user input via the `specify workflow resume <run_id>` CLI argument, so a value like `../escape` lets
  `state_path.exists()` probe arbitrary paths and `json.load` parse a state file from outside `.specify/workflows/runs/`.

  ## Reproducer

  ```python
  import json, tempfile
  from pathlib import Path
  from specify_cli.workflows.engine import RunState

  with tempfile.TemporaryDirectory() as td:
      proj = Path(td) / "proj"; proj.mkdir(); (proj / ".specify").mkdir()
      runs = proj / ".specify" / "workflows" / "runs"; runs.mkdir(parents=True)

      # Plant a state.json *outside* the runs/ directory.
      escape = proj / ".specify" / "workflows" / "escape"; escape.mkdir()
      (escape / "state.json").write_text(json.dumps({
          "run_id": "pwned",
          "workflow_id": "attacker-owned",
          "status": "created",
      }))

      state = RunState.load("../escape", proj)
      print(state.workflow_id)  # 'attacker-owned' — traversal succeeded
  ```

  ## Impact

  - **File-existence oracle**: `state_path.exists()` reveals whether arbitrary paths the process can read exist.
  - **Arbitrary JSON read**: a `state.json` planted anywhere reachable from `runs_dir / <traversal>` is parsed and the
  workflow resumes under attacker-chosen `workflow_id` / step state.
  - **Persistence**: a subsequent `state.save()` writes back to the traversed location, persisting the corruption across
  resume cycles.

  Reachable from:
  - `specify workflow resume <run_id>` (CLI argument from `__init__.py:4239`)

  ## The change

  - Extract the existing `r'^[a-zA-Z0-9][a-zA-Z0-9_-]*$'` regex into a class-level `_RUN_ID_PATTERN` and a single
  `_validate_run_id` classmethod so `__init__` and `load` cannot drift.
  - Call `_validate_run_id(run_id)` at the top of `load` — *before* the path is built. A malicious value raises
  `ValueError` cleanly, which the CLI already handles at `__init__.py:4243` to surface a friendly `"Error: Invalid run_id
  ..."` message.
  - `__init__` refactored to call the same helper, so there is exactly one definition of "valid `run_id`" in the codebase.

  ## Precedent

  `src/specify_cli/agents.py:435-438` (`_ensure_within_directory`) already guards extension-install paths against the same
  threat model with `os.path.normpath` + `Path.is_relative_to`. This change brings `RunState` in line with that precedent.

  ## Tests

  `tests/test_workflows.py::TestRunState`:

  - `test_load_rejects_path_traversal` — parametrized over 9 attack vectors: `../escape`, `..`, `../../etc/passwd`,
  `foo/bar`, `foo\bar`, `.hidden`, `-flag`, `foo\x00bar`, and `""`. Each test plants a `state.json` outside the legitimate
  `runs/` directory so a missing guard would surface as a successful load rather than the ambiguous `FileNotFoundError`.
  - `test_init_and_load_share_validation` — asserts `__init__`, `load`, and the helper all reject the same representative
  malformed ID so future changes can't silently drift.

  ```
  $ python -m pytest tests/test_workflows.py::TestRunState -v
  13 passed in 2.79s
  $ python -m ruff check src/specify_cli/workflows/engine.py
  All checks passed!
  ```

  ## Notes for maintainers

  - Happy to move this to a private security advisory if you'd prefer to coordinate disclosure. Filed as a regular PR
  because the threat model is local (requires either CLI argument control or env-var control on the target machine, neither
   of which is a network-reachable attack surface) — but I'll defer to your preferred process.
  - No behaviour change for any valid `run_id`. The exact charset accepted before this PR is the exact charset accepted
  after.
  - The same regex also gates `SPECKIT_WORKFLOW_RUN_ID` (env-var override) via `__init__`'s call to the new helper, so the
  env-var override path is now consistently validated too.

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -298,7 +298,15 @@ class RunState:
         workflow_id: str = "",
         project_root: Path | None = None,
     ) -> None:
-        self.run_id = run_id or str(uuid.uuid4())[:8]
+        # ``run_id is None`` (omitted) → auto-generate. An explicit empty
+        # string is *not* the same as "omitted" and must be validated like
+        # any other caller-provided value — otherwise ``__init__("")``
+        # would silently substitute a UUID while ``load("")`` rejects, and
+        # the two entry points would diverge on the empty-string vector.
+        if run_id is None:
+            self.run_id = str(uuid.uuid4())[:8]
+        else:
+            self.run_id = run_id
         self._validate_run_id(self.run_id)
         self.workflow_id = workflow_id
         self.project_root = project_root or Path(".")
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -265,6 +265,33 @@ def _validate_steps(
 class RunState:
     """Manages workflow run state for persistence and resume."""
 
+    # ``run_id`` is interpolated into a filesystem path (``runs/<run_id>``)
+    # by both ``save()`` and ``load()``. Constrain it to a charset that
+    # cannot contain path separators (``/`` ``\``), parent-directory
+    # segments (``..``), or NULs — anything that could escape the
+    # ``.specify/workflows/runs/`` directory or be mis-interpreted by the
+    # filesystem. The first-character anchor blocks IDs that start with
+    # ``-`` (which would be mistaken for a CLI flag in error messages
+    # and shell completions).
+    _RUN_ID_PATTERN = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9_-]*$")
+
+    @classmethod
+    def _validate_run_id(cls, run_id: str) -> None:
+        """Raise ``ValueError`` if ``run_id`` is not a safe path component.
+
+        This is the single source of truth for what counts as a valid
+        ``run_id``. ``__init__`` calls it to reject malformed IDs at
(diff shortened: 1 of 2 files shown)


mnriem (approved):

mnriem:
Thank you!

PR #2826, opened 2026-06-03
refactor: move preset command handlers to presets/_commands.py (PR-6/8)

darion-yaphet:
## Summary

PR-6 of the `__init__.py` split (continues PR-5's integration co-location). Moves the `preset` command group out of `src/specify_cli/__init__.py` into a co-located `presets/_commands.py`, trimming `__init__.py` toward a thin registration layer.

Two commits:

1. **`presets.py` → `presets/` package** — pure structural move so command handlers can live beside the preset domain logic (mirrors `integrations/`). `presets.py` becomes `presets/__init__.py`; package-relative imports rebased one level deeper. Public import surface (`from specify_cli.presets import ...`) unchanged.
2. **Move handlers to `presets/_commands.py`** — the `preset_app` / `preset_catalog_app` Typer groups and all 12 handlers (`list`, `add`, `remove`, `search`, `resolve`, `info`, `set-priority`, `enable`, `disable`, `catalog list/add/remove`) move out; `__init__.py` registers them via `_register_preset_cmds(app)`.

`__init__.py` drops ~620 lines in the preset region.

## Design notes

- Handlers lazy-import root helpers (`_require_specify_project`, `get_speckit_version`, `_locate_bundled_preset`, `_display_project_path`) via `from .. import` inside each function, so test monkeypatching of `specify_cli.<helper>` keeps working — same pattern PR-5 used.
- `_locate_bundled_preset` kept as an explicit re-export in `__init__.py` for that resolution path.

## Test plan

- [x] `python -c "import specify_cli"` — OK
- [x] `specify preset --help` — full command surface intact
- [x] `uvx ruff check src/` — All checks passed
- [x] Full suite: **3344 passed, 40 skipped**
- [x] No patch-path updates needed (lazy imports preserve monkeypatch targets)


mnriem (changes requested):
Please pull in upstream/main and resolve conflicts

darion-yaphet:
**Scope change**: the URL-hardening commits previously on this branch have been extracted to #2911 so they can be reviewed and merged independently. This PR is now back to a pure code-move (2 commits).

While splitting, we found the moved copy of `preset_add` in `presets/_commands.py` was stale — it predated the GitHub release-asset resolution (`resolve_github_release_asset_api_url` + `Accept: application/octet-stream` header) that landed on `main`, so merging the old branch state would have silently reverted that behavior. The move commit has been amended so `_commands.py` now matches `main`'s `__init__.py` line-for-line (only import paths differ). Full suite: 3711 passed, 45 skipped.

Suggested merge order: #2911 first, then this PR rebased on top.

mnriem:
Setting to draft so it does NOT get merged before PR #2911

mnriem (approved):

PR #2843, opened 2026-06-04
test(workflows): cover executable override fallback preflight

WOLIKIMCHENG:
## Description

After #2539 landed, workflow command and prompt dispatch already include the `impl.key` -> `build_exec_args()[0]` fallback path during CLI availability preflight.

This updates #2843 to focus on regression coverage for the executable override case that originally motivated the change: when `impl.key` is not available on PATH, but `SPECKIT_INTEGRATION_<KEY>_EXECUTABLE` points to an available executable, workflow preflight should allow dispatch and the spawned argv should use the resolved override.

The remaining diff adds coverage for this path in both workflow command and prompt dispatch.

## Testing

- [x] Ran `git diff --check origin/main...HEAD`
- [x] Ran `git diff --check`
- [x] Ran `.venv/bin/pytest tests/test_workflows.py::TestCommandStep::test_dispatch_uses_executable_override_for_fallback_preflight tests/test_workflows.py::TestPromptStep::test_dispatch_uses_executable_override_for_fallback_preflight -v`
- [x] Ran `.venv/bin/pytest tests/test_workflows.py::TestCommandStep tests/test_workflows.py::TestPromptStep -v`
- [x] Ran `.venv/bin/pytest tests/test_workflows.py -v`
- [x] Ran `UV_CACHE_DIR=/tmp/uv-cache-spec-kit-pr2843 uvx ruff check src/ tests/test_workflows.py`
- [ ] Tested locally with `uv run specify --help` (not applicable; no CLI help path changed)
- [ ] Tested with a sample project (covered by workflow step regression tests)

### DIFF
--- a/tests/test_workflows.py
+++ b/tests/test_workflows.py
@@ -657,6 +657,47 @@ class TestCommandStep:
         # Claude is a SkillsIntegration so uses /speckit-specify
         assert "/speckit-specify login" in call_args[0][0][2]
 
+    def test_dispatch_uses_executable_override_for_fallback_preflight(self, tmp_path, monkeypatch):
+        """Command preflight falls back to build_exec_args() argv[0]."""
+        from unittest.mock import MagicMock, patch
+        from specify_cli.workflows.steps.command import CommandStep
+        from specify_cli.workflows.base import StepContext, StepStatus
+
+        monkeypatch.setenv("SPECKIT_INTEGRATION_CLAUDE_EXECUTABLE", "/opt/claude")
+        seen_which: list[str] = []
+
+        def fake_which(name: str) -> str | None:
+            seen_which.append(name)
+            return name if name == "/opt/claude" else None
+
+        step = CommandStep()
+        ctx = StepContext(
+            inputs={"name": "login"},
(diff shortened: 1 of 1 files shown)


mnriem (changes requested):

WOLIKIMCHENG:
Thanks for the review. I updated this branch against the current `origin/main` after #2539 landed, and narrowed the remaining change to regression coverage only.

Since the implementation now already includes the `impl.key` -> `build_exec_args()[0]` fallback path, this PR now focuses on covering the executable override case that originally motivated the change: when `impl.key` is not available on PATH, but `SPECKIT_INTEGRATION_<KEY>_EXECUTABLE` points to an available executable, workflow preflight should allow dispatch and the spawned argv should use the resolved override.

The updated tests cover this path for both workflow command and prompt dispatch. I also reran the focused workflow tests, the full `tests/test_workflows.py` file, `git diff --check`, and ruff on `src/` plus the updated test file.

mnriem (approved):

mnriem:
Thank you!

PR #2863, opened 2026-06-05
fix: register preset-provided namespaced commands

### LINKED ISSUE #2862: Bug: preset commands with 3-part names (speckit.<domain>.<cmd>) silently skipped during registration
## Summary

Preset commands that follow the `speckit.<domain>.<cmd>` naming convention are silently skipped during installation if `.specify/extensions/<domain>/` does not exist as a directory. This means any preset that introduces **new** commands under a three-part `speckit.*.*` namespace can never register them — the filter intended for extension command overrides is incorrectly applied to preset-provided commands.

Affected preset: [spec-kit-extended-flow](https://github.com/markuswondrak/spec-kit-extended-flow) — all three of its commands (`speckit.extendedflow.reviewer`, `speckit.extendedflow.documentation`, `speckit.extendedflow.documentation-init`) are dropped.

## Steps to Reproduce

1. Install any preset that declares new commands with a three-part `speckit.<domain>.<cmd>` name, where `<domain>` is **not** an installed extension ID.
2. Run `specify preset add --dev <path>` or install from a ZIP.
3. Check the agent command directories (e.g. `.opencode/commands/`, `.claude/commands/`).

**Expected:** The preset's command files are written to all detected agent directories.  
**Actual:** No command files are created. `specify preset add` reports success but silently drops all three-part commands.

Running the workflow later fails with:
```
Error: Command not found: "speckit.extendedflow.reviewer"
```

## Root Cause

Three locations in `src/specify_cli/presets.py` share identical filter logic:

| Location | Method | Line (v0.9.4) |
|---|---|---|
| Command registration | `_register_commands()` | 618–629 |
| Skill registration | `_register_skills()` | 1237–1247 |
| Post-install reconciliation | `install_from_directory()` | 1589–1602 |

All three contain:

```python
# Filter out extension command overrides if the extension isn't installed.
extensions_dir = self.project_root / ".specify" / "extensions"
filtered = []
for cmd in command_templates:
    parts = cmd["name"].split(".")
    if len(parts) >= 3 and parts[0] == "speckit":
        ext_id = parts[1]
        if not (extensions_dir / ext_id).is_dir():
            continue          # ← silently skips the command
    filtered.append(cmd)
```

The intent is correct for **extension command overrides** — commands a preset provides to wrap/replace a command that lives inside an extension (e.g. `speckit.git.feature` should only register when the `git` extension is installed). But the filter cannot distinguish:

- **Extension command overrides** — wrap an existing extension command; should be skipped when the extension is absent.
- **Preset-provided commands** — brand-new commands introduced by the preset; should always register.

Both use three-part names, so all three-part names are gated on an extension directory that will never exist for a purely preset-provided command.

## Impact

- Any preset that introduces new agent commands under a `speckit.<domain>.<cmd>` namespace is broken out of the box.
- Failure is completely silent — no warning, no error, install reports success.
- All downstream workflow steps that invoke those commands fail at runtime.

## Workaround (until upstream fix)

Create an empty stub directory to satisfy the filter, then reinstall:

```bash
mkdir -p .specify/extensions/extendedflow
specify preset remove spec-kit-extended-flow
specify preset add --dev ../spec-kit-extended-flow
```

The filter only calls `is_dir()` — an empty directory passes.

## Suggested Fix

The filter needs to distinguish extension command overrides from preset-provided commands. One clean approach: check whether the `<domain>` segment matches the name of any installed extension. If not, the command is preset-provided and should register unconditionally. A minimal, backward-compatible patch in all three locations:

```python
for cmd in command_templates:
    parts = cmd["name"].split(".")
    if len(parts) >= 3 and parts[0] == "speckit":
        ext_id = parts[1]
        ext_dir = extensions_dir / ext_id
        if not ext_dir.is_dir():
            # Only skip if this command overrides an extension command (i.e.
            # an extension with this ID exists elsewhere / was once installed).
            # Pure preset-provided commands (no matching extension) always register.
            if not any(
                (self.project_root / ".specify" / "extensions" / ext_id).parent.exists()
                for _ in [None]
            ):
                pass  # falls through to filtered.append(cmd)
            # For a cleaner fix: gate only on cmd.get('strategy') != 'replace'
            # (extension overrides typically use wrap/prepend/append).
    filtered.append(cmd)
```

The simplest safe heuristic: **if no extension directory with that ID exists anywhere in the installed extension set, treat the command as preset-provided and always register it.**

## Spec-Kit Version

**v0.9.4** (latest release). The same filter code is present in `main` (v0.9.5.dev0) — **not yet fixed**.

## References

- Affected preset: https://github.com/markuswondrak/spec-kit-extended-flow
- Affected file: `src/specify_cli/presets.py`, lines 618–629, 1237–1247, 1589–1602

wondr-wclabs:
## Summary

Fixes #2862.

Preset command registration was treating every `speckit.<domain>.<command>` name as an extension command override. That made valid preset-provided commands, such as `speckit.extendedflow.reviewer`, disappear whenever `.specify/extensions/<domain>/` did not exist.

This changes the missing-extension guard so it applies only when the preset is actually targeting an extension command:

- the template explicitly declares `replaces: speckit.<ext-id>.<command>`; or
- the template uses a composition strategy (`prepend`, `append`, or `wrap`) that needs a lower-priority base command.

A default `replace` command with a three-part name and no `replaces` entry is now treated as preset-provided and is registered normally. The existing extension-override behavior is preserved by making the override tests explicit through `replaces`.

## Validation

- `uv sync --extra test`
- `uv run python -m pytest tests/test_presets.py -k "extension_command or three_part_preset_command" -q` — 6 passed
- `uv run python -m pytest tests/test_presets.py -q` — 273 passed
- `uv run specify --help`
- `git diff --check`
- `uv run python -m pytest -q` — 3576 passed, 45 skipped

## AI assistance disclosure

This PR was developed with AI assistance from Codex/ChatGPT for codebase inspection, implementation, and validation. I reviewed the resulting diff and test output before submitting.

### DIFF
--- a/presets/ARCHITECTURE.md
+++ b/presets/ARCHITECTURE.md
@@ -67,11 +67,11 @@ When a preset is installed with `type: "command"` entries, the `PresetManager` r
 flowchart TD
     A["specify preset add my-preset"] --> B{Preset has type: command?}
     B -- No --> Z["done (templates only)"]
-    B -- Yes --> C{Extension command?}
-    C -- "speckit.myext.cmd\n(3+ dot segments)" --> D{Extension installed?}
+    B -- Yes --> C{Targets extension override?}
+    C -- "replaces speckit.myext.cmd\nor composition strategy" --> D{Extension installed?}
     D -- No --> E["skip (extension not active)"]
     D -- Yes --> F["register command"]
-    C -- "speckit.specify\n(core command)" --> F
+    C -- "preset-provided command\nor core command" --> F
     F --> G["detect agent directories"]
     G --> H[".claude/commands/"]
     G --> I[".gemini/commands/"]
@@ -89,7 +89,7 @@ flowchart TD
 
 ### Extension safety check
 
-Command names follow the pattern `speckit.<ext-id>.<cmd-name>`. When a command has 3+ dot segments, the system extracts the extension ID and checks if `.specify/extensions/<ext-id>/` exists. If the extension isn't installed, the command is skipped — preventing orphan files referencing non-existent extensions.
+Preset-provided commands can use namespaced command IDs such as `speckit.extendedflow.reviewer`, so dot-count alone does not make a command an extension override. The missing-extension guard applies when a command explicitly declares `replaces: speckit.<ext-id>.<cmd-name>` or when it uses a composition strategy (`prepend`, `append`, or `wrap`) that needs a lower-priority base command. In those cases, the system extracts the extension ID and checks if `.specify/extensions/<ext-id>/` exists. If the extension isn't installed, the command is skipped — preventing orphan files referencing non-existent extensions.
 
 Core commands (e.g. `speckit.specify`, with only 2 segments) are always registered.
 
--- a/src/specify_cli/presets.py
+++ b/src/specify_cli/presets.py
@@ -587,6 +587,40 @@ class PresetManager:
 
         return True
 
+    def _should_skip_missing_extension_command(
+        self,
+        command_template: Dict[str, Any],
+    ) -> bool:
+        """Return True when a preset command targets an absent extension.
+
+        Three-part command names are not enough to identify extension
+        overrides because presets can define their own namespaced commands,
+        such as ``speckit.extendedflow.reviewer``. Only commands that
+        explicitly replace an extension command, or composition strategies
+        that need a lower-priority base command, are gated on extension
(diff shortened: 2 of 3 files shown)


mnriem:
A preset is not supposed to deliver commands. If you are delivering commands you should be delivering an extension, not a preset

PR #2866, opened 2026-06-05
Add Claude no-model-invocation integration option

### LINKED ISSUE #2654: Allow opting out of model invocation for installed skills (large monorepo skill-budget pain)
## Problem

Since v0.8.4 ([PR #2099](https://github.com/github/spec-kit/pull/2099), closing [#2098](https://github.com/github/spec-kit/issues/2098)), `specify init --ai claude` writes `disable-model-invocation: false` on every generated SKILL.md, and the toggle was removed as "redundant" because the new default opts skills into auto-invocation. There is no opt-out flag.

This is a real problem in large monorepos that already register many domain skills. Claude Code reserves a small fraction of context for the skill-listing budget; when that budget is exhausted, lower-priority skills get truncated and Claude loses visibility on them entirely. Adding the 11 spec-kit skills as model-invocable pushes us over the limit and silently drops other skills we use far more often.

Two specific pain points:

1. **Skill-listing budget exhaustion.** With ~50 skills already in play (CI helpers, MR workflows, service tooling, etc.), the spec-kit 11 push us past the default `skillListingBudgetFraction`. The cheapest fix is to make spec-kit skills user-invocable only, so they cost nothing in the listing.
2. **Over-triggering on generic prompts.** Several spec-kit skill descriptions (`speckit-implement`, `speckit-tasks`, `speckit-plan`) are generic enough that Claude auto-invokes them on prompts that have nothing to do with SDD — e.g. someone asking to "implement" or "plan" something unrelated.

## Workaround

We've patched our internally bundled copy to set `disable-model-invocation: true` on all 11 skills. But every fresh `specify init` reverts this, so we either lose the fix on upgrade or have to maintain a long-running patch.

## Proposal

Add a `--no-model-invocation` flag to `specify init` (and persist to the init-options file) that emits `disable-model-invocation: true` in generated SKILL.md frontmatter. Default behaviour stays as-is (model-invocable). Opt-in users get manual `/speckit.*` invocation only.

Alternatively, a per-skill override in a config file would also solve it.

Happy to send a PR if the direction is welcome.

wondr-wclabs:
Fixes #2654.

## Summary

Adds a Claude-specific `--no-model-invocation` integration option so users can install Spec Kit's Claude skills as user-invocable only:

```bash
specify init my-project --integration claude --integration-options="--no-model-invocation"
```

The option defaults to the current behavior, so generated Claude skills still use `disable-model-invocation: false` unless the flag is explicitly supplied.

## Rationale

The main issue in #2654 is Claude's skill-listing budget in larger monorepos. Only changing the core Claude `setup()` path would leave skills generated later by the bundled `agent-context` extension, extension commands, or preset overrides as model-invocable, which partially defeats the opt-out. This threads parsed integration options through shared skill post-processing so the same Claude flag is applied consistently to:

- core Spec Kit Claude skills generated during init/upgrade
- bundled extension skills such as `speckit-agent-context-update`
- extension and preset-generated SKILL.md files

The raw and parsed integration options are also mirrored into `init-options.json`, while `integration.json` remains the structured integration-state source used by install/upgrade/switch flows. That lets later skill writers read the active option without needing to know the integration-state schema directly.

## Validation

- `uv sync --extra test`
- `uv run python -m pytest tests/integrations/test_integration_claude.py -q`
- `uv run python -m pytest tests/test_extension_skills.py -k "no_model_invocation or skill_md_has_parseable_yaml" -q`
- `uv run python -m pytest tests/integrations/test_integration_subcommand.py -q`
- `uv run python -m pytest tests/test_extension_skills.py -q`
- `uv run python -m pytest tests/integrations/test_integration_copilot.py tests/integrations/test_integration_base_skills.py -q`
- `uv run python -m pytest tests/test_presets.py -q`
- `uv run specify --help`
- `git diff --check`
- `uv run python -m pytest -q` (3609 passed, 45 skipped)

## AI disclosure

This PR was developed with assistance from OpenAI Codex. I reviewed the implementation and validation results before submitting.

### DIFF
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -335,7 +335,30 @@ class CommandRegistrar:
             description,
             f"{source_id}:{source_file}",
         )
-        return self.render_frontmatter(skill_frontmatter) + "\n" + body
+        skill_content = self.render_frontmatter(skill_frontmatter) + "\n" + body
+
+        from . import load_init_options
+        from .integrations import get_integration
+
+        init_options = load_init_options(project_root)
+        parsed_options = (
+            init_options.get("integration_parsed_options")
+            if isinstance(init_options, dict)
+            else None
+        )
+        if not isinstance(parsed_options, dict):
+            parsed_options = None
+        integration = get_integration(agent_name)
+        if integration is not None and hasattr(
+            integration,
+            "post_process_skill_content",
+        ):
+            skill_content = integration.post_process_skill_content(
+                skill_content,
+                parsed_options=parsed_options,
+            )
+
+        return skill_content
 
     @staticmethod
     def build_skill_frontmatter(
--- a/src/specify_cli/commands/init.py
+++ b/src/specify_cli/commands/init.py
@@ -459,6 +459,10 @@ def register(app: typer.Typer) -> None:
                 from ..integrations.base import SkillsIntegration as _SkillsPersist
                 if isinstance(resolved_integration, _SkillsPersist) or getattr(resolved_integration, "_skills_mode", False):
                     init_opts["ai_skills"] = True
+                if integration_options:
+                    init_opts["integration_options"] = integration_options
+                    if integration_parsed_options:
+                        init_opts["integration_parsed_options"] = integration_parsed_options
                 save_init_options(project_path, init_opts)
 
                 # --- agent-context extension (bundled, auto-installed) ---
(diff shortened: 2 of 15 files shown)


mnriem (changes requested):
As this is Claude specific it should not touch any other files than Claude integration specific files

PR #2882, opened 2026-06-06
fix: self-heal agent-context extension

### LINKED ISSUE #2881: [Bug]: agent-context extension is only installed by 'specify init' — other paths write an inert config that stops updating context files at v0.12.0
## Bug Description

`specify init` is the only code path that installs and registers the bundled `agent-context` extension (`commands/init.py:464`). The `specify integration install`, `integration switch`, and `integration upgrade` commands write/refresh its config file (`.specify/extensions/agent-context/agent-context-config.yml`) **without** installing the extension.

So a project managed through the integration commands ends up with an inert config: the `.yml` exists, but the extension is absent on disk and unregistered. Context files still get updated today only because the deprecated inline updater runs while the extension is missing. When inline updates are removed in **v0.12.0**, a project in this state stops updating its context files.

> Note: re-running `specify init --here --force` **does** back-fill the extension (the init block is idempotent via `is_installed`), so a full re-init recovers. The gap is that the integration commands don't, and they silently drop an inert config in the meantime — which is easy to miss.

## Steps to Reproduce

1. Use a project without the `agent-context` extension (e.g. initialized before PR #2546).
2. Run `specify integration install <agent>` (or `switch` / `upgrade`).
3. Observe: `.specify/extensions/agent-context/agent-context-config.yml` is written, but `.specify/extensions/.registry` has no `agent-context` entry and `.specify/extensions/agent-context/` contains only the `.yml` (no `extension.yml`/`scripts/`/`commands/`).

## Expected Behavior

`integration install` / `switch` / `upgrade` should ensure the agent-context extension is installed and registered (the same back-fill `specify init` already does), or not write its config when the extension is absent. A project should never be left with a config that nothing reads.

## Actual Behavior

The config is written unconditionally (`_save_agent_context_config` does `mkdir`+`write` with no registration check); the extension is installed only by `specify init`. The deprecation notice claims the extension manages context, but the integration commands leave it absent — an inert config now, and frozen context files after v0.12.0 unless the user happens to re-run `init --force`.

## Specify CLI Version

0.9.3 (root cause present on `main` @ `7106858`)

## AI Agent

Claude Code

## Additional Context

References on `main` (@ `7106858`):
- `commands/init.py:464` — only installer of `agent-context` (idempotent; re-init back-fills it).
- `integrations/_install_commands.py` — install path: no extension install; only calls `_update_init_options_for_integration`.
- `integrations/_migrate_commands.py` — `switch` only re-registers already-installed extensions (`register_enabled_extensions_for_agent`, `:278`); `upgrade` (`:339`) runs shared-infra + config update but never installs the extension.
- `integrations/_helpers.py:308-321` — writes the config even when, per its own comment, "the extension is not installed".
- `__init__.py:300` (`_save_agent_context_config`) — `mkdir`+`write`, no registration check.
- `integrations/base.py:587` — inline path still runs while the extension is absent (masks the bug until v0.12); `:699` prints the deprecation notice.

Reproduced on a real 0.9.3 project: registry lists only `git`; `agent-context/` holds just the 101-byte config.

Related: #2879 (multi-file divergence); #2319 / #2320 (`--force` shared-infra overwrite — different scope, predates this extension).

nkgotcode:
## Description

Moves the `agent-context` extension repair path out of integration helpers and into the top-level CLI project startup path. Older projects now restore the bundled extension on normal CLI invocation while integration commands only update the active context config.

Fixes #2881.

## Testing

- `uv run python -m pytest tests/test_agent_config_consistency.py -q`
- `uv run python -m pytest tests/extensions/test_extension_agent_context.py::TestExtensionEnabledGate tests/extensions/test_extension_agent_context.py::TestAgentContextSelfHeal tests/integrations/test_integration_subcommand.py::TestIntegrationInstall::test_install_backfills_agent_context_extension_when_missing tests/integrations/test_integration_subcommand.py::TestIntegrationSwitch::test_switch_backfills_agent_context_extension_when_missing tests/integrations/test_integration_subcommand.py::TestIntegrationUpgrade::test_upgrade_backfills_agent_context_extension_when_missing -q`

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -377,13 +377,18 @@ def _self_heal_agent_context_extension(project_root: Path) -> None:
 
     bundled_ac = _locate_bundled_extension("agent-context")
     if bundled_ac is None:
-        raise ValueError("bundled agent-context extension not found")
+        return
 
-    ext_mgr.install_from_directory(
-        bundled_ac,
-        get_speckit_version(),
-        force=ext_mgr.registry.is_installed("agent-context"),
-    )
+    try:
+        ext_mgr.install_from_directory(
+            bundled_ac,
+            get_speckit_version(),
+            force=ext_mgr.registry.is_installed("agent-context"),
+        )
+    except Exception:
+        if existing_cfg is not None:
+            _save_agent_context_config(project_root, existing_cfg)
+        return
 
     if existing_cfg is not None:
         _save_agent_context_config(project_root, existing_cfg)
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -109,6 +109,8 @@ def callback(
     version: bool = typer.Option(False, "--version", "-V", callback=_version_callback, is_eager=True, help="Show version and exit."),
 ):
     """Show banner when no subcommand is provided."""
+    if ctx.invoked_subcommand != "init":
+        _self_heal_agent_context_extension(Path.cwd())
     if ctx.invoked_subcommand is None and "--help" not in sys.argv and "-h" not in sys.argv:
         show_banner()
         console.print(Align.center("[dim]Run 'specify --help' for usage information[/dim]"))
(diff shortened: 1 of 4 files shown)


mnriem (changes requested):
The `agent-context` extension should self-heal at the extension-system level — not get backfilled by integration flows. Integrations are consumers of the config API, not owners of extension lifecycle.

How it should work:
1. `agent-context` auto-enables and self-heals at a top-level CLI hook (or during `specify init`)
2. Integration layer just writes to `agent-context-config.yml`, assumes the extension is there
3. Older projects get fixed on any CLI invocation, not only on integration install/switch/upgrade

Can you move the backfill logic out of `_helpers.py` and into the extension system instead?

nkgotcode:
Moved this to the top-level CLI startup path in `aac1f2c`: non-`init` commands now call `_self_heal_agent_context_extension(Path.cwd())`, and integrations only update `agent-context-config.yml`.

Rebased onto current `main`; the branch is now mergeable. Validation passed:
- `uv run python -m pytest tests/test_agent_config_consistency.py -q` -> 28 passed
- `uv run python -m pytest tests/extensions/test_extension_agent_context.py::TestExtensionEnabledGate tests/extensions/test_extension_agent_context.py::TestAgentContextSelfHeal tests/integrations/test_integration_subcommand.py::TestIntegrationInstall::test_install_backfills_agent_context_extension_when_missing tests/integrations/test_integration_subcommand.py::TestIntegrationSwitch::test_switch_backfills_agent_context_extension_when_missing tests/integrations/test_integration_subcommand.py::TestIntegrationUpgrade::test_upgrade_backfills_agent_context_extension_when_missing -q` -> 9 passed
- `git diff --check origin/main...HEAD` -> exit 0

mnriem:

nkgotcode:
Addressed the review feedback in nkgotcode/spec-kit@3196cbc.\n\nValidated locally:\n- `uv sync --extra test`\n- `uv run pytest tests/extensions/test_extension_agent_context.py::TestAgentContextSelfHeal tests/extensions/test_extension_agent_context.py::TestExtensionConfigWriters -q`\n- `uv run pytest tests/integrations/test_integration_subcommand.py::TestIntegrationInstall::test_install_backfills_agent_context_extension_when_missing tests/integrations/test_integration_subcommand.py::TestIntegrationInstall::test_install_non_default_refreshes_init_options_version_only tests/integrations/test_integration_subcommand.py::TestIntegrationUpgrade::test_upgrade_non_default_refreshes_init_options_version_only -q`\n- `uv run specify --help`

nkgotcode:
Addressed the follow-up review notes in `b6d77be`: self-heal now skips when the bundled extension is unavailable, restores prior config on install failure, and remains side-effect free for help output.

Validated locally:
- `uv run pytest tests/extensions/test_extension_agent_context.py::TestAgentContextSelfHeal tests/extensions/test_extension_agent_context.py::TestExtensionConfigWriters -q` -> 14 passed
- `uv run pytest tests/integrations/test_integration_subcommand.py::TestIntegrationInstall::test_install_backfills_agent_context_extension_when_missing tests/integrations/test_integration_subcommand.py::TestIntegrationInstall::test_install_non_default_refreshes_init_options_version_only tests/integrations/test_integration_subcommand.py::TestIntegrationUpgrade::test_upgrade_non_default_refreshes_init_options_version_only -q` -> 3 passed
- `uv run specify --help` -> exit 0
- `git diff --check origin/main...HEAD` -> exit 0

mnriem:
Closed by #2882

mnriem:
Actually closed by #2885

mnriem:
Thank you for working on this. In spite of another PR making it in before this one. Sometimes that happens and with the influx of so many issues and PRs this sometimes happens. Again thank you!

PR #2884, opened 2026-06-06
fix: ensure installed files are owner-writable regardless of source permissions

rayhem:
## Description

`shutil.copy2` and `shutil.copytree` (which uses `copy2` by default) propagate permission bits from source to destination. When the source lives on a read-only filesystem — the Nix store, any read-only mount, or a permission-restricted directory — installed files land at `0o444` and directories at `0o555`. Any subsequent write to `.specify/` (re-running `specify init`, upgrading, editing installed config) then fails with `PermissionError`.

An install operation should produce owner-writable destinations. The installed file's mtime should also reflect when it was installed, not when the bundled asset was built — so the `copy2` → `copyfile` change is correct on both counts.

I encountered this problem attempting to package spec-kit on NixOS. The repository gets added read-only to the Nix store. `specify` assumes it can write to the files it creates, but the copy utilities preserve the read-only permissions resulting in an error.

## Testing

## AI Disclosure

Fix devised and generated by Claude Opus 4.6

### DIFF
--- a/src/specify_cli/_utils.py
+++ b/src/specify_cli/_utils.py
@@ -149,13 +149,13 @@ def handle_vscode_settings(sub_item, dest_file, rel_path, verbose=False, tracker
             else:
                 log("Skipped merge (preserved existing settings)", "yellow")
         else:
-            shutil.copy2(sub_item, dest_file)
+            shutil.copyfile(sub_item, dest_file)
             log("Copied (no existing settings.json):", "blue")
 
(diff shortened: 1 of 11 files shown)


mnriem (changes requested):
Please add some positive and negative tests

mnriem:

mnriem:

mnriem (changes requested):

PR #2885, opened 2026-06-06
fix(integrations): don't write an inert agent-context config when the extension is absent

### LINKED ISSUE #2881: [Bug]: agent-context extension is only installed by 'specify init' — other paths write an inert config that stops updating context files at v0.12.0

PascalThuet:
Fixes #2881.

Per your review, I dropped the back-fill approach (no more auto-install on install/switch/upgrade) and cut this down to the underlying bug.

`integration install`/`switch`/`upgrade` go through `_update_init_options_for_integration`, which wrote `agent-context-config.yml` whenever the config file already existed and created it from defaults otherwise. A project without the extension ended up with an inert config nothing reads. It now writes that config only when the extension is actually registered (`registry.is_installed("agent-context")`). Nothing gets installed or back-filled, and `init` is unchanged.

Tests (`pytest tests/extensions tests/integrations tests/test_agent_config_consistency.py`): 2381 passed, 15 skipped. Added unit + end-to-end coverage for both cases (extension absent, extension present).

Left out, can split into follow-ups:
- `is_installed` doesn't check the `enabled` flag, so a disabled-but-installed extension still gets its config updated.
- `init` still writes the config if the extension install itself fails (pre-existing).

### DIFF
--- a/src/specify_cli/integrations/_helpers.py
+++ b/src/specify_cli/integrations/_helpers.py
@@ -285,11 +285,11 @@ def _update_init_options_for_integration(
     falls back to the class-level defaults.
     """
     from .. import (
-        _AGENT_CTX_EXT_CONFIG,
         _update_agent_context_config_file,
         load_init_options,
         save_init_options,
     )
+    from ..extensions import ExtensionManager
     from .base import SkillsIntegration
     opts = load_init_options(project_root)
     opts["integration"] = integration.key
@@ -307,20 +307,17 @@ def _update_init_options_for_integration(
 
     # Update the agent-context extension config BEFORE init-options.json
     # so a failure here doesn't leave init-options partially updated.
-    ext_cfg_path = project_root / _AGENT_CTX_EXT_CONFIG
-    if ext_cfg_path.exists():
+    #
+    # Only touch the config when the agent-context extension is installed and
+    # registered.  Keying off the registry (rather than the config file's
+    # presence) means a project without the extension isn't handed an inert
+    # config that nothing reads, and a stale file left by an older version
+    # isn't perpetuated (see #2881).
+    if ExtensionManager(project_root).registry.is_installed("agent-context"):
         _update_agent_context_config_file(
             project_root,
             integration.context_file,
             preserve_markers=True,
-        )
-    elif integration.context_file:
-        # Extension config doesn't exist yet (extension not installed).
-        # Write defaults so scripts have something to read.
-        _update_agent_context_config_file(
-            project_root,
-            integration.context_file,
-            preserve_markers=False,
         )
 
     save_init_options(project_root, opts)
--- a/tests/extensions/test_extension_agent_context.py
+++ b/tests/extensions/test_extension_agent_context.py
@@ -335,8 +335,10 @@ class TestExtensionConfigWriters:
     def test_update_init_options_writes_context_file_to_ext_config(self, tmp_path):
         from specify_cli import _update_init_options_for_integration
 
-        # Pre-create the extension config so _update_init_options_for_integration
-        # updates it (rather than skipping it when ext config doesn't exist yet).
+        # The extension config is only managed when the extension is installed/
+        # registered; register it and pre-create its config so the call updates
+        # it (an absent extension is left alone — see #2881).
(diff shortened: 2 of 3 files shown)


mnriem (changes requested):
We are phasing using agent context files completely and the extension will only be available as an opt-in for single agent install. We are NOT going to support it for multi install and it will be entirely an opt-in at 0.0.12. So please limit to scope to that

PascalThuet:
Thanks. Dropped the back-fill, so there's no auto-install on any path now. This just stops install/switch/upgrade from leaving an inert `agent-context-config.yml` when the extension isn't installed: the write now keys off `registry.is_installed("agent-context")` instead of the config file existing. `init` is unchanged. Can narrow it further if you'd prefer.

mnriem (commented):
Thanks for the investigation. The scenario this addresses — agent-context config written when the extension is absent — only occurs when a user upgrades the CLI but skips the documented project update step (`specify init --here --force --integration <agent>`). Our upgrade guide already covers this two-step process, and the worst case is an inert YAML file that causes no harm.

Additionally, agent-context is moving to fully opt-in at 0.0.12, so adding defensive infrastructure around it is working against the direction we're heading.

PascalThuet:
Makes sense. If the inert config is harmless and agent-context is going fully opt-in at 0.0.12, this isn't worth the added surface. Closing this out. Thanks for the context.

PR #2901, opened 2026-06-09
fix(extensions): make mandatory-hook dispatch contract self-contained

### LINKED ISSUE #2730: [Bug]: PR #2713's EXECUTE_COMMAND directive emits correctly but is never executed in pure-Claude-Code invocations
### Follow-up to #2688

PR #2713 (shipped in v0.8.15) restructured the after_*/before_* hook dispatch in core command SKILLs with a `## Mandatory Post-Execution Hooks` H2 + directive language + `## Done When` checklist. Empirically, **the emission half works**: the agent now reliably emits

> EXECUTE_COMMAND: my-extension.my-hook-target

in its response stream during `/speckit-plan` invocations on v0.8.15+.

**The execution half does not work in pure Claude Code (no `specify`-CLI wrapper).** Nothing in Claude Code's slash-command runtime watches the agent's stdout for `EXECUTE_COMMAND:` lines and dispatches them. The agent emits the directive, waits ~6 seconds, and returns control without ever invoking the underlying hook script. The result is functionally identical to #2688's original failure mode (the registered hook does not run) — just with a more visible symptom (the directive shows in the agent's output, so the user can SEE that something should have happened).

### Reproduction

The repro is structural and doesn't depend on any particular extension. A minimal stub is enough:

1. `specify init . --integration claude` on spec-kit ≥ 0.8.15.
2. Hand-create a minimal local extension at `.specify/extensions/repro/`:
   - `extension.yml`:
     ```yaml
     schema_version: "1.0"
     extension:
       id: repro
       name: "Hook Dispatch Repro"
       version: "0.0.1"
     requires:
       speckit_version: ">=0.8.15"
     provides:
       commands:
         - name: repro.before-plan
           file: commands/before-plan.md
           description: "Hook dispatch repro target."
     hooks:
       before_plan:
         command: repro.before-plan
         optional: false
         description: "Should run before /speckit-plan."
     ```
   - `commands/before-plan.md`: any content; the command's body never runs in this repro so the contents don't matter beyond being present.
3. Run `specify extension add --dev .specify/extensions/repro` so the hook is registered in `.specify/extensions.yml` and a SKILL is generated under `.claude/skills/repro-before-plan/`.
4. From inside Claude Code (NOT via `specify run` or any CLI wrapper), invoke `/speckit-plan` against any feature directory.
5. Observe: the agent emits the H2 block with `EXECUTE_COMMAND: repro.before-plan` correctly, then stalls. The hook command body never runs. `/speckit-plan` then proceeds (or fails, depending on what the hook was supposed to do) as if no hook were registered.

Empirical log grep on a real session against a `before_plan` hook that's supposed to write a file:

```
grep -E 'EXECUTE_COMMAND|tool_dispatch_start tool=Bash.*<hook-script>' debug.log
# → emits EXECUTE_COMMAND line, zero matching Bash dispatches
```

### Why this is distinct from #2104

#2104 asked for **auto-run** as a new feature and was closed because the workflow engine is the right answer for deterministic dispatch. This issue is narrower: the **existing best-effort dispatch contract documented in PR #2713 doesn't actually dispatch** in the most common Claude Code invocation pattern. I am not asking for auto-run; I am asking that the contract the SKILL describes actually completes in the documented happy path.

### Hypothesis

The v0.8.15 SKILL implicitly assumes a platform-level stdout watcher that parses `EXECUTE_COMMAND:` lines. That watcher exists when `specify` is the orchestrator (`auto_execute_hooks: true` in `.specify/extensions.yml` is presumably the flag that activates it). It does not exist when the agent's slash command is invoked directly in Claude Code, which is the default invocation pattern for ~all extension users.

### Proposed Resolution Paths (one of)

1. **Strengthen the SKILL contract.** Add an explicit instruction immediately after the `EXECUTE_COMMAND:` directive emission: *"You MUST then invoke the Bash tool with the underlying hook script path to actually execute the command. Emitting the directive alone is not sufficient — the directive is a ceremony signal, not a dispatcher."* This is the lowest-cost fix and makes the contract self-contained.

2. **Document the prerequisite explicitly.** Add a note to `extensions/EXTENSION-API-REFERENCE.md` that `EXECUTE_COMMAND:` directives require an orchestrator process (e.g. `specify run`) and will not be dispatched by pure Claude Code slash invocations. Extension authors can then advertise the dispatch limitation in their READMEs.

3. **Ship a Claude Code hook.** Bundle a small `settings.json` hook (CLAUDE_CODE_HOOKS_DIR) that watches assistant message text for `EXECUTE_COMMAND: <command>` lines and dispatches them via the Bash tool — closing the gap at the platform level without depending on the agent doing two things.

(1) is the cleanest from an extension-author perspective. (3) is the most robust but adds non-trivial machinery.

### Component

Specify CLI (commands, templates) / Extensions / Agent Integration

### AI Agent (if applicable)

Claude Code (Claude Opus 4.7) — but pattern is structural; same failure mode is likely with Cursor / Codex / Gemini CLI when invoked directly without a spec-kit CLI orchestrator.

### Use Cases

Every extension that registers an `after_*` or `before_*` hook with `optional: false` and expects it to fire during agent-direct slash invocations. Concrete classes:

- Spec-format adapters whose `before_plan` materializes a transient `spec.md` from a non-default spec store, so `/speckit-plan` has something to read; without dispatch, `/speckit-plan` has nothing to read and silently bails.
- Enterprise linting / AC-format validation pre-implement gates (mentioned in #1788).
- Pre-implementation policy / compliance checks.

### References

- #2688 — the original bug; PR #2713's fix closed the emission half
- PR #2713 — the structural restructuring that landed in v0.8.15
- #2104 — closed feature request for auto-run; relevant context, **not** a duplicate
- #1788 — original hook feature request

Quratulain-bilal:
Fixes #2730

## Problem

The `EXECUTE_COMMAND:` directive emitted for `optional: false` hooks is meant as a dispatch signal. In agent-direct invocations (a slash command run inside Claude Code, Cursor, Codex, etc. - the default pattern for extension users), nothing watches for it. The agent emits the directive, waits, and moves on - the hook never runs, and the failure is completely silent.

I could not find any code path in this repository that consumes the directive: `format_hook_message()`'s only caller is `check_hooks_for_event()`, which itself has no production callers. So in practice the bundled command templates are what drive hook emission - the agent reads `.specify/extensions.yml` and emits the block by following the template instructions. That makes the templates the right place to fix this.

## Changes

- `templates/commands/*.md` (all 10 command templates): added one instruction line after each mandatory-hook block (pre- and post-execution, 20 sites). It tells the agent that emitting the block alone does not run the hook - it must actually invoke the hook and wait for completion before continuing.
- The instruction deliberately does not hardcode `/{command}`. It tells the agent to run the hook the way it would run the command itself in the current agent/session, and notes the invocation may differ from the literal `{command}` id shown in the block (e.g. skills-mode agents run it as `/skill:speckit-...` or `$speckit-...`). This keeps it correct outside the default slash-command form.

## Scope

This is the minimal version of an earlier, broader change, per maintainer feedback to keep the surface small. The example blocks are left untouched; only a single instruction line is added per mandatory-hook block. No runtime code, API reference, or tests are changed - the fix is entirely in the bundled template instructions the agent follows.

## AI disclosure

This PR was authored with AI assistance (Claude Code); the change and this description are under my direction and review.

### DIFF
--- a/templates/commands/analyze.md
+++ b/templates/commands/analyze.md
@@ -45,6 +45,7 @@ You **MUST** consider the user input before proceeding (if not empty).
 
     Wait for the result of the hook command before proceeding to the Goal.
     ```
+    After emitting the block above you MUST actually invoke the hook and wait for it to finish before continuing. Run it the same way you would run the command yourself in this agent/session (the invocation may differ from the literal `{command}` id shown above, e.g. a skills-mode agent runs it as `/skill:speckit-...` or `$speckit-...`). Emitting the block alone does not run the hook.
 - If no hooks are registered or `.specify/extensions.yml` does not exist, skip silently
 
 ## Goal
@@ -228,6 +229,7 @@ After reporting, check if `.specify/extensions.yml` exists in the project root.
     Executing: `/{command}`
     EXECUTE_COMMAND: {command}
     ```
+    After emitting the block above you MUST actually invoke the hook and wait for it to finish before continuing. Run it the same way you would run the command yourself in this agent/session (the invocation may differ from the literal `{command}` id shown above, e.g. a skills-mode agent runs it as `/skill:speckit-...` or `$speckit-...`). Emitting the block alone does not run the hook.
 - If no hooks are registered or `.specify/extensions.yml` does not exist, skip silently
 
 ## Operating Principles
--- a/templates/commands/checklist.md
+++ b/templates/commands/checklist.md
@@ -66,6 +66,7 @@ You **MUST** consider the user input before proceeding (if not empty).
 
     Wait for the result of the hook command before proceeding to the Execution Steps.
     ```
+    After emitting the block above you MUST actually invoke the hook and wait for it to finish before continuing. Run it the same way you would run the command yourself in this agent/session (the invocation may differ from the literal `{command}` id shown above, e.g. a skills-mode agent runs it as `/skill:speckit-...` or `$speckit-...`). Emitting the block alone does not run the hook.
 - If no hooks are registered or `.specify/extensions.yml` does not exist, skip silently
 
 ## Execution Steps
@@ -363,4 +364,5 @@ Check if `.specify/extensions.yml` exists in the project root.
     Executing: `/{command}`
     EXECUTE_COMMAND: {command}
     ```
+    After emitting the block above you MUST actually invoke the hook and wait for it to finish before continuing. Run it the same way you would run the command yourself in this agent/session (the invocation may differ from the literal `{command}` id shown above, e.g. a skills-mode agent runs it as `/skill:speckit-...` or `$speckit-...`). Emitting the block alone does not run the hook.
 - If no hooks are registered or `.specify/extensions.yml` does not exist, skip silently
--- a/templates/commands/clarify.md
+++ b/templates/commands/clarify.md
@@ -49,6 +49,7 @@ You **MUST** consider the user input before proceeding (if not empty).
 
     Wait for the result of the hook command before proceeding to the Outline.
     ```
+    After emitting the block above you MUST actually invoke the hook and wait for it to finish before continuing. Run it the same way you would run the command yourself in this agent/session (the invocation may differ from the literal `{command}` id shown above, e.g. a skills-mode agent runs it as `/skill:speckit-...` or `$speckit-...`). Emitting the block alone does not run the hook.
 - If no hooks are registered or `.specify/extensions.yml` does not exist, skip silently
 
 ## Outline
@@ -251,6 +252,7 @@ Check if `.specify/extensions.yml` exists in the project root.
     Executing: `/{command}`
     EXECUTE_COMMAND: {command}
     ```
+    After emitting the block above you MUST actually invoke the hook and wait for it to finish before continuing. Run it the same way you would run the command yourself in this agent/session (the invocation may differ from the literal `{command}` id shown above, e.g. a skills-mode agent runs it as `/skill:speckit-...` or `$speckit-...`). Emitting the block alone does not run the hook.
   - **Optional hook** (`optional: true`):
     ```
     ## Extension Hooks
--- a/templates/commands/constitution.md
+++ b/templates/commands/constitution.md
@@ -46,6 +46,7 @@ You **MUST** consider the user input before proceeding (if not empty).
 
     Wait for the result of the hook command before proceeding to the Outline.
     ```
+    After emitting the block above you MUST actually invoke the hook and wait for it to finish before continuing. Run it the same way you would run the command yourself in this agent/session (the invocation may differ from the literal `{command}` id shown above, e.g. a skills-mode agent runs it as `/skill:speckit-...` or `$speckit-...`). Emitting the block alone does not run the hook.
 - If no hooks are registered or `.specify/extensions.yml` does not exist, skip silently
 
 ## Outline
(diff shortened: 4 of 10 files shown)


mnriem (commented):
Thanks for the contribution. A couple of questions before we can move forward:

1. **Orchestrator claim**: The PR body references `specify run` and "CLI orchestrators that watch agent output" as existing consumers of the `EXECUTE_COMMAND:` directive. Can you point to the code path where this dispatch actually happens today? If no orchestrator currently consumes the directive, the backward-compatibility framing is misleading and the PR description should be corrected.

2. **AI disclosure**: Was this PR authored with AI assistance? If so, please disclose per contributing guidelines.

The underlying fix (agent self-dispatches mandatory hooks in agent-direct sessions) addresses a real problem, but the bar is high for sweeping template changes and we need confidence the framing accurately reflects the codebase.

Quratulain-bilal:
closed this by accident while editing the description - reopened.

  on the orchestrator question: you're right, I rechecked and couldn't find anything in the repo that actually consumes
  EXECUTE_COMMAND today. execute_hook() has no production callers and auto_execute_hooks gets written as a default but
  never read. I picked that framing up from the issue author's wording in #2730 without verifying it myself. fixed the PR
  description.

  and yes, this was built with AI assistance (Claude Code) - code, tests and description, reviewed by me. added the
  disclosure to the PR body.

  if the template changes are too broad I can cut it down to just the format_hook_message() change + tests and do the
  templates as a follow-up.

mnriem:

mnriem:
Yes, the template changes are way too broad. We should look into a fix with a smaller surface

Quratulain-bilal:
makes sense, agree it's too broad.

  while trying to shrink it I found the reason it spread: format_hook_message() in extensions.py isn't actually reached at runtime  its only caller is check_hooks_for_event(), which nothing in the CLI or agent path calls. the hook block is emitted by the agent following the command-template instructions (it reads .specify/extensions.yml and outputs the block itself), so the templates are where the behavior actually lives  hence all 9.

  two smaller options:
   one line in the mandatory-hook instruction of each template, no example-block edits - still 9 files but a tiny diff each
  or fix just plan as the pattern, then the rest in a follow-up

  any preference on direction?

mnriem:
Pick one and lets see how it looks

Quratulain-bilal:
went with option 1 - one instruction line after each mandatory-hook block, example blocks untouched. rebased on current main so the diff is clean: 9 files, 18 lines.

Quratulain-bilal:
@mnriem this is ready for another look when you have time - went with the minimal option (one line per template, examples untouched, 9 files / 18 lines). no rush, just flagging it's been sitting since the rebase.

mnriem:

Quratulain-bilal:
addressed both Copilot points:

- the dispatch line no longer leans on the literal `/{command}` id. it now tells the agent to run the hook the way it'd run the command itself in this agent/session, and calls out that the invocation can differ from the `{command}` id (e.g. `/skill:speckit-...` or `$speckit-...` in skills mode). so the id-vs-invocation distinction is explicit now.
- fixed the PR description - it was stale from the earlier broad version and claimed runtime/API-ref/test changes that aren't in this minimal cut. the change is now purely the template instruction lines, and the description says exactly that.

also rebased on current main, diff is 10 files / 20 lines.

mnriem (approved):

mnriem:
Thank you!

Issue #804, opened 2025-10-09
Support Claude Code Plugins


peterHoburg:
Anthropic just released plugins for claude code. 

TLDR plugins are a way to bundle commands + agents + hooks into a single package with a native installer built into Claude Code. 

https://www.anthropic.com/news/claude-code-plugins

MartyBonacci:
Hey Peter, I just made this plugin for Claude Code directly from spec-kit....but haven't fully tested it. Let me know how it works for you. https://github.com/MartyBonacci/specswarm

MartyBonacci:
In Claude Code...

/plugin marketplace add MartyBonacci/specswarm 

/plugin install speckit

MartyBonacci:
Just made an update to the Claude Code plugin. I missed that the scripts got dropped from the plugin. Modified the commands to handle the operations that the scripts were doing.

amondnet:
I am migrating Gemini extensions to Claude Code plugins, including Spec-kit.

https://plugins.pleaseai.dev/?install=spec-kit

MartyBonacci:
@amondnet , is pleaseai.dev your marketplace? Are you the owner?
It has a lot of great plugins in it.

grhaonan:
Any official plug-in released from spec-kit yet??

MartyBonacci:
@grhaonan it was progressing so slow that I just made my own Claude Code plugin based on spec-kit https://github.com/MartyBonacci/specswarm

amondnet:
https://www.notion.so/notiondevs/Notion-Skills-for-Claude-28da4445d27180c7af1df7d8615723d0

> ### Spec to Implementation 🚀
> 
> Turns product or tech specs into concrete Notion tasks that Claude Code can implement. Breaks down spec pages into detailed implementation plans with clear tasks, acceptance criteria, and progress tracking to guide development from requirements to completion.

[notion-spec-to-implementation.zip](https://github.com/user-attachments/files/22991048/notion-spec-to-implementation.zip)

PR #2905, opened 2026-06-09
fix(presets): create skill overrides when skills_source is marketplace

### LINKED ISSUE #2906: Bug: preset skill overrides not written when skills_source is marketplace and ai_skills is false
## Summary

When a spec-kit project uses `"skills_source": "marketplace"` in `.specify/integration.json` AND has `"ai_skills": false` in `.specify/init-options.json`, installing a preset that provides `type: "command"` entries (e.g. `specify preset add jira`) does **not** write the corresponding `.claude/skills/<skill-name>/SKILL.md` override file. The preset appears installed (registry updated, files in `.specify/presets/`) but the local skill override that should shadow the marketplace plugin skill is silently skipped.

## Root cause

Two places in `src/specify_cli/presets.py`:

1. `_get_skills_dir()` has an early-return guard: `if not ai_skills_enabled and agent != "kimi": return None` — this causes the method to return `None` for marketplace projects, so `_register_skills()` is never reached.

2. `_register_skills()` computes `create_missing_skills = ai_skills_enabled and agent_config.get("extension") != "/SKILL.md"` — when `ai_skills: false` this is always `False`, so no skill directory is created even when the preset command override would require one.

## Expected behaviour

Running `specify preset add <preset>` in a `skills_source: marketplace` project should write `.claude/skills/<skill-name>/SKILL.md` for each command the preset overrides, so that the local skill shadows the marketplace plugin skill.

## Steps to reproduce

1. Initialise a project with `specify init --here --integration claude --ai claude` (sets `ai_skills: false` by default with marketplace source)
2. Install a preset: `specify preset add jira --from https://github.com/luno/spec-kit-preset-jira/archive/refs/heads/main.zip`
3. Observe: `.specify/presets/jira/` exists and registry shows `registered_skills: ["speckit-taskstoissues"]`, but `.claude/skills/speckit-taskstoissues/SKILL.md` is never created

## Fix

A fix is already in PR #2905. Reference: https://github.com/github/spec-kit/pull/2905

echarrod:
## Summary

- **Bug:** When a project configured `"skills_source": "marketplace"` in `.specify/integration.json` and `"ai_skills": false` in `init-options.json`, running `specify preset add <preset>` appeared to succeed (registry updated, files placed in `.specify/presets/`) but silently skipped writing the local `.claude/skills/<skill-name>/SKILL.md` override file.
- **Root cause:** Two guards in `PresetManager` both keyed on `ai_skills_enabled`:
  1. `_get_skills_dir()` returned `None` early for non-kimi agents when `ai_skills` was false, so `_register_skills()` bailed before reaching the `create_missing_skills` check.
  2. `create_missing_skills` was itself gated on `ai_skills_enabled`, so even when a skills dir existed the new skill file would not be created.
- **Fix:** Both checks now read `skills_source` from `integration.json` and treat `ai_skills` as effectively enabled when `skills_source == "marketplace"`. This enables the local SKILL.md to shadow the marketplace plugin version, which is the whole point of a preset override in a marketplace-integrated project.

Fixes #2906

## Test plan

- [x] New test `TestPresetSkills::test_skill_created_when_ai_skills_false_but_marketplace_source` covers the exact failure scenario: `ai_skills: false` + `skills_source: marketplace` + preset with a `type: command` entry → SKILL.md IS created.
- [x] All existing `TestPresetSkills` tests (19 total) continue to pass - `ai_skills: false` without `skills_source: marketplace` is unchanged.
- [x] `test_skill_not_updated_when_ai_skills_disabled` confirms the non-marketplace `ai_skills: false` behaviour is preserved.

### DIFF
--- a/src/specify_cli/integrations/claude/__init__.py
+++ b/src/specify_cli/integrations/claude/__init__.py
@@ -24,6 +24,15 @@ ARGUMENT_HINTS: dict[str, str] = {
     "taskstoissues": "Optional filter or label for GitHub issues",
 }
 
+# Per-command frontmatter overrides for skills that should run in a forked
+# subagent context. Read-only analysis commands are good candidates: the
+# heavy reads (spec/plan/tasks artefacts) collapse to a short summary,
+# so isolating them keeps the main conversation context clean.
+# See https://code.claude.com/docs/en/skills#run-skills-in-a-subagent
+FORK_CONTEXT_COMMANDS: dict[str, dict[str, str]] = {
+    "analyze": {"context": "fork", "agent": "general-purpose"},
+}
+
 
 class ClaudeIntegration(SkillsIntegration):
     """Integration for Claude Code skills."""
@@ -190,6 +199,11 @@ class ClaudeIntegration(SkillsIntegration):
             hint = ARGUMENT_HINTS.get(stem, "")
             if hint:
                 updated = self.inject_argument_hint(updated, hint)
+
+            fork_config = FORK_CONTEXT_COMMANDS.get(stem)
+            if fork_config:
+                for key, value in fork_config.items():
+                    updated = self._inject_frontmatter_flag(updated, key, value)
 
             if updated != content:
                 path.write_bytes(updated.encode("utf-8"))
--- a/src/specify_cli/presets.py
+++ b/src/specify_cli/presets.py
@@ -1111,6 +1111,20 @@ class PresetManager:
         that callers can fall back gracefully.
         """
         from . import resolve_active_skills_dir, _print_cli_warning
+        from .integrations._helpers import _read_integration_json
+
+        # For marketplace integrations, presets must still create local skill overrides
+        # to shadow the marketplace plugin version, bypassing the ai_skills guard.
+        integration_data = _read_integration_json(self.project_root)
+        if integration_data.get("skills_source") == "marketplace":
+            from . import load_init_options, _get_skills_dir as _module_get_skills_dir
+            opts = load_init_options(self.project_root) or {}
+            agent = opts.get("ai")
+            if not isinstance(agent, str) or not agent:
+                return None
+            skills_dir = _module_get_skills_dir(self.project_root, agent)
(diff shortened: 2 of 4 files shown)


mnriem:
**Hey @echarrod** — thanks for the detailed write-up! I took a look but have some concerns.

**The `skills_source: "marketplace"` feature doesn't exist.** I searched `main` for `skills_source` across all Python, JSON, YAML, and Markdown files and got zero hits. There's no code that reads or writes a `"skills_source": "marketplace"` field in `integration.json` — not in the state schema, not in `_write_integration_json`, and not in any init flow.

The repro steps in #2906 also reference `specify init --here --integration claude --ai claude`, but `--ai` isn't a valid flag on `specify init` — only `--integration` exists.

Could you point me to where `skills_source: "marketplace"` gets written today? Without the upstream feature that produces this state, the fix guards against a condition that can't occur. If this is intended for a future feature, it should land after that feature ships.

**PRs should deliver a single focused change.** This PR bundles two unrelated things: a preset/marketplace fix and the `context: fork` injection for the `analyze` skill. The fork context change looks fine on its own and belongs in `ClaudeIntegration`, but it should be its own PR so each change can be reviewed, reverted, and bisected independently.

echarrod:
Thanks @mnriem - you're right on both counts.

**On `skills_source: "marketplace"`:** that field is written by our downstream install tooling. We distribute the spec-kit skills through a private Claude Code plugin marketplace, so `ai_skills` is `false` and skills come from the plugin rather than being materialised locally per repo. It doesn't exist in upstream spec-kit, so - exactly as you say - this guard protects a state upstream can't produce. Native Claude Code plugin support was requested upstream in #804 but closed as not-planned, which is why we handle distribution downstream. Rather than carry a consumer-side guard with no upstream producer, we'll handle the preset-override behaviour in our own install tooling too. If that direction is ever revisited upstream, the consumer side can follow.

**On bundling:** agreed - the forked-subagent change is already tracked separately in #2511, so it shouldn't have been bundled here. Dropping it.

Closing this PR; thanks for the careful review.

PR #2911, opened 2026-06-10
fix(presets): harden preset URL installs against unsafe redirects

darion-yaphet:
## Summary

Hardens `specify preset add --from <url>` downloads against redirect-based attacks. Extracted from #2826 so the security fix can be reviewed and merged independently of the refactor.

- **Validate redirect targets**: the final URL after redirects must satisfy the same policy as the initial URL (HTTPS, or HTTP only for localhost/loopback). A `redirect_validator` hook is added to `authentication.http.open_url` so every hop is checked, not just the final response.
- **Reject malformed download URLs**: URLs without a valid hostname are rejected before any network I/O.
- **Prevent transport downgrade for credentialed redirects**: `_StripAuthOnRedirect` no longer allows a redirect to carry the request to a weaker transport when credentials were attached.
- Loopback detection now uses `ipaddress.ip_address(...).is_loopback` instead of a hardcoded host list.

## Relationship to #2826

These commits originally lived on the `refactor/split-init-pr6` branch, mixed with the preset command refactor. They are rebased here onto `main` (applied to `__init__.py`, where the preset handlers currently live) so they carry no refactor dependency. #2826 will be rebased to drop them and return to a pure code-move.

## Test plan

- [x] New tests for redirect validation, malformed URL rejection, and credentialed downgrade prevention (`tests/test_presets.py`, `tests/test_authentication.py`)
- [x] Full suite: 3717 passed, 45 skipped


mnriem:
Thank you!

PR #2933, opened 2026-06-11
chore(catalog): bump jira-sync to v0.2.1 (multi-spec + dry-run fixes)

ashbrener:
Bumps the `jira-sync` community catalog entry from **v0.2.0 → v0.2.1** (patch bug-fix release).

The cataloged v0.2.0 was tagged *before* two bug fixes landed, so the listed version still carries the multi-spec collision bug. v0.2.1 fixes:

- **Multi-spec phase-Subtask identity collision** — phase Subtasks were matched by the `task-phase:N` label alone; across a multi-spec repo every spec's "Phase N" collided on the same Subtask. The find is now parent-scoped, so each spec mirrors to its own Subtask (self-healing on a plain reconcile). Also guards an empty-phase `taskList` that triggered a Jira 400 INVALID_INPUT, and surfaces field-level write errors.
- **Dry-run of an unmirrored spec no longer spuriously exits 3** — the `DRY-0` placeholder key 404'd on a live Jira during post-create reconciles; the existence-check read is now skipped for that placeholder (scoped to `DRY-0`; real keys still fail closed).

Changes: `version` → `0.2.1`, `download_url` → the `v0.2.1` source ZIP. JSON validated; no other entries touched.


ashbrener:
Superseded by #2952 (bumps jira-sync straight to v0.3.0, which includes these v0.2.1 fixes plus two new features). Closing to avoid two open PRs against the same catalog entry — please review #2952 instead. Thanks!

PR #2947, opened 2026-06-12
feat: add /speckit-delta — constitution-driven vision-gap analysis

jzhao401:
### What this adds

Spec Kit's loop today covers the life of a single feature — specify → clarify → plan → tasks → implement. The decision *which feature to build next* lives outside the toolkit. This PR closes that loop with a new core command:

**`/speckit-delta`** reads the constitution's **Vision & Direction** section, inspects the current repo state (code + merged specs), computes the gap between the two, and recommends the next feature — with a handoff straight into `/speckit-specify`.

Concretely:

- `templates/commands/delta.md` — the command template (200 lines; registers across integrations like every core command).
- `templates/constitution-template.md` — adds a **Vision & Direction** section (North Star, Target Users & Value, Long-Term Objectives, Non-Goals). Long-Term Objectives are required to be *observable* — phrased so progress can be assessed by inspecting repo state — which is what makes the gap computation meaningful.
- `templates/plan-template.md` — adds a **Vision Alignment Check** gate next to the existing Constitution Check, so each plan is validated against where the project is *going*, not only against its principles.
- `templates/commands/constitution.md` — instructs the constitution command to populate Vision (inferring from README/docs when user input is silent; `TODO(VISION)` markers instead of bare placeholders).
- One-line registrations (`SKILL_DESCRIPTIONS`, `_FALLBACK_CORE_COMMAND_NAMES`, claude `ARGUMENT_HINTS`) plus `tests/test_delta_and_vision.py` (14 tests) and expected-file inventory updates in the integration suites.

### Why in core rather than a preset

The command is fully generic — it contains no org-specific content and works for any project whose constitution states a vision. We've been running it in production across our repos (currently distributed as a preset command) and the loop-closing step has proven to be the part of spec-driven development the core pipeline doesn't yet cover. Upstreaming makes it available at bare `specify init`.

### Backward compatibility

Purely additive. Existing projects see no change until they regenerate their constitution; projects without a Vision section get inference + `TODO(VISION)` guidance rather than failures.

### Testing

`tests/test_delta_and_vision.py` (14 new tests) plus the full touched suites — extensions, CLI, claude/copilot/generic and base integration suites — pass locally on top of current `main` (485 tests across the touched files).


mnriem:
Please deliver this as a community extension and then we can host it in the community catalog. Thanks!

PR #2949, opened 2026-06-12
fix: register enabled extensions for agent on integration use/upgrade

PascalThuet:
## What

Enabled extension command/skill registration now happens only when an integration is selected or refreshed:

- `specify integration use <agent>`
- `specify integration switch <agent>`
- `specify integration upgrade <agent>`

Plain `specify integration install <agent>` installs only that integration core command artifacts. This follows maintainer feedback in #2949 and avoids extension side effects for secondary integrations until they are selected.

## Behavior

The shared `_register_extensions_for_agent` helper remains best-effort: registration failures warn via `_print_cli_warning` and do not roll back the surrounding `use`/`switch`/`upgrade` transaction.

`use` now registers enabled extensions after the default integration is changed. `switch` and `upgrade` continue to register enabled extensions after their own transactions commit.

## Tests

- `test_install_defers_extension_commands_until_use`
- `test_install_does_not_register_disabled_extensions`
- `test_upgrade_backfills_extension_commands_for_agent`
- `test_install_skills_mode_secondary_agent_defers_extension_artifacts`
- `test_upgrade_non_active_agent_preserves_active_agent_skills`

Targeted local run:

- `uv run --extra test pytest tests/integrations/test_integration_subcommand.py -q`
- `uvx ruff check src/specify_cli/integrations/_install_commands.py src/specify_cli/integrations/_query_commands.py src/specify_cli/integrations/_helpers.py tests/integrations/test_integration_subcommand.py`
- `git diff --check origin/main...HEAD`

Closes #2886

Issue #2886, opened 2026-06-06
[Bug]: integration install/upgrade don't register extension commands for the new agent (only switch does)


PascalThuet:
## Bug Description

`specify integration switch` re-registers enabled extensions' commands for the target agent (`register_enabled_extensions_for_agent`). `specify integration install` and `specify integration upgrade` do not. `integration.setup()` only installs the integration's own core commands, not extension commands.

So when a second integration is added to a project, the new agent is missing the extension commands the first agent has (e.g. the bundled agent-context extension's `/speckit.agent-context.update`, and any other installed extension's commands).

## Steps to Reproduce

1. `specify init --here --integration claude`
2. `specify integration install codex`
3. Inspect `.specify/extensions/.registry`: the agent-context entry's `registered_commands` contains only `claude`, not `codex`.

Reproduced on `main` @ `7106858`:
```
registered_commands: { "claude": ["speckit.agent-context.update"] }
```
No `codex` key; codex has no rendered `speckit.agent-context.update` command.

## Expected Behavior

Installing/upgrading an integration registers the enabled extensions' commands for that integration's agent, same as `switch` does — so every installed agent has the same extension commands.

## Actual Behavior

Only `switch` registers extension commands for an agent. A second integration added via `install` (or refreshed via `upgrade`) never gets them, so its extension commands are silently absent.

## Question for maintainers

The fix is to call `register_enabled_extensions_for_agent(<agent>)` from `install` and `upgrade` (as `switch` already does). Before I open a PR, one decision is yours:

**Should `install`/`upgrade` register *all* enabled extensions for the new agent (full parity with `switch`), or only a subset?**

- **Full parity (recommended):** every installed agent ends up with every enabled extension's commands. Consistent with `switch`; smallest, most predictable change. Downside: `install`/`upgrade` start writing more files than today.
- **Subset / opt-in:** narrower blast radius, but special-cases which extensions propagate to which agents — more config, easy to get inconsistent.

If full parity is fine, I'll send a PR adding the `register_enabled_extensions_for_agent` call to both paths with tests.

## Code references (`main` @ `7106858`)

- `integrations/_migrate_commands.py` — `switch` calls `register_enabled_extensions_for_agent(target)`; `upgrade` does not.
- `integrations/_install_commands.py` — `install` does not call it.
- `integrations/base.py` — `setup()` installs the integration's own commands only.

## Specify CLI Version

main @ `7106858`

## AI Agent

Codex CLI

[...]

PR #2951, opened 2026-06-12
fix: isolate per-extension failures so one bad extension can't drop the rest

### LINKED ISSUE #2950: register_enabled_extensions_for_agent has no per-extension error isolation: one failing extension silently drops the rest
## Summary

`ExtensionManager.register_enabled_extensions_for_agent` iterates over enabled extensions with **no per-extension error isolation**. If registering one extension raises, the loop aborts and the exception propagates, so every *subsequent* enabled extension is skipped — the agent ends up with only a prefix of its extensions.

The callers wrap the whole call in a single best-effort `try/except` (`_register_extensions_for_agent` in `src/specify_cli/integrations/_helpers.py`), so the wholesale abort surfaces as **one warning while the command still exits 0**. The partial state is silent.

## Affected paths

- `integration switch` (pre-existing).
- `integration install` / `integration upgrade` — routed through this batch path by #2886 / #2949, so the blast radius now includes those commands too.

## Root cause

`src/specify_cli/extensions.py`, `register_enabled_extensions_for_agent`:

```python
for ext_id, metadata in self.registry.list().items():
    if not metadata.get("enabled", True):
        continue
    manifest = self.get_extension(ext_id)
    if manifest is None:
        continue
    ext_dir = self.extensions_dir / ext_id
    updates = {}
    if agent_config and not skills_mode_active:
        registered = registrar.register_commands_for_agent(agent_name, manifest, ext_dir, self.project_root)
        ...
    registered_skills = self._register_extension_skills(manifest, ext_dir)
    ...
    self.registry.update(ext_id, updates)
```

There is no `try/except` around the per-extension body, so an exception from `register_commands_for_agent` / `register_commands` (e.g. an `OSError` writing a command file, or a path collision where a command directory is expected) or from `_register_extension_skills` propagates out of the loop and skips the remaining extensions.

Note the loop already handles the *non-raising* degraded cases, so those are **not** triggers: a manifest that fails to load is skipped (`get_extension(...) is None → continue`), and an empty registration result is handled by clearing stale state. The gap is specifically genuine exceptions raised mid-loop.

## Reproduction

With two or more enabled extensions, force the first one iterated to raise during registration (e.g. an `OSError` from a command-file write). Trigger registration for an agent (`integration install <agent>` / `switch` / `upgrade`): only a single warning is printed, the later healthy extensions are not registered for that agent, and the command exits 0.

A deterministic version is in the fix's regression test (`test_one_failing_extension_does_not_abort_the_rest`): it patches the first extension's registration to raise and asserts the later one still registers.

## Expected

A failure registering one extension should warn and **continue** to the rest (per-extension isolation), so a single bad extension cannot silently drop the others.

## Fix

PR #2951 wraps the per-extension loop body in `try/except`, warns per failing extension, and continues; the caller-level best-effort catch remains as a backstop.

Surfaced by review of #2949 (follow-up to #2886). Distinct from #2948, which is about *where* extension artifacts render (skills vs commands) for non-active skills-mode agents, not error isolation.

PascalThuet:
## What

`ExtensionManager.register_enabled_extensions_for_agent` iterates over enabled extensions with no per-extension error isolation. If registering one extension raised (e.g. an `OSError` writing a command file), the loop aborted and the exception propagated, so every *subsequent* enabled extension was silently skipped. The callers wrap the whole call in a single best-effort `try/except`, so the wholesale abort surfaced as one warning while the command still exited 0 — leaving the agent with only a prefix of its extensions registered.

This path is reached by `integration switch` and (since #2949 / #2886) `integration install` and `integration upgrade`.

## Fix

Wrap the per-extension loop body in `try/except`: on failure, warn (naming the failing extension) and `continue` to the rest. The caller-level best-effort catch remains as a backstop.

Note the loop already guards the *non-raising* degraded cases — a manifest that fails to load is skipped (`get_extension(...) is None → continue`) and an empty registration result clears stale state — so this only adds isolation for genuine exceptions mid-loop.

## Test

`test_one_failing_extension_does_not_abort_the_rest` installs two enabled extensions, forces the first-iterated one's registration to raise, and asserts the later extension still registers and the call does not propagate. Confirmed it fails without the fix.

Full suite: 3726 passed, 45 skipped locally.

Closes #2950

### DIFF
--- a/src/specify_cli/extensions.py
+++ b/src/specify_cli/extensions.py
@@ -1742,11 +1742,31 @@ class ExtensionManager:
                     if new_registered != registered_commands:
                         updates["registered_commands"] = new_registered
 
-                registered_skills = self._register_extension_skills(manifest, ext_dir)
-                if registered_skills:
-                    existing_skills = self._valid_name_list(metadata.get("registered_skills", []))
-                    merged_skills = list(dict.fromkeys(existing_skills + registered_skills))
-                    updates["registered_skills"] = merged_skills
+                try:
+                    registered_skills = self._register_extension_skills(manifest, ext_dir)
+                except Exception as skills_err:
+                    # Skills are a companion artifact.  If command registration
+                    # already succeeded, still persist it so later cleanup can
+                    # find those command files.
+                    from . import _print_cli_warning
+
+                    _print_cli_warning(
+                        "register extension skills for",
+                        "extension",
+                        ext_id,
+                        skills_err,
+                        continuing=(
+                            "Continuing with command registration for this "
+                            "extension and the remaining extensions."
+                        ),
+                    )
+                else:
+                    if registered_skills:
+                        existing_skills = self._valid_name_list(
+                            metadata.get("registered_skills", [])
+                        )
+                        merged_skills = list(dict.fromkeys(existing_skills + registered_skills))
+                        updates["registered_skills"] = merged_skills
 
                 if updates:
                     self.registry.update(ext_id, updates)
--- a/src/specify_cli/extensions.py
+++ b/src/specify_cli/extensions.py
@@ -1716,37 +1716,53 @@ class ExtensionManager:
                 continue
 
             ext_dir = self.extensions_dir / ext_id
-            updates: Dict[str, Any] = {}
 
-            if agent_config and not skills_mode_active:
-                registered = registrar.register_commands_for_agent(
-                    agent_name, manifest, ext_dir, self.project_root
+            # Isolate per-extension failures: one extension that fails to
+            # register (e.g. an OSError writing a command file) must not abort
+            # registration of the remaining enabled extensions for this agent.
+            try:
+                updates: Dict[str, Any] = {}
+
+                if agent_config and not skills_mode_active:
+                    registered = registrar.register_commands_for_agent(
+                        agent_name, manifest, ext_dir, self.project_root
+                    )
+                    registered_commands = metadata.get("registered_commands", {})
+                    if not isinstance(registered_commands, dict):
+                        registered_commands = {}
+                    new_registered = copy.deepcopy(registered_commands)
+                    if registered:
+                        new_registered[agent_name] = registered
+                    else:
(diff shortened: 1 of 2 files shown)


PascalThuet:
Updated on behalf of @PascalThuet by Codex (GPT-5).

Commit: 5cdbc5fbe438105b73eb0639300a8eb7faae1a82

What changed:
- Addressed Copilot review 4506765356 by isolating `_register_extension_skills()` failures from successful command registration inside `register_enabled_extensions_for_agent`.
- Successful command registrations are now still persisted to `registered_commands` even when companion skill generation raises, preserving later cleanup/unregister behavior.
- Added regression coverage for the command-registration-succeeds / skill-registration-fails case.
- Merged current `origin/main` into the PR branch. The merge completed without manual conflicts.

Validation run locally:
- `uv run pytest tests/test_extension_skills.py -q` -> 53 passed
- `uvx ruff check src/specify_cli/extensions.py tests/test_extension_skills.py` -> passed
- `git diff --check` -> passed

GitHub now reports the PR as mergeable. `gh pr checks` currently reports no checks on the branch.

mnriem:

PascalThuet:
Updated on behalf of @PascalThuet by Codex (GPT-5).

Commit: c899cd32a27e631bf7fb0e7b4c7c7f5f04656a0a

What changed:
- Addressed the remaining Copilot feedback by replacing the skills-registration warning with wording that is accurate for both command-mode and skills-mode runs.
- Updated the regression assertion to use a stable substring from the new generic warning.
- Rebasing onto current `origin/main` resolved the merge conflict in `src/specify_cli/extensions.py`; GitHub now reports the PR as mergeable and `behind_by: 0` against `main`.

Validation run locally:
- `uv run pytest tests/test_extension_skills.py -q` -> 54 passed
- `uvx ruff check src/specify_cli/extensions.py tests/test_extension_skills.py` -> passed
- `git diff --check origin/main...HEAD` -> passed

`gh pr checks 2951 --repo github/spec-kit` currently reports no checks on the branch.

mnriem (approved):

PR #2957, opened 2026-06-12
fix: fail loudly when a fan-out 'items' expression does not resolve to a list

### LINKED ISSUE #2956: [Bug]: fan-out silently coerces a non-list 'items' to [] — zero-instance runs with no error
### Bug Description

`FanOutStep.execute` silently coerces a non-list `items` value to `[]` (`src/specify_cli/workflows/steps/fan_out/__init__.py:23-26`):

```python
items_expr = config.get("items", "[]")
items = evaluate_expression(items_expr, context)
if not isinstance(items, list):
    items = []
```

When the `items:` expression is mis-wired or unresolvable (for example templated from a step output key that doesn't exist), the fan-out completes with `item_count: 0` and `results: []` and the workflow sails on. The wiring error surfaces — if at all — steps later, as a confusing downstream failure with no pointer back to the fan-out.

We hit this in a real pipeline: a fan-out templated `items` from a shell step's output (shell steps expose only `stdout`/`stderr`/`exit_code`, so the template resolved to nothing), and the run produced a vacuous zero-instance result with no error anywhere near the cause.

### Steps to Reproduce

1. Author a workflow with a fan-out whose `items` expression resolves to a non-list, e.g.:

```yaml
workflow:
  id: repro
  name: fan-out silent empty
  version: "1.0.0"
steps:
  - id: emit
    type: shell
    run: "echo '[1, 2, 3]'"
  - id: fan
    type: fan-out
    items: "{{ steps.emit.output.items }}"   # resolves to None — key doesn't exist
    step:
      id: worker
      type: shell
      run: "echo {{ item }}"
```

2. `specify workflow run repro`

### Expected Behavior

The fan-out step fails loudly, naming the expression that did not resolve to a list — a non-list here is a wiring error, not data. (An explicit **empty list** is a legitimate data case and should remain valid.)

### Actual Behavior

`Status: completed`, `item_count: 0`, zero instances executed, no error.

### Specify CLI Version

main @ 1b0556c (also present in 0.9.2)

### AI Agent

n/a (workflow engine bug, agent-independent)

### Operating System

macOS 14 (not OS-specific)

### Python Version

3.11

---

**Note on the existing test:** `tests/test_workflows.py::TestFanOutStep::test_execute_non_list_items_resolves_empty` currently pins the silent behavior, so it may have been deliberate. If silent-empty is intentional for some use case, an opt-in knob (e.g. `allow_empty: true` / `strict: false`) preserving fail-loud as the default would also solve the underlying problem. I have a small PR ready either way — it updates that test to the fail-loud contract and adds an explicit empty-list-is-valid case.

**AI disclosure (per CONTRIBUTING.md):** this issue and the accompanying PR were authored with AI assistance (Claude), based on a real-pipeline failure investigation; all code and claims were verified by running the repo's test suite locally.

doquanghuy:
## Description

Fixes #2956.

`FanOutStep.execute` silently coerced any non-list `items` result to `[]`, so a mis-wired or unresolvable `items:` expression produced a vacuous zero-instance run with no error anywhere near the cause. This PR makes the step fail loudly instead, with an error naming the expression and the resolved type. An explicit empty **list** remains valid input (legitimately-empty data still completes with `item_count: 0`).

The previous behavior was pinned by `test_execute_non_list_items_resolves_empty`; that test is updated to the fail-loud contract, and a new `test_execute_empty_list_items_is_valid` locks the empty-list-stays-valid distinction. If silent-empty was intentional for some use case, happy to rework this as an opt-in knob (`allow_empty:`/`strict:`) with fail-loud as default — see the issue for that alternative.

The failure output keeps the step's four output keys (`items`/`max_concurrency`/`step_template`/`item_count`) so downstream readers of a failed run's state don't hit missing keys.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` — 208 passed (the new non-list test fails red against current `main`, passes with the fix; verified both directions)
- [ ] Tested with a sample project (covered by the unit tests above; the repro workflow from #2956 now fails at the fan-out step with the new error)

## AI Disclosure

Code, tests, and this description were authored with AI assistance (Claude), driven by a real-pipeline failure investigation; everything was verified by running the repo's test suite and ruff locally in both red (unfixed) and green (fixed) directions.

### DIFF
--- a/src/specify_cli/workflows/steps/fan_out/__init__.py
+++ b/src/specify_cli/workflows/steps/fan_out/__init__.py
@@ -22,11 +22,27 @@ class FanOutStep(StepBase):
     def execute(self, config: dict[str, Any], context: StepContext) -> StepResult:
         items_expr = config.get("items", "[]")
         items = evaluate_expression(items_expr, context)
-        if not isinstance(items, list):
-            items = []
-
         max_concurrency = config.get("max_concurrency", 1)
         step_template = config.get("step", {})
+
+        if not isinstance(items, list):
+            # A non-list here is a wiring error (the expression did not
+            # resolve to a collection); silently fanning out over zero
+            # items hides it. An explicit empty list remains valid input.
+            return StepResult(
+                status=StepStatus.FAILED,
+                error=(
+                    f"Fan-out step {config.get('id', '?')!r}: 'items' must "
+                    f"resolve to a list, got {type(items).__name__} from "
+                    f"{items_expr!r}."
+                ),
+                output={
+                    "items": [],
+                    "max_concurrency": max_concurrency,
+                    "step_template": step_template,
+                    "item_count": 0,
+                },
+            )
 
         return StepResult(
             status=StepStatus.COMPLETED,
--- a/tests/test_workflows.py
+++ b/tests/test_workflows.py
@@ -1475,9 +1475,9 @@ class TestFanOutStep:
         assert result.output["item_count"] == 2
         assert result.output["max_concurrency"] == 3
 
-    def test_execute_non_list_items_resolves_empty(self):
+    def test_execute_non_list_items_fails_loudly(self):
         from specify_cli.workflows.steps.fan_out import FanOutStep
-        from specify_cli.workflows.base import StepContext
+        from specify_cli.workflows.base import StepContext, StepStatus
 
         step = FanOutStep()
         ctx = StepContext()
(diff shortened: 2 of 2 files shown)


doquanghuy:
@mnriem when you have a moment, would appreciate a review — happy to adjust anything.

mnriem (approved):

mnriem:
Thank you!

PR #2961, opened 2026-06-12
feat(workflows): add from_json expression filter

### LINKED ISSUE #2960: [Feature]: a from_json expression filter so step outputs can become typed values
### Problem Statement

Step outputs that are captured as strings can never become typed values inside workflow templates. The expression evaluator's filter set is `default` / `join` / `map` / `contains` (`src/specify_cli/workflows/expressions.py:15-51`) — there is no way to parse JSON.

Concrete case: a shell step emits a JSON items list on stdout (`output.stdout` is captured), but a downstream `fan-out` step's `items:` can never consume it — the template yields a string, and `FanOutStep` requires a list. Today there is **no** mechanism in the engine for a step that runs external code to hand a collection to a later step (workflow inputs are scalar-only as well).

### Proposed Solution

One arg-less pipe filter:

```yaml
items: "{{ steps.emit.output.stdout | from_json }}"
```

Semantics: parse-or-raise. Invalid JSON or non-string input raises a clear `ValueError` (`from_json: invalid JSON: …` / `from_json: expected a JSON string, got int`) — never a silent passthrough, since a parse failure here means the pipeline wiring is wrong and silence would hide it (same philosophy as #2956).

I have a small PR ready: the filter + dispatch + docstring mention + three tests (valid JSON parses to the typed value; invalid JSON raises; non-string raises).

### Alternatives Considered

A declared `outputs:` schema (or `output_file:`) on shell/command steps — a more structural fix to the same gap, but a much bigger API surface; I'm filing that separately as a design proposal so the two can be weighed independently. `from_json` is the smallest unblock and remains useful even if a declared-outputs mechanism lands later.

### Component

Workflow engine (expression evaluator)

### AI Agent (if applicable)

n/a

### Use Cases

- Feeding `fan-out` `items:` from a step that computes the collection at runtime (extract-style steps).
- Reading any structured tool output (JSON on stdout) into conditions or later step args.

---

**AI disclosure (per CONTRIBUTING.md):** this issue and the accompanying PR were authored with AI assistance (Claude); verified by running the repo's test suite locally.

doquanghuy:
## Description

Fixes #2960.

Adds one arg-less pipe filter to the sandboxed expression evaluator:

```yaml
items: "{{ steps.emit.output.stdout | from_json }}"
```

`_filter_from_json` parses a JSON string into its typed value (list/dict/scalar). Semantics are parse-or-raise: invalid JSON or non-string input raises a clear `ValueError` — never a silent passthrough, since a parse failure means the pipeline wiring is wrong and silence would hide it. The module docstring's filter list is updated; no other filter or evaluator behavior is touched.

This is the smallest unblock for feeding `fan-out` `items:` (or any condition/arg) from a step that computes a collection at runtime. It composes with — and doesn't preclude — a future declared-`outputs:` mechanism, which I'm proposing separately.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` — `tests/test_workflows.py` 210 passed
- [x] Three new tests in `TestExpressions` (valid JSON → typed value; invalid JSON raises; non-string raises) — all three are red against current `main`, green with the filter (verified both directions)
- [ ] Tested with a sample project (covered by the evaluator tests; exercised manually with the #2956 repro workflow rewired through `| from_json`)

## AI Disclosure

Code, tests, and this description were authored with AI assistance (Claude); verified by running the repo's test suite and ruff locally in both red and green directions.

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -280,7 +280,7 @@ Steps can reference inputs and previous step outputs using `{{ expression }}` sy
 | `steps.specify.output.file`    | Output from a previous step          |
 | `item`                         | Current item in a fan-out iteration  |
 
-Available filters: `default`, `join`, `contains`, `map`.
+Available filters: `default`, `join`, `contains`, `map`, `from_json`.
 
 Example:
 
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -158,6 +158,18 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
         value = _evaluate_simple_expression(parts[0].strip(), namespace)
         filter_expr = parts[1].strip()
 
+        # `from_json` is strict and takes no arguments. Match the filter name
+        # tolerant of whitespace and reject any parenthesized form
+        # (`from_json()`, `from_json('x')`, `from_json ()`) so a mis-wired
+        # template fails loudly instead of silently returning the unparsed
+        # value.
+        if filter_expr.split("(", 1)[0].strip() == "from_json":
+            if "(" in filter_expr:
+                raise ValueError(
+                    "from_json: filter takes no arguments (use '| from_json')"
+                )
+            return _filter_from_json(value)
+
         # Parse filter name and argument
         filter_match = re.match(r"(\w+)\((.+)\)", filter_expr)
         if filter_match:
@@ -175,8 +187,6 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
         filter_name = filter_expr.strip()
         if filter_name == "default":
             return _filter_default(value)
-        if filter_name == "from_json":
-            return _filter_from_json(value)
         return value
 
     # Boolean operators — parse 'or' first (lower precedence) so that
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -1,7 +1,8 @@
 """Sandboxed expression evaluator for workflow templates.
 
 Provides a safe Jinja2 subset for evaluating expressions in workflow YAML.
-No file I/O, no imports, no arbitrary code execution.
+Templates cannot perform file I/O, import modules, or run arbitrary code —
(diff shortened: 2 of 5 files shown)


doquanghuy:
@mnriem when you have a moment, would appreciate a review — happy to adjust anything.

mnriem:

doquanghuy:
@mnriem Thanks for the review — addressed the Copilot feedback:
- `from_json` is now strict: any parenthesized form (`from_json()` or `from_json('x')`) raises a clear error instead of silently returning the unparsed value.
- Renamed the mis-named test to `test_filter_from_json_parses_object` (the JSON under test is an object), and added coverage for the strict no-arguments behavior.

Full suite green; `ruff check src/` clean. Ready for another look.

mnriem (changes requested):

doquanghuy:
@mnriem Pushed `0eba146` addressing the latest Copilot comment:

- Documented the new `from_json` filter in the `workflows/ARCHITECTURE.md` filter table with the suggested example `{{ steps.emit.output.stdout | from_json }}`.
- Preempted the same gap in the other two filter enumerations so all user-facing docs match the evaluator: `workflows/README.md` and `docs/reference/workflows.md`.

The earlier strictness fix (parenthesized forms like `from_json()` / `from_json('x')` raise instead of silently passing through) and the test rename are unchanged.

mnriem:

doquanghuy:
@mnriem Pushed `fd896a3` for the latest Copilot round:

- **Strictness now rejects trailing tokens, not just parens.** Typos like `| from_json)` or `| from_json extra` previously fell through to the unknown-filter path and silently returned the unparsed value. The check now matches the leading filter token and requires the whole filter to be exactly `from_json`, so every mis-wired form raises. Extended the rejection test to cover the trailing-token cases (7 forms total).
- **Docstring fixed.** `No file I/O, no imports` was misleading after the `json` import; reworded to state the actual sandbox guarantee (templates cannot do file I/O, import modules, or run arbitrary code).

Full `tests/test_workflows.py` green (211 passed).

mnriem (approved):

mnriem:
Thank you!

Issue #2964, opened 2026-06-12
[Feature]: surface gate detail in the workflow run/resume --json outcome


doquanghuy:
### Problem Statement

The machine-readable outcome of `workflow run`/`resume` (`--json`) is `{run_id, workflow_id, status, current_step_id, current_step_index}`. When a run pauses at a **gate**, that payload is indistinguishable from any other pause — and the gate's `message`, `options`, and (after an interactive choice) `choice` exist only in the gate step's internal output (`steps/gate/__init__.py` stores them, including `output.choice`), never in the structured outcome. An orchestrator driving Spec Kit programmatically has to parse the human-facing stdout to learn "this is a review gate, here are the options".

### Proposed Solution

Two small pieces (this is an API-surface proposal — would appreciate direction before polishing):

1. Record each step's `type` in the run state's step results (one line in the engine's `step_data`).
2. When the run sits at a gate, add a `gate` block to the `--json` payload:

```json
{
  "status": "paused",
  "current_step_id": "review",
  "gate": {
    "step_id": "review",
    "message": "Approve the thing?",
    "options": ["approve", "reject"],
    "choice": null
  }
}
```

`choice` populates when the outcome ends **at** the gate with a decision recorded — e.g. an interactive rejection with `on_reject: abort` yields a `failed` payload whose gate block carries `"choice": "reject"`, and an `on_reject: retry` pause carries the rejected choice. (A mid-flow approval proceeds past the gate, so the final payload has no gate block — by design.) Non-gate runs and runs that end elsewhere are unchanged (no `gate` key).

I have a small reference PR ready.

### Alternatives Considered

- A generic `paused_step: {id, type, output}` block (more information, larger surface, leaks step internals).
- A dedicated `gate-outcome` status value (changes the status enum — more disruptive than an additive block).
- Status quo: orchestrators parse stdout (fragile, and impossible under `--json` where the stream is redirected).

### Component

Workflow engine / CLI (`--json` outcome)

### AI Agent (if applicable)

n/a

### Use Cases

- External orchestrators (CI, parent-pipeline coordinators) detecting "human review needed" and presenting the gate's options without scraping terminal output.
- Distinguishing a review-gate pause from other waiting states programmatically.

---

**AI disclosure (per CONTRIBUTING.md):** this issue and the accompanying reference PR were authored with AI assistance (Claude); verified by running the repo's test suite locally.

marcelsafin:
[...]

PR #2963, opened 2026-06-12
feat(workflows): opt-in output_format: json exposes parsed shell stdout as output.data

doquanghuy:
## Description

**Reference implementation for #2962 — for discussion, direction welcome.**

Adds an opt-in `output_format: json` to shell steps: when declared, stdout is parsed and exposed under `output.data` (the raw `stdout`/`stderr`/`exit_code` keys are unchanged, so there is no merge/clobber ambiguity), letting later steps consume typed values — e.g. a fan-out's `items: "{{ steps.emit.output.data.items }}"`. A parse failure fails the step with a clear error (declaring the format is a contract; silence would hide wiring bugs). Without the key, behavior is byte-identical — fully backward-compatible. `validate()` rejects unknown formats.

The issue lists `output_file:` and a declared named-`outputs:` schema as alternatives — happy to rework toward either if you prefer that direction.

**Coordination:** aware of #2443 touching the same file; this diff is isolated to the structured-output addition and I'm happy to rebase in whichever order suits.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` — full suite 3729 passed
- [x] Four new tests in `TestShellStep`: data exposed on valid JSON; invalid stdout fails the step; no-flag backcompat (no `data` key); validate rejects unknown formats — the three behavior tests are red against current `main`, green with the change (verified both directions)
- [x] `uvx ruff check src/` — clean
- [x] Tested locally with `uv run specify --help`
- [ ] Tested with a sample project (covered by the step-level tests)

## AI Disclosure

- [ ] I **did not** use AI assistance for this contribution
- [x] I **did** use AI assistance (describe below)

Code, tests, and this description were authored with AI assistance (Claude); verified by running the repo's test suite and ruff locally in both red and green directions.

PR #2959, opened 2026-06-12
fix: non-zero exit code when a workflow run ends failed or aborted

doquanghuy:
## Description

Fixes #2958.

`workflow run` and `workflow resume` ended with exit code 0 even when the run finished `Status: failed`, so scripts/CI had to parse stdout to detect failure. This PR adds a single status→exit-code mapping (`_run_outcome_exit_code`: `failed`/`aborted` → 1; `completed`/`paused` → 0 — paused is a waiting state, not a failure) and applies it at the four epilogues (run/resume × text/`--json`). The `--json` payload itself is unchanged — it still prints first; only the process exit code changes.

**Behavior change, flagged plainly:** the old behavior was deliberately pinned by `tests/test_workflow_run_without_project.py::test_workflow_run_failing_yaml_without_project` (`exit_code == 0` asserted on a failed run). That pin is updated to the new contract with an explanatory comment. Anyone scripting against exit-0-on-failed would be affected — the issue discusses the trade-off; happy to gate this behind a flag instead if you'd rather not change the default.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` — full suite 3728 passed
- [x] New `TestWorkflowRunExitCodes` (3 CLI-level tests): failed→1 (text + `--json`), completed→0; the two failure-path tests are red against current `main`, green with the fix (verified both directions)
- [x] `uvx ruff check src/` — clean
- [x] Tested locally with `uv run specify --help`
- [ ] Tested with a sample project (covered by the CLI-level tests, which drive real workflow YAML files end-to-end through `CliRunner`)

## AI Disclosure

- [ ] I **did not** use AI assistance for this contribution
- [x] I **did** use AI assistance (describe below)

Code, tests, and this description were authored with AI assistance (Claude); verified by running the repo's test suite and ruff locally in both red (unfixed) and green (fixed) directions.

PR #2965, opened 2026-06-12
feat: surface gate detail in the workflow run/resume --json payload

doquanghuy:
## Description

**Reference implementation for #2964 — for discussion, direction welcome.**

When a run pauses at a gate, the `--json` outcome now carries a `gate` block (`step_id` / `message` / `options` / `choice`) so orchestrators can detect "human review needed" and present the options without parsing the human-facing stream. Two small pieces:

1. The engine records each step's `type` in the run state's step results (one added line in `step_data` — previously the type was not recoverable from state).
2. `_workflow_run_payload` adds the `gate` block via a `_gate_outcome` helper when the run's current step is a gate. `choice` populates when the outcome ends at the gate with a decision recorded (e.g. an interactive rejection with `on_reject: abort` → a `failed` payload carrying `"choice": "reject"`; an `on_reject: retry` pause likewise). A mid-flow approval proceeds past the gate, so the block clears — by design. Non-gate runs and runs that end elsewhere are unchanged — no `gate` key, payload byte-identical to today.

The issue lists alternatives (a generic `paused_step` block; a dedicated status value) — happy to rework toward either.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` — full suite 3727 passed
- [x] Two new CLI-level tests (`TestWorkflowRunGateOutcomeJson`): a gate pause carries the exact block (CliRunner stdin is non-TTY, so the gate pauses); a completed run has no `gate` key — the gate-pause test is red against current `main`, green with the change (verified both directions)
- [ ] Tested with a sample project (covered by the CLI-level tests, which drive a real gate workflow through `workflow run --json`)

## AI Disclosure

Code, tests, and this description were authored with AI assistance (Claude); verified by running the repo's test suite and ruff locally in both red and green directions.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -2145,15 +2145,30 @@ def _gate_outcome(state: Any) -> dict[str, Any] | None:
     output = step.get("output") or {}
     # `message` and `options` may be non-string YAML literals in an unvalidated
     # workflow (GateStep coerces neither for the payload), so normalise both
-    # here for a stable JSON schema: message → str, options → list[str].
+    # here for a stable JSON schema: message → str, options → list[str] | None.
     message = output.get("message")
-    options = output.get("options")
     return {
         "step_id": state.current_step_id,
         "message": None if message is None else str(message),
-        "options": [str(o) for o in options] if isinstance(options, list) else options,
+        "options": _normalize_gate_options(output.get("options")),
         "choice": output.get("choice"),
     }
+
+
+def _normalize_gate_options(options: Any) -> list[str] | None:
+    """Normalise a gate's ``options`` to a stable ``list[str]`` (or ``None``).
+
+    A valid gate stores a list, but an unvalidated workflow could leave a
+    scalar or tuple. ``None`` stays ``None`` (no options); a list/tuple maps
+    each element through ``str``; any other scalar becomes a single-element
+    list — so the emitted JSON schema is always ``list[str] | None``. A bare
+    string is treated as one option, never iterated character-by-character.
+    """
+    if options is None:
+        return None
+    if isinstance(options, (list, tuple)):
+        return [str(o) for o in options]
+    return [str(options)]
 
 
 def _run_outcome_exit_code(status_value: str) -> int:
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -2143,15 +2143,17 @@ def _gate_outcome(state: Any) -> dict[str, Any] | None:
     if not isinstance(step, dict) or not _is_gate_step(step):
         return None
     output = step.get("output") or {}
-    # `message` and `options` may be non-string YAML literals in an unvalidated
-    # workflow (GateStep coerces neither for the payload), so normalise both
-    # here for a stable JSON schema: message → str, options → list[str] | None.
+    # `message`, `options`, and `choice` may be non-string YAML literals in an
+    # unvalidated workflow (GateStep coerces none of them for the payload), so
+    # normalise all three for a stable JSON schema: message → str, options →
+    # list[str] | None, choice → str | None (None means no decision yet).
     message = output.get("message")
+    choice = output.get("choice")
     return {
         "step_id": state.current_step_id,
         "message": None if message is None else str(message),
         "options": _normalize_gate_options(output.get("options")),
-        "choice": output.get("choice"),
+        "choice": None if choice is None else str(choice),
     }
 
 
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -2105,6 +2105,24 @@ def _workflow_run_payload(state: Any) -> dict[str, Any]:
     return payload
 
 
+def _is_gate_step(step: dict[str, Any]) -> bool:
+    """Whether a recorded step result is a gate.
+
+    Prefers the persisted ``type`` field, but when it is absent — a run paused
+    by an older version, whose step record predates ``type`` being stored —
+    falls back to the gate's unique output signature: only ``GateStep`` writes
+    an ``on_reject`` key. A record carrying a *different* known ``type`` is not
+    a gate, so the fallback applies only when ``type`` is missing entirely.
+    """
+    step_type = step.get("type")
+    if step_type == "gate":
+        return True
+    if step_type:
+        return False
+    output = step.get("output")
+    return isinstance(output, dict) and "on_reject" in output
+
+
 def _gate_outcome(state: Any) -> dict[str, Any] | None:
     """Gate detail for the structured outcome, when the run rests at a gate.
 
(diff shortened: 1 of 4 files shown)


doquanghuy:
@mnriem when you have a moment, would appreciate your thoughts on the direction here — the issue lists the alternatives considered, and I'm happy to rework toward whichever shape fits Spec Kit best.

mnriem:

doquanghuy:
@mnriem Thanks for the review — addressed the Copilot feedback:
- `_gate_outcome()` now only surfaces the `gate` block while the run is actually paused (guards on `status == paused`). Since `RunState.current_step_id` isn't cleared on completion, a completed/failed run whose last step was a gate no longer leaks stale gate detail in run/resume/status `--json`.
- Hardened the `_run_json` test helper to assert CLI success before JSON-parsing, and added direct coverage for the suppression guard.

Full suite green; `ruff check src/` clean. Ready for another look.

doquanghuy:
@mnriem Pushed `5fd0f85` addressing the latest Copilot round:

- **Abort path now surfaces the gate block.** `_gate_outcome()` emits the gate detail for `aborted` runs too, not only `paused`. Abort is the *only* path that sets `ABORTED` (gate rejection with `on_reject: abort`) and it leaves `current_step_id` on that gate, so an orchestrator can read the recorded `choice` for the stop. `completed`/`failed` stay suppressed.
- **Stable JSON schema.** `message` is coerced to a string — `GateStep` only coerces it for expression interpolation, so a non-string YAML literal could otherwise leak into the payload.
- **Tests:** added a CLI-level aborted-path test (`test_gate_abort_carries_gate_block`, asserts `status == aborted` and `choice == reject`), a message-coercion test, and extended the suppression test to allow `aborted`. Shared the run helper via `_invoke_json` to avoid duplicating invoke boilerplate.

mnriem:

doquanghuy:
@mnriem Pushed `5f60408` for the latest Copilot round:

- The gate-abort test parsed stdout without first asserting the CLI exited cleanly, so an invoke failure would have surfaced as an opaque JSON decode error. It now routes through `_run_json` (which asserts `exit_code == 0` before parsing), and I dropped the now-redundant `_invoke_json` helper — a gate abort emits the payload and returns, so the run exits cleanly.

Full `tests/test_workflows.py` green (212 passed).

doquanghuy:
@mnriem Pushed `3e303fb` addressing the latest Copilot round:

- **Run-helper assertion message** now uses `result.output` instead of `result.stdout`. Under `--json`, step output is redirected off stdout, so a failing run's useful diagnostics live on `result.output`; the JSON parse still reads stdout. This also brings `_run_json` in line with the other CLI tests in the file.
- **`StepContext.steps` docstring** updated from the old 5-key entry shape to the canonical 7-key shape the engine actually persists (`type`, `integration`, `model`, `options`, `input`, `output`, `status`), so step authors and debuggers see the real record.

Full `tests/test_workflows.py` green (212 passed); ruff clean.

Heads up: GitHub now shows this branch as conflicting with `main` — that's from #2959 and #2963 having merged (they touch the same run-command / engine.py / test regions). I've left the branch as-is rather than rebase a public PR unprompted; happy to rebase onto current `main` and resolve if you'd like, just say the word.

doquanghuy:
@mnriem Rebased onto current `main` and resolved the conflict (force-pushed `24d6e85`). The branch is now `MERGEABLE`.

What the rebase touched:
- **Conflict was a clean "keep both"**: #2959 added `TestWorkflowRunExitCodes` exactly where this PR adds `TestWorkflowRunGateOutcomeJson` — both classes are preserved in full. `_workflow_run_payload` now (correctly) calls `_gate_outcome` *and* keeps `_run_outcome_exit_code`; the `step_data["type"]` field this PR relies on sits cleanly alongside the rest.
- **One real semantic interaction with #2959:** a gate abort now emits the `--json` payload and then exits non-zero (aborted → 1). I gave the test helper an `expected_exit` param (default 0) so the abort test asserts exit 1 while paused/completed stay at 0 — single shared helper, no duplicated invoke boilerplate.

All prior Copilot fixes are intact (paused/aborted guard, message coercion, `result.output` assert message, 7-key `StepContext.steps` docstring). Full `specify_cli` test suite green locally (**3879 passed**); ruff clean.

doquanghuy:
@mnriem Pushed `d322cf5` for the latest Copilot comment:

- **Backward-compatible gate detection.** A run paused by an older version has no persisted step `type`, so `_gate_outcome` would have dropped its gate block on resume. Added `_is_gate_step`: it prefers the `type` field, and when `type` is absent falls back to the gate's unique output signature (`on_reject`, which only `GateStep` writes). A record carrying a *different* known `type` is still not treated as a gate.
- **Options normalized to strings.** Mirroring the existing `message` coercion, `options` is now normalized to a list of strings so an unvalidated workflow with non-string options can't destabilize the JSON schema.
- **Tests:** options coercion, type-less gate detection (resume path), and a type-less non-gate negative case.

Full `tests/test_workflows.py` green (291 passed); ruff clean.

doquanghuy:
@mnriem Pushed `a97cd18` for the latest Copilot comment:

- **Non-list `options` now normalized too.** The previous pass only mapped a `list`, returning the raw value for any other shape — so a scalar or tuple leaked through unnormalized, contradicting the `list[str]` intent. Extracted `_normalize_gate_options`: `None` stays `None`; a list/tuple maps each element through `str`; any other scalar becomes a single-element list (a bare string is treated as one option, never iterated character-by-character). The emitted schema is now always `list[str] | None`.
- **Tests** extended to cover list, tuple, bare string, numeric scalar, and `None`.

Full `tests/test_workflows.py` green (291 passed); ruff clean.

doquanghuy:
@mnriem Pushed `c4297a0` for the latest Copilot round:

- **`choice` now normalized too.** `_gate_outcome` normalized `message` and `options` but passed `choice` through as-is — an unvalidated gate can record a non-string `choice`, contradicting the stable-schema rationale. It is now coerced to `str | None` (None still means "no decision yet"), consistent with the other two fields. Added a focused choice-coercion test (None / string / non-string).
- **Portable plain-gate test.** The no-gate test workflow used `run: "true"`, which fails under `cmd.exe` on Windows (ShellStep uses `shell=True`). Switched to the cross-platform `run: "exit 0"`, matching the exit-code suite's workflows. (Scoped to the one occurrence this PR introduced; two pre-existing `run: "true"` lines elsewhere are upstream and left untouched to keep the diff tight.)

Full `tests/test_workflows.py` green (292 passed); ruff clean.

doquanghuy:
@mnriem Pushed `be970e7` addressing the latest Copilot round (pyproject.toml / CHANGELOG.md).

Both findings traced to a single stray commit on the branch: `chore: bump version to 0.11.2` (authored by `github-actions[bot]`), which had leaked in and bumped `pyproject.toml` `0.11.2.dev0` → `0.11.2` plus added a full 0.11.2 release-note block to `CHANGELOG.md`. That's release automation, not part of this feature.

- **Rebased onto current `upstream/main` and dropped the bot bump commit.** Upstream has since cut 0.11.2 and moved to `0.11.3.dev0`; the branch now inherits that verbatim. The PR diff no longer touches `pyproject.toml` or `CHANGELOG.md` at all — it's back to just the gate-JSON feature (4 files: `__init__.py`, `workflows/base.py`, `workflows/engine.py`, `tests/test_workflows.py`).
- All prior Copilot fixes are intact (paused/aborted guard, message/options/choice coercion, backward-compat gate detection, `result.output` assert message, 7-key `StepContext.steps` docstring, portable `exit 0` plain-gate test).

Full `tests/test_workflows.py` green (296 passed).

mnriem:
Thank you!

PR #2988, opened 2026-06-15
fix: write Codex dev skills as files

### LINKED ISSUE #2930: [Bug]: Codex does not recognize skills with symbolic link generated by `specify extension add --dev`
Codex fails to detect or load skills that are linked into the project using symbolic links created by the `specify extension add --dev` command.

### Steps to Reproduce

1. Run:
   ```bash
   specify extension add --dev <extension-name>
   ```
   This creates a symbolic link to the extension in the project.
2. Attempt to use the skill provided by the extension in Codex.
3. Observe that Codex does not recognize or load the skill.

### Expected Behavior

Codex should correctly resolve symbolic links and recognize skills added via `specify extension add --dev`.

### Actual Behavior

- Skills added via symbolic links are not recognized.
- Codex behaves as if the extension is missing or not installed.

### Specify CLI Version

0.10

### AI Agent

Codex CLI

### Operating System

Windows

### Python Version

3.12

### Error Logs

```shell

```

### Additional Context

- Skills work correctly when installed normally (without symbolic links).
- The issue seems specific to symbolic link resolution.

sjh9714:
## Description

Fixes #2930.

`specify extension add --dev` created symlinked Codex `SKILL.md` files, but Codex currently does not load those symlinked skills. This keeps dev-mode symlinks for other agents while writing Codex `SKILL.md` outputs as regular files in both the agent command registration path and the extension skill registration path.

Normal non-dev installs are unchanged, and existing Copilot dev symlink behavior remains covered by the existing test.

## Testing

Commands run:

- `uv run --extra test pytest tests/test_extensions.py -k "add_dev_writes_codex_skills_as_files" -q` (failed before the fix, passed after)
- `uv run --extra test pytest tests/test_extensions.py -q`
- `uv sync --extra test && uv run pytest -q`
- `uv run specify --help`
- `git diff --check`

## AI Disclosure

Codex assisted with issue scouting, code-path inspection, and local review. I reviewed the final diff and ran the listed checks locally.

### DIFF
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -37,6 +37,8 @@ def _build_agent_configs() -> dict[str, Any]:
             # when register_commands() resolves __SPECKIT_COMMAND_*__ tokens.
             if "invoke_separator" not in config:
                 config["invoke_separator"] = integration.invoke_separator
+            if integration.dev_no_symlink:
+                config["dev_no_symlink"] = True
             configs[key] = config
     return configs
 
@@ -714,6 +716,7 @@ class CommandRegistrar:
                 output_name,
                 agent_config["extension"],
                 link_outputs,
+                agent_config,
             )
 
             if agent_name == "copilot":
@@ -788,6 +791,7 @@ class CommandRegistrar:
                     alias_output_name,
                     agent_config["extension"],
                     link_outputs,
+                    agent_config,
                 )
                 if agent_name == "copilot":
                     self.write_copilot_prompt(project_root, alias)
@@ -804,9 +808,10 @@ class CommandRegistrar:
         output_name: str,
         extension: str,
         link_outputs: bool,
+        agent_config: dict[str, Any] | None = None,
     ) -> None:
         """Write a rendered agent artifact, optionally as a dev-mode symlink."""
-        if not link_outputs or (agent_name == "codex" and extension == "/SKILL.md"):
+        if not link_outputs or (agent_config or {}).get("dev_no_symlink"):
             dest_file.write_text(content, encoding="utf-8")
             return
 
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -934,6 +934,16 @@ class CommandRegistrar:
             self._active_skills_agent(project_root)
             if create_missing_active_skills_dir else None
         )
+        active_skills_dir: Optional[Path] = None
+        if active_skills_agent:
+            active_skills_config = self.AGENT_CONFIGS.get(active_skills_agent)
+            if (
+                active_skills_config
+                and active_skills_config.get("extension") == "/SKILL.md"
+            ):
+                active_skills_dir = self._resolve_agent_dir(
+                    active_skills_agent, active_skills_config, project_root,
+                )
         active_created_skills_dir: Optional[Path] = None
         for agent_name, agent_config in self.AGENT_CONFIGS.items():
             active_skills_output = (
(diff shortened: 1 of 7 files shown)


mnriem:
Thanks for tracking this down — the bug in #2930 is real and the behavior here is correct: Codex can't load symlinked `SKILL.md` files, so writing them as real files is the right outcome.

My concern is with *how* it's special-cased rather than the fix itself. The change hardcodes `agent_name == "codex"` in `CommandRegistrar._write_registered_output` (`agents.py`) and `selected_ai != "codex"` in `_register_extension_skills` (`extensions.py`). Both are generic, agent-agnostic code paths, and we have an integration architecture specifically so per-agent knowledge lives in one declarative place. `AGENT_CONFIGS` is already derived from `INTEGRATION_REGISTRY` via `_build_agent_configs()`, and `CodexIntegration` is a clean self-contained class — "Codex can't use dev symlinks" really belongs there, not in shared plumbing. (The existing `agent_name == "copilot"` carve-out is the same debt; I'd rather not grow it.)

Could we route this through integration metadata instead? Roughly:

1. Add `dev_no_symlink = True` to `CodexIntegration` and propagate it into the config dict in `_build_agent_configs()` the same way `invoke_separator` already is.
2. In `_write_registered_output`, swap the string check for a metadata lookup:
   ```python
   if not link_outputs or agent_config.get("dev_no_symlink"):
   ```
3. In `_register_extension_skills`, replace `selected_ai != "codex"` with `link_outputs and not agent_config.get("dev_no_symlink")`.

This follows the existing `inject_name` convention (opt-in flag, absent means the established default of symlinking), keeps the constraint owned by the Codex integration as the single source of truth, and means any future agent with the same limitation just sets the flag — no edits to shared code. The test you added is great and would carry over almost unchanged.

mnriem (changes requested):
Please resolve conflicts

mnriem:
Please resolve test & lint errors

sjh9714:
Resolved the conflict and the dev-mode regression after rebasing on current `main`.

The follow-up keeps the Codex no-symlink policy on integration metadata and also prevents inactive agents that share the same skills directory from overwriting Codex dev skills as symlinks.

Verified with:
- `.venv/bin/python -m pytest tests/test_extensions.py::TestExtensionAddCLI::test_add_dev_replaces_existing_codex_skill_symlink tests/test_extensions.py::TestExtensionAddCLI::test_add_dev_writes_codex_skills_as_files -q`
- `.venv/bin/python -m pytest tests/test_extension_skills.py -k "shared_skills_dir or codex_dev_skill_registration" -q`
- `.venv/bin/python -m pytest tests/test_extensions.py -q`
- `.venv/bin/python -m pytest tests/test_extension_skills.py tests/test_agent_config_consistency.py -q`
- `.venv/bin/python -m pytest -q`
- `uv run --with ruff ruff check src/specify_cli/agents.py tests/test_extensions.py tests/test_extension_skills.py tests/test_agent_config_consistency.py`
- `git diff --check`

mnriem:

sjh9714:
Addressed the Copilot feedback in `42fec17`.

I removed the broad suffix-based `_is_extension_skill_dev_symlink()` check and now only refresh the exact expected dev cache symlink for the current extension. I also added a regression test that keeps an unrelated user-owned symlink untouched.

Verified with:
- `.venv/bin/python -m pytest tests/test_extension_skills.py -k "codex_dev_skill_registration" -q`
- `.venv/bin/python -m pytest tests/test_extension_skills.py tests/test_extensions.py::TestExtensionAddCLI::test_add_dev_replaces_existing_codex_skill_symlink tests/test_extensions.py::TestExtensionAddCLI::test_add_dev_writes_codex_skills_as_files -q`
- `.venv/bin/python -m pytest tests/test_extensions.py tests/test_extension_skills.py tests/test_agent_config_consistency.py -q`
- `uv run --with ruff ruff check src/specify_cli/extensions/__init__.py tests/test_extension_skills.py`
- `git diff --check`

mnriem (approved):

mnriem:
Thank you!

PR #3015, opened 2026-06-17
feat: interactive approval flow for community extensions

DyanGalih:
## Description
This PR introduces an interactive approval flow when users attempt to install an extension from a non-approved catalog (like the `community` catalog). Previously, users were met with a hard error instructing them to manually edit `.specify/extension-catalogs.yml`. 

## Changes Made
- **Interactive Prompt:** `extension_add` now prompts the user with a warning panel if the extension's catalog is not approved. If approved interactively, it updates the config and proceeds with installation.
- **Config Persistence:** Added `approve_catalog_install` in `ExtensionCatalog` to cleanly serialize and write the updated active catalogs to YAML, preserving the full catalog stack (including defaults).
- **Consistency Updates:** 
  - Updated `extension_search` and `extension_info` CLI output to guide users to run `specify extension add` instead of manual file editing.
  - Replaced `yaml.dump` with `yaml.safe_dump` in all catalog-related commands for consistency and safety.
- **Developer QoL:** Added `__main__.py` to `specify_cli` so the package can be invoked locally using `python -m specify_cli` without global installation collisions.
- **Tests:** Added 7 new test cases covering prompt ordering, cancellation, duplicate approvals, and fallback behaviors.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -1086,40 +1086,45 @@ def extension_add(
                         console.print(f"  {REINSTALL_COMMAND}")
                         raise typer.Exit(1)
 
-                    # If a different approved source exists, use it instead of prompting.
-                    installable_info = catalog.get_installable_extension_info(resolved_id)
-                    if installable_info is not None:
-                        ext_info = installable_info
-
                     # Enforce install_allowed policy only when no approved source exists.
                     if not ext_info.get("_install_allowed", True):
-                        catalog_name = ext_info.get("_catalog_name", "community")
-                        console.print()
-                        console.print(
-                            Panel(
-                                f"[bold]'{ext_info['name']}' is available in the '{catalog_name}' catalog "
-                                f"but installation is not allowed from that catalog.[/bold]\n\n"
-                                f"Approve installation from '{catalog_name}' for this project?\n"
-                                "This will update .specify/extension-catalogs.yml so future installs "
-                                "from that catalog are allowed.",
-                                title="[bold yellow]Catalog Approval Required[/bold yellow]",
-                                border_style="yellow",
-                                padding=(1, 2),
+                        # If a different approved source exists, use it instead of prompting.
+                        installable_info = catalog.get_installable_extension_info(resolved_id)
+                        if installable_info is not None:
+                            ext_info = installable_info
+                        else:
+                            catalog_name = ext_info.get("_catalog_name", "community")
+                            console.print()
+                            console.print(
+                                Panel(
+                                    f"[bold]'{ext_info['name']}' is available in the '{catalog_name}' catalog "
+                                    f"but installation is not allowed from that catalog.[/bold]\n\n"
+                                    f"Approve installation from '{catalog_name}' for this project?\n"
+                                    "This will update .specify/extension-catalogs.yml so future installs "
+                                    "from that catalog are allowed.",
+                                    title="[bold yellow]Catalog Approval Required[/bold yellow]",
+                                    border_style="yellow",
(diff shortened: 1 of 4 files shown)


DyanGalih:
We have addressed the PR review feedback! 🚀 

**Updates:**
1. **Preserving User-Level Configs:** We updated `approve_catalog_install` to base the serialization on the currently active stack instead of a hardcoded minimal default. This ensures we don't accidentally "freeze" or discard any custom user-level catalogs from `~/.specify/extension-catalogs.yml` into the project config when an approval takes place.
2. **Simplified Approval Prompts:** We implemented `get_installable_extension_info` to automatically select the first *approved* source for a given extension. If you have an approved version of the extension in another catalog, the CLI will now gracefully install it without reprompting you for approval on the blocked catalog.
3. **Test Formatting:** Converted the very long `with patch(...)` blocks in `test_extensions.py` into multi-line parenthesized context managers for better readability, as requested.

All test suites have been run locally and are passing successfully! Let me know if there's anything else!

DyanGalih:
I have pushed an additional update to address the remaining feedback points from the review:

1. **Path Checks & Config Safety**: Reordered `approve_catalog_install` so path containment checks (`is_relative_to`) and symlink safety checks run *before* reading or creating directories. Also removed duplicate computation of `config_path`.
2. **Dangling Symlink Logic**: In `extensions.py`, replaced `exists()` with `is_symlink()` first when removing backup directories, avoiding `rmtree()` crashes on broken dangling symlinks.
3. **Graceful CLI Exit on ValidationError**: In `specify_cli/__init__.py`, wrapped the `approve_catalog_install` call with `try...except ValidationError` to catch issues (like symlinked catalogs) and gracefully print an error message instead of showing a raw traceback.
4. **Optimized Extension Lookups**: Delayed calling `catalog.get_installable_extension_info()` until *after* checking if the current extension is blocked (`_install_allowed` == False), skipping unnecessary fetches on the happy path.
5. **Test Mocking**: Added the missing `get_installable_extension_info` mocks in the test blocks of `test_extensions.py` so we don't accidentally perform real catalog fetches in those tests.
6. **Note on Prereleases**: The `prereleases=True` flag in the specifier check remains intentional to preserve the ability for contributors developing on `main` (like `0.11.1.dev0`) to successfully test extension installations without failing compatibility tests, which was the original goal of that specific change!

All tests are passing locally. Thanks for the thorough review! ✨

mnriem:
Note this is not necessary at all as a user can add the community catalog as an authorized read/write catalog by using the catalog system. Note that this would mean the entire catalog is available for install so BUYER BEWARE. So can we close this?

DyanGalih:
Thanks for the feedback. I agree that a user can already make the community catalog installable through the catalog system, but that is a broader trust decision than the specific dead-end I was trying to solve here.

The problem I hit as a first-time user was that the install fails with a message pointing to , but that file is not visible or explained in the normal flow. So the PR is meant to make the blocked install recoverable in-place, and the approval path creates or updates that config automatically instead of asking the user to hand-edit it.

I do think there is still a discoverability/docs gap, and I’m happy to follow up separately if we want the first-run messaging to make that clearer.

DyanGalih:
Correction on my last comment: the missing path is `.specify/extension-catalogs.yml`. The point still stands that the PR makes that file creation/update automatic during approval, so a first-time user does not need to know about it up front.

mnriem:
Sorry, but I am not going to apply this PR. I am OK documenting it better. The key here is that nothing from the community catalog should ever be applied without actually looking at the extension/preset/workflow/workflow step sources

Issue #1464, opened 2026-01-11
update-agent-context.sh breaks symlinks


przadka:
## Bug

`update-agent-context.sh` doesn't check if the target agent file (e.g., `CLAUDE.md`) is a symlink before writing to it. This converts the symlink into a regular file.

## Steps to Reproduce

1. Create `AGENTS.md` with your agent instructions
2. Create symlink: `ln -s AGENTS.md CLAUDE.md`
3. Run `/speckit.plan` which calls `update-agent-context.sh claude`
4. Check `CLAUDE.md` - it's now a regular file instead of a symlink

## Expected Behavior

The script should either:
- Detect symlinks and update the target file instead
- Skip updating symlinked files
- Warn the user that a symlink will be converted

## Workaround

Manually restore the symlink after running the script:
```bash
rm CLAUDE.md && ln -s AGENTS.md CLAUDE.md
```

carter003:
you can remove #templates/commands/plan.md line 14-16

```
agent_scripts:
  sh: scripts/bash/update-agent-context.sh __AGENT__
  ps: scripts/powershell/update-agent-context.ps1 -AgentType __AGENT__
```

PR #3017, opened 2026-06-17
Add regression coverage for agent-context symlink updates

Jiandong7:
## Summary
- add regression coverage for `update-agent-context.sh` when the configured context file is a symlink
- assert the symlink is preserved and the managed section is written through to the target file

## Validation
- `python -m pytest tests\extensions\test_extension_agent_context.py -q` (34 passed, 1 skipped locally because working bash is unavailable)
- `git diff --check`

Addresses #1464


mnriem (changes requested):

PR #3074, opened 2026-06-20
fix: fail loudly on an unknown workflow expression filter

### LINKED ISSUE #3073: [Bug]: workflow expression evaluator silently ignores unknown filters
### Bug Description

In the workflow expression evaluator, an **unknown pipe filter is silently ignored** — the value passes through unfiltered and the run completes with no error. Only five filters are registered (`default`, `join`, `map`, `contains`, `from_json`); any other filter name (a typo, or an unsupported one a user reasonably reaches for, e.g. `length`) is dropped silently, so the author gets a wrong value with no signal that the filter did nothing.

Root cause: `src/specify_cli/workflows/expressions.py` — the filter dispatch falls through to `return value` when no recognized filter matches (both the no-arg form and the unrecognized `name(arg)` form). This is the same "unknown-filter path" that the strict `from_json` handling already guards against for its own forms; it is unguarded in general.

### Steps to Reproduce

1. Use an unregistered filter in a workflow expression, e.g. a shell step:
   ```yaml
   - id: s
     type: shell
     run: "echo count={{ [1, 2, 3] | length }}"
   ```
2. `specify workflow run <workflow>`
3. The step prints `count=[1, 2, 3]` and the run completes — the `length` filter was silently ignored. (`evaluate_expression("{{ [1,2,3] | length }}", ctx)` returns `[1, 2, 3]`, not an error or a count.)

### Expected Behavior

An unregistered filter should fail loudly — a clear error naming the offending filter (and the valid ones) — so a typo or unsupported filter surfaces immediately instead of corrupting downstream expressions/conditions/gates silently. This matches the project's existing stance: `from_json` already raises on its mis-wired forms specifically to avoid "silently falling through to the unknown-filter path and returning the unparsed value," and `fan-out` `items:` was made fail-loud in #2957 for the same reason.

### Actual Behavior

The unknown filter is silently dropped (`return value`); the expression yields the unfiltered value and the run reports success. No error, no warning.

### Specify CLI Version

0.11.3

### AI Agent

Claude Code

### Operating System

macOS 14.5

### Python Version

Python 3.11.15

### Error Logs

```
(none — that is the bug: no error is raised; the filter is silently ignored)
```

### Additional Context

Found while exercising workflow expressions end-to-end. Happy to send a small PR that makes the unknown-filter path raise (reusing the existing `ValueError` style from `from_json`), preserving all five registered filters.

**AI disclosure:** this issue was investigated and written with AI assistance (Claude Code); I reproduced and verified the behavior against the current tree myself.

doquanghuy:
## Description

Fixes #3073.

The expression evaluator's filter dispatch fell through to `return value` for any **unregistered filter**, so a typo'd or unsupported filter — e.g. `{{ items | length }}` — rendered the value **unchanged with no error**, and the run completed. A silent wrong result, with nothing pointing at the cause.

This makes the unknown-filter path **fail loudly**: an unrecognized filter raises a clear `ValueError` naming the offending filter and the valid ones, reusing the same strict-`ValueError` style already used for `from_json` (which raises on its mis-wired forms precisely to avoid "silently falling through to the unknown-filter path and returning the unparsed value"). It closes that path in general, consistent with the fail-loud direction of #2957 (`fan-out items`) and #2961 (`from_json`).

- The five registered filters (`default` / `join` / `map` / `contains` / `from_json`) are **unchanged** — no behavior change for any valid usage.
- Both forms of an unknown filter are caught: the no-arg form (`| length`) and the `name(arg)` form (`| upper('x')`), which previously also fell through silently.

**Backward-compat note (calling it out honestly):** this turns a previously **silent** author error into a **loud** one. Any workflow that today relies on an unknown filter being silently dropped would now error. That passthrough was never intentional (the `from_json` strictness comment already names it as a path to avoid), and a scan of the repo's own workflow/template YAMLs found no `{{ … | <unknown-filter> }}` usage, and the full suite stays green — but flagging it so you can weigh the direction. Happy to rework as a warning-instead-of-raise, or gate it, if you'd prefer.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` — full suite **3913 passed, 59 skipped**
- [x] Tested with a sample project (if applicable) — ran a workflow whose shell step used `{{ [1,2,3] | length }}`: pre-fix it printed `count=[1, 2, 3]` and completed; post-fix the run fails loudly with `unknown filter 'length'`.

Tests in `TestExpressions` (`tests/test_workflows.py`):
- `test_filter_unknown_name_raises` — `| length` raises `ValueError`.
- `test_filter_unknown_name_with_args_raises` — `| upper('x')` raises (the `name(arg)` form).
- `test_registered_filter_unsupported_form_raises` — a registered filter in an unsupported form (`| join`, `| map` with no arg) raises a *misused-filter* message, not "unknown filter".
- `test_registered_filters_unaffected` — regression: all five registered filters still work, now including `map('attr')`.

The two unknown-filter tests are red against `main` (DID NOT RAISE), green with the fix.

## AI Disclosure

This change was authored with Claude Code: the AI drafted the fix and the tests and ran the suite. I discovered the behavior through end-to-end workflow-expression testing, verified the root cause and the red-against-main / green-with-fix tests myself, and reviewed the diff. The fix deliberately reuses the existing `from_json` error style rather than inventing a new one.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -192,7 +192,18 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
         filter_name = filter_expr.strip()
         if filter_name == "default":
             return _filter_default(value)
-        return value
+        # No recognized filter matched — an unknown filter name, or a known
+        # filter used in an unsupported form. Fail loudly rather than silently
+        # returning the unfiltered value: a passthrough turns a mis-typed or
+        # unsupported filter into a wrong result with no signal. Mirrors the
+        # strict `from_json` handling above.
+        leading_name = re.match(r"\w+", filter_expr)
+        name = leading_name.group(0) if leading_name else filter_expr
+        raise ValueError(
+            f"unknown filter '{name}': expected one of default('x'), "
+            "join('sep'), map('attr'), contains('s'), or from_json "
+            f"(got '| {filter_expr}')"
+        )
 
     # Boolean operators — parse 'or' first (lower precedence) so that
     # 'a or b and c' is evaluated as 'a or (b and c)'.
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -215,8 +215,8 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
         leading_name = re.match(r"\w+", filter_expr)
         name = leading_name.group(0) if leading_name else filter_expr
         expected = (
-            "expected one of default('x'), join('sep'), map('attr'), "
-            "contains('s'), or from_json"
+            "expected one of default or default('x'), join('sep'), "
+            "map('attr'), contains('s'), or from_json"
         )
         if name in _REGISTERED_FILTERS:
             raise ValueError(
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -12,6 +12,19 @@ import re
 from typing import Any
 
 
+# The filters the expression evaluator recognizes. Used to tell a
+# *registered* filter used in an unsupported form (e.g. `| join` with no
+# argument) apart from a genuinely unknown filter name, so each raises an
+# error that names the real problem.
+_REGISTERED_FILTERS: tuple[str, ...] = (
+    "default",
+    "join",
+    "map",
+    "contains",
+    "from_json",
+)
+
+
 # -- Custom filters -------------------------------------------------------
 
 def _filter_default(value: Any, default_value: Any = "") -> Any:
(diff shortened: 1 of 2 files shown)


doquanghuy:
@mnriem this is the fix for #3073 — fail-loud on unknown expression filters, reusing the from_json ValueError style; the five registered filters are unchanged and the full suite stays green. Happy to adjust the direction (warn instead of raise, or gate it) if you'd prefer.

doquanghuy:
@mnriem — pushed `d37e440` addressing the two Copilot review comments:

1. **Misleading "unknown filter" for a *registered* filter in an unsupported form** (`expressions.py`). A registered filter used with the wrong syntax — e.g. `| join` or `| map` with no argument — hit the same error as a genuinely unknown name and was reported as `unknown filter 'join'`, which is wrong (the filter is registered; the *form* isn't). The no-match path now splits: a name in the recognized set raises `filter 'join' used in an unsupported form (got '| join'): expected …`, while a truly unknown name keeps `unknown filter '<name>': …`. A new `_REGISTERED_FILTERS` constant drives the distinction.

2. **`map` was asserted-unaffected but untested.** `test_registered_filters_unaffected` now exercises `map('attr')` (`{{ rows | map('id') }} == ["a","b"]`) alongside the other four. Added `test_registered_filter_unsupported_form_raises` covering the `| join` / `| map` no-arg path.

The five registered filters are unchanged; full suite **3913 passed, 59 skipped**, `ruff check` clean. Happy to tune the unsupported-form wording if you'd prefer a per-filter expected form.

doquanghuy:
@mnriem pushed `e7e64ba` — the filter-error hint now lists the no-arg `default` form too (`default or default('x')`), per the latest Copilot note. Suite green (3913 passed).

mnriem:
Thank you!

PR #3079, opened 2026-06-21
fix(workflows): validate requires keys and reject phantom permissions gate

zied-jlassi:
## Description

A workflow's `requires:` block was parsed (`WorkflowDefinition.requires`) but its keys were never validated, so a typo or an unsupported key was silently ignored.

Most importantly, an author could write:

```yaml
requires:
  permissions:
    shell: true
```

expecting a runtime capability gate — but no such gate exists. A `shell` step always runs with the user's privileges, so this declaration gives a false sense of sandboxing. (This came up in #2440, where it was understandably assumed that such a declaration was already enforced.)

This PR makes `validate_workflow` honest about `requires`:

- Only the recognised keys are accepted: `speckit_version`, `integrations`.
- Any unknown key is rejected (a typo surfaces at validation time instead of being silently dropped) — same spirit as the recent "fail loudly on unknown" hardening.
- `requires.permissions` is rejected with an explicit message pointing authors at a `gate` step for approval, so nobody mistakes it for a security boundary.

It does **not** add any per-step permission system or runtime prompt — `requires` stays advisory. The model comment and the docs (`docs/reference/workflows.md`, `workflows/PUBLISHING.md`) are updated to say so plainly.

## Testing

- [x] Ran existing tests with `pytest` (full suite green, no regressions)
- [x] Added tests: valid recognised keys, non-mapping `requires`, unknown key, and `requires.permissions`

Validation is reached on the user-facing paths (`workflow add` / `info` / `run`), so the new errors actually surface.

## AI Disclosure

AI assistance (an AI coding agent) was used for the initial code review that surfaced the issue, and to help draft the implementation, tests, and documentation wording. All changes were reviewed and verified by me (red→green tests, full suite, `ruff`) before submitting.

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -270,7 +270,7 @@ specify workflow run speckit -i spec="Build a kanban board with drag-and-drop ta
 | `fan-out`    | Dispatch a step for each item in a list          |
 | `fan-in`     | Aggregate results from a fan-out step            |
 
-> **Security note:** a `shell` step runs a local command with **your** privileges. There is no capability sandbox — `requires` (e.g. `requires.permissions`) is an advisory pre-condition block, not a runtime gate, so it does **not** restrict what a step can do. Review any catalog or downloaded workflow before running it, and use a `gate` step to require explicit approval before sensitive or destructive shell commands.
+> **Security note:** a `shell` step runs a local command with **your** privileges. There is no capability sandbox — `requires` is an advisory pre-condition block (spec-kit version, integrations), not a runtime gate, so it does **not** restrict what a step can do. In particular there is no `requires.permissions` capability gate: it is rejected by validation precisely because it would imply a sandbox that does not exist. Review any catalog or downloaded workflow before running it, and use a `gate` step to require explicit approval before sensitive or destructive shell commands.
 
 ## Expressions
 
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -270,6 +270,8 @@ specify workflow run speckit -i spec="Build a kanban board with drag-and-drop ta
 | `fan-out`    | Dispatch a step for each item in a list          |
 | `fan-in`     | Aggregate results from a fan-out step            |
 
+> **Security note:** a `shell` step runs a local command with **your** privileges. There is no capability sandbox — `requires` (e.g. `requires.permissions`) is an advisory pre-condition block, not a runtime gate, so it does **not** restrict what a step can do. Review any catalog or downloaded workflow before running it, and use a `gate` step to require explicit approval before sensitive or destructive shell commands.
+
 ## Expressions
 
 Steps can reference inputs and previous step outputs using `{{ expression }}` syntax:
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -58,7 +58,12 @@ class WorkflowDefinition:
         # are not a security boundary. In particular there is no
         # ``requires.permissions`` capability gate: shell steps always run with
         # the user's privileges.
-        self.requires: dict[str, Any] = data.get("requires", {})
+        #
+        # Holds the raw parsed value, so before ``validate_workflow`` runs it may
+        # be a non-mapping (``None`` for a bare ``requires:``, a list for
+        # ``requires: []``, etc.); typed ``Any`` rather than ``dict[str, Any]``
+        # to avoid implying it is always a mapping at this point.
+        self.requires: Any = data.get("requires", {})
 
         # Inputs
         self.inputs: dict[str, Any] = data.get("inputs", {})
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -91,14 +91,14 @@ class WorkflowDefinition:
 # ID format: lowercase alphanumeric with hyphens
 _ID_PATTERN = re.compile(r"^[a-z0-9][a-z0-9-]*[a-z0-9]$|^[a-z0-9]$")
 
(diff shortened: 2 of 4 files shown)


mnriem (changes requested):

mnriem:

zied-jlassi:
Thanks @mnriem — addressed the Copilot feedback in the latest commit:

1. **Truthiness vs `None`** (`engine.py`) — validation now guards on `if definition.requires is not None:` instead of truthiness, so a falsy-but-non-mapping value (`requires: []` / `requires: ''`) is reported instead of silently skipped; a bare `requires:` (YAML null) is still treated as omitted. Added a regression test (`test_requires_empty_sequence_is_rejected_as_non_mapping`).
2. **Docs example** (`workflows.md`) — reworded the security note so `requires.permissions` is presented as *rejected/unsupported* rather than as a valid example of `requires`.
3. **Spelling** — standardised on US spelling to match the surrounding code (`_RECOGNIZED_REQUIRES_KEYS`, "recognized" in messages/comments).
4. **Test assertion** — the permissions-rejection test now asserts on specific markers (`requires.permissions` + the `gate` guidance) instead of the permissive `"permissions" + "not"`.

Validation: `tests/test_workflows.py` 292 passed (the 5 `requires` tests included). No new `ruff` findings on `src/`. Scope unchanged — still validation + honest docs, no runtime gate.

mnriem:
The core fix is solid — validating `requires` keys and rejecting `requires.permissions` with a pointer to a `gate` step is a real improvement, and the tests cover it well.

One change needed before this lands: **drop `tools` and `mcp` from the recognized set.** They're out of scope for this PR, and the justification doesn't hold for workflows:

- `PUBLISHING.md` (L59–62) and the workflow docs only ever document `speckit_version` and `integrations`. This PR's own `PUBLISHING.md` edit is what *introduces* the "recognised keys: …, `tools`, `mcp`" claim.
- `requires.tools` exists only in a **different schema** — the integration descriptor in `integrations/catalog.py` (~L709), not workflow `requires`.
- `requires.mcp` doesn't appear anywhere as a requires key (the only `mcp` hits are an unrelated `cursor-agent --approve-mcps` flag). Nothing resolves it for workflows or integrations.

So for workflows, `tools`/`mcp` have no docs (pre-PR), no schema, and no consumer. Whitelisting them undercuts the PR's own thesis: an author could write `requires.tools:`, pass validation, and have nothing read or enforce it — the same false-affordance trap you're closing for `permissions`.

Please:

1. Set `_RECOGNIZED_REQUIRES_KEYS = {"speckit_version", "integrations"}`.
2. Remove the `tools`/`mcp` additions from the `PUBLISHING.md` line.
3. Update the code comment so it no longer claims `tools`/`mcp` are "resolved by the bundler" for workflows.

Expanding the recognized `requires` surface is a separate decision and shouldn't ride along with this fix. If there's a future need for `tools`/`mcp` in workflow `requires`, that should go through its own issue/PR with docs and a real consumer.

zied-jlassi:
Thanks @mnriem — agreed, and you're right about the scope. I've dropped `tools` and `mcp` from `_RECOGNIZED_REQUIRES_KEYS`, so workflow `requires` now only recognizes `speckit_version` and `integrations`.

Root cause on my side: I'd sourced those two keys from the **bundle manifest** `requires` schema (`bundler/models/manifest.py`, resolved in `bundler/services/resolver.py`), which is a different schema from the workflow `requires` this PR validates. I also reverted the `PUBLISHING.md` line that introduced the `tools`/`mcp` claim — with those gone the existing docs are accurate again, which also resolves the four inline doc-consistency comments.

Validation: `pytest -k "requires or Validation"` green, ruff clean.

mnriem:

zied-jlassi:
Thanks @mnriem — addressed the latest Copilot feedback:

1. **`engine.py` annotation** — `WorkflowDefinition.requires` was typed `dict[str, Any]`, but it holds the raw parsed value, so before `validate_workflow()` runs it can be a non-mapping (`None` for a bare `requires:`, a list for `requires: []`, etc.). Retyped it as `Any` and added a comment noting that `validate_workflow()` enforces the mapping shape, so editors/type-checkers aren't misled into assuming it's always a dict. (`3f32bed`)
2. **PR description** — reconciled it with the implementation: the recognised `requires` keys are now listed as just `speckit_version` and `integrations` (dropped the stale `tools`/`mcp`, which we already removed from `_RECOGNIZED_REQUIRES_KEYS` per your earlier note).

Validation: `pytest tests/test_workflows.py` → 292 passed, `ruff check` clean. The annotation change is type-only (module uses `from __future__ import annotations`), no runtime behaviour change.

mnriem:

zied-jlassi:
Thanks @mnriem — addressed the Copilot feedback in `30191ec`.

`requires` now validates the same way as `inputs`: I dropped the `is not None` guard and check `isinstance(definition.requires, dict)` directly. A bare `requires:` (YAML null) is no longer treated as an omitted block — it now surfaces as `'requires' must be a mapping (or omitted).`, so a stray `requires:` line fails loudly instead of being silently ignored. A genuinely omitted `requires` still defaults to `{}` and stays valid.

Tests: added `test_requires_yaml_null_is_rejected_as_non_mapping` and a `test_requires_omitted_is_valid` guard (so the YAML-null rejection doesn't over-correct into flagging the omitted case), and refreshed the now-stale docstring on the empty-list test.

Verified in an isolated container against the repo's own toolchain:
- `uvx ruff check src/` → All checks passed!
- `uv run pytest tests/test_workflows.py` → 294 passed

mnriem (approved):

mnriem:
Thank you!

PR #3080, opened 2026-06-21
feat(extensions): verify catalog archive sha256 before install

zied-jlassi:
## Description

Extension and preset archives are downloaded over HTTPS and unpacked (with Zip-Slip protection), but the downloaded bytes were never checked against a known digest. Trust rested entirely on TLS plus the integrity of the third-party release host.

This is a defense-in-depth / supply-chain hardening change, not an exploit against spec-kit's own code. But the threat model is real: the publishing guide notes that maintainers do **not** audit extension code, so if a third-party release is tampered with or swapped (compromised release, hijacked `download_url`), the altered archive would be installed silently. Consumer-side integrity is the only remaining defense.

This PR lets a catalog entry pin an optional `sha256` digest:

```json
"download_url": "https://.../my-ext.zip",
"sha256": "<hex digest of the archive>"
```

- When present, the downloaded archive is verified **before it is written to disk** and installed; a mismatch aborts with a clear error.
- When absent, install behaves exactly as before (a `DEBUG` line records that the download was unverified), so this is fully backwards compatible.
- The check runs on **both** download paths — `ExtensionCatalog.download_extension` and `PresetCatalog.download_pack` — through a single shared helper (`shared_infra.verify_archive_sha256`) so the two cannot drift apart.

The optional `sha256` field is documented in both publishing guides.

## Testing

- [x] Ran existing tests with `pytest` (full suite green, no regressions)
- [x] Added tests: shared-helper unit tests (match / mismatch / `sha256:` prefix / case-insensitive / no-digest DEBUG) and download-path tests for matching, mismatch, and no-digest on both extensions and presets

## AI Disclosure

AI assistance (an AI coding agent) was used for the security review that identified the missing integrity check, and to help draft the implementation, tests, and documentation. All changes were reviewed and verified by me (red→green tests including path-parity across both download paths, full suite, `ruff`) before submitting.

### DIFF
--- a/extensions/EXTENSION-PUBLISHING-GUIDE.md
+++ b/extensions/EXTENSION-PUBLISHING-GUIDE.md
@@ -320,6 +320,7 @@ A: Extensions should be free and open-source. Commercial support/services are al
   "author": "string (required)",
   "version": "string (required, semver)",
   "download_url": "string (required, valid URL)",
+  "sha256": "string (optional, SHA-256 hex digest of the archive at download_url; verified before install)",
   "repository": "string (required, valid URL)",
   "homepage": "string (optional, valid URL)",
   "documentation": "string (optional, valid URL)",
--- a/presets/PUBLISHING.md
+++ b/presets/PUBLISHING.md
@@ -185,6 +185,7 @@ Edit `presets/catalog.community.json` and add your preset.
       "author": "Your Name",
       "version": "1.0.0",
       "download_url": "https://github.com/your-org/spec-kit-preset-your-preset/archive/refs/tags/v1.0.0.zip",
+      "sha256": "OPTIONAL: SHA-256 hex digest of the archive above; verified before install",
       "repository": "https://github.com/your-org/spec-kit-preset-your-preset",
       "license": "MIT",
       "requires": {
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -31,6 +31,7 @@ from .._invocation_style import is_dollar_skills_agent, is_slash_skills_agent
 from .._utils import dump_frontmatter, relative_extension_path_violation
 from ..catalogs import CatalogEntry as BaseCatalogEntry
 from ..catalogs import CatalogStackBase
+from ..shared_infra import verify_archive_sha256
 
 _FALLBACK_CORE_COMMAND_NAMES = frozenset(
     {
@@ -2616,6 +2617,10 @@ class ExtensionCatalog(CatalogStackBase):
                 download_url, timeout=60, extra_headers=extra_headers
             ) as response:
                 zip_data = response.read()
+
+            verify_archive_sha256(
+                zip_data, ext_info.get("sha256"), extension_id, ExtensionError
+            )
 
             zip_path.write_bytes(zip_data)
             return zip_path
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -31,6 +31,7 @@ from ..extensions import REINSTALL_COMMAND, ExtensionRegistry, normalize_priorit
 from .._init_options import is_ai_skills_enabled
 from ..integrations.base import IntegrationBase
 from .._utils import dump_frontmatter
+from ..shared_infra import verify_archive_sha256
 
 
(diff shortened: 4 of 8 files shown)


mnriem:

zied-jlassi:
Thanks @mnriem — went through the Copilot feedback. Here's what I changed and what I checked, honestly:

**Applied — the substantive one (`shared_infra.verify_archive_sha256`):**
- The digest parsing used `split(':', 1)[-1]`, which strips *any* prefix — so `md5:<valid-sha256-hex>` was silently accepted as a valid SHA-256. Now only a literal `sha256:` prefix (case-insensitive) is stripped, the value must be exactly 64 hex chars (clear error otherwise), and the comparison uses `hmac.compare_digest` (constant-time).
- Added two tests: a malformed digest and a non-`sha256:` prefix — both of which the previous code accepted. (Verified the old behaviour empirically before/after.)

**Checked but left as-is (Copilot false positives — happy to change if you prefer):**
- `__enter__ = lambda s: s` in the download test helpers does *not* raise `TypeError`: `MagicMock` binds an assigned function with `self`, so the `with` block enters fine (those 12 tests pass). I still switched to the canonical `__enter__.return_value = resp` form since it's clearer and arity-independent.
- `zipfile` in the `test_presets.py` helper is already imported at module level (line 19, used in ~9 tests), so there's no `NameError`; I kept it consistent with the file's existing convention rather than adding a redundant local import.

**On framing:** to be straight about severity — this is **defense-in-depth / supply-chain hardening**, not an exploitable vulnerability. Transport is already HTTPS; this adds optional consumer-side integrity pinning (`sha256` is opt-in, installs without it are unchanged). If you'd rather not carry the feature, no problem — feel free to close.

Validation: `test_shared_infra_integrity.py` 7 passed, `test_extensions.py` + `test_presets.py` 600 passed. No new `ruff` findings on `src/`.

mnriem (changes requested):

zied-jlassi:
Thanks @mnriem — addressed the remaining Copilot point and rebased.

1. **A declared-but-empty `sha256` no longer silently skips verification.** `verify_archive_sha256` guarded on `if not expected:`, so an empty digest (e.g. `sha256: ""` reaching the helper via `...get("sha256")`) disabled the integrity check instead of surfacing the authoring error. It now guards on `expected is None`, so only an absent digest skips; blank/whitespace/bare-`sha256:` values fall through to the 64-hex validation and are rejected. Added a regression test (`test_blank_declared_digest_is_rejected`).
2. **Rebased on `main`** to clear the merge conflict (the `extensions.py` → package refactor); resolved the import in `extensions/__init__.py` the same way the sibling `presets/__init__.py` imports the helper.

Validation: integrity + extensions + presets suites green (640 passed), ruff clean.

mnriem:

zied-jlassi:
Thanks @mnriem — addressed the Copilot comment on `shared_infra.py:20`:

The comment described `_SHA256_HEX_RE` as matching "64 lowercase" hex characters, but `verify_archive_sha256` lowercases the declared value (`expected_hex = raw.lower()`) before matching, so an uppercase digest is in fact accepted and normalized rather than rejected. Updated the comment to say so explicitly, pointing at the `raw.lower()` normalization. (`4e03399`)

Comment-only change, no behaviour change. Validation: `pytest tests/test_shared_infra_integrity.py` → 8 passed, `ruff check` clean.

mnriem:

zied-jlassi:
Thanks @mnriem — addressed the Copilot feedback in `ef31d4c`.

Added `test_download_pack_without_sha256_skips_verification`: it exercises the backwards-compatible path where a catalog entry has no `sha256` (`pack_info.get("sha256")` is `None`), asserting `download_pack` still downloads instead of aborting. This mirrors the existing extensions coverage (`test_download_extension_without_sha256_still_succeeds`) so the integrity helper can't silently become mandatory for presets via a future regression.

Verified in an isolated container against the repo's own toolchain:
- `uvx ruff check src/` → All checks passed!
- `uv run pytest tests/test_presets.py` → 318 passed (the three `sha256` cases: match / mismatch / absent)

mnriem (approved):

mnriem:
Thank you!

Issue #3086, opened 2026-06-22
specify init does not copy templates/commands/ to .specify/templates/commands/ — wrap composition strategy fails


djimit:
### Problem

During `specify init`, the `install_shared_infra` function in `shared_infra.py` copies template files from the bundled `templates/` directory to `.specify/templates/`. However, it only processes **files** at the top level (`src.is_file()` check), skipping subdirectories — specifically `templates/commands/`.

This means the core command templates (e.g., `implement.md`, `specify.md`, `plan.md`, etc.) are installed directly to the agent directory (e.g., `.opencode/commands/speckit.implement.md`) via the integration system, but they are **not** kept in `.specify/templates/commands/` for the preset resolver to find.

### Impact

Preset authors who use the `wrap` composition strategy for command overrides (e.g., wrapping `speckit.implement` with a preamble) cannot resolve the core command as a base layer. The `PresetResolver.collect_all_layers()` method looks for core commands at:

```python
c = self.templates_dir / "commands" / f"{template_name}.md"  # or stem fallback
```

Since `.specify/templates/commands/` doesn't exist after `specify init`, the resolver finds no base layer, composition fails, and the command is silently not registered (`registered_commands` stays empty `{}`).

### Reproduction

```bash
specify init --here --integration opencode --script sh --ignore-agent-tools
# .specify/templates/commands/ does NOT exist
# .opencode/commands/speckit.implement.md exists (installed by integration)

# Create a preset with a wrap strategy for speckit.implement
specify preset add --dev /path/to/my-preset
# registered_commands is {} — the wrap composition failed silently
```

### Workaround

Manually copy command templates after init:

```bash
mkdir -p .specify/templates/commands
cp <spec-kit-source>/templates/commands/*.md .specify/templates/commands/
```

### Suggested fix

In `shared_infra.py`, the `install_shared_infra` function should also copy the `templates/commands/` subdirectory (with the same overwrite/manifest logic). Something like:

```python
# After the top-level template files loop:
commands_src = templates_src / "commands"
if commands_src.is_dir():
    dest_commands = dest_templates / "commands"
    if _ensure_or_bucket_dir(dest_commands):
        for src_path in commands_src.iterdir():
            if not src_path.is_file() or src_path.name.startswith("."):
                continue
            dst = dest_commands / src_path.name
            rel = dst.relative_to(project_path).as_posix()
            # ... same overwrite/manifest logic as above ...
```

[...]

PR #3091, opened 2026-06-22
fix(presets): use _repo_root() for bundled-core source-checkout fallback (#3086)

mnriem:
## Summary

Assessment and fix for #3086.

The issue claims `specify init` not copying `templates/commands/` into `.specify/templates/commands/` causes `wrap`-strategy presets to fail. That root cause is **incorrect**: `PresetResolver` has an explicit Priority‑5 fallback (`_find_bundled_core`, plus the matching tier‑5 branch in `resolve()`) specifically designed to locate the core base layer when `.specify/templates/commands/` is absent. The missing copy is a layout/cosmetic gap, not a functional one.

However, while verifying, I found a **real** bug that *does* break `wrap` presets — but only in **source/editable installs**, and for a different reason:

Both fallbacks computed the repo root as `Path(__file__).parent.parent.parent`. After `presets.py` was moved to `presets/__init__.py` (#2826), that chain is one level short — it resolves to `src/` and looks for `src/templates/commands/<cmd>.md`, which never exists. So the base layer was never found.

Wheel installs were unaffected because they take the `core_pack` branch.

## Fix

Use the shared `_repo_root()` helper (from `_assets.py`) in both `resolve()` and `_find_bundled_core()` instead of a hand-rolled `.parent` chain, so the path stays correct regardless of module nesting.

## Tests

- Added two regression tests asserting the bundled-core fallback resolves `speckit.implement` via both `collect_all_layers()` and `resolve()` when `.specify/templates/commands/` has no matching file.
- Full preset suite (312 tests) passes.

## Why not the issue's suggested fix

Copying `templates/commands/` into `.specify/templates/commands/` during `specify init` touches the core SDD install/manifest/upgrade flow to solve a non-problem (the fallback already covers wheel installs). This PR keeps the change isolated to the preset resolver and does not alter the SDD process.

Refs #3086

### DIFF
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -2703,7 +2703,7 @@ class PresetResolver:
         # (source-checkout / editable install).  This is the canonical home for
         # speckit's built-in command/template files and must always be checked
         # so that strategy:wrap presets can locate {CORE_TEMPLATE}.
-        from specify_cli import _locate_core_pack  # local import to avoid cycles
+        from specify_cli import _locate_core_pack, _repo_root  # local import to avoid cycles
         _core_pack = _locate_core_pack()
         if _core_pack is not None:
             # Wheel install path
@@ -2723,7 +2723,7 @@ class PresetResolver:
                 return candidate
         else:
             # Source-checkout / editable install: templates live at repo root
-            repo_root = Path(__file__).parent.parent.parent
+            repo_root = _repo_root()
             if template_type == "template":
                 candidate = repo_root / "templates" / f"{template_name}.md"
             elif template_type == "command":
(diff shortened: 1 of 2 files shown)


mnriem:
Good catch — the `def test_resolve_extension_command_via_manifest_skips_oserror_manifests` line had been dropped, orphaning that test's body inside `test_resolve_command_falls_back_to_bundled_core`. Restored the definition in c52ee42; pytest now collects it again (suite goes 312 → 313, all green).

_Posted on behalf of @mnriem by GitHub Copilot CLI (model: Claude Opus 4.8)._

mnriem:
Moved both tier-5 fallback regression tests out of `TestResolveCore` and into `TestPresetResolver` in 129514c, since they exercise `collect_all_layers()` and `resolve()` rather than `resolve_core()`. Suite still green (313 passing).

_Posted on behalf of @mnriem by GitHub Copilot CLI (model: Claude Opus 4.8)._

PR #3123, opened 2026-06-23
fix(scripts): send check-prerequisites.ps1 errors to stderr

### LINKED ISSUE #3122: [Bug]: check-prerequisites.ps1 writes errors to stdout, corrupting --json output
### Bug Description

the powershell `check-prerequisites.ps1` writes its validation errors and run-hints with `Write-Output`, which goes to stdout. this script is normally run with `-Json` and the agent parses its stdout as json. when validation fails (e.g. plan.md is missing), the error text lands on stdout instead of json, so the consumer gets something like `ERROR: plan.md not found ...` where it expects a json object, and parsing breaks.

the bash counterpart doesn't have this - it sends the same errors to stderr (`>&2`), keeping stdout clean. the sibling powershell scripts are also fine: `setup-tasks.ps1` and `common.ps1`'s `Get-FeaturePathsEnv` both use `[Console]::Error.WriteLine`. only `check-prerequisites.ps1` is the odd one out.

### Steps to Reproduce

1. in a spec-kit project with a feature dir but no `plan.md`, run:
   `pwsh .specify/scripts/powershell/check-prerequisites.ps1 -Json`
2. capture stdout only (e.g. redirect stderr away).
3. observe stdout contains the error string instead of json.

compare with bash:
`bash .specify/scripts/bash/check-prerequisites.sh --json` - stdout stays empty, the error goes to stderr.

### Expected Behavior

errors and run-hints go to stderr; stdout only ever carries the json (or the documented text payload). both shells behave the same.

### Actual Behavior

the six error/hint lines (check-prerequisites.ps1 lines 86, 88, 93, 95, 101, 103 on main) use `Write-Output`, so they go to stdout and pollute the `-Json` output.

### Additional Context

suggested fix: switch those six lines to `[Console]::Error.WriteLine`, matching the bash script and the other powershell scripts. small diff.

Quratulain-bilal:
closes #3122

`check-prerequisites.ps1` wrote its validation errors and run-hints with `Write-Output`, so they went to stdout. the script is usually run with `-Json` and its stdout is parsed as json by the agent, so when validation fails (e.g. missing plan.md) the parser gets an error string instead of json and breaks.

the bash counterpart already sends these to stderr (`>&2`), and the sibling powershell scripts (`setup-tasks.ps1`, `common.ps1`'s `Get-FeaturePathsEnv`) already use `[Console]::Error.WriteLine`. this switches the six error/hint lines in `check-prerequisites.ps1` to `[Console]::Error.WriteLine` so stdout stays clean and the two shells behave the same.

no behaviour change on the success path - only the failure paths, which still `exit 1`.

note: i used an ai assistant to help investigate and prepare this change. disclosing per CONTRIBUTING.md.

### DIFF
--- a/scripts/powershell/check-prerequisites.ps1
+++ b/scripts/powershell/check-prerequisites.ps1
@@ -83,24 +83,24 @@ if ($PathsOnly) {
 
 # Validate required directories and files
 if (-not (Test-Path $paths.FEATURE_DIR -PathType Container)) {
-    Write-Output "ERROR: Feature directory not found: $($paths.FEATURE_DIR)"
+    [Console]::Error.WriteLine("ERROR: Feature directory not found: $($paths.FEATURE_DIR)")
     $specifyCommand = Format-SpecKitCommand -CommandName 'specify' -RepoRoot $paths.REPO_ROOT
-    Write-Output "Run $specifyCommand first to create the feature structure."
+    [Console]::Error.WriteLine("Run $specifyCommand first to create the feature structure.")
     exit 1
 }
 
 if (-not (Test-Path $paths.IMPL_PLAN -PathType Leaf)) {
-    Write-Output "ERROR: plan.md not found in $($paths.FEATURE_DIR)"
+    [Console]::Error.WriteLine("ERROR: plan.md not found in $($paths.FEATURE_DIR)")
     $planCommand = Format-SpecKitCommand -CommandName 'plan' -RepoRoot $paths.REPO_ROOT
-    Write-Output "Run $planCommand first to create the implementation plan."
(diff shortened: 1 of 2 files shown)


mnriem:
Can you also address the requested test changes as per #3122 bug assessment?

Quratulain-bilal:
done - tightened the failure-path tests to assert stdout stays clean (empty/valid json) and the error only shows on stderr, instead of the old combined stdout+stderr check. added cases for the missing plan.md and missing tasks.md (-RequireTasks) paths too. confirmed the two error-routing tests fail on the pre-fix script and pass after the fix.

mnriem:
Thank you!

PR #3130, opened 2026-06-23
fix(scripts): match bash case-sensitivity for acronym retention in PowerShell branch names

### LINKED ISSUE #3131: [Bug]: PowerShell create-new-feature keeps lowercase short words in branch names (case-insensitive -match)
### Bug Description

`create-new-feature` keeps a short (<3 char) word in the generated branch name only when that word appears in **uppercase** in the description, so acronyms like `AI` or `ML` survive while ordinary short words are dropped. the code comment says as much.

the bash script does this with a case-sensitive `grep`:

```bash
elif echo "$description" | grep -q "${word^^}"; then
```

the powershell script uses `-match`, which is **case-insensitive by default**:

```powershell
} elseif ($Description -match "$($word.ToUpper())") {
```

because the word being tested was lowercased earlier, `-match` against its uppercased form always matches, so on powershell every short non-stop word is kept, not just acronyms.

### Steps to Reproduce

1. on powershell, run `create-new-feature.ps1 -DryRun "go AI now"`
2. compare with bash `create-new-feature.sh --dry-run "go AI now"`

### Expected Behavior

both shells produce the same branch name. `go` (lowercase 2-char word) is dropped, `AI` (uppercase acronym) is kept -> `001-ai-now`.

### Actual Behavior

powershell keeps the lowercase `go` too -> `001-go-ai-now`, while bash gives `001-ai-now`. same input, different branch name depending on the shell.

| description | bash | powershell |
|---|---|---|
| `go AI now` | `001-ai-now` | `001-go-ai-now` |
| `go to the pub` | `001-pub` | `001-go-pub` |

### Specify CLI Version

main (current)

### AI Agent

n/a (script-level)

### Operating System

cross-platform (powershell)

### Additional Context

suggested fix: use `-cmatch` (case-sensitive) instead of `-match` in `scripts/powershell/create-new-feature.ps1`, matching the bash `grep`. small diff.

Quratulain-bilal:
closes #3131

## what

`create-new-feature` keeps a short (<3 char) word in the generated branch name only when that word appears in **uppercase** in the description — the idea is to preserve acronyms like `AI` or `ML` while dropping ordinary short words. The code comment says exactly this.

the bash script does the check with a case-sensitive `grep`:

```bash
elif echo "$description" | grep -q "${word^^}"; then
```

but the PowerShell script used `-match`, which is **case-insensitive by default**:

```powershell
} elseif ($Description -match "$($word.ToUpper())") {
```

since the word being tested is itself lowercased earlier, `-match` against its uppercased form always matches — so on PowerShell *every* short non-stop word was kept, not just acronyms.

## impact

same description, different branch name depending on the shell:

| description | bash | powershell (before) |
|---|---|---|
| `go AI now` | `001-ai-now` | `001-go-ai-now` |
| `go to the pub` | `001-pub` | `001-go-pub` |

`go` is a 2-char non-stop word that is lowercase, so it should be dropped (and is, on bash). on PowerShell it leaked into the branch name.

## fix

switch `-match` to `-cmatch` so the comparison is case-sensitive, matching the bash `grep`. one-character change plus a clarifying comment.

## tests

added parity tests (`TestShortWordRetentionBash` / `...PowerShell`) covering a dropped lowercase short word and a kept uppercase acronym. verified the PowerShell case fails on the pre-fix script (`go AI now` -> `001-go-ai-now`) and passes after.

---

note: i used an ai assistant to help investigate and prepare this change. disclosing per CONTRIBUTING.md.

### DIFF
--- a/scripts/powershell/create-new-feature.ps1
+++ b/scripts/powershell/create-new-feature.ps1
@@ -111,8 +111,11 @@ function Get-BranchName {
         # Keep words that are length >= 3 OR appear as uppercase in original (likely acronyms)
         if ($word.Length -ge 3) {
             $meaningfulWords += $word
-        } elseif ($Description -match "\b$($word.ToUpper())\b") {
-            # Keep short words if they appear as uppercase in original (likely acronyms)
+        } elseif ($Description -cmatch "\b$($word.ToUpper())\b") {
+            # Keep short words only if they appear as uppercase in original (likely
+            # acronyms). Use -cmatch so the comparison is case-sensitive, matching the
+            # bash script's case-sensitive grep; -match would be case-insensitive and
+            # would keep every short word.
             $meaningfulWords += $word
         }
     }
--- a/tests/test_timestamp_branches.py
+++ b/tests/test_timestamp_branches.py
@@ -869,6 +869,52 @@ class TestPowerShellDryRun:
         assert "DRY_RUN" not in data, f"DRY_RUN should not be in normal JSON: {data}"
 
 
+# ── Short-Word / Acronym Branch-Name Tests ──────────────────────────────────
+
+
+def _branch_from_output(stdout: str) -> str | None:
+    for line in stdout.splitlines():
+        if line.startswith("BRANCH_NAME:"):
+            return line.split(":", 1)[1].strip()
+    return None
+
+
+SHORT_WORD_CASES = [
+    # description, expected branch — "go" (lowercase short word) is dropped,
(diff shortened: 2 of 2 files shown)


Quratulain-bilal:
@mnriem whenever you get a chance - this one's ready: small case-sensitivity fix with parity tests for both shells. happy to adjust if anything looks off.

mnriem:
Thank you!

PR #3132, opened 2026-06-23
harden: reject shell=True in run_command

PascalThuet:
Part of splitting #2442 into smaller, dedicated PRs. Re-derived against current `main` (the #2442 branch is ~20 commits behind).

## What
- `run_command()` now rejects `shell=True` with `ValueError` instead of forwarding it to `subprocess.run`. The `shell` parameter is kept (default `False`) for signature / keyword-caller compatibility.
- Enable ruff `S602`/`S604`/`S605` so any future `shell=True` (or `os.system`/`popen`) must be acknowledged with an explicit `# noqa`.
- Annotate the one intentional shell sink (the workflow `ShellStep`) with `# noqa: S602`.
- Document the shell-step arbitrary-code-execution risk in `workflows/PUBLISHING.md`.

## Why
`run_command(shell=True)` would run its argv through a shell (word-splitting, globbing, injection surface). No call site needs that mode; failing loudly prevents a future caller from silently enabling it.

## Validation
- `tests/test_utils.py::test_run_command_rejects_shell_execution_compatibly` passes.
- `ruff check --select S602,S604,S605 src tests/test_utils.py` is clean.

Independent of the other split PRs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,3 +74,13 @@ precision = 2
 show_missing = true
 skip_covered = false
 
+[tool.ruff.lint]
+# Lock in subprocess security posture: any reintroduction of shell=True
+# (or os.system / popen2) must be acknowledged with an explicit `# noqa`
+# pointing at the rule, making the deviation visible in review.
+extend-select = [
+    "S602",  # subprocess-popen-with-shell-equals-true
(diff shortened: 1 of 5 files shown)


mnriem:
Thank you!

PR #3134, opened 2026-06-23
fix(workflows): preserve commas inside quoted list-literal elements

jawwad-ali:
## Description

The workflow simple-expression evaluator parsed list literals with a naive `inner.split(",")`:

```python
# src/specify_cli/workflows/expressions.py
items = [_evaluate_simple_expression(i.strip(), namespace) for i in inner.split(",")]
```

That splits on commas **inside quoted strings** (and nested brackets), so a list element containing a comma is silently broken into multiple elements:

- `{{ ["a, b", "c"] }}` → `["a", "b", "c"]` (3 items) instead of `["a, b", "c"]` (2 items)
- `{{ ["x, y, z"] }}` → `["x", "y", "z"]` instead of `["x, y, z"]`

This silently corrupts any list expression with a comma inside a quoted element — most importantly a `fan-out` step's `items:`, where it changes the number of iterations and the data passed to each one, with no error raised.

### Fix

Split list-literal elements on **top-level commas only**, ignoring commas inside quotes or nested brackets, via a small `_split_top_level_commas` helper. Per-element evaluation is unchanged (elements can still be variables, nested lists, etc.), and plain/empty lists behave exactly as before.

## Testing

- `uv run pytest tests/test_workflows.py` — **295 passed, 1 skipped**; the only failures on my Windows host are the pre-existing `os.symlink` `WinError 1314` privilege tests (unrelated to this change).
- New test `tests/test_workflows.py::TestExpressions::test_list_literal_preserves_quoted_commas` covers quoted commas, nested lists, and the existing plain/empty cases. It **fails on `main`** (`["a", "b", "c"]`) and **passes** with this fix.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI located the naive split, implemented the quote/bracket-aware helper, and wrote the tests; I confirmed the corruption reproduces on `main` and that the fix resolves it (and is a no-op for plain lists) before submitting. I'll disclose if any review responses are AI-assisted as well.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -146,6 +146,40 @@ def _build_namespace(context: Any) -> dict[str, Any]:
     return ns
 
 
+def _split_top_level_commas(text: str) -> list[str]:
+    """Split *text* on commas that are not inside quotes or nested brackets.
+
+    Used for list-literal elements so a quoted element containing a comma
+    (e.g. ``["a, b", "c"]``) is not split mid-string, and nested lists/calls
+    (e.g. ``[[1, 2], 3]``) are kept intact.
+    """
+    parts: list[str] = []
+    buf: list[str] = []
+    quote: str | None = None
+    depth = 0
+    for ch in text:
+        if quote is not None:
+            buf.append(ch)
+            if ch == quote:
+                quote = None
+        elif ch in ("'", '"'):
+            quote = ch
+            buf.append(ch)
+        elif ch in "([{":
+            depth += 1
+            buf.append(ch)
+        elif ch in ")]}":
+            depth = max(0, depth - 1)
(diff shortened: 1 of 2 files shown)


jawwad-ali:
Thanks for the review! Addressed in 70b6ef7 — extended `test_list_literal_preserves_quoted_commas` to also assert single-quoted elements (`['a, b', 'c']`, `['p, q, r']`) and a nested-list-of-lists case (`[[1, 2], [3, 4]]`) alongside the existing double-quoted/nested coverage. `ruff` clean; the test passes.

_AI disclosure: this response and the test additions were prepared with Claude Code (Claude Opus 4.8) under my direction; I reviewed the diff before pushing._

mnriem:
Thank you!

PR #3135, opened 2026-06-23
fix: render valid TOML when a command body contains backslashes

jawwad-ali:
## Description

`CommandRegistrar.render_toml_command()` renders a command body into a multiline **basic** TOML string (`"""..."""`). Basic strings process backslash escape sequences, but the body was emitted raw — so any backslash in the body breaks the output:

- A Windows path like `C:\Users\...` contains `\U`, which TOML reads as an 8-digit unicode escape → the agent's TOML loader fails with `Invalid hex value` and the generated **Gemini / Tabnine** command file is completely unparseable.
- A body ending in a single `\` becomes a TOML line-continuation that silently swallows the closing newline.

Reproduced with `tomllib`:

```python
render_toml_command({}, r"Run C:\Users\dev\tool.exe", "src")
# -> prompt = """ ... C:\Users\dev\tool.exe ... """
# tomllib.loads(...) -> TOMLDecodeError: Invalid hex value
```

### Fix

Route bodies containing a backslash away from the basic-string branch: prefer the multiline **literal** form (`'''...'''`), which does not process escapes, and fall back to the escaped basic string (`_render_basic_toml_string`) when both triple-quote styles are present. This mirrors the backslash escaping already done by `base.py`'s `TomlIntegration`. One-line condition change; behavior is unchanged for backslash-free bodies.

## Testing

- `uv run pytest tests/test_extensions.py tests/integrations/test_integration_gemini.py tests/integrations/test_integration_tabnine.py` — **381 passed, 3 skipped** (no regressions; the existing 3 `render_toml_command` tests still pass).
- New tests in `tests/test_extensions.py`: a Windows-path body, a trailing-backslash body, and a backslash + both-triple-quote-styles fallback — each asserts `tomllib.loads()` succeeds and the body round-trips. The first two **fail on `main`** (`TOMLDecodeError` / dropped backslash) and pass with this fix.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI located the unescaped basic-string branch, confirmed the existing `TomlIntegration` escaping precedent, implemented the routing fix, and wrote the tests; I reproduced the `tomllib` parse failure on `main` and verified the fix + no regressions in the Gemini/Tabnine suites before submitting. I'll disclose if any review responses are AI-assisted as well.

### DIFF
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -236,9 +236,14 @@ class CommandRegistrar:
         toml_lines.append(f"# Source: {source_id}")
         toml_lines.append("")
 
-        # Keep TOML output valid even when body contains triple-quote delimiters.
-        # Prefer multiline forms, then fall back to escaped basic string.
-        if '"""' not in body:
+        # Keep TOML output valid even when body contains triple-quote delimiters
+        # or backslashes. Prefer multiline forms, then fall back to escaped basic
+        # string. A multiline *basic* string ("""...""") processes backslash escape
+        # sequences, so a body containing a backslash (e.g. a Windows path
+        # ``C:\\Users\\...`` whose ``\\U`` reads as an invalid unicode escape) would
+        # produce unparseable TOML — route those to the *literal* form ('''...'''),
+        # which does not process escapes, or to the escaped basic string.
+        if '"""' not in body and "\\" not in body:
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3143, opened 2026-06-24
Update preset composition strategy reference

WOLIKIMCHENG:
## Summary
- update the presets reference to describe the implemented composition strategies
- clarify that templates and commands support prepend, append, and wrap
- note that scripts support replace and wrap with `$CORE_SCRIPT`

## Testing
- git diff --check
- npx markdownlint-cli2 docs/reference/presets.md

### DIFF
--- a/docs/reference/presets.md
+++ b/docs/reference/presets.md
@@ -139,7 +139,7 @@ catalogs:
 
 Presets can provide command files, template files (like `plan-template.md`), and script files. Each file name is evaluated independently against the priority stack, so different files can come from different layers.
 
-Templates and scripts are looked up from the stack when Spec Kit needs them. Commands use the same stack for replacement and composition, but are materialized into detected agent directories instead of being re-resolved by agents. During preset install, Spec Kit registers command files for the preset being installed; post-install and post-removal reconciliation then recalculates affected command names from the active stack and writes the effective content. Agents do not re-resolve the stack each time they run a command.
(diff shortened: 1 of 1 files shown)


mnriem (changes requested):

WOLIKIMCHENG:
Thanks for the review. Updated the File Resolution section to separate template/script runtime stack lookup from command composition during preset install, before commands are registered. This should address the Copilot feedback.

mnriem:

WOLIKIMCHENG:
Thank you for the follow-up review. I updated the File Resolution text again to avoid implying that command composition always happens before registration. The section now separates the supported composition strategies from when command content is materialized: preset install and post-install/post-removal reconciliation write the effective command content into detected agent directories, while agents do not re-resolve the stack each time a command runs. I appreciate the clarification.

mnriem:

WOLIKIMCHENG:
Thank you for the additional review and clarification. I updated the File Resolution section to make the per-file stack behavior explicit for command, template, and script files before describing their timing differences.

It now clarifies that templates and scripts are looked up when Spec Kit needs them, while commands use the same stack for replacement and composition but are materialized into detected agent directories: preset install registers the command files, and post-install/post-removal reconciliation recalculates affected command names from the active stack and writes the effective content. I also tightened the disable/remove FAQ so it does not imply that disabling a preset stops previously registered commands.

I appreciate the continued guidance.

mnriem:

WOLIKIMCHENG:
Thanks for the review. I updated the wording to clarify that reconciliation recomputes and writes the effective command content for affected command names, rather than recalculating the command names themselves.

mnriem (approved):

mnriem:
Thank you!

PR #3144, opened 2026-06-24
test: isolate integration test home

PascalThuet:
## What

Isolate integration-test HOME/USERPROFILE and XDG directories to each test tmp path.

## Why

Some integrations intentionally write to user-global locations such as `~/.hermes/skills`. Without a shared isolation fixture, inherited/base integration tests can write to the real developer or CI home and also block safe parallel execution later.

## Validation

- `uv sync --extra test`
- `uvx ruff check tests/integrations/conftest.py`
- `git diff --check`
- `.venv/bin/python -m pytest tests/integrations/test_integration_hermes.py -q` (`34 passed`)
- `.venv/bin/python -m pytest tests/integrations --durations=30 -q` (`2688 passed, 1 skipped in 100.07s`)

## Disclosure

Prepared by Codex (model: GPT-5) on behalf of @PascalThuet.


mnriem:
Good instinct, but drop the XDG vars — they're not read anywhere — and either align with the existing Path.home monkeypatch or add an assertion so this actually fails loudly when someone forgets

mnriem (changes requested):

mnriem (approved):

mnriem:
Thank you!

PR #3141, opened 2026-06-23
harden: secure extension and preset archive downloads

PascalThuet:
Originally rebased onto `main` at `4d3a428`. The published head now includes a validated merge of current `main` (`0117a7b`). #3140 and #3671 are merged, so this PR reuses their shared URL, redirect, bounded-read, and checksum primitives instead of carrying the #3140 commit.

## What

- Replace path-only ZIP extraction for extensions and presets with bounded extraction that rejects traversal, symlinks, ambiguous normalized paths, file/descendant conflicts, non-portable Windows names, oversized members, oversized totals, and excessive entry counts.
- Preflight ZIP32 central directories and local headers before `ZipFile` materializes entries, including declared/actual entry counts, a central-directory size cap, and archive-level or per-entry ZIP64 indicators.
- Bound extension/preset catalog JSON, package downloads, direct URL installs, and extension-update manifest reads.
- Fully validate extension-update manifests, compatibility, normalized catalog/package version equality, and install conflicts before removing the installed extension.
- Complete recoverable backups before the destructive update boundary, including registered skill directories; rollback removes newly created skills and restores the original registry, commands, and skills.
- Keep extension-update manifest selection consistent with extraction across slash, case, and Unicode normalization while preserving root-manifest priority over a distinct nested manifest.
- Reject malformed URL types/hosts/ports and catalog-controlled output filenames that could escape the requested download directory.
- Verify a catalog SHA-256 only when the entry provides one; checksum presence remains backward-compatible and optional.
- Escape untrusted archive/download values in errors so terminal control sequences are not emitted.

## Security and compatibility boundaries

- Downloaded archives are capped at 50 MiB; extracted archives at 512 entries, 10 MiB per member, 50 MiB total, and a 4 MiB central directory.
- Archive paths are capped at 255 UTF-8 bytes per component and 4096 UTF-8 bytes overall.
- ZIP64 is deliberately rejected, including per-entry central/local ZIP64 markers and extra fields.
- Bounded extraction accepts ZIP32 STORE and DEFLATE entries. BZIP2 and LZMA entries are rejected before `ZipFile` because the standard-library read path cannot enforce the requested decompressed-output bound for those methods.
- SHA-256 validation is not catalog-wide integrity until catalogs publish hashes consistently.

## Validation

- Full suite on the published merged head (`ab2a825`, `main` at `0117a7b`): `5641 passed, 172 skipped`.
- Pre-merge full branch suite: `5427 passed, 166 skipped`.
[...]

PR #3140, opened 2026-06-23
harden: bound HTTP reads and enforce strict redirects

PascalThuet:
Part of splitting #2442 into smaller, dedicated PRs. Rebased onto current `main`.

## What
- Add shared bounded-response and URL-scheme helpers in `src/specify_cli/_download_security.py`.
- Enforce strict redirect safety for both authenticated and unauthenticated requests while preserving the existing redirect validator, GHES handling, auth stripping, and malformed-URL behavior.
- Bound GitHub release metadata, self-upgrade metadata, and Azure DevOps token response reads.
- Route the Azure DevOps token POST through the guarded redirect opener.
- Keep HTTP test fakes cursor-aware and reuse shared helpers where possible.

## Why
Unbounded remote JSON reads expose a memory-denial-of-service surface. Unsafe redirect downgrades can also expose credentials or silently weaken transport. The change closes both paths while preserving the existing happy path and current `main` validation behavior.

## Conflict and review resolution
- Preserved the configurable GitHub metadata limit and redirect validator added on `main` while switching its implementation to the shared chunked reader.
- Preserved current GHES shape validation and malformed-redirect handling.
- Made strict redirect handling unconditional, including the no-token fallback used by catalogs and downloads.
- Updated the Azure AD network-error test to intercept the guarded opener rather than bare `urlopen`.
- Reused the shared opener-routing fixture in affected extension and integration-catalog tests.
- Removed obsolete duplicate HTTP fake classes rather than carrying parallel helpers.

## Validation
- `uvx ruff check src tests` — passed.
- HTTP/security and affected-caller tests — 1,675 passed.
- Full suite — 4,786 passed, 153 skipped.
- `git diff --check` — passed.

Disclosure: Updated on behalf of @PascalThuet by OpenAI Codex (model: GPT-5, autonomous).

PR #3145, opened 2026-06-24
harden: bound remaining catalog reads + lock in no-unbounded-read invariant

PascalThuet:
Part of splitting #2442 into smaller, dedicated PRs. Re-derived against current `main`. **Final PR of the runtime-hardening stack.**

> **Stacked on #3141** (which is itself stacked on #3140). Its diff includes those commits — review/merge #3140 → #3141 → this, in order.

## What
- Bound the remaining catalog JSON reads via `read_response_limited(max_bytes=MAX_JSON_CATALOG_BYTES)` and pass `strict_redirects=True`:
  - `workflows/catalog.py`, `integrations/catalog.py`, `bundler/services/adapters.py`, `commands/bundle/__init__.py`.
- `__init__.py` `workflow_add` / `workflow_step_add`: bounded reads + `strict_redirects=True` + shared `is_https_or_localhost_http` URL validation.
- `extensions/_commands.py`: bound the `extension add --from <url>` ZIP download — the last unbounded network read (surfaced by the gate below).

## The invariant
Adds `tests/test_download_security.py::test_remote_downloads_do_not_use_unbounded_response_reads`: an AST scan over the whole `src/specify_cli` tree that fails CI if any `response.read()` is unbounded, with a 3-entry allowlist for the local-file `get_hash` reads (extensions/presets/integrations). This only goes green once **#3140 + #3141 + this** have converted every remote read — it's the lock that keeps the bound from silently regressing.

## Validation
- `test_download_security` (incl. the AST gate), `test_workflows`, `test_integration_catalog`, `test_bundler_adapters`, `test_upgrade`, `test_extensions` — **792 passed**.
- Bundle integration tests — 10 passed. `ruff check` clean.
- Existing catalog/workflow tests pass with stream mocks made cursor-based (so they exercise the bounded reads) and `strict_redirects` threaded through the fake openers.

This completes the runtime hardening. The Bandit + detect-secrets CI gates land in a separate PR on the #2/#3 track.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -39,6 +39,9 @@ import typer
 from rich.panel import Panel
 from rich.align import Align
 from rich.table import Table
+from ._download_security import (
+    is_https_or_localhost_http,
+)
 from .shared_infra import (
     install_shared_infra as _install_shared_infra_impl,
     refresh_shared_templates as _refresh_shared_templates_impl,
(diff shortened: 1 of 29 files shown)


mnriem:
Converted to a draft as it is waiting for #3141

mnriem:
Thanks for your contributions here, @PascalThuet — a quick note on review prioritization. You currently have **4 open pull requests**, above the three-open-PR guideline in [CONTRIBUTING](https://github.com/github/spec-kit/blob/main/CONTRIBUTING.md). This isn't a freeze — your open PRs stay in the queue and can still be reviewed and merged. But beyond three, additional submissions may be placed behind other contributors' work, so the most effective path is to **consolidate where you can, or tell us the few you'd most like prioritized**, and I'll focus on those first. I'm adding the `author-over-cap` label here purely as a tracking marker; no action needed beyond letting us know your priority order.

PR #3146, opened 2026-06-24
test: reduce registry manifest test repetition

PascalThuet:
## What

Reduce the expensive multi-install manifest contract from one CLI setup per safe-integration pair to one aggregate setup per multi-install-safe rotation.

Each rotation places a different safe integration first, so every safe integration still exercises the `specify init --integration ...` path at least once. The test then reuses the resulting manifests for the pairwise manifest-disjoint assertions instead of repeating equivalent setup for every pair.

## Stack

Stacked on #3144. Because this fork cannot push the stack base branch to `github/spec-kit` (`403`), this PR targets `main` for upstream visibility and will include the #3144 diff until #3144 lands. After #3144 is merged, this PR diff should shrink to only `tests/integrations/test_registry.py`.

## Validation

- `git diff --check`
- `.venv/bin/python -m pytest tests/integrations/test_registry.py -q` (`443 passed in 15.18s`)

Earlier validation before the final merge from `origin/main`:

- `uv sync --extra test`
- `uvx ruff check tests/integrations/test_registry.py`
- `.venv/bin/python -m pytest tests/integrations/test_registry.py --durations=30 -q` (`1035 passed in 34.86s`)
- `.venv/bin/python -m pytest tests/integrations --durations=30 -q` before the final naming/comment cleanup (`2500 passed, 1 skipped in 236.69s`)

## Disclosure

Prepared by Codex (model: GPT-5) on behalf of @PascalThuet.


PascalThuet:
Pushed `ca07e0e` for the review feedback on `test_safe_integrations_have_disjoint_manifests`. The assertion is unchanged; this is clarity plus a guard.

- Documented why both orders run. Manifest contents don't depend on install order: each safe integration writes only to its own disjoint directories and always records what it writes, so installing the whole set into one project gives the same per-integration manifests either way. `forward`/`reverse` doesn't vary the manifests; it just routes a different integration through `init` vs `integration install` (forward inits the first key, reverse the last). The old comment didn't say this and read as if order mattered.
- Added a `len(ordered_keys) >= 2` precondition. With fewer than two safe integrations the contract is vacuous and `ordered_keys[0]` would fail on an empty set, so a shrunken registry now fails here instead of passing silently.
- Built the manifest map with a comprehension and pulled out `integrations_dir`. Cosmetic.

Validation: `ruff check` clean; `pytest tests/integrations/test_registry.py` -> `1035 passed in 6.48s`.

(Disclosure: drafted and verified with an AI coding agent; reviewed by me.)

mnriem (changes requested):

mnriem:

mnriem:

mnriem (approved):

PR #3186, opened 2026-06-26
feat(cli): honor SPECIFY_INIT_DIR in the specify CLI project resolver

PascalThuet:
## Description

Follow-up to #2892, per the go-ahead on #2834.

`SPECIFY_INIT_DIR` only worked in the shell scripts; the Python CLI ignored it, so you still had to `cd` into the member project to run `specify integration` / `extension` / `workflow`. This makes the CLI's project resolver honor it too, with the same validation rules as the shell: relative to cwd, must exist and contain `.specify/`, hard error, no fallback. `init` is unchanged (it creates `.specify/`).

No `--project` flag, per your note that the env var is the exception, not the rule.

## Testing

- [x] Tested locally with `specify --help`
- [x] Ran the full test suite (4555 passed, 66 skipped)
- [x] Tested with a sample project

New tests in `tests/test_init_dir_cli.py` cover targeting from outside the project, path normalization, and the no-fallback error cases.

## AI Disclosure

AI agent used to audit the call sites, draft the change, and review the diff; reviewed by me.


PascalThuet:
One design point worth a call before merge: how `SPECIFY_INIT_DIR` should treat a symlinked `.specify/` at the target. The surfaces differ today, and the split is deliberate but worth confirming.

- `integration` / `extension` / `workflow` follow a symlinked `.specify` (matching the shell resolver, which uses `cd && pwd` + `[ -d .specify ]` and follows symlinks).
- `bundle` and `workflow run <file>` refuse it, since they build/traverse and already enforce symlink confinement on the cwd path.

The tension: uniform behavior and shell parity can't both hold. Making the first group refuse symlinks would diverge from the shell contract #2892 established; making the second group follow them would weaken the bundler's write confinement. So I kept each surface aligned with its existing cwd-path stance and documented the split in `core.md`.

If you'd rather have one rule, I can make it uniformly strict (refuse a symlinked `.specify` everywhere in the CLI, accepting that the CLI becomes stricter than the shell). Which way do you want it?

mnriem:
Let's go with what you have — keep the split. I want to reframe *why*, though, because "shell parity" undersells it.

The rule I'd like documented isn't "match the shell," it's:

**`SPECIFY_INIT_DIR` changes *where* the project is, not *how* a surface treats symlinks.**

Each surface's override path should mirror its existing cwd-path stance, which is exactly what you've done:

- `bundle` / `workflow run <file>` already refuse a symlinked `.specify` on the cwd path because they traverse and write — refusing it under the override too keeps write confinement intact.
- `integration` / `extension` / `workflow` already follow it on the cwd path (read/config only, no escape risk) — so they follow it under the override too.

So the difference between the two groups isn't an inconsistency to apologize for; it's the *same* per-surface symlink policy we already ship, just relocated. Going uniformly strict would either make a surface's override path stricter than its own cwd path (a new inconsistency), or change the cwd behavior of the read surfaces — which is scope creep and would break a legitimate pattern (a shared `.specify` symlinked into several member projects, which our monorepo guide makes plausible). The security upside on read-only surfaces is negligible, so I don't think the trade is worth it.

One small ask before merge: please reword the `core.md` "Symlinked project roots" note to lead with the invariant above rather than the shell-parity framing — something like "the override relocates the project root; each surface keeps its existing symlink stance (write surfaces refuse a symlinked `.specify`, read surfaces follow it)." With that wording tweak this is good to merge.

PascalThuet:
Done. Reworded the `core.md` note to lead with the invariant (the override relocates the project root; each surface keeps its cwd-path symlink stance). Pushed in `9b2ba6f`.

mnriem (changes requested):

PascalThuet:
Addressed both Copilot notes in `80316e6` (docstrings only): reworded the `_project.py` error-message line to "mirror" the shell wording rather than "match" it (the CLI renders a Rich `Error:` line, the shell a plain `ERROR:`), and documented that `find_project_root` can raise `typer.Exit` / `BundlerError` under the `SPECIFY_INIT_DIR` override when `start` is None, so the `Path | None` signature isn't surprising to direct callers.

mnriem:

mnriem:
Please address Copilot feedback. If not applicable, please explain why. And revert changes to CHANGELOG.md as it is automatically generated

mnriem:

PascalThuet:
Posted on behalf of @PascalThuet by OpenAI Codex (model: GPT-5).

Addressed the latest Copilot review round in commit eed7ff34850df854f2d913027da09a9dff349d5f, then merged origin/main in 276edcc370d744ada490b9c8b36b2e90578e55dd. SPECIFY_INIT_DIR validation errors now go through err_console so JSON stdout stays clean, _require_specify_project imports the resolver directly from _project, and the missing-project error now mentions SPECIFY_INIT_DIR with the standardized "Spec Kit" wording.

CHANGELOG.md is not modified in this PR; the manual changelog entry was reverted because the file is generated.

Validation: targeted pytest suite passed locally: 349 passed.

mnriem:

PascalThuet:
Posted on behalf of @PascalThuet by OpenAI Codex (model: GPT-5).

Addressed the follow-up Copilot review in commit 55b667cf8bda80cc317319c6bc409f735d3b47c7, then resolved the new conflict with origin/main in merge commit ae5c31774cd514d369f525a99b6ea4b541e61b5d. The workflow command move from main is preserved, SPECIFY_INIT_DIR support for standalone workflow YAML runs is carried into src/specify_cli/workflows/_commands.py, workflow storage guard errors now go through err_console so --json stdout stays clean, and the unused integration-helper re-export was removed.

CHANGELOG.md is not part of the PR diff against origin/main; the only changelog content present is from merging generated mainline history.

Validation: targeted pytest suite passed locally: 609 passed.

mnriem (approved):

mnriem:
Thank you!

PR #3190, opened 2026-06-26
fix: stop check-prerequisites --paths-only from writing feature.json (#3025)

### LINKED ISSUE #3025: [Bug]: get_feature_paths persists .specify/feature.json during read-only path resolution (check-prerequisites.sh --paths-only)

Noor-ul-ain001:
## What

`check-prerequisites --paths-only` (and PowerShell `-PathsOnly`) is documented as pure, read-only path resolution. But when `SPECIFY_FEATURE_DIRECTORY` is set, `get_feature_paths` / `Get-FeaturePathsEnv` called the persist routine and rewrote `.specify/feature.json` — dirtying the working tree and overwriting a pinned feature directory during what should be a no-op.

Fixes #3025.

## How

Opt out of the write at the resolver boundary with an **explicit flag**, rather than a global env back-channel (per maintainer feedback on #3053):

- **bash**: `get_feature_paths` accepts a leading `--no-persist` flag that skips `_persist_feature_json`; `check-prerequisites.sh` passes it in `--paths-only` mode.
- **PowerShell**: `Get-FeaturePathsEnv` gains a `-NoPersist` switch that skips `Save-FeatureJson`; `check-prerequisites.ps1` passes it in `-PathsOnly` mode.

Normal (non-paths-only) invocations are unchanged and still persist the override, so future sessions without the env var keep working.

## Tests

Added regression tests (bash + PowerShell):
- `--paths-only` / `-PathsOnly` leaves a pinned `feature.json` untouched even when `SPECIFY_FEATURE_DIRECTORY` differs from the pinned value, while still honoring the override in the output.
- Normal mode still persists the override (guards against regressing write behavior).

## Notes

This is an alternative to #3053, which originally used a `SPECIFY_NO_PERSIST_FEATURE_JSON` env var; that approach was flagged in review as a leaky global back-channel. This PR uses explicit local flags/parameters at the call site instead.

### DIFF
--- a/scripts/bash/check-prerequisites.sh
+++ b/scripts/bash/check-prerequisites.sh
@@ -78,8 +78,14 @@ done
 SCRIPT_DIR="$(CDPATH="" cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/common.sh"
 
-# Get feature paths
-_paths_output=$(get_feature_paths) || { echo "ERROR: Failed to resolve feature paths" >&2; exit 1; }
+# Get feature paths.
+# In --paths-only mode this is pure resolution, so pass --no-persist to opt out
+# of the feature.json write side effect (issue #3025).
+if $PATHS_ONLY; then
+    _paths_output=$(get_feature_paths --no-persist) || { echo "ERROR: Failed to resolve feature paths" >&2; exit 1; }
+else
+    _paths_output=$(get_feature_paths) || { echo "ERROR: Failed to resolve feature paths" >&2; exit 1; }
+fi
 eval "$_paths_output"
 unset _paths_output
 
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -152,6 +152,15 @@ _persist_feature_json() {
 }
 
 get_feature_paths() {
+    # Read-only callers (e.g. check-prerequisites.sh --paths-only) pass
+    # --no-persist so pure path resolution never writes .specify/feature.json,
+    # which would dirty the working tree or overwrite a pinned value (issue #3025).
+    local no_persist=false
+    if [[ "${1:-}" == "--no-persist" ]]; then
+        no_persist=true
+        shift
+    fi
+
     # Split decl/assignment so a SPECIFY_INIT_DIR validation failure in
     # get_repo_root propagates as a hard error instead of being masked by `local`.
     local repo_root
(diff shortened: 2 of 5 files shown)


mnriem (changes requested):

mnriem (changes requested):
Please address Copilot feedback and add the requested test

Noor-ul-ain001:
@mnriem  All feedback has been incorporated. Commit 2e7e49c1 adds the 
requested `test_ps_normal_mode_still_persists_feature_json` test to 
achieve feature parity with bash coverage. All 4 regression tests are 
now in place and comprehensive. Please re-review when available.

Noor-ul-ain001:
@mnriem this is ready for another look. The Copilot feedback has been fully addressed:

- Commit `2e7e49c1` adds the requested `test_ps_normal_mode_still_persists_feature_json` test — the PowerShell counterpart to the bash `normal mode still persists` guard, so both languages now assert that the default path still writes `feature.json` (guarding against `-NoPersist`/`--no-persist` ever being applied unconditionally).
- Copilot's latest re-review reports **0 new comments** and both review threads are resolved.

All four regression tests are in place (bash ×2, PowerShell ×2). Could you dismiss the stale change request and re-review when you have a moment? Thanks!

mnriem (approved):

mnriem:
Thank you!

PR #3196, opened 2026-06-27
fix(scripts): honor explicit -Number 0 in PowerShell create-new-feature (parity with bash)

jawwad-ali:
## Description

`create-new-feature.ps1` could not honor an explicit `-Number 0`. The parameter used `[long]$Number = 0` as **both** the default and the auto-detect sentinel:

```powershell
if ($Number -eq 0) { $Number = (Get-HighestNumberFromSpecs ...) + 1 }
```

So `-Number 0` was indistinguishable from *not supplied* and silently auto-incremented. The bash twin keys off whether the value is **non-empty** (`[ -z "$BRANCH_NUMBER" ]`), so `--number 0` is honored and yields `FEATURE_NUM=000`.

Result — identical input, different output across platforms:

| input | bash | PowerShell (before) |
|-------|------|---------------------|
| `--number 0` | `000-feature` | `001-feature` (auto) |

### Fix

Use `$PSBoundParameters.ContainsKey('Number')` to detect whether `-Number` was actually supplied, so an explicit value (including `0`) is honored and only a missing `-Number` auto-detects — mirroring bash's empty-check. This also aligns the `-Timestamp`+`-Number` warning, which bash emits for `--number 0 --timestamp` (non-empty check) but PowerShell previously skipped.

## Testing

- New regression tests:
  - `TestSequentialBranch::test_explicit_number_zero_is_honored` (bash) — pins the canonical `--number 0` -> `000` behavior.
  - `TestSequentialBranchPowerShell::test_explicit_number_zero_is_honored_matching_bash` — **fails before** this change (PS auto-detected), passes after.
- Verified end-to-end with Windows PowerShell 5.1: `-Number 0` -> `000-zero` (even with `specs/003-*` present), `-Number 5` -> `005-zero`, and no `-Number` still auto-detects (`004-zero` with `specs/003-*`). bash confirmed `--number 0` -> `000-zero`.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI located the default-vs-sentinel ambiguity across the bash/PowerShell twins and drafted the `ContainsKey` fix plus regression tests; I reproduced the differing `FEATURE_NUM` under Windows PowerShell 5.1 and bash, confirmed auto-detect and explicit non-zero paths are unchanged, and reviewed the diff before submitting.

### DIFF
--- a/scripts/powershell/create-new-feature.ps1
+++ b/scripts/powershell/create-new-feature.ps1
@@ -142,8 +142,10 @@ if ($ShortName) {
     $branchSuffix = Get-BranchName -Description $featureDesc
 }
 
-# Warn if -Number and -Timestamp are both specified
-if ($Timestamp -and $Number -ne 0) {
+# Warn if -Number and -Timestamp are both specified. Use ContainsKey (not
+# `-ne 0`) so an explicit `-Number 0` is also detected, matching the bash twin's
+# `[ -n "$BRANCH_NUMBER" ]` check.
+if ($Timestamp -and $PSBoundParameters.ContainsKey('Number')) {
     Write-Warning "[specify] Warning: -Number is ignored when -Timestamp is used"
     $Number = 0
 }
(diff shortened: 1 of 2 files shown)


jawwad-ali:
Note for reviewers: #2036 (open since March, currently conflicting + changes-requested) proposes a broader change to *validate* `--number` against existing branches/specs. This PR is intentionally orthogonal and minimal — it only fixes the PowerShell↔bash divergence where an explicit `-Number 0` is silently ignored. Happy to coordinate/rebase if #2036 is revived, but the parity fix here stands on its own.

mnriem:
Thank you!

PR #3197, opened 2026-06-27
fix(workflows): make expression operator/literal parsing quote-aware

jawwad-ali:
## Description

The workflow expression evaluator (`_evaluate_simple_expression`) split on operator keywords with naive `str.find`/`split`, so a keyword **inside a quoted operand** was mistaken for an operator. Two concrete bugs:

1. **Keyword inside a quoted string** — `{{ inputs.mode == 'read and write' }}` split on the inner ` and ` and evaluated as `(inputs.mode == 'read) and (write')` -> wrong result.
2. **Greedy literal short-circuit** — `'a' == 'b'` matched `startswith("'") and endswith("'")` and was stripped to the garbage truthy string `a' == 'b`. So `{{ 'done' == 'failed' }}` evaluated **truthy** and gated the wrong branch in `if`/`switch`/`while`/fan-out steps.

The docstring already lists comparisons, `and`/`or`/`not`, `in`/`not in`, and string literals as supported, so silent mis-parsing of common human-readable strings ("approve or reject", "read and write", "in progress") contradicts the contract.

### Fix

- Add a quote/bracket-aware `_find_top_level(text, token)` helper — a direct sibling of the existing `_split_top_level_commas` (added in #3134) — and use it for the ` or `/` and `/comparison/` in `/` not in ` splits, so operators inside quoted operands are skipped.
- Tighten the literal short-circuit to fire only when the opening quote's matching close is the **final** character (otherwise fall through to operator parsing).

Pipe-filter handling is untouched (it runs first and already worked for single filters).

## Testing

- `uvx ruff check` clean.
- New `TestExpressions::test_operator_splitting_is_quote_aware` — fails before / passes after; covers keyword-in-literal (both operand sides), membership, literal-vs-literal equality, single-literal preservation, plus regression guards (plain comparisons, real `and`/`or`, `| default('a and b')`).
- Full `tests/test_workflows.py`: **305 passed**, 1 skipped. (5 pre-existing `symlink` tests fail only on Windows due to the `SeCreateSymbolicLink` privilege — identical on clean `main` with this change stashed; they pass on CI Linux.)
- Verified each case against the real module on `main` (b7e67f5).

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI located the quote-naive splitting + greedy literal short-circuit, mirrored the existing `_split_top_level_commas` pattern, and drafted the regression tests; I reproduced every mis-parse against the real evaluator, confirmed the regression guards (filter args, real boolean operators, list literals) still pass, and reviewed the diff before submitting.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -180,6 +180,35 @@ def _split_top_level_commas(text: str) -> list[str]:
     return parts
 
 
+def _find_top_level(text: str, token: str) -> int:
+    """Return the index of the first occurrence of *token* in *text* that lies
+    outside any quoted string or nested bracket, or ``-1`` if there is none.
+
+    Used so operator/keyword splitting (``and``/``or``/``in``/comparisons) does
+    not match a separator that appears *inside* a quoted operand -- e.g. the
+    ``and`` in ``mode == 'read and write'`` or the ``or`` in ``'approve or reject'``.
+    """
+    quote: str | None = None
+    depth = 0
+    i = 0
+    n = len(text)
+    while i < n:
+        ch = text[i]
+        if quote is not None:
+            if ch == quote:
+                quote = None
+        elif ch in ("'", '"'):
+            quote = ch
+        elif ch in "([{":
+            depth += 1
+        elif ch in ")]}":
+            depth = max(0, depth - 1)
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3198, opened 2026-06-27
fix(scripts): emit 'Copied plan template' status in setup-plan.ps1 (parity with bash)

jawwad-ali:
## Description

`setup-plan.sh` prints a status message after copying the plan template on a fresh run:

```bash
cp "$TEMPLATE" "$IMPL_PLAN"
if $JSON_MODE; then echo "Copied plan template to $IMPL_PLAN" >&2
else echo "Copied plan template to $IMPL_PLAN"; fi
```

The PowerShell twin `setup-plan.ps1` emitted **nothing** on the successful-copy path — only the `Plan already exists` skip message and the `Plan template not found` warning had equivalents. So on a first run the two scripts had a divergent status-output contract (the bash behavior is even covered by a test, `test_setup_plan_json_parseable_on_first_run`, which asserts `Copied plan template` on stderr).

### Fix

Emit the same message after `WriteAllText`, routed exactly like the sibling `Plan already exists` message — `[Console]::Error.WriteLine` in `-Json` mode so stdout stays pure JSON, `Write-Output` in text mode. Mirrors the bash wording and stream routing. One source file.

## Testing

- New `test_ps_setup_plan_copied_message_on_stderr_in_json_mode` mirrors the existing bash `test_setup_plan_json_parseable_on_first_run`: asserts `-Json` first-run stdout is parseable JSON **and** `Copied plan template` is on stderr.
- Verified fail-before / pass-after locally with Windows PowerShell 5.1: before, stderr is empty and the test fails; after, the message is on stderr and stdout remains pure JSON (`{...,"BRANCH":""}`). Existing setup-plan tests still pass.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI located the missing status message across the bash/PowerShell twins and drafted the fix plus a mirroring regression test; I confirmed the bash wording/stream routing, verified fail-before/pass-after under Windows PowerShell 5.1 (stdout stays pure JSON), and reviewed the diff before submitting.

### DIFF
--- a/scripts/powershell/setup-plan.ps1
+++ b/scripts/powershell/setup-plan.ps1
@@ -40,6 +40,13 @@ if (Test-Path $paths.IMPL_PLAN -PathType Leaf) {
         $content = [System.IO.File]::ReadAllText($template)
         $utf8NoBom = New-Object System.Text.UTF8Encoding($false)
         [System.IO.File]::WriteAllText($paths.IMPL_PLAN, $content, $utf8NoBom)
+        # Emit the copy status like the bash twin (setup-plan.sh); route to stderr
+        # in -Json mode so stdout stays pure JSON, matching the sibling messages.
+        if ($Json) {
+            [Console]::Error.WriteLine("Copied plan template to $($paths.IMPL_PLAN)")
+        } else {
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3210, opened 2026-06-29
fix(catalogs): reject host-less catalog URLs in base and preset validators

### LINKED ISSUE #3209: Host-less catalog URLs (e.g. https://:8080) pass URL validation in base and preset validators
## Summary

Two catalog URL validators reject host-less URLs incorrectly: they check `parsed.netloc` instead of `parsed.hostname`. `netloc` is truthy for URLs like `https://:8080` (port only) or `https://user@` (userinfo only), which have no actual host. So these slip past validation even though the error message promises "a valid URL with a host".

The workflow, step, and bundler validators already use `parsed.hostname` (which is `None` for these inputs), so this is an inconsistency between sibling validators - the base/preset ones are the stragglers.

## Affected code

- `src/specify_cli/catalogs.py` `CatalogStackBase._validate_catalog_url` (the shared base; `IntegrationCatalog` inherits it)
- `src/specify_cli/presets/__init__.py` `PresetCatalog._validate_catalog_url`

## Reproduction

```python
from urllib.parse import urlparse
p = urlparse("https://:8080")
print(bool(p.netloc), p.hostname)   # True None  <- netloc truthy, no host
```

With the current code, `_validate_catalog_url("https://:8080")` passes (no error raised), then a later fetch fails with a confusing message instead of being rejected up front.

## Correct behavior

Already used elsewhere in the same codebase - `workflows/catalog.py`, the step catalog, and `bundler/commands_impl/catalog_config.py` all check `parsed.hostname`. `bundler` even has a comment explaining why `.hostname` is the right field. The fix aligns the two stragglers with that.

---

Quratulain-bilal:
closes #3209

## what

`CatalogStackBase._validate_catalog_url` (`src/specify_cli/catalogs.py`) and `PresetCatalog._validate_catalog_url` (`src/specify_cli/presets/__init__.py`) checked `parsed.netloc` to enforce "a valid URL with a host". but `netloc` is truthy for host-less URLs:

```python
urlparse("https://:8080").netloc   # ":8080"  (truthy)
urlparse("https://:8080").hostname # None
urlparse("https://user@").netloc   # "user@"  (truthy)
urlparse("https://user@").hostname # None
```

so those URLs passed validation despite having no host, contradicting the error message. switched both to `parsed.hostname`, which is `None` in exactly these cases.

## why this is the right field

the other catalog validators already do this:

- `src/specify_cli/workflows/catalog.py` (all 4 copies) check `parsed.hostname`
- `src/specify_cli/bundler/commands_impl/catalog_config.py` checks `parsed.hostname` and has a comment explaining credentials/ports/IPv6 are why `.hostname` is correct
- `src/specify_cli/presets/_commands.py` (the URL-install path) also validates on hostname

this PR aligns the two stragglers with that established pattern; no behavior change for valid URLs.

## tests

added parametrized regression tests for port-only (`https://:8080`, `https://:0`) and userinfo-only (`https://user@`, `https://user:pw@`) URLs in `tests/integrations/test_integration_catalog.py` and `tests/test_presets.py`. verified they fail on the pre-fix code (8 failures) and pass after. full preset/integration/workflow/extension suites: 1058 passed, 1 skipped.

---

### DIFF
--- a/src/specify_cli/catalogs.py
+++ b/src/specify_cli/catalogs.py
@@ -78,7 +78,10 @@ class CatalogStackBase:
                 f"Catalog URL must use HTTPS (got {parsed.scheme}://). "
                 "HTTP is only allowed for localhost."
             )
-        if not parsed.netloc:
+        # Check hostname, not netloc: netloc is truthy for host-less URLs like
+        # "https://:8080" or "https://user@", so the host guarantee this error
+        # promises would not actually hold. hostname is None in those cases.
+        if not parsed.hostname:
             raise cls._error("Catalog URL must be a valid URL with a host.")
 
     def _load_catalog_config(self, config_path: Path) -> list[CatalogEntry] | None:
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -1861,7 +1861,10 @@ class PresetCatalog:
                 f"Catalog URL must use HTTPS (got {parsed.scheme}://). "
                 "HTTP is only allowed for localhost."
             )
-        if not parsed.netloc:
+        # Check hostname, not netloc: netloc is truthy for host-less URLs like
+        # "https://:8080" or "https://user@", so the host guarantee this error
+        # promises would not actually hold. hostname is None in those cases.
+        if not parsed.hostname:
             raise PresetValidationError(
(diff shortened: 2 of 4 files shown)


mnriem:
Thank you!

PR #3224, opened 2026-06-29
feat(workflows): honor max_concurrency in fan-out via a bounded thread pool

### LINKED ISSUE #3222: [Feature]: fan-out ignores max_concurrency — items run sequentially
### Summary

The workflow engine's `fan-out` step accepts a `max_concurrency` field (added with the engine in #2158 / #2142) but the engine never honors it. `FanOutStep.execute` records the value in the step output, and `WorkflowEngine._execute_steps` then runs the per-item templates in a plain sequential `for` loop (`src/specify_cli/workflows/engine.py`, the `if step_type == "fan-out":` branch):

```python
for item_idx, item_val in enumerate(result.output["items"]):
    context.item = item_val
    ...
    self._execute_steps([item_step], context, state, registry, step_offset=-1)
```

So a workflow that sets `max_concurrency: 4` still runs its items one at a time. A fan-out item typically dispatches a `command` step that shells out to an agent CLI — a blocking subprocess that releases the GIL — so there is real wall-clock parallelism available that the engine currently leaves on the table.

### Proposed direction

Honor `max_concurrency` by dispatching items on a bounded `concurrent.futures.ThreadPoolExecutor` when it is `> 1`, and taking the existing sequential path when it is `<= 1` (the default), so existing workflows are unchanged. Keep it opt-in, default-sequential, and order-preserving:

- results assembled in **item order** regardless of completion order (fan-in reads them positionally, so the join is unaffected),
- the `parentId:templateId:index` per-item id grammar unchanged,
- halt-on-first-failure preserved, and
- `state.json` persistence made concurrency-safe (it currently serializes the live `step_results` dict via a plain `open("w")`, which a concurrent fan-out can both interleave on disk and mutate mid-serialization).

I have a small PR ready.

### AI disclosure (per CONTRIBUTING.md)

This issue and the accompanying PR were authored with AI assistance (Claude Code), from a fan-out concurrency investigation; the change was verified by running the repo's test suite and ruff locally.

cc @mnriem — would appreciate your thoughts on the approach when you have a moment.

doquanghuy:
## Description

Closes #3222.

The `fan-out` step has carried a `max_concurrency` field since the workflow engine landed (#2158), but the engine ignored it: `_execute_steps` ran fan-out items in a sequential `for` loop and `max_concurrency` was only recorded in the step output. This honors it.

A new `WorkflowEngine._run_fan_out` runs items on a bounded `ThreadPoolExecutor` when `max_concurrency > 1`, and takes the existing sequential path when `<= 1` (the default) — so existing workflows are byte-for-byte unchanged. Results are always assembled in **item order** (a preallocated slot per item, collected in submission order), never completion order, so fan-in — which reads them positionally — is unaffected. The `parentId:templateId:index` id grammar and halt-on-first-failure are preserved; `max_concurrency` is coerced with `int()`: a value that cannot be coerced (`None`, a non-numeric string) or that coerces to `<= 1` runs sequentially, while a numeric string like `"4"` or a float like `4.0` is honored.

Fan-out items are I/O-bound — each typically dispatches a `command` step that spawns a blocking agent-CLI subprocess, which releases the GIL — so a thread pool yields real wall-clock parallelism.

Two concurrency care points:

- **Per-item context isolation** — each item runs against its own `dataclasses.replace(context, item=…)`, so `context.item` is never clobbered across threads; the shared `steps` dict is written only on the disjoint `parent:template:index` key.
- **State persistence** — `RunState.save()` previously serialized the live `step_results` dict via a plain `open("w")`, so a concurrent fan-out could both interleave on-disk writes and mutate the dict mid-`json.dump` (`dictionary changed size during iteration`). `save()` is now held under a per-run lock and written atomically (temp file + `os.replace`), and per-item result recording goes through a small `record_step_result` helper under that lock. Sequential runs see only an uncontended lock.

A genuine exception escaping an item (as opposed to a normal step `FAILED`, which sets the run status) cancels outstanding work and re-raises, so the run is marked failed rather than reporting a vacuous completion.

## Testing

- [x] Ran the workflow suite with `.venv/bin/python -m pytest tests/test_workflows.py` — 325 passed, including 15 new `TestFanOutConcurrency` cases: K≤1 sequential parity, item-order under forced reverse completion (event chain, no sleeps), real parallelism, `max_concurrency` coercion (0 / negative / None / non-int / string), per-thread item isolation, halt-on-failure prefix, and first-exception cancel + re-raise.
- [x] Ran the full suite — green except 3 pre-existing failures in unrelated branch-name tests (`test_timestamp_branches` / git extension) that fail identically on `main`.
- [x] `uvx ruff check src/ tests/test_workflows.py` — clean
- [ ] Tested with a sample project (covered by the unit tests above)

## AI Disclosure

Code, tests, and this description were authored with AI assistance (Claude Code), from a fan-out concurrency investigation; everything was verified by running the repo's test suite and ruff locally.

@mnriem — would appreciate your review when you have a moment. Happy to swap the `save()` atomicity for a narrower lock if you'd prefer a smaller change.


doquanghuy:
Thanks @Copilot — all five points addressed at root cause (latest: d4479ed):

1. **Run-ahead after halt**: replaced the submit-all-up-front loop with a **sliding submission window** (≤ `max_workers` in flight) that stops launching new items once the run is halting.
2. **Prefix could drop the halting item**: halt is now attributed **per item** from each item's own recorded result (replaying the sequential break condition, honoring `continue_on_error`/`aborted`) rather than the shared run status a later concurrent item may have flipped — so the returned prefix includes the actual halting item, matching the sequential path. (An item that fails before recording, e.g. an unknown step type, is attributed too, since every item runs the same template.)
3. **Unlocked output mutation**: the parent fan-out output update now routes through a new `RunState.set_step_output()` under the run lock, so it can't race a concurrent `save()`.
4. **Docstring vs `int()` coercion**: the docstring now accurately describes the behavior — numeric strings/floats are honored; only non-coercible or `<= 1` runs sequentially.
5. **Flaky timing test**: switched to a monotonic clock with a looser threshold while keeping a clear gap vs the serialized baseline.

Coverage added: concurrent halt-includes-halting-item, continue_on_error-does-not-truncate, and unknown-template-type-matches-sequential. Full workflows suite green (328 passed), `ruff check` clean, 30× stress on the concurrent path.

@mnriem could you give this a review when you get a chance? 🙏

doquanghuy:
Thanks @Copilot — second-pass comments addressed at root cause in ce352a3:

1. **`append_log` race (`engine.py:388`)**: `append_log` now serializes its `log_entries` append + `log.jsonl` write under a dedicated `RunState._log_lock`, so concurrent fan-out workers can't interleave/corrupt log lines. It's a separate lock from the state `_lock` (logging shouldn't contend with `save()`), and since `append_log` is never called while `_lock` is held, the two never nest.
2. **`run_item` returns via outer closure (`engine.py:1004`)**: now reads back through the `item_ctx` it executed against (`item_ctx.steps`) instead of the outer `context.steps` — same dict today, but clearer and robust if `StepContext` copying ever stops sharing by reference.
3. **PR description vs `int()` coercion (`engine.py:990`)**: honoring numeric strings/floats is the intended contract (locked by `test_string_max_concurrency_is_honored`), so I updated the **PR description** to describe the real `int()` behavior (the docstring already does).
4. **Flaky timing test (`tests:2075`)**: `test_concurrency_is_real` now proves parallelism **deterministically** with a `threading.Barrier(n)` — sequential execution can't clear it (times out → `BrokenBarrierError`), so there's no wall-clock threshold to tune or flake.
5. **Singleton step instances (`engine.py:1054`)**: documented the **StepBase thread-safety contract** — `STEP_REGISTRY` holds one shared instance per type, so concurrent fan-out invokes `execute()` on the same object; implementations must be stateless/thread-safe (the built-ins already are). I chose documenting the contract over instantiating a fresh step per execution to avoid a broader behavioral change to step construction — happy to switch to fresh-instance-per-exec instead if you'd prefer that, @mnriem.

Full workflows suite green (328 passed), `ruff check` clean, 30× stress on the concurrency tests.

@mnriem would appreciate your review when you have a chance 🙏

doquanghuy:
Thanks @Copilot — both addressed at root cause in 6f03222:

1. **`save()` stamps `updated_at` outside the lock (`engine.py:437`)**: moved the `updated_at = …` assignment inside the `with self._lock:` block, so the timestamp matches the snapshot the thread actually serializes and concurrent savers no longer race on it.
2. **Cancel-semantics overstatement (`engine.py:986`)**: reworded the docstring — on a halt, only items that **haven't started** are cancelled; items **already running** are allowed to finish but their outputs are ignored (`Future.cancel()` can't stop running work, and the `ThreadPoolExecutor` joins on exit).

Full workflows suite green (328 passed), `ruff check` clean.

@mnriem thanks again for merging #3225 — would appreciate your review on this one too when you have a moment 🙏

doquanghuy:
Thanks @Copilot — addressed at root cause in f78db82:

**Concurrent `on_step_start` (`engine.py:1034`)**: the callback (a `console.print` lambda from the CLI) was invoked from worker threads during concurrent fan-out and could interleave/garble output. It's now serialized under a new `WorkflowEngine._callback_lock` for the duration of the call, so concurrent workers can't interleave it (uncontended for sequential runs). This also protects any other non-thread-safe `on_step_start` implementation.

Full workflows suite green (328 passed), `ruff check` clean.

@mnriem would appreciate your review when you have a chance 🙏

doquanghuy:
Thanks @Copilot — addressed at root cause in 73ced6a:

**Worker exception traceback (`engine.py:1104`)**: the concurrent fan-out no longer stashes the exception in `first_exc` to re-raise after the loop. It now re-raises in place with a bare `raise` (after cancelling outstanding futures), preserving the original traceback, and the `first_exc` variable is gone. The `ThreadPoolExecutor` `__exit__` still joins any already-running workers before the exception escapes.

Full workflows suite green (328 passed), `ruff check` clean. `test_first_exception_cancels_and_reraises` still verifies the cancel + re-raise path.

@mnriem would appreciate your review when you have a moment 🙏

doquanghuy:
Thanks @Copilot — all three addressed at root cause in 56caa29:

1. **Redundant unlocked output write (`engine.py:958`)**: removed the `context.steps[step_id]["output"] = …` writes. `context.steps[step_id]` is the *same dict object* that `set_step_output()` updates under the run lock, so the direct unsynchronized mutation was redundant — the lock now covers it on the shared object.
2. **Status clobbered by a later in-flight item (`engine.py:1116`)**: `_run_fan_out` now derives the halting item's run status per item (`item_halt_status`, replacing the bool `item_halted`) and **restores `state.status` to it after the pool joins**, so a later concurrent item (e.g. FAILED) can't overwrite an earlier halting item's status (e.g. PAUSED). Added a test that an earlier PAUSED item wins over a later FAILED one.
3. **Unbounded pool (`engine.py:1039`)**: `workers = min(max_concurrency, len(items))` plus an early return for empty items, so a user-controlled `max_concurrency` can't over-allocate threads.

Full workflows suite green (329 passed), `ruff check` clean, 30× stress on the concurrency tests.

@mnriem would appreciate your review when you have a moment 🙏

mnriem:
Please address Copilot feedback. If not applicable, please explain why. And please resolve conflicts

doquanghuy:
@mnriem done on both counts:

**Conflicts resolved** — merged latest `upstream/main` into the branch (merge `3462666`). The only real conflict was in `tests/test_workflows.py`, where this PR's `TestFanOutConcurrency` and the now-merged #3225 `TestFanInWaitForValidation` were both added at the same spot; kept both. PR is now mergeable.

**Copilot feedback addressed** (`b385796`):
- **`engine.py:808` & `:938` — unlocked `context.steps` writes during resume**: on a resume run `StepContext` is built with `steps=state.step_results` (same dict), so the two direct `context.steps[...] = ...` writes mutated the shared dict outside `RunState._lock` and could race `save()`. Both now route through a new `_record_result` helper that mirrors into `context.steps` only when it's a *distinct* object (a fresh run), and otherwise relies solely on `record_step_result`'s locked write — so no unlocked write to the shared dict.

Full workflows suite green (340 passed, including the merged fan-in tests), `ruff check` clean. The earlier rounds of Copilot feedback are all addressed in the commits above too. CI shows `action_required` — I believe it needs a maintainer to approve the workflow run on this fork PR. 🙏

mnriem (approved):

mnriem:
Thank you!

PR #3225, opened 2026-06-29
fix(workflows): reject a fan-in wait_for that names an unknown step at validation

### LINKED ISSUE #3223: [Bug]: fan-in silently joins empty when wait_for names an unknown step
### Bug Description

`FanInStep` validates only that `wait_for` is a non-empty list; it never checks that the ids exist. At runtime `execute` does `context.steps.get(step_id, {})` for each id (`src/specify_cli/workflows/steps/fan_in/__init__.py`):

```python
results = []
for step_id in wait_for:
    step_data = context.steps.get(step_id, {})
    results.append(step_data.get("output", {}))
...
return StepResult(status=StepStatus.COMPLETED, output=resolved_output)
```

So a `wait_for` that names a step which does not exist — a typo, or a step declared *after* the fan-in (whose results cannot exist when the fan-in runs) — contributes `{}` to the join, and the step reports `COMPLETED`. The wiring error is silent and surfaces, if at all, as a confusing empty downstream result.

### Steps to Reproduce

1. Author a workflow whose `fan-in` waits on an id that is never declared:

```yaml
workflow:
  id: repro
  name: fan-in unknown wait_for
  version: "1.0.0"
steps:
  - id: collect
    type: fan-in
    wait_for: [tasks]   # no step with id 'tasks' exists (typo for a real id)
```

2. `specify workflow run repro`

### Expected Behavior

The workflow fails at validation, naming the offending id — mirroring #2957 (fail loud when a fan-out `items` expression does not resolve to a list). A step that is *declared* but only conditionally executed (inside an un-taken `if`/`switch` branch) is a legitimate empty-join case and must remain valid.

### Actual Behavior

`Status: completed`, the unknown id silently contributes an empty `{}` to `fan_in.results`, no error.

### Specify CLI Version

main @ b7e67f5

### AI Agent

n/a (workflow engine bug, agent-independent)

### Operating System

macOS (not OS-specific)

### Python Version

3.11

---

**AI disclosure (per CONTRIBUTING.md):** this issue and the accompanying PR were authored with AI assistance (Claude Code), from a fan-out/fan-in audit; all code and claims were verified by running the repo's test suite locally.

cc @mnriem — small validation-only fix; would appreciate your review when you have a moment.

doquanghuy:
## Description

Closes #3223. Sibling to #2957 (fail loud when a fan-out `items` expression does not resolve to a list).

A `fan-in` whose `wait_for` names a step id that isn't declared — a typo, or a forward reference to a step declared after the fan-in — currently aggregates an empty result and reports `COMPLETED`, masking the authoring error. `validate_workflow` now flags it: each `wait_for` id must name a step declared at or before the fan-in, checked against the `seen_ids` set `_validate_steps` already maintains.

A step that is declared but only conditionally executed (e.g. inside an `if`/`switch` branch) is still "seen", so a legitimately-empty join stays valid — only genuinely unknown or forward-referenced ids are rejected. This is validation-time, so there is no runtime behavior change for correctly-authored workflows.

## Testing

- [x] Ran the workflow suite with `.venv/bin/python -m pytest tests/test_workflows.py` — 314 passed, including 4 new `TestFanInWaitForValidation` cases: unknown id rejected, declared-earlier id passes, conditionally-declared (if-branch) id passes, forward reference rejected.
- [x] `uvx ruff check src/ tests/test_workflows.py` — clean

## AI Disclosure

Code, tests, and this description were authored with AI assistance (Claude Code), from a fan-out/fan-in audit; everything was verified by running the repo's test suite and ruff locally.

@mnriem — small validation-only change; would appreciate your review when you have a moment.

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -296,6 +296,23 @@ def _validate_steps(
                     f"boolean, got {type(coe).__name__}."
                 )
 
+        # Fan-in: every wait_for id must reference a step declared at or before
+        # this point. An id not yet seen is either a typo (unknown step) or a
+        # forward reference (the target runs after this fan-in, so its results
+        # cannot exist yet) — both are wiring errors that previously surfaced as
+        # a silent empty result + COMPLETED. A step that is declared but only
+        # conditionally executed (e.g. inside an if/switch branch) is still
+        # "seen" here, so a legitimately-empty result at runtime stays valid.
+        if step_type == "fan-in":
+            wait_for = step_config.get("wait_for")
+            if isinstance(wait_for, list):
+                for wid in wait_for:
+                    if isinstance(wid, str) and wid not in seen_ids:
+                        errors.append(
+                            f"Fan-in step {step_id!r}: 'wait_for' references "
+                            f"unknown or not-yet-declared step id {wid!r}."
+                        )
+
         # Recursively validate nested steps
         for nested_key in ("then", "else", "steps"):
             nested = step_config.get(nested_key)
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -307,7 +307,24 @@ def _validate_steps(
             wait_for = step_config.get("wait_for")
             if isinstance(wait_for, list):
                 for wid in wait_for:
-                    if isinstance(wid, str) and wid not in seen_ids:
+                    if not isinstance(wid, str):
+                        # A non-string entry (e.g. YAML `wait_for: [123]`) can
+                        # never match a real step id, so the join is silently
+                        # empty at runtime — surface it as a wiring error.
+                        errors.append(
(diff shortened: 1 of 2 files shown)


doquanghuy:
Thanks @Copilot for the review — both points addressed at root cause in d9c867b:

- **Self-reference**: a fan-in's own id is already in `seen_ids` by the time `wait_for` is checked, so `wait_for: [<self>]` passed validation while still producing a silent empty join at runtime. Now rejected explicitly.
- **Non-string entries**: `wait_for: [123]` was skipped by the `isinstance(str)` guard and silently validated even though it can never match a step id. Now flagged as a wiring error.

Added coverage: `test_self_reference_is_rejected`, `test_non_string_wait_for_entry_is_rejected`.

@mnriem would you mind taking a look when you have a moment? 🙏

mnriem (approved):

mnriem:
Thank you!

PR #3227, opened 2026-06-29
fix: reject host-less catalog URLs in base and preset validators (#3209)

### LINKED ISSUE #3209: Host-less catalog URLs (e.g. https://:8080) pass URL validation in base and preset validators

Noor-ul-ain001:
## What

Two catalog URL validators rejected host-less URLs incorrectly because they checked `parsed.netloc` instead of `parsed.hostname`. `netloc` is truthy for URLs like `https://:8080` (port only) or `https://user@` (userinfo only), which have **no host** — so they slipped past validation even though the error message promises "a valid URL with a host". The bad URL was then accepted and only failed later with a confusing fetch error.

Fixes #3209.

## How

Switch both stragglers to `parsed.hostname` (which is `None` for these inputs), aligning them with the sibling validators that already do this correctly:

- `src/specify_cli/catalogs.py` — `CatalogStackBase._validate_catalog_url` (inherited by `IntegrationCatalog`)
- `src/specify_cli/presets/__init__.py` — `PresetCatalog._validate_catalog_url`

The workflow (`workflows/catalog.py`), step, and bundler (`bundler/commands_impl/catalog_config.py`) validators already check `.hostname` — the bundler even documents *why*. This is purely an alignment fix; HTTPS/localhost behavior is unchanged.

```python
from urllib.parse import urlparse
p = urlparse("https://:8080")
print(bool(p.netloc), p.hostname)   # True None  <- netloc truthy, no host
```

## Tests

Added parametrized regression tests for **both** validators covering port-only (`https://:8080`, `https://:8080/catalog.json`) and userinfo-only (`https://user@`, `https://user:pass@`) URLs:

- `tests/integrations/test_integration_catalog.py::TestCatalogURLValidation::test_hostless_url_rejected`
- `tests/test_presets.py::TestPresetCatalog::test_validate_catalog_url_hostless_rejected`

Verified the new tests **fail on `main`** (host-less URLs wrongly accepted) and **pass with the fix**. Existing HTTPS / HTTP-rejected / localhost-allowed / missing-host tests still pass. `ruff` clean

### DIFF
--- a/src/specify_cli/catalogs.py
+++ b/src/specify_cli/catalogs.py
@@ -78,7 +78,11 @@ class CatalogStackBase:
                 f"Catalog URL must use HTTPS (got {parsed.scheme}://). "
                 "HTTP is only allowed for localhost."
             )
-        if not parsed.netloc:
+        # Use .hostname (not .netloc) so host-less URLs like "https://:8080"
+        # (port only) or "https://user@" (userinfo only) are rejected: netloc
+        # is truthy for those even though there is no host. This matches the
+        # workflow/step/bundler catalog validators (issue #3209).
+        if not parsed.hostname:
             raise cls._error("Catalog URL must be a valid URL with a host.")
 
     def _load_catalog_config(self, config_path: Path) -> list[CatalogEntry] | None:
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -1861,7 +1861,11 @@ class PresetCatalog:
                 f"Catalog URL must use HTTPS (got {parsed.scheme}://). "
                 "HTTP is only allowed for localhost."
             )
-        if not parsed.netloc:
+        # Use .hostname (not .netloc) so host-less URLs like "https://:8080"
+        # (port only) or "https://user@" (userinfo only) are rejected: netloc
+        # is truthy for those even though there is no host. This matches the
+        # workflow/step/bundler catalog validators (issue #3209).
+        if not parsed.hostname:
             raise PresetValidationError(
                 "Catalog URL must be a valid URL with a host."
             )
--- a/tests/integrations/test_integration_catalog.py
+++ b/tests/integrations/test_integration_catalog.py
@@ -67,6 +67,20 @@ class TestCatalogURLValidation:
         with pytest.raises(IntegrationCatalogError, match="valid URL"):
             IntegrationCatalog._validate_catalog_url("https:///no-host")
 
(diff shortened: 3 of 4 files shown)


mnriem (changes requested):
Please resolve conflicts by pulling in upstream/main

Noor-ul-ain001:
@mnriem conflicts resolved — merged `upstream/main` into the branch (commit `3bb6627`) and GitHub now reports the PR as mergeable.

Notes on the resolution:
- `upstream/main` had since landed an equivalent `parsed.hostname` fix, so the production code in `catalogs.py` and `presets/__init__.py` was byte-identical on both sides. I kept main's canonical comment wording and tagged it with #3209.
- In the two test files I unioned the parametrized host-less cases from both sides (`https://:8080`, `https://:8080/catalog.json`, `https://:0`, `https://user@`, `https://user:pass@`) so no coverage is lost, and adopted main's method name/docstring.

Affected tests pass locally (`tests/test_presets.py`, `tests/integrations/test_integration_catalog.py`). Ready for re-review whenever you have a moment — thanks!

mnriem (approved):

mnriem:
Thank you!

PR #3230, opened 2026-06-29
fix(scripts): warn when spec template is missing in create-new-feature.ps1 (parity with bash)

jawwad-ali:
## Description

When no spec template can be resolved, the bash `create-new-feature.sh` warns before falling back to an empty spec:

```bash
else
    echo "Warning: Spec template not found; created empty spec file" >&2
    touch "$SPEC_FILE"
fi
```

The PowerShell twin `create-new-feature.ps1` created the empty file **silently** (no warning), so on Windows a missing or broken template tree gave the user no signal:

```powershell
} else {
    New-Item -ItemType File -Path $specFile -Force | Out-Null
}
```

### Fix

Emit the same warning on stderr before creating the empty spec file, matching the bash wording and stream (keeps stdout/JSON output pure). One source file.

## Testing

- New `TestSequentialBranchPowerShell::test_missing_spec_template_warns_matching_bash`: with no resolvable template, asserts `Spec template not found` on stderr, stdout still parseable JSON, and the empty spec file created. **Fails before** this change (PowerShell was silent), passes after.
- Verified end-to-end with Windows PowerShell 5.1: the warning lands on stderr while stdout stays pure JSON.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI located the missing-template warning divergence across the bash/PowerShell twins and drafted the fix plus a regression test; I verified the bash wording/stream and confirmed the warning lands on stderr with stdout staying pure JSON under Windows PowerShell 5.1, and reviewed the diff before submitting.

### DIFF
--- a/scripts/powershell/create-new-feature.ps1
+++ b/scripts/powershell/create-new-feature.ps1
@@ -211,6 +211,10 @@ if (-not $DryRun) {
             $utf8NoBom = New-Object System.Text.UTF8Encoding($false)
             [System.IO.File]::WriteAllText($specFile, $content, $utf8NoBom)
         } else {
+            # Match the bash twin (create-new-feature.sh): warn on stderr that no
+            # spec template was found before creating an empty spec file, so the
+            # missing-template signal is not silently swallowed on Windows.
+            [Console]::Error.WriteLine("Warning: Spec template not found; created empty spec file")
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3232, opened 2026-06-29
fix(workflows): make pipe-filter detection quote-aware in expression evaluator

jawwad-ali:
## Description

The simple-expression evaluator detected a filter pipe with a naive substring check:

```python
if "|" in expr:
    parts = expr.split("|", 1)
```

So a literal `|` **inside a quoted operand** was mistaken for a filter separator. For example `{{ inputs.x == 'a|b' }}` split into `inputs.x == 'a` and `b'`, then tried to parse `b'` as a filter and raised a spurious `ValueError: unknown filter` instead of comparing the string.

This is the same class of bug as the operator-splitting fix in #3197 (which added the quote/bracket-aware `_find_top_level` helper) — the pipe check was the one branch left using a naive split.

### Fix

Detect the pipe with `_find_top_level(expr, "|")` so only a **top-level** pipe (outside quotes/brackets) is treated as a filter separator. A `|` inside a quoted operand now falls through to normal operator/literal parsing.

## Testing

- `uvx ruff check` clean; full `TestExpressions` passes (31 tests).
- New `test_pipe_detection_is_quote_aware`: `inputs.x == 'a|b'`, membership against a pipe-containing literal, and a single `'a|b|c'` literal all parse correctly — **fails before** (spurious ValueError), passes after. Regression guards confirm real filters still work, including a pipe inside a filter argument (`default('a|b')`).
- Verified each case against the real evaluator on current main.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI identified the lone non-quote-aware pipe split (sibling of #3197) and reused the existing `_find_top_level` helper; I reproduced the spurious ValueError against the real evaluator, confirmed filters (including pipe-in-argument) still work, and reviewed the diff before submitting.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -230,11 +230,13 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
     if expr[:1] in ("'", '"') and expr.find(expr[0], 1) == len(expr) - 1:
         return expr[1:-1]
 
-    # Handle pipe filters
-    if "|" in expr:
-        parts = expr.split("|", 1)
-        value = _evaluate_simple_expression(parts[0].strip(), namespace)
-        filter_expr = parts[1].strip()
+    # Handle pipe filters. Detect the pipe at the top level only, so a literal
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3233, opened 2026-06-29
fix(workflows): gate validate() must not crash on non-string options

jawwad-ali:
## Description

`GateStep.validate()` reports non-string `options` as an error, but then — when `on_reject` is `abort`/`retry` (`abort` is the default) — still runs the reject-choice check:

```python
if on_reject in ("abort", "retry") and isinstance(options, list):
    if not any(o.lower() in reject_choices for o in options):
```

For a non-string option (e.g. `options: [123]`), `o.lower()` raises `AttributeError: 'int' object has no attribute 'lower'`, which **escapes `validate()`** and breaks `validate_workflow`'s documented contract — *"return a list of error messages"*, never raise. A single malformed gate option crashes validation of the whole workflow instead of being reported.

### Fix

Run the option-text check only when every option is a string (the non-string case is already reported by the earlier check). One guard clause.

## Testing

- `uvx ruff check` clean; `TestGateStep` passes (14 tests).
- New `test_validate_non_string_options_does_not_raise`: `options: [123]` (default `on_reject`) and `options: [True], on_reject: retry` now **return** the "must be strings" error instead of raising. **Fails before** (AttributeError), passes after.
- Verified against the real `GateStep` on current main; valid string options still produce the correct reject-choice error.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI traced the uncaught `AttributeError` from `o.lower()` through `validate_workflow`'s never-raise contract; I reproduced the crash against the real `GateStep`, confirmed string options still validate correctly, and reviewed the diff before submitting.

### DIFF
--- a/src/specify_cli/workflows/steps/gate/__init__.py
+++ b/src/specify_cli/workflows/steps/gate/__init__.py
@@ -194,7 +194,14 @@ class GateStep(StepBase):
                 f"Gate step {config.get('id', '?')!r}: 'on_reject' must be "
                 f"'abort', 'skip', or 'retry'."
             )
-        if on_reject in ("abort", "retry") and isinstance(options, list):
+        # Only inspect option text when every option is a string; otherwise the
+        # `o.lower()` below would raise AttributeError on a non-string option
+        # (already reported above) and break validate_workflow's never-raise contract.
+        if (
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3234, opened 2026-06-29
docs(workflows): add the built-in 'init' step type to the Step Types table

jawwad-ali:
## Description

The **Step Types** table in `docs/reference/workflows.md` lists `command`, `prompt`, `shell`, `gate`, `if`, `switch`, `while`, `do-while`, `fan-out`, and `fan-in` — but omits **`init`**, which is a registered built-in step.

`init` is registered in `workflows/__init__.py` (`_register_builtin_steps` → `_register_step(InitStep())`) and documented in `steps/init/__init__.py` as bootstrapping a project (equivalent to `specify init`) from within a workflow. So the reference table was missing a real, usable step type.

### Fix

Add the `init` row to the table. Docs-only.

## Testing

- `markdownlint-cli2 docs/reference/workflows.md` → 0 errors.
- Cross-checked the table against the registry (`_register_builtin_steps`): with `init` added, all ten registered built-ins are now documented.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI cross-referenced the documented Step Types table against the runtime step registry and found `init` missing; I confirmed `InitStep` is registered and its documented purpose, and reviewed the diff before submitting.

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -262,6 +262,7 @@ specify workflow run speckit -i spec="Build a kanban board with drag-and-drop ta
 | `command`    | Invoke a Spec Kit command (e.g., `speckit.plan`) |
 | `prompt`     | Send an arbitrary prompt to the AI coding agent  |
 | `shell`      | Execute a shell command and capture output       |
+| `init`       | Bootstrap a project (like `specify init`)        |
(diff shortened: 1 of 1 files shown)


mnriem:
Thank you!

PR #3235, opened 2026-06-29
fix(bundle): send command errors to stderr so --json stdout stays parseable

jawwad-ali:
## Description

The `bundle` command group documents its output contract as:

> `--json` emits machine-readable data on **stdout**; human logs go to **stderr**/console.

and its `_fail()` helper's docstring says *"Print an actionable error to **stderr** and exit non-zero."* But `_fail` called `console.print(...)`, and the shared `console` (`_console.console = Console(highlight=False)`) writes to **stdout**. Every `bundle` subcommand routes failures through `_fail` (`search`, `info`, `list`, `install`, `update`, `remove`, `validate`, `build`, `init`), so under `--json` an error printed `Error: ...` Rich text onto **stdout**, corrupting the JSON stream a consumer parses.

### Fix

Add a stderr-bound `err_console` to `_console.py` (the module documented as the single source of Rich `Console` instances) and use it in `_fail`. stdout now carries only the JSON payload.

## Testing

- `uvx ruff check` clean; full `tests/contract/test_bundle_cli.py` passes (24 tests; the existing outside-project guidance test still passes).
- New `test_fail_writes_error_to_stderr_not_stdout`: asserts `_fail` writes to stderr and **not** stdout. **Fails before** (message was on stdout), passes after.
- Verified directly: `_fail('boom')` now writes `Error: boom` to stderr with stdout empty.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI traced `_fail` → `console` → stdout against the module's stated `--json`/stderr contract; I reproduced the stdout contamination, confirmed the existing suite still passes, and reviewed the diff before submitting.

### DIFF
--- a/src/specify_cli/_console.py
+++ b/src/specify_cli/_console.py
@@ -34,6 +34,10 @@ TAGLINE = "GitHub Spec Kit - Spec-Driven Development Toolkit"
 
 console = Console(highlight=False)
 
+# Stderr-bound console for error/diagnostic output, so human-facing messages
+# never contaminate stdout (which carries machine-readable ``--json`` payloads).
+err_console = Console(stderr=True, highlight=False)
+
 class StepTracker:
     """Track and render hierarchical steps without emojis, similar to Claude Code tree output.
     Supports live auto-refresh via an attached refresh callback.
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3241, opened 2026-06-29
fix(scripts): route 'Plan template not found' per --json in setup-plan.ps1 (parity with bash)

jawwad-ali:
## Description

When no plan template resolves, the bash `setup-plan.sh` emits:

```bash
if $JSON_MODE; then echo "Warning: Plan template not found" >&2
else echo "Warning: Plan template not found"; fi
```

The PowerShell twin used `Write-Warning "Plan template not found"`, which renders `WARNING: Plan template not found` on PowerShell's **warning stream** — diverging from bash in both **wording** (`WARNING:` vs `Warning:`) and **routing** (warning stream vs stderr/stdout per mode). It is also inconsistent with the sibling "Copied plan template" message in the same block, which #3198 just aligned to the per-`--json` pattern.

### Fix

Route the message the same way as bash and the sibling: `[Console]::Error.WriteLine` under `-Json` (so stdout stays pure JSON), `Write-Output` otherwise, with bash's exact wording. Completes the parity started in #3198.

## Testing

- `uvx ruff check` clean; `tests/test_ps1_encoding.py` green (`.ps1` stays ASCII).
- New `test_ps_setup_plan_template_not_found_warning_matches_bash`: with no resolvable template, `-Json` mode now emits `Warning: Plan template not found` on stderr with stdout still parseable JSON. **Fails before** (it was `WARNING:` on the warning stream), passes after.
- Verified with Windows PowerShell 5.1; existing setup-plan tests still pass.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI spotted the leftover `Write-Warning` divergence beside the `#3198`-aligned sibling; I confirmed the bash wording/stream and verified fail-before/pass-after under Windows PowerShell 5.1, and reviewed the diff before submitting.

### DIFF
--- a/scripts/powershell/setup-plan.ps1
+++ b/scripts/powershell/setup-plan.ps1
@@ -48,7 +48,14 @@ if (Test-Path $paths.IMPL_PLAN -PathType Leaf) {
             Write-Output "Copied plan template to $($paths.IMPL_PLAN)"
         }
     } else {
-        Write-Warning "Plan template not found"
+        # Match the bash twin's wording and stream routing (stderr in -Json so
+        # stdout stays pure JSON, stdout otherwise), consistent with the sibling
+        # "Copied plan template" message above.
+        if ($Json) {
+            [Console]::Error.WriteLine("Warning: Plan template not found")
+        } else {
+            Write-Output "Warning: Plan template not found"
+        }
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3244, opened 2026-06-29
chore: align CI Python matrix with devguide lifecycle + fix bash 3.2 portability

mnriem:
## Summary

Aligns Spec Kit's CI Python matrix with the [Python release lifecycle](https://devguide.python.org/versions/), and fixes the bash 3.2 (macOS) portability issues this surfaced.

**Policy:**
- **Floor stays at `requires-python = ">=3.11"`** — the oldest non-EOL security release. Security-only versions (3.11, 3.12) are supported by claim and fixed reactively, not gated on a wide per-commit matrix. (No change to `pyproject.toml` — upstream already requires 3.11.)
- **Test matrix runs only the bugfix (maintenance) releases** — `3.13` and `3.14` — which are the versions most likely to surface real differences.
- **`ruff` lint job points at the latest interpreter (`3.14`)** so packaging/lint is exercised on the newest release.
- **Added `macos-latest`** to the OS matrix (now ubuntu/windows/macos × 3.13/3.14) so macOS regressions are caught.

## bash 3.2 portability fixes

Adding `macos-latest` surfaced two pre-existing incompatibilities with **bash 3.2**, which macOS still ships as the system `/bin/bash`:

1. **`update-agent-context.sh`** embeds Python heredocs inside `$(...)` command substitution. bash 3.2 mis-parses an apostrophe in a heredoc body nested in `$(...)`, failing with an `unexpected EOF` quote-matching error. Removed the apostrophes from the affected `$()`-nested heredoc body and documented the constraint inline to prevent regressions.
2. **`create-new-feature-branch.sh`** and **`create-new-feature.sh`** used the bash 4+ `${word^^}` uppercase expansion, which errors as a *bad substitution* on bash 3.2 and caused short uppercase acronyms (e.g. `GO`) to be dropped from derived branch names. Replaced with a portable `tr '[:lower:]' '[:upper:]'` pipeline.

## Validation

- Full test suite passes under **bash 3.2.57** (verified locally on macOS by forcing `bash` → `/bin/bash`).
- `shellcheck --severity=error` clean on all changed scripts.
- All `.sh` files parse under bash 3.2 (`bash -n`).

## Notes

- 3.13 enters security-only status in Oct 2026, at which point the matrix rolls to 3.14/3.15.

---
This PR was prepared by GitHub Copilot (model: Claude Opus 4.8) on behalf of @mnriem.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
         with:
-          python-version: "3.13"
+          python-version: "3.14"
 
       - name: Run ruff check
         run: uvx ruff check src/
@@ -30,8 +30,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        python-version: ["3.11", "3.12", "3.13"]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ["3.13", "3.14"]
     steps:
       - name: Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
--- a/extensions/agent-context/scripts/bash/update-agent-context.sh
+++ b/extensions/agent-context/scripts/bash/update-agent-context.sh
@@ -59,6 +59,13 @@ case "$(uname -s 2>/dev/null || true)" in
 esac
 
 # Parse extension config once; emit context files as JSON, followed by marker strings.
+#
+# NOTE (bash 3.2 / macOS portability): the embedded Python heredocs below run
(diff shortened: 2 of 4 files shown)


mnriem:
Addressed both review comments in b247298:

- **`create-new-feature.sh` (non-portable `\b`):** replaced the `\b...\b` grep word-boundary with `grep -qw` (BSD grep treats `\b` as a backspace, so the acronym-preservation branch could silently miss). It now matches its twin `create-new-feature-branch.sh`, and the description is piped via `printf '%s'` instead of `echo`. I also switched the twin's acronym check to `printf '%s'` so both scripts are byte-for-byte identical in that branch.
- **`update-agent-context.sh` (awkward comment):** reworded the self-seeding comment to be clearer while staying apostrophe-free (the bash 3.2 `$()`-nested-heredoc constraint).

Verified under bash 3.2.57 (full bash-script suite green) and `shellcheck --severity=error`.

Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8).

PR #3250, opened 2026-06-30
Fix macOS/BSD Bash 3.2 compatibility issues in shell scripts

aoright:
This resolves 19 test failures when running the test suite on macOS. There were two main issues: 1. Syntax errors in update-agent-context.sh caused by nested double-quoted command substitutions with tr '\' character escaping, causing mismatched single quotes in Bash 3.2. 2. Bad substitutions in feature-naming scripts caused by using Bash 4.0+ uppercase parameter expansion ${word^^} which is unsupported in macOS default Bash 3.2. The fix replaces the expansion with standard tr '[:lower:]' '[:upper:]' and cleans up nested double-quoted assignments.


mnriem (changes requested):
Please address Copilot feedback. If not applicable, please explain why. And please resolve conflicts

Issue #865, opened 2025-10-13
Spec Kit slash commands don't work in Copilot


GC-Scott:
I'm not sure if the issue is with Spec Kit or GitHub Copilot. When I try to load a project up in Copilot, no `/speckit.*` commands are available. They work just fine in Claude Code. I've run `specify init` in each case, choosing Copilot (as well as Claude Code).

I've seen this issue on two different machines and three different projects. I've upgraded to the most recent versions of specify and copilot, but the problem persists.

josecoelho:
I upgraded on `Claude` 

I had to run `specify init --here --force --ai claude`

Then remove the old commands manually.
The `--force` was the magic.

varunr89:
I am assuming you mean Copilot CLI. It does looks like Copilot CLI does not have support for slash commands. You can use speckit with Copilot chat (the 

<img width="739" height="1219" alt="Image" src="https://github.com/user-attachments/assets/1d17f970-8dc2-44a2-aba9-c29e0f1c0bac" />

localden:
@GC-Scott are you opening the project folder when going to Copilot? And as @varunr89 called out - are you using the GitHub Copilot CLI by chance?

GC-Scott:
I've been using the Copilot CLI, since I've never seen the VS Code extension try to modify code as autonomously as Spec Kit enables. But I've verified that I see the slash commands in the VS Code extension. The CLI *does* support slash commands, as hitting `/` brings up a menu of options, but I guess it doesn't support speckit's slash commands.

Thanks for your help!

varunr89:
The issue is with Github Copilot CLI still doesn't support 'custom' slash commands like Claude Code or Codex CLI; i would be surprised if it's not being added soon though.  If you want to use github copilot with spec kit you have to use copilot chat. Honestly I have found Github Copilot CLI to be lacking in features and the chat to be a much better experience. However, I also do prefer CLI tools so mostly use Claude Code.

sh1man999:
not working /speckit
<img width="748" height="1002" alt="Image" src="https://github.com/user-attachments/assets/2ffdad2a-1896-46f8-9dff-21a86e1ce5b4" />

localden:
@sh1man999 please make sure that you are opening the project folder to see the commands.

jeroaranda:
> [@sh1man999](https://github.com/sh1man999) please make sure that you are opening the project folder to see the commands.

What does it mean opening the project folder? should it work with free copilot?

[...]

PR #3256, opened 2026-06-30
feat(copilot): warn before skills default rollout

WOLIKIMCHENG:
## Description

Adds a deprecation warning for Copilot's legacy markdown scaffold when users initialize Copilot without `--skills`.

This keeps the current `.github/agents/*.agent.md` + `.github/prompts/` default behavior intact, while pointing users to opt in early with `--integration-options "--skills"` before skills mode becomes the default in a future release.

Scope of this change:
- keep Copilot markdown mode as the default
- warn when default Copilot setup uses the legacy markdown scaffold
- keep explicit Copilot `--skills` installs warning-free
- add direct setup and CLI init coverage for the warning behavior

## Testing

Focused validation run:
- `.venv/bin/python -m pytest tests/integrations/test_integration_copilot.py -q`
- `.venv/bin/python -m pytest tests/integrations/test_cli.py -q`
- `.venv/bin/python -m pytest tests/integrations/test_integration_subcommand.py -q`
- `uvx ruff check src/specify_cli/integrations/copilot/__init__.py tests/integrations/test_integration_copilot.py`
- `git diff --check`

All focused checks above passed locally.

### DIFF
--- a/src/specify_cli/integrations/copilot/__init__.py
+++ b/src/specify_cli/integrations/copilot/__init__.py
@@ -5,9 +5,10 @@ Copilot has several unique behaviors compared to standard markdown agents:
 - Each command gets a companion ``.prompt.md`` file in ``.github/prompts/``
 - Installs ``.vscode/settings.json`` with prompt file recommendations
 
-When ``--skills`` is passed via ``--integration-options``, Copilot scaffolds
-commands as ``speckit-<name>/SKILL.md`` directories under ``.github/skills/``
-instead.  The two modes are mutually exclusive.
+Copilot now defaults to skills mode, scaffolding commands as
+``speckit-<name>/SKILL.md`` directories under ``.github/skills/``.
+The legacy ``.agent.md`` + ``.prompt.md`` layout remains available for
+explicit non-skills installs.
 """
 
 from __future__ import annotations
@@ -126,8 +127,8 @@ class CopilotIntegration(IntegrationBase):
             IntegrationOption(
                 "--skills",
                 is_flag=True,
-                default=False,
-                help="Scaffold commands as agent skills (speckit-<name>/SKILL.md) instead of .agent.md files",
+                default=True,
+                help="Scaffold commands as agent skills (default for Copilot)",
             ),
         ]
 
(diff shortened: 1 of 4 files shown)


mnriem:
Thanks for tackling this, @WOLIKIMCHENG. Defaulting Copilot to skills mode is a reasonable long-term direction, but I'd like to push back on both the **rollout** and the **justification**.

**1. Drop "Fixes #865" — this PR doesn't fix it.**
Slash commands in Copilot do work; #865 isn't an actual defect in the markdown/`.agent.md` path, and nothing in this change addresses it. Please remove the `Fixes #865` reference entirely (don't auto-close that issue on merge). The default flip needs to stand on its own merits, not on an unrelated, non-reproducing report.

**2. The legacy mode becomes unreachable, not just non-default.**
Flag options in `_parse_integration_options` (`_helpers.py`) can only ever be set to `True` when present — there's no negation, and `IntegrationOption.default` isn't seeded into `parsed_options`. Behavior is driven solely by `parsed_options.get("skills", True)` in `setup()`. After this change there's **no CLI path** that produces the `.agent.md` layout — it survives only because tests call `setup(skills=False)` directly. We don't want to remove that mode.

**3. This needs a deprecation cycle.**
Main is currently at `0.12.x` (latest release `v0.12.1`). Proposed plan:
- **Now (`0.12`/`0.13`):** keep markdown as the default but emit a **deprecation warning** on Copilot init telling users skills mode is becoming the default and how to opt in early. There's precedent in this same file — see the `SPECKIT_ALLOW_ALL_TOOLS` deprecation warning.
- **Flip the default ~2 minor (`Y`) releases later (~`0.14.0`)**, giving users a heads-up window.

So this PR shouldn't merge the flip yet — let's split it into "add deprecation warning now" and "flip default later."

**4. Add an explicit `--agents` opt-out, and preserve existing users on flip.**
When skills becomes the default, users need a way back to the legacy layout — please add `--agents` (negation support in the option parser) selecting the `.github/agents/*.agent.md` + `.github/prompts/` + `.vscode/settings.json` scaffold. And the flip must respect `.specify/init-options.json`: markdown installs don't set `ai_skills: true`, so re-init/upgrade has to read and rewrite/normalize that file for existing `--agents` users so they aren't silently migrated onto skills.

WOLIKIMCHENG:
Thanks for the detailed guidance. That makes sense.

I revised the PR to remove the default flip and keep the current markdown layout as the default. The PR now only adds a deprecation warning for default Copilot legacy markdown setup and points users to `--integration-options "--skills"` as the early opt-in path.

I also removed the `Fixes #865` framing from the PR body. The default flip, explicit `--agents` opt-out, and `.specify/init-options.json` preservation/migration behavior can be handled in a later PR with the rollout window you described.

mnriem (changes requested):

WOLIKIMCHENG:
Addressed the Copilot feedback by changing the skills-mode warning assertion to check only that the legacy markdown deprecation warning is absent. The focused Copilot integration test file passes locally.

mnriem (approved):

PR #3261, opened 2026-06-30
fix(workflows): make pipe-filter split quote/bracket-aware

### LINKED ISSUE #3260: [Bug]: workflow expression evaluator mis-parses a pipe inside a quoted string as a filter
## Summary

In the workflow expression evaluator (`src/specify_cli/workflows/expressions.py`), the pipe-filter split is not quote/bracket-aware. A `|` that appears inside a quoted operand is wrongly treated as a filter separator, so an expression like `{{ inputs.mode == 'read|write' }}` raises `ValueError: unknown filter 'write'` and crashes the whole expression.

## Root cause

`_evaluate_simple_expression` handles pipes with a naive check:

```python
if "|" in expr:
    parts = expr.split("|", 1)
```

The `and`/`or`/`in`/comparison splits all go through `_find_top_level()`, which skips separators inside quotes or nested brackets (its docstring calls this out for `mode == 'read and write'`). The pipe split was the one separator that skipped that protection, so the `|` in a quoted literal is parsed as a filter pipe.

## Reproduction

```python

ctx = StepContext(inputs={"mode": "read|write"})
evaluate_expression("{{ inputs.mode == 'read|write' }}", ctx)
# ValueError: unknown filter 'write': expected one of default ...
```

Expected: `True` (the comparison should apply to the whole string literal).

## Impact

Any workflow gate/`when` condition or template that compares against (or contains) a string literal with a `|` crashes at evaluation. It also breaks `join('|')` where the separator argument is itself a pipe.

## Suggested fix

Split at the first *top-level* pipe via `_find_top_level(expr, "|")`, matching how the other operators are already handled.

---

Quratulain-bilal:
closes #3260

## what

`_evaluate_simple_expression` split on the first `|` with a naive `if "|" in expr` / `expr.split("|", 1)`. unlike the and/or/in/comparison splits (which use `_find_top_level` to ignore separators inside quoted operands or nested brackets), the pipe split had no such guard. so a `|` inside a quoted string was treated as a filter pipe:

```python
evaluate_expression("{{ inputs.mode == 'read|write' }}", ctx)
# before: ValueError: unknown filter 'write'
# after:  True
```

this crashed any workflow gate/`when` condition or template comparing against a string literal containing `|`.

## fix

use `_find_top_level(expr, "|")` so only a top-level pipe starts a filter, mirroring the existing operator handling. one-spot change. as a side effect it also fixes `join('|')` (pipe as the separator argument), which previously mis-split.

## tests

added `test_pipe_splitting_is_quote_aware` in `tests/test_workflows.py`, mirroring the existing `test_operator_splitting_is_quote_aware`. verified it fails on the pre-fix code (`ValueError: unknown filter 'write'`) and passes after. full `tests/test_workflows.py`: 313 passed, 1 skipped.

note: this is separate from #3208/#3228 (which fix the multi-expression `fullmatch` path in `evaluate_expression`); this touches the pipe split in `_evaluate_simple_expression`, a different spot.

---

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -230,11 +230,14 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
     if expr[:1] in ("'", '"') and expr.find(expr[0], 1) == len(expr) - 1:
         return expr[1:-1]
 
-    # Handle pipe filters
-    if "|" in expr:
-        parts = expr.split("|", 1)
-        value = _evaluate_simple_expression(parts[0].strip(), namespace)
-        filter_expr = parts[1].strip()
+    # Handle pipe filters. Split at the first *top-level* ``|`` so a pipe inside
+    # a quoted operand (e.g. the ``|`` in ``mode == 'read|write'``) or nested
+    # brackets is not mistaken for a filter separator. A naive ``"|" in expr``
+    # check would mis-parse such operands and raise a spurious "unknown filter".
+    pipe_idx = _find_top_level(expr, "|")
+    if pipe_idx != -1:
+        value = _evaluate_simple_expression(expr[:pipe_idx].strip(), namespace)
+        filter_expr = expr[pipe_idx + 1:].strip()
 
         # `from_json` is strict: it takes no arguments and tolerates no
         # trailing tokens. Match on the leading filter name and require the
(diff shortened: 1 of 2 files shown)


Quratulain-bilal:
closing as duplicate — this exact fix already landed in #3232 (same root cause, same _find_top_level approach). sorry for the noise.

PR #3263, opened 2026-06-30
fix(integrations): cline hook note collapses onto instruction at EOF

jawwad-ali:
## Description

`ClineIntegration._inject_hook_command_note()` inserts a dot-to-hyphen note before the hook-output instruction. Its regex matches the line terminator with an alternation that includes `$`:

```python
r"(?m)^(\s*)(- For each executable hook, output the following[^\r\n]*)(\r\n|\n|$)"
```

When the instruction is the **final line of a file with no trailing newline**, the third group (`eol`) captures the empty string. The integration then emits:

```python
eol = m.group(3)          # ""  on a trailing-newline-less last line
return indent + NOTE.rstrip("\n") + eol + indent + instruction + eol
```

so the note text and the instruction are concatenated onto a **single line** (e.g. `...replace dots with hyphens.- For each executable hook...`).

The sibling `agy` integration already guards this exact case with `eol = m.group(3) or "\n"` and an explanatory comment — `cline` is the lone outlier.

### Fix

Default `eol` to `"\n"` when empty, mirroring the `agy` twin verbatim (same one-line guard + comment).

## Testing

- New `test_cline_hook_instruction_injection_no_trailing_newline`: feeds the instruction with no trailing newline and asserts the instruction stays on its own line (`"\n- For each executable hook..."`). **Fails before** (note mashed onto the instruction), **passes after**.
- `uvx ruff check` clean; the two cline hook-injection tests pass.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI spotted the `agy`-vs-`cline` parity gap; I confirmed the empty-`eol` collapse, verified the fail-before/pass-after test, and reviewed the diff before submitting.

### DIFF
--- a/src/specify_cli/integrations/cline/__init__.py
+++ b/src/specify_cli/integrations/cline/__init__.py
@@ -96,7 +96,11 @@ class ClineIntegration(MarkdownIntegration):
         def repl(m: re.Match[str]) -> str:
             indent = m.group(1)
             instruction = m.group(2)
-            eol = m.group(3)
+            # ``eol`` is empty when the regex matched via ``$`` because the
+            # instruction was the final line of a file with no trailing
+            # newline. Default to ``\n`` so the note never collapses onto
+            # the same line as the instruction.
+            eol = m.group(3) or "\n"
             return (
                 indent
                 + _HOOK_COMMAND_NOTE.rstrip("\n")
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3266, opened 2026-06-30
fix(integrations): add zed to discovery catalog.json

jawwad-ali:
## Description

`integrations/catalog.json` is the discovery manifest for Spec Kit integrations. The `zed` integration is registered in `INTEGRATION_REGISTRY`, registrar-aligned, and covered by the registry tests — but it is the **only one of the 34 registered integrations missing from the catalog**:

```python
set(INTEGRATION_REGISTRY) - {"generic"} - set(catalog["integrations"])  # -> {'zed'}
```

So `zed` is undiscoverable through the catalog while every sibling (including the other skills-based editors) is listed.

### Fix

- Add the `"zed"` entry to `integrations/catalog.json`, mirroring the sibling skills entries (e.g. `agy`, `lingma`) — `name: "Zed"`, `description: "Zed editor skills-based integration"`, `tags: ["ide", "skills"]`, consistent with the module's own "Zed editor integration — skills-based agent" description.
- Add a `TestCatalogParity` regression test asserting every registered integration (except the `generic` fallback) appears in the catalog, so this can't silently drift again.

## Testing

- New `test_every_registered_integration_is_in_catalog`: **fails before** (`missing: ['zed']`, verified by stashing the catalog change), **passes after**.
- Full `tests/integrations/test_registry.py` passes (434 tests); `uvx ruff check` clean; catalog.json validated as well-formed JSON.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI computed the registry-vs-catalog set difference; I confirmed `zed` was the sole omission, matched the sibling entry shape, proved fail-before via stash, and reviewed the diff before submitting.

### DIFF
--- a/integrations/catalog.json
+++ b/integrations/catalog.json
@@ -299,6 +299,15 @@
       "author": "spec-kit-core",
       "repository": "https://github.com/github/spec-kit",
       "tags": ["cli", "skills", "z-ai"]
+    },
+    "zed": {
+      "id": "zed",
+      "name": "Zed",
+      "version": "1.0.0",
+      "description": "Zed editor skills-based integration",
+      "author": "spec-kit-core",
+      "repository": "https://github.com/github/spec-kit",
+      "tags": ["ide", "skills"]
     }
   }
 }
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3270, opened 2026-06-30
fix(workflows): engine loop cap ignores bool max_iterations

jawwad-ali:
## Description

The workflow engine's while/do-while loop-cap guard:

```python
max_iters = step_config.get("max_iterations")
if not isinstance(max_iters, int) or max_iters < 1:
    max_iters = 10
...
for _loop_iter in range(max_iters - 1):
```

does not fall back to the default for a **boolean** `max_iterations`. Since `isinstance(True, int)` is `True` and `True < 1` is `False`, a `max_iterations: true` slips past the guard and the loop runs `range(True - 1) == range(0)` — capping at a **single iteration** instead of the default 10.

This is the same bool-is-int trap already fixed in the `while`/`do-while` *validators* (#3237); the engine's runtime path had the same gap, and `WorkflowEngine.execute()` does not auto-validate, so this guard is the only line of defence at run time. The function's sibling `continue_on_error is True` handling already special-cases bools.

### Fix

```python
if (
    isinstance(max_iters, bool)
    or not isinstance(max_iters, int)
    or max_iters < 1
):
    max_iters = 10
```

## Testing

- New `test_loop_with_bool_max_iterations_uses_default_cap`: an unvalidated `from_string` workflow with `max_iterations: true` and a condition that stays truthy; asserts the tick counter reaches **10** (default cap). **Fails before** (`'1' == '10'` — capped at one iteration, verified by source-stash), **passes after**.
- `uvx ruff check` clean; the existing `test_while_loop_runs_to_max_when_condition_stays_true` (int max_iterations) stays green.

## AI Disclosure

Found and fixed with Claude Code (Claude Opus 4.8) under my direction. AI flagged the runtime gap left by the validator-only #3237 fix; I confirmed the single-iteration cap, proved fail-before via source-stash, and reviewed the diff before submitting.

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -982,7 +982,16 @@ class WorkflowEngine:
                     from .expressions import evaluate_condition
 
                     max_iters = step_config.get("max_iterations")
-                    if not isinstance(max_iters, int) or max_iters < 1:
+                    # A bool is an int in Python (isinstance(True, int) is True
+                    # and True == 1), so a bool max_iterations would slip past
+                    # the int check and cap the loop at range(0)==1 iteration
+                    # instead of the default. Exclude bools, mirroring the
+                    # while/do-while validators and the continue_on_error guard.
+                    if (
+                        isinstance(max_iters, bool)
+                        or not isinstance(max_iters, int)
+                        or max_iters < 1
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3275, opened 2026-06-30
fix(extensions): restore core-command discovery broken by #3014 module move

### LINKED ISSUE #3274: Bug: extensions core-command discovery is dead — off-by-one paths after the #3014 module move
## Summary

`specify_cli.extensions._load_core_command_names()` never discovers the bundled command templates. Both candidate paths are off by one directory, so the function silently returns the hardcoded `_FALLBACK_CORE_COMMAND_NAMES` on every call — in both wheel installs and source checkouts.

## Root cause

The function resolves its command dirs with bespoke `Path(__file__)` math:

```python
candidate_dirs = [
    Path(__file__).parent / "core_pack" / "commands",
    Path(__file__).resolve().parent.parent.parent / "templates" / "commands",
]
```

These were correct when the code lived at `src/specify_cli/extensions.py` (anti-shadowing guard added in #1994). The refactor that moved it into the package at `src/specify_cli/extensions/__init__.py` (#3014, `826e193` — diff shows `{extensions.py => extensions/__init__.py}`) pushed the file one directory deeper but did not update the parent counts. They now resolve to:

- wheel → `specify_cli/extensions/core_pack/commands` — real path is `specify_cli/core_pack/commands`
- source → `src/templates/commands` — real path is repo-root `templates/commands`

Neither exists, so `commands_dir.is_dir()` is always false and the loop falls through to the fallback set.

This is the same off-by-one class @mnriem identified for the presets loader (`parent.parent.parent` after the #2826 `presets.py` → `presets/__init__.py` move, re #3086), but in the extensions module — which the preset-scoped fix does not touch.

## Reproduction

From a source checkout:

```python
from specify_cli.extensions import _load_core_command_names, _FALLBACK_CORE_COMMAND_NAMES
# Discovery is meant to read repo-root templates/commands. Instead:
assert _load_core_command_names() == _FALLBACK_CORE_COMMAND_NAMES  # passes — discovery is dead
```

Both candidate dirs resolve to non-existent paths (verifiable by printing them).

## Impact (honest assessment)

Currently **latent**, not user-visible: `_FALLBACK_CORE_COMMAND_NAMES` happens to equal the 10 real command stems today, so nothing breaks right now.

The risk is silent drift. `CORE_COMMAND_NAMES` (derived from this function) guards extensions against shadowing core command names (#1994). With dynamic discovery dead, that guard depends entirely on someone hand-editing the fallback whenever a core command is added or removed — which has already happened: `converge` was manually appended to `_FALLBACK_CORE_COMMAND_NAMES` in #3001 (`0c29d89`). A future add/remove that forgets the fallback would silently desync the guard.

## Proposed fix

Delegate path resolution to the canonical `_assets` resolvers (`_locate_core_pack` / `_repo_root`), exactly as `presets/__init__.py` already does. They are anchored to the package root, so discovery is immune to future module moves. Add regression tests that pin live discovery (they fail on the current code and pass after the fix).

---
*Filed by @v-dhruv with assistance from GitHub Copilot (model: Claude Opus 4.8). The investigation and proposed patch are agent-generated and human-directed.*

v-dhruv:
## What

Fix the off-by-one path resolution in `specify_cli.extensions._load_core_command_names()` so it actually discovers the bundled core command templates instead of silently returning the hardcoded fallback set. Fixes #3274.

## Why

When this code moved from `src/specify_cli/extensions.py` to `src/specify_cli/extensions/__init__.py` (#3014, `826e193`), the file went one directory deeper but the bespoke `Path(__file__)` parent counts were not updated. Both candidate dirs now resolve to non-existent paths:

- wheel -> `specify_cli/extensions/core_pack/commands` (real: `specify_cli/core_pack/commands`)
- source -> `src/templates/commands` (real: repo-root `templates/commands`)

So `_load_core_command_names()` always falls through to `_FALLBACK_CORE_COMMAND_NAMES`. It only looks healthy because the fallback currently equals the 10 real command stems. `CORE_COMMAND_NAMES` (built from this function) guards extensions against shadowing core commands (#1994); with discovery dead, that guard depends on the fallback being hand-maintained -- `converge` was manually added to it in #3001 (`0c29d89`).

Same off-by-one class @mnriem diagnosed for the presets loader (re #3086, #2826), but in a module the preset-scoped fix does not touch.

## How

- Delegate path resolution to the canonical `_assets` resolvers (`_locate_core_pack` / `_repo_root`), exactly as `presets/__init__.py` does. These are anchored to the package root, so discovery survives future module moves.
- Add `tests/test_extensions.py::TestCoreCommandDiscovery` covering the source, wheel, and fallback branches, plus a sentinel test proving discovery returns the real bundled set rather than the fallback.

## Verification

- New regression tests fail on the pre-fix code (the sentinel test fails with a clean assertion -- it returns the fallback sentinel instead of the bundled set) and pass after the fix.
- `tests/test_extensions.py`: 328 passed, 3 skipped.
- Broader relevant suites (imports, extensions, presets, integration registry/manifest): 887 passed. The only failures are 6 pre-existing Windows symlink-privilege errors in `test_manifest.py` that fail identically on the unmodified base (unrelated to this change).

## Behavior change

None for end users today (the fallback already matches the real commands). This restores the intended dynamic discovery and removes the silent manual-maintenance dependency on the fallback list.

---
*Authored with assistance from GitHub Copilot (model: Claude Opus 4.8), acting autonomously under the direction of @v-dhruv. Commits carry `Assisted-by:` trailers and the change is agent-generated; please review accordingly.*

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -26,6 +26,7 @@ import yaml
 from packaging import version as pkg_version
 from packaging.specifiers import InvalidSpecifier, SpecifierSet
 
+from .._assets import _locate_core_pack, _repo_root
 from .._init_options import is_ai_skills_enabled
 from .._invocation_style import is_dollar_skills_agent, is_slash_skills_agent
 from .._utils import dump_frontmatter, relative_extension_path_violation, version_satisfies
@@ -62,11 +63,21 @@ def _load_core_command_names() -> frozenset[str]:
     Prefer the wheel-time ``core_pack`` bundle when present, and fall back to
     the source checkout when running from the repository. If neither is
     available, use the baked-in fallback set so validation still works.
+
+    Path resolution is delegated to the canonical ``_assets`` helpers
+    (``_locate_core_pack`` / ``_repo_root``), which are anchored to the package
+    root rather than this module's location. The previous bespoke
+    ``Path(__file__)`` math was correct only while this code lived at
+    ``specify_cli/extensions.py``; once it moved into the
+    ``specify_cli/extensions/`` package the relative parents fell one level
+    short, so discovery always missed the real command dirs and silently fell
+    through to ``_FALLBACK_CORE_COMMAND_NAMES`` below.
     """
-    candidate_dirs = [
-        Path(__file__).parent / "core_pack" / "commands",
-        Path(__file__).resolve().parent.parent.parent / "templates" / "commands",
-    ]
+    candidate_dirs: list[Path] = []
+    core_pack = _locate_core_pack()
+    if core_pack is not None:
+        candidate_dirs.append(core_pack / "commands")
+    candidate_dirs.append(_repo_root() / "templates" / "commands")
 
     for commands_dir in candidate_dirs:
         if not commands_dir.is_dir():
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -205,6 +205,108 @@ class TestNormalizePriority:
         assert normalize_priority(True, default=5) == 5
 
 
+# ===== Core Command Discovery Tests =====
+
+
+class TestCoreCommandDiscovery:
+    """Regression tests for dynamic core-command discovery.
+
+    ``_load_core_command_names()`` must read the bundled command templates via
+    the canonical ``_assets`` resolvers. A prior off-by-one in this module's
+    bespoke ``Path(__file__)`` math (introduced when the code moved from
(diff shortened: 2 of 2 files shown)


dhruv-15-03:
Superseded by #3292, which contains the identical commit opened from the correct author account (@dhruv-15-03). Closing this one to avoid a duplicate. Apologies for the noise.

v-dhruv:
Superseded by #3292, which contains the identical commit opened from the correct author account (@dhruv-15-03). Closing this duplicate. Apologies for the noise.

Issue #3283, opened 2026-06-30
Add `py:` lines to command templates' `scripts:` frontmatter


mnriem:
Part of #3277. Depends on the core + extension ports.

## Tasks

- Add a `py:` line to the `scripts:` block in every `templates/commands/*.md` (and extension templates).
- Verify `{SCRIPT}` substitution renders correct Python invocations for the `py` script type.

## Acceptance

- Rendering a command with `--script py` produces a valid Python invocation string.

marcelsafin:
I'd like to take this one, as a follow-up to the ports in #3386/#3387/#3400. Plan: add `py:` lines to the `scripts:` frontmatter in all command templates (core + extensions), pointing at the Python twins, and extend the rendering tests to assert `--script py` produces valid invocation strings.

Issue #3278, opened 2026-06-30
Add `py` script type & interpreter resolution


mnriem:
Part of #3277 (foundation step).

**Why:** Everything else builds on the CLI knowing about a Python variant and being able to invoke it portably.

## Tasks

- Add `"py": "Python"` to `SCRIPT_TYPE_CHOICES` in `src/specify_cli/_agent_config.py`.
- Verify `process_template()` (`src/specify_cli/integrations/base.py`) extracts `py:` from frontmatter via the existing generic `script_type` logic (should need no change — confirm).
- Update `install_scripts()` to copy `.py` files (and handle the `.sh`-only chmod path appropriately).
- Add an interpreter-resolution helper (prefer project venv → `python3` → `python`) for the `{SCRIPT}` invocation string.
- Update `VALID_SCRIPT_TYPES` consumers: `src/specify_cli/workflows/steps/init/`, `src/specify_cli/integrations/_helpers.py`, `src/specify_cli/commands/init.py`.

## Acceptance

- `specify init x --script py` selects the Python variant without error.
- Unit tests cover the new choice.

dhruv-15-03:
Hi @mnriem — I'd like to help move this epic forward, starting with this foundation issue. I've read through #3277 / #3278 / #3279 and the relevant code.

My read of #3278's scope:

- Add `"py": "Python"` to `SCRIPT_TYPE_CHOICES` in `_agent_config.py` — this propagates automatically to the validators in `commands/init.py`, `integrations/_helpers.py`, and `VALID_SCRIPT_TYPES` in `workflows/steps/init/__init__.py`.
- Extend `install_scripts()` to copy `.py` files, keeping the `chmod +x` scoped to the `.sh` path.
- Add an interpreter-resolution helper (project venv → `python3` → `python`) for the `{SCRIPT}` invocation string.
- Confirm `process_template()` already extracts `py:` frontmatter via the generic `script_type` logic (looks like it should need no change — I'll verify).
- Unit tests for the new choice + interpreter resolution.

Before I open a PR: since #3278 and the #3279 PoC are the keystone that sets the conventions for the rest of the epic, would you prefer to own the foundation + PoC yourself and delegate the bulk ports (#3280–#3282) once the pattern lands? I'm happy to take whichever you'd rather hand off — I just don't want to cut across your plan for the pieces that define the pattern. If #3278 is open for me, I'll pick it up.

[...]

Issue #3277, opened 2026-06-30
Unify workflow scripts under a single Python implementation per command


mnriem:
## Goal

Replace the paired bash (`.sh`) + PowerShell (`.ps1`) scripts with **one Python script per occurrence**, eliminating dual-maintenance and the `jq`/manual-JSON divergence. The `specify` CLI is already Python, so the interpreter is guaranteed present — no new runtime dependency.

## Scope

~4,500 lines across ~10 script pairs:

- **Core** `scripts/{bash,powershell}/` — `common`, `check-prerequisites`, `create-new-feature`, `setup-plan`, `setup-tasks`
- **`extensions/agent-context/scripts/`** — `update-agent-context`
- **`extensions/git/scripts/`** — `git-common`, `initialize-repo`, `auto-commit`, `create-new-feature-branch`

## Strategy

Incremental. Introduce `py` as a **third** script type alongside `sh`/`ps`, port one script end-to-end as a proof of concept with output-parity tests, then migrate the rest. Only after parity is proven do we consider deprecating `sh`/`ps`.

## Contract to preserve

Agents parse script stdout (`FEATURE_DIR:…`, `AVAILABLE_DOCS:…`, JSON shapes, and the `--json` / `--paths-only` / `--require-tasks` / `--include-tasks` flags). **Behavioral parity is the acceptance bar for every port.**

## Sub-issues

- [x] #3278 — Add `py` script type & interpreter resolution _(foundation)_
- [x] #3279 — PoC: port `check-prerequisites` to Python + parity tests _(depends on #3278)_
- [x] #3280 — Port core `common` + `create-new-feature` + `setup-plan` + `setup-tasks` _(depends on #3279)_
- [x] #3281 — Port `agent-context` extension scripts _(depends on #3279)_
- [x] #3282 — Port `git` extension scripts _(depends on #3279)_
- [x] #3283 — Update templates' `scripts:` frontmatter to add `py:` lines _(depends on #3280–#3282)_
- [x] #3284 — Docs, AGENTS.md, and migration/deprecation plan _(depends on #3283)_

## Out of scope (for now)

Removing `.sh`/`.ps1` files. They stay until `py` reaches full parity and adoption.

mnriem:
All seven sub-issues (#3278–#3284) are closed and every in-scope deliverable is merged to `main`:

- `py` script type + interpreter resolution in the CLI (#3278)
[...]

PR #3285, opened 2026-06-30
feat(cli): add `py` script type & Python interpreter resolution (#3278)

mnriem:
## Summary

Implements #3278 — the foundation step toward unifying workflow scripts under a single Python implementation per command (#3277). Adds a third script variant (`py`) alongside `sh`/`ps`.

## Changes

- **New choice:** Add `"py": "Python"` to `SCRIPT_TYPE_CHOICES` in `_agent_config.py`. All `VALID_SCRIPT_TYPES` consumers (the init workflow step, `commands/init.py`, and `integrations/_helpers.py`) derive from this mapping, so they accept `py` without further change.
- **`process_template()` (`integrations/base.py`):** Confirmed the existing generic `script_type` logic extracts `py:` from frontmatter unchanged. When the script type is `py`, the `{SCRIPT}` expansion is now prefixed with a resolved Python interpreter so `.py` scripts run portably (notably on Windows, where `.py` files aren't directly executable). `sh`/`ps` go down the identical path as before.
- **Interpreter resolution:** New `IntegrationBase.resolve_python_interpreter()` helper — prefers a project venv (returned **relative to the project root**: `.venv/bin/python` or `.venv/Scripts/python.exe`), then `python3` on PATH, then `python` on PATH, then the running interpreter (`sys.executable`), falling back to `python3` only if even that is unavailable. `project_root` is threaded through the `process_template()` callers so the venv preference works. When the resolved interpreter path contains whitespace (e.g. Windows `Program Files`) it is quoted so the `{SCRIPT}` invocation isn't split into multiple args.
- **`install_scripts()`:** Copied `.py` files are now marked executable (previously `.sh`-only; `.ps1` behavior unchanged).

## Out of scope

Adding `py:` lines to the templates' `scripts:` frontmatter is tracked separately in #3283. Until then, `{SCRIPT}` is left literal for `py` (existing graceful fallback when no matching frontmatter line exists).

## Tests

Added positive **and** negative unit tests:
- Interpreter resolution: prefers `python3`, falls back to `python`, falls back to `sys.executable` when PATH resolution fails (and to `python3` only if that is empty), prefers POSIX/Windows venv (relative path), ignores a missing venv.
- Interpreter quoting: a whitespace-containing interpreter path is quoted; a whitespace-free one is not.
- `process_template` with `py`: prefixes interpreter + rewrites path; `sh` is **not** prefixed; venv is used when `project_root` is passed.
- `install_scripts`: `.py`/`.sh` made executable, non-script file is **not**.
- New choice present in `SCRIPT_TYPE_CHOICES` / `VALID_SCRIPT_TYPES`; unknown variant rejected.

## Verification

- `specify init x --script py` completes without error and records `"script": "py"`.
- Full suite: **3860 passed, 4 skipped**. `ruff check` clean on changed files.

---

🤖 This PR was authored autonomously by GitHub Copilot (model: Claude Opus 4.8) on behalf of @mnriem.

### DIFF
--- a/src/specify_cli/_agent_config.py
+++ b/src/specify_cli/_agent_config.py
@@ -17,4 +17,8 @@ AGENT_CONFIG: dict[str, dict[str, Any]] = _build_agent_config()
 
 DEFAULT_INIT_INTEGRATION = "copilot"
 
-SCRIPT_TYPE_CHOICES: dict[str, str] = {"sh": "POSIX Shell (bash/zsh)", "ps": "PowerShell"}
+SCRIPT_TYPE_CHOICES: dict[str, str] = {
+    "sh": "POSIX Shell (bash/zsh)",
+    "ps": "PowerShell",
+    "py": "Python",
+}
--- a/src/specify_cli/integrations/base.py
+++ b/src/specify_cli/integrations/base.py
@@ -626,6 +626,11 @@ class IntegrationBase(ABC):
             # executable on Windows).
             if script_type == "py":
                 interpreter = IntegrationBase.resolve_python_interpreter(project_root)
+                # Quote the interpreter if it contains whitespace (e.g. an
+                # absolute ``sys.executable`` path under Windows
+                # ``Program Files``) so it isn't split into multiple args.
+                if any(ch.isspace() for ch in interpreter):
+                    interpreter = f'"{interpreter}"'
                 script_command = f"{interpreter} {script_command}"
             content = content.replace("{SCRIPT}", script_command)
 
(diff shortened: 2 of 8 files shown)


mnriem:
Addressed both review comments in 24c0e2f:

- **`resolve_python_interpreter()` portability:** now returns the venv interpreter as a path **relative to the project root** (`.venv/bin/python` or `.venv/Scripts/python.exe`) instead of an absolute/joined path, so the generated `{SCRIPT}` invocation stays runnable from the repo root regardless of where the project lives. Updated the corresponding tests to assert the relative path.
- **`install_scripts()` docstring:** corrected to state that both `.sh` and `.py` scripts are made executable.

Full suite: 3860 passed, 4 skipped; `ruff check` clean.

Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8, autonomous).

PR #3287, opened 2026-07-01
fix(extensions): resolve core-command dirs via _assets helpers (#3274)

### LINKED ISSUE #3274: Bug: extensions core-command discovery is dead — off-by-one paths after the #3014 module move

Noor-ul-ain001:
Closes #3274.

## Problem

`specify_cli.extensions._load_core_command_names()` never discovers the bundled command templates. It computed its candidate command dirs with bespoke `Path(__file__)` arithmetic that was correct when the code lived at `src/specify_cli/extensions.py`, but the #3014 move into the package (`extensions.py` → `extensions/__init__.py`) pushed the file one directory deeper without updating the `.parent` counts. Both candidates now resolve to non-existent paths:

| env | computed | real |
| --- | --- | --- |
| wheel | `specify_cli/extensions/core_pack/commands` | `specify_cli/core_pack/commands` |
| source | `src/templates/commands` | repo-root `templates/commands` |

Neither exists, so `commands_dir.is_dir()` is always false and every call silently returns `_FALLBACK_CORE_COMMAND_NAMES`.

## Why it matters

Latent today (the fallback happens to equal the 10 real command stems), but the shadowing guard from #1994 that consumes `CORE_COMMAND_NAMES` now depends entirely on someone hand-editing the fallback whenever a core command is added or removed — which already happened once (`converge` was manually appended in #3001). A future add/remove that forgets the fallback would silently desync the guard.

This is the same off-by-one class @mnriem fixed for the presets loader, but in the extensions module.

## Fix

Delegate path resolution to the canonical `_locate_core_pack` / `_repo_root` resolvers in `_assets` — the same ones the presets and bundle loaders already use. They are anchored to the package root, so discovery survives future module moves. (`_assets` is stdlib-only with zero internal imports, so no cycle risk.)

## Tests

The existing `test_core_command_names_match_bundled_templates` can't catch this — the fallback equals the live tree, so a dead loader and a working one look identical. Added three tests that point the resolvers at a temp tree with *different* command names, proving discovery reads from disk:

- `test_load_core_command_names_discovers_from_source_checkout`
- `test_load_core_command_names_prefers_wheel_core_pack`
- `test_load_core_command_names_falls_back_when_nothing_found`

The first two fail on the pre-fix code (verified by reverting the source change).

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -26,6 +26,7 @@ import yaml
 from packaging import version as pkg_version
 from packaging.specifiers import InvalidSpecifier, SpecifierSet
 
+from .._assets import _locate_core_pack, _repo_root
 from .._init_options import is_ai_skills_enabled
 from .._invocation_style import is_dollar_skills_agent, is_slash_skills_agent
 from .._utils import dump_frontmatter, relative_extension_path_violation
@@ -62,14 +63,28 @@ def _load_core_command_names() -> frozenset[str]:
     Prefer the wheel-time ``core_pack`` bundle when present, and fall back to
     the source checkout when running from the repository. If neither is
     available, use the baked-in fallback set so validation still works.
+
+    Path resolution is delegated to the canonical ``_assets`` resolvers
+    (``_locate_core_pack`` / ``_repo_root``) — the same ones the presets and
+    bundle loaders use — rather than bespoke ``Path(__file__)`` arithmetic.
+    Hand-counted ``.parent`` chains silently broke discovery once already: the
+    #3014 move of this module from ``specify_cli/extensions.py`` to
+    ``specify_cli/extensions/__init__.py`` pushed the file one directory deeper
+    without updating the counts, so both candidates resolved to non-existent
+    paths and every call fell through to the fallback (#3274). The shared
+    resolvers are anchored to the package root, so discovery survives future
+    module moves.
     """
+    core_pack = _locate_core_pack()
     candidate_dirs = [
-        Path(__file__).parent / "core_pack" / "commands",
-        Path(__file__).resolve().parent.parent.parent / "templates" / "commands",
+        # Wheel install: force-include maps templates/commands → core_pack/commands.
+        core_pack / "commands" if core_pack is not None else None,
+        # Source checkout / editable install: repo-root templates/commands.
+        _repo_root() / "templates" / "commands",
     ]
 
     for commands_dir in candidate_dirs:
-        if not commands_dir.is_dir():
+        if commands_dir is None or not commands_dir.is_dir():
             continue
 
         command_names = {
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3312, opened 2026-07-02
fix(scripts): fall through to grep/sed when python3 is a broken stub in feature.json parser

### LINKED ISSUE #3304: Windows/Git Bash: python3 App Execution Alias stub defeats feature.json parser fallback in common.sh (sh variant)
### Summary

On a stock Windows machine (Git for Windows only — no jq, no real Python on `PATH`), every bash-variant script that reads `.specify/feature.json` fails with `ERROR: Feature directory not found ...` immediately after a successful `specify init --script sh`. The cause is that the JSON parser fallback chain in `common.sh` selects `python3` based on **existence** (`command -v`), but on Windows `python3` resolves to the **Microsoft Store App Execution Alias stub**, which exists on `PATH` yet fails at runtime. The chain then swallows the runtime failure and returns an empty value, and the `grep/sed` last-resort parser is unreachable because the chain is `if/elif/else` on availability, not on parse success.

### Environment

- Spec Kit `v0.12.3` (installed via `uv tool`/`uvx` from the git tag)
- Windows 11, Git for Windows (bash 5.2.37, MSYS), run under Git Bash
- No `jq` installed (Git for Windows does not ship one)
- No real Python on `PATH`; `C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\python3.exe` stub present (Windows default)

### Reproduction

```bash
# on the machine described above, in Git Bash
uvx --from "git+https://github.com/github/spec-kit.git@v0.12.3" \
  specify init target --integration generic \
  --integration-options "--commands-dir .agent/commands" \
  --script sh --ignore-agent-tools
cd target && git init -b main . && git add -A && git commit -m scaffold
bash .specify/scripts/bash/create-new-feature.sh --json --short-name demo "demo feature"
# ^ succeeds, persists .specify/feature.json = {"feature_directory":"specs/001-demo"}
bash .specify/scripts/bash/setup-plan.sh --json
# ERROR: Feature directory not found. Set SPECIFY_FEATURE_DIRECTORY or ensure
#        .specify/feature.json contains feature_directory.
```

`feature.json` exists and contains the key — it just can't be read:

```bash
$ command -v python3
/c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python3
$ python3 -c "print(42)"; echo exit=$?
Python was not found; run without arguments to install from the Microsoft Store, ...
exit=49
```

### Root cause

`scripts/bash/common.sh`, `read_feature_json_feature_directory()` (v0.12.3, ~lines 95–120):

```bash
if command -v jq >/dev/null 2>&1; then
    ... jq ...
elif command -v python3 >/dev/null 2>&1; then   # <- stub passes this test
    if ! _fd=$(python3 -c "..." 2>/dev/null); then  # <- fails at runtime (exit 49)
        _fd=''                                       # <- failure swallowed
    fi
else
    ... grep/sed fallback ...                        # <- unreachable
fi
```

`create-new-feature.sh` masks the problem on its own persist path (it falls back to `printf` for writing), so the failure only appears one script later, which makes it look like state corruption rather than a parser issue.

Notably, CI runners (GitHub Actions `windows-latest` etc.) never see this: they ship both `jq` and a real `python3`. It only bites real end-user Windows machines — which are exactly the machines the sh-variant + Git Bash path is meant to serve.

### Suggested fix

Any of these would do (first is smallest):

1. Runtime-probe instead of existence-probe: `python3 -c 'print(0)' >/dev/null 2>&1` before committing to the python3 branch (the stub also fails this, exit 49).
2. Fall through the chain on parse failure rather than returning empty (`jq` → `python3` → `grep/sed` as successive *attempts*, not exclusive branches).
3. Prefer `python` over `python3` on MSYS/Windows, or check both.

The same existence-vs-runtime pattern may exist in other scripts that shell out to `python3`; worth a quick grep.

### Workaround (for anyone else hitting this)

Install jq, disable the `python3` App Execution Alias, or expose a real interpreter as `python3` (e.g. one managed by `uv python install`).

Quratulain-bilal:
fixes #3304

## what

`read_feature_json_feature_directory` in `scripts/bash/common.sh` chose its JSON parser by *availability*:

```bash
if command -v jq ...; then ...
elif command -v python3 ...; then ...   # stub passes this
else grep/sed fallback fi               # unreachable once python3 exists
```

on windows `python3` usually resolves to the microsoft store app execution alias stub. it satisfies `command -v python3` but fails at runtime (`exit 49`, "Python was not found..."). the `elif` selected it, the failure was swallowed to `_fd=''`, and the grep/sed last resort was never reached. so a valid `.specify/feature.json` read as empty, and every script that resolves a feature this way (`setup-plan.sh`, `setup-tasks.sh`, `check-prerequisites.sh`, ...) errored with `Feature directory not found` right after a successful `specify init --script sh` — as reported.

## how

select by *parse success*, not availability: try jq, then python3 only if the result is still empty, then grep/sed only if still empty. a parser that exists but yields nothing now falls through instead of ending the chain.

the write path (`_persist_feature_json`) already uses jq-or-`printf` with no python3, so it wasn't affected; only the read path needed this.

## verified

reproduced end-to-end with git-for-windows bash, a stub `python3` (exit 49) first on PATH, and no jq:

- **before**: `bash setup-plan.sh --json` → `ERROR: Feature directory not found` (exit 1)
- **after**: same command → resolves the feature and writes `plan.md` (exit 0), emits the expected JSON paths

added `test_setup_plan_survives_broken_python3_stub`: writes a `python3` stub that exits 49, prepends it to PATH, and asserts `setup-plan.sh` still succeeds via the grep/sed fallback. it uses the same `@requires_bash` harness as the existing tests in that file (skips where a suitable bash isn't present).

### DIFF
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -97,17 +97,26 @@ read_feature_json_feature_directory() {
     local fj="$repo_root/.specify/feature.json"
     [[ -f "$fj" ]] || { printf '%s' ''; return 0; }
 
+    # Try parsers in order (jq -> python3 -> grep/sed), falling through on
+    # failure. Selection is by *parse success*, not mere availability: on
+    # Windows `python3` commonly resolves to the Microsoft Store App Execution
+    # Alias stub, which passes `command -v` but fails at runtime (exit 49), so
+    # an availability-gated `elif` would pick python3, swallow its failure, and
+    # never reach the grep/sed fallback -- leaving feature.json unreadable even
+    # though it is valid (issue #3304).
     local _fd=''
     if command -v jq >/dev/null 2>&1; then
         if ! _fd=$(jq -r '.feature_directory // empty' "$fj" 2>/dev/null); then
             _fd=''
         fi
-    elif command -v python3 >/dev/null 2>&1; then
+    fi
+    if [[ -z "$_fd" ]] && command -v python3 >/dev/null 2>&1; then
         # Use Python so pretty-printed/multi-line JSON still parses correctly.
         if ! _fd=$(python3 -c "import json,sys; d=json.load(open(sys.argv[1])); v=d.get('feature_directory'); print(v if v else '')" "$fj" 2>/dev/null); then
             _fd=''
         fi
-    else
+    fi
+    if [[ -z "$_fd" ]]; then
         # Last-resort single-line grep/sed fallback. The `|| true` guards against
         # grep returning 1 (no match) aborting under `set -e` / `pipefail`.
         _fd=$( { grep -E '"feature_directory"[[:space:]]*:' "$fj" 2>/dev/null || true; } \
--- a/tests/test_setup_plan_feature_json.py
+++ b/tests/test_setup_plan_feature_json.py
@@ -159,9 +159,23 @@ def test_setup_plan_survives_broken_python3_stub(plan_repo: Path) -> None:
     )
     stub.chmod(0o755)
 
+    # A stub jq that shadows any real jq on PATH and also fails, so the parser
+    # cannot short-circuit on jq and must reach the broken python3 stub and then
+    # fall through to grep/sed. Without this, a runner that has jq installed
+    # would parse feature.json via jq and never exercise the fallback this test
+    # is meant to cover.
+    jq_stub = stub_dir / "jq"
+    jq_stub.write_text(
+        "#!/bin/sh\n"
+        'echo "jq: simulated failure" >&2\n'
+        "exit 1\n",
+        encoding="utf-8",
+    )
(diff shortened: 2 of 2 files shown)


mnriem (changes requested):

Quratulain-bilal:
pushed 2345ac4 which adds a failing jq stub next to the python3 stub, so the chain is forced through jq -> python3 -> grep/sed no matter what the runner has installed. verified locally that with both stubs failing, grep/sed still extracts the pretty-printed feature_directory.

mnriem (approved):

mnriem:
Thank you!

PR #3323, opened 2026-07-03
fix(workflows): compare non-numeric strings lexicographically instead of returning False

### LINKED ISSUE #3321: Bug: workflow expression ordering comparisons return False for non-numeric strings
## summary

in the workflow expression evaluator, ordering comparisons (`<`, `>`, `<=`, `>=`) between non-numeric strings silently return `False`. dates, version tags, and names never compare correctly.

## root cause

`_safe_compare` in `src/specify_cli/workflows/expressions.py` coerces *both* operands to a number before comparing:

```python
try:
    if isinstance(left, str):
        left = float(left) if "." in left else int(left)
    if isinstance(right, str):
        right = float(right) if "." in right else int(right)
except (ValueError, TypeError):
    return False
```

any non-numeric string fails `int()`/`float()`, hits the `except`, and returns `False` for the whole comparison. so a legitimate string ordering is lost.

## reproduction

```python

ctx = StepContext(inputs={"d": "2026-01-01"})
# want True (jan is before feb); actual: False
evaluate_expression("{{ inputs.d < '2026-02-01' }}", ctx)

ctx = StepContext(inputs={"name": "beta"})
# want True; actual: False
evaluate_expression("{{ inputs.name > 'alpha' }}", ctx)
```

`==`/`!=` are unaffected (they don't go through `_safe_compare`); only ordering ops.

## impact

any workflow `when:` / gate condition that orders strings — comparing an iso date, a semver-ish tag, or a name — evaluates as `False` regardless of the real order, so steps are skipped (or run) incorrectly with no error.

## proposed fix

coerce to a number only when *both* operands look numeric; otherwise compare the original values. two strings then order lexicographically like python, two numeric strings still compare as numbers (`"10" > "9"`), and a number vs a non-numeric string stays incomparable (`False`).

happy to open a PR (have one ready with a regression test).

Quratulain-bilal:
fixes #3321

## what

ordering comparisons (`<`, `>`, `<=`, `>=`) between non-numeric strings in workflow expressions silently returned `False`. `_safe_compare` coerced both operands to int/float unconditionally, so any non-numeric string (iso date, version tag, name) failed coercion and the whole comparison returned `False`:

```python
ctx = StepContext(inputs={"d": "2026-01-01"})
evaluate_expression("{{ inputs.d < '2026-02-01' }}", ctx)  # was False, want True
```

## how

coerce to a number only when *both* operands look numeric; otherwise compare the original values. so:
- two strings order lexicographically like python (`"2026-01-01" < "2026-02-01"` is True)
- two numeric strings still compare as numbers (`"10" > "9"` is True, not lexical False)
- a number vs a non-numeric string stays incomparable and yields False

`==`/`!=` are untouched (they never went through `_safe_compare`).

## tests

added `test_ordering_comparison_of_non_numeric_strings` (dates, plain strings, numeric strings, number-vs-string). verified it fails on the current code (returns False) and passes with the fix; full `TestExpressions` passes.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -408,15 +408,34 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
     return _resolve_dot_path(namespace, expr)
 
 
+def _coerce_number(value: Any) -> Any:
+    """Return *value* as int/float if it is a numeric string, else unchanged."""
+    if isinstance(value, str):
+        try:
+            return float(value) if "." in value else int(value)
+        except ValueError:
+            return value
+    return value
+
+
 def _safe_compare(left: Any, right: Any, op: str) -> bool:
-    """Safely compare two values, coercing types when possible."""
-    try:
-        if isinstance(left, str):
-            left = float(left) if "." in left else int(left)
-        if isinstance(right, str):
-            right = float(right) if "." in right else int(right)
-    except (ValueError, TypeError):
-        return False
+    """Compare two values for ordering, coercing numeric strings when possible.
+
+    Numeric coercion is applied only when *both* operands look numeric, so a
+    pair like ``"10"`` and ``"9"`` compares as numbers (10 > 9). When either
+    side is a non-numeric string, both fall back to their original values and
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3331, opened 2026-07-04
fix(bundler): resolve catalog search at highest-precedence source before filtering

### LINKED ISSUE #3330: Bug: catalog search can return a shadowed lower-precedence entry that install never uses
## summary

`specify integration search` (via `CatalogStack.search()`) can return a bundle entry from a **lower-precedence** catalog even when a higher-precedence catalog defines the same bundle id. `resolve()`/install always use the higher-precedence entry, so search advertises a name/version/source the user can never actually install.

## root cause

`src/specify_cli/bundler/services/catalog_stack.py`, `search()`:

```python
for source in self._sources:               # highest precedence first
    for bundle_id, entry in self._entries_for(source).items():
        if bundle_id in seen:
            continue
        if needle and not _matches(entry, needle):   # <-- id NOT claimed on a non-match
            continue
        seen[bundle_id] = ResolvedBundle(...)
```

an id is only added to `seen` when its entry matches the query. so if the highest-precedence entry for an id does **not** match, the id stays unclaimed, and a lower-precedence entry with the same id can match on a later iteration and be returned. that contradicts the docstring ("Each bundle id appears once, resolved at its highest-precedence source") and disagrees with `resolve()`.

## reproduction

bundle id `shared` exists in two catalogs — high-precedence (name "Alpha Tool") and low-precedence (name "Searchable Widget"):

- `resolve("shared")` -> high-precedence entry (Alpha Tool)
- `search("widget")` -> returns the low-precedence "Searchable Widget" entry, which install would never use

## impact

search results can misrepresent a shadowed bundle: wrong name, version, author, and source catalog versus what `install` actually resolves. a user searching by a term that only matches the shadowed copy sees an entry they cannot install as shown.

## proposed fix

resolve each id to its highest-precedence entry first, then filter the resolved set by the query. search then agrees with `resolve()`: a query matching only a shadowed lower-precedence copy returns nothing.

Quratulain-bilal:
fixes #3330

## what

`CatalogStack.search()` claimed a bundle id in `seen` only when the entry matched the query. so when the highest-precedence entry for an id did **not** match, a lower-precedence entry with the same id could match and be returned instead — even though `resolve()`/install always use the highest-precedence entry. search advertised a bundle the user could never actually install, contradicting the method's own docstring ("resolved at its highest-precedence source").

## how

resolve every id to its highest-precedence entry first, then filter the resolved set by the query. search now agrees with `resolve()`.

## reproduction (fixed)

id `shared` in two catalogs — high-precedence "Alpha Tool", low-precedence "Searchable Widget":

- before: `search("widget")` returned the shadowed low-precedence "Searchable Widget"
- after: `search("widget")` returns nothing (that entry is not installable); `search("alpha tool")` returns the high-precedence entry

## tests

added `test_search_does_not_surface_a_shadowed_lower_precedence_entry`. verified it fails on the current code (returns the shadowed entry) and passes with the fix; the existing `test_search_dedupes_by_precedence_and_filters` and the rest of the file still pass.

### DIFF
--- a/src/specify_cli/bundler/services/catalog_stack.py
+++ b/src/specify_cli/bundler/services/catalog_stack.py
@@ -88,17 +88,25 @@ class CatalogStack:
         Results are sorted by bundle id for deterministic output.
         """
         needle = query.strip().lower()
-        seen: dict[str, ResolvedBundle] = {}
+        # Resolve each id to its highest-precedence entry FIRST, then filter by
+        # the query. Claiming an id only when it matches would let a lower-
+        # precedence entry with the same id surface when the highest-precedence
+        # one doesn't match the query — but that shadowed entry is not what
+        # `resolve()`/install would use, so search would advertise a bundle
+        # (name, version, author) the user can never actually get.
+        resolved: dict[str, ResolvedBundle] = {}
         for source in self._sources:
             for bundle_id, entry in self._entries_for(source).items():
-                if bundle_id in seen:
+                if bundle_id in resolved:
                     continue
-                if needle and not _matches(entry, needle):
-                    continue
-                seen[bundle_id] = ResolvedBundle(
+                resolved[bundle_id] = ResolvedBundle(
                     entry=entry.with_provenance(source), source=source
                 )
-        return [seen[k] for k in sorted(seen)]
+        return [
+            resolved[k]
+            for k in sorted(resolved)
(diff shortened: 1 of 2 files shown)


mnriem (approved):

mnriem:
Thank you!

PR #3333, opened 2026-07-04
fix(bundler): reject host-less catalog URLs in adapters (use hostname, not netloc)

### LINKED ISSUE #3332: Bug: bundler adapters accept host-less catalog URLs (https://:8080), unlike catalogs.py
## summary

`_validate_remote_url` in `src/specify_cli/bundler/services/adapters.py` accepts host-less catalog URLs like `https://:8080` and `https://user@`. its docstring says it "mirrors `specify_cli.catalogs` URL validation", but that site was fixed to use `hostname` in #3210/#3227 while this bundler twin still checks `netloc`.

## root cause

```python
if not parsed.netloc:
    raise BundlerError(f"Catalog '{source_id}' URL must be a valid URL with a host: {url}")
```

`urlparse("https://:8080")` gives `netloc=":8080"` (truthy) but `hostname=None`. so the guard passes even though there is no host. same for `https://user@`, `https://user:pw@`.

## reproduction

```python
from specify_cli.bundler.services.adapters import _validate_remote_url
_validate_remote_url("team", "https://:8080")     # accepted; should reject
_validate_remote_url("team", "https://user@")     # accepted; should reject
```

## impact

this is a pre-flight validation that runs before any network fetch. a host-less URL slips past the "must be a valid URL with a host" guarantee the error message promises, deferring the failure to a less clear point (or a malformed request). the sibling `specify_cli.catalogs` validator already rejects these since #3210; the bundler path should match.

## proposed fix

check `parsed.hostname` (None for host-less URLs) instead of `parsed.netloc`, mirroring the catalogs.py fix.

happy to open a PR (have one ready with regression tests).

Quratulain-bilal:
fixes #3332

## what

`_validate_remote_url` in `bundler/services/adapters.py` guarded on `parsed.netloc`, which is truthy for host-less URLs (`https://:8080`, `https://user@`) even though they carry no host — so they passed the "must be a valid URL with a host" check. the docstring says it mirrors `specify_cli.catalogs` validation, but that site was fixed to use `hostname` in #3210/#3227 and this twin was missed.

## how

check `parsed.hostname` (None for host-less URLs) instead of `parsed.netloc`, matching catalogs.py. this guard runs before any network call, so it's a pre-flight safety check.

## reproduction (fixed)

```python
_validate_remote_url("team", "https://:8080")   # before: accepted -> after: rejected
_validate_remote_url("team", "https://user@")   # before: accepted -> after: rejected
_validate_remote_url("team", "https://example.com:8080/c.json")  # still accepted
```

## tests

added parametrized regression tests for the host-less forms (`https://:8080`, `https://:0`, `https://user@`, `https://user:pw@`, `https://:8080/catalog.json`) plus a valid host+port sanity case. verified they fail on the current code and pass with the fix; the full adapters test file pas

### DIFF
--- a/src/specify_cli/bundler/services/adapters.py
+++ b/src/specify_cli/bundler/services/adapters.py
@@ -75,7 +75,11 @@ def _validate_remote_url(source_id: str, url: str) -> None:
             f"Catalog '{source_id}' URL must use HTTPS (got {parsed.scheme}://). "
             "HTTP is only allowed for localhost."
         )
-    if not parsed.netloc:
+    # Check hostname, not netloc: netloc is truthy for host-less URLs like
+    # "https://:8080" or "https://user@...", so requiring netloc would let
+    # those through even though they carry no host. hostname is None in those
+    # cases. Mirrors the fix in ``specify_cli.catalogs`` (#3210).
+    if not parsed.hostname:
         raise BundlerError(
             f"Catalog '{source_id}' URL must be a valid URL with a host: {url}"
         )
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3335, opened 2026-07-04
fix(workflows): match gate reject option case-insensitively

Noor-ul-ain001:
## Summary

A gate step authored with capitalised options — e.g. `options: [Approve, Reject]` — passes `validate()` but never aborts on rejection. Picking **Reject** silently falls through to the approval path and runs downstream steps.

## Root cause

`validate()` accepts a reject option **case-insensitively**:

```python
reject_choices = {"reject", "abort"}
if not any(o.lower() in reject_choices for o in options):
    errors.append(...)
```

So `Reject` is a valid reject option and the workflow validates fine. But `_prompt` echoes the option's **original casing**, and `execute()` compared it **case-sensitively**:

```python
if choice in ("reject", "abort"):   # "Reject" != "reject" → False
```

The two checks disagree: validation lets `Reject` through as a reject option, then the runtime treats the same `Reject` pick as approval. `on_reject: abort` is silently skipped and downstream steps run.

## Fix

Lower-case `choice` before the reject comparison so the runtime agrees with the validation that admitted the option:

```python
if choice.lower() in ("reject", "abort"):
```

## Test

`test_gate_reject_matches_case_insensitively` drives a gate with `options: [Approve, Reject]`, forces a TTY, stubs `_prompt` to return the capitalised `"Reject"`, and asserts the run reaches `RunStatus.ABORTED` and the downstream step never executes. Fails before the fix, passes after.

All 26 `test_workflows.py` gate tests pass.

### DIFF
--- a/src/specify_cli/workflows/steps/gate/__init__.py
+++ b/src/specify_cli/workflows/steps/gate/__init__.py
@@ -73,7 +73,14 @@ class GateStep(StepBase):
         choice = self._prompt(self._compose_prompt(message, show_file), options)
         output["choice"] = choice
 
-        if choice in ("reject", "abort"):
+        # Match rejection case-insensitively. ``_prompt`` echoes the option's
+        # original casing, and ``validate`` accepts a reject option
+        # case-insensitively (``o.lower() in {"reject", "abort"}``), so a gate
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3339, opened 2026-07-05
fix(workflows): apply chained expression filters left-to-right

Noor-ul-ain001:
## Summary

Filter chaining is broken in the workflow expression evaluator. `_evaluate_simple_expression` splits at the **first** top-level `|` and treats the entire remainder as one filter, so any expression with more than one filter raises `ValueError`.

This makes the canonical `map` use case impossible: `map` returns a **list**, and `join` is the only filter that renders a list to a string — the two are meant to be chained.

```python

ctx = StepContext(inputs={"rows": [{"name": "a"}, {"name": "b"}]})

# want "a, b"; actual: ValueError: filter 'join' used in an unsupported form
evaluate_expression("{{ inputs.rows | map('name') | join(', ') }}", ctx)
```

A single filter (`| map('name')` or `| join(', ')` alone) works, so chaining is the only broken axis — but it affects every registered filter.

## Root cause

```python
pipe_idx = _find_top_level(expr, "|")
if pipe_idx != -1:
    value = _evaluate_simple_expression(expr[:pipe_idx].strip(), namespace)
    filter_expr = expr[pipe_idx + 1:].strip()   # <- whole tail, incl. later `|`
    ...
    filter_match = re.match(r"(\w+)\((.+)\)", filter_expr)  # chokes on the next `|`
```

For `map('name') | join(', ')`, `filter_expr` becomes the full string and the greedy `(.+)` argument capture spans the second filter, so parsing fails.

## Fix

- Split the pipe chain into top-level segments (quote/bracket aware, so a literal `|` inside a quoted argument such as `join(' | ')` or an operand like `'a|b'` is **not** treated as a separator).
- Fold each filter over the running value left-to-right.
- The single-filter logic is extracted **verbatim** into a new `_apply_filter` helper, so all existing strict handling is preserved and now applies to every link in the chain:
  - `from_json` arity/trailing-token strictness
  - "registered filter used in an unsupported form" vs "unknown filter" messages

## Tests

Added to `TestExpressions`:
- `test_chained_filters_apply_left_to_right` — `map → join`, a three-link `map → join → contains`, and `default → contains`.
- `test_chained_filter_error_in_later_link_raises` — a mis-wired filter anywhere in the chain still fails loudly.
- `test_pipe_in_quoted_arg_is_not_a_filter_separator` — literal `|` inside quotes stays intact.

Test-the-test verified: the two chaining tests fail without the source change and pass with it. Full `TestExpressions` (37 tests) passes; the only failures in `tests/test_workflows.py` are the pre-existing Windows symlink-guard tests (unrelated, require elevation).

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -244,7 +244,7 @@ def _interpolate_expressions(template: str, namespace: dict[str, Any]) -> str:
 
 def _split_top_level(text: str, sep: str) -> list[str]:
     """Split *text* on each occurrence of *sep* that lies outside any quoted
-    string or nested bracket.
+    string or nested brackets.
 
     Used to break a filter chain (``a | map('x') | join(',')``) into its
     individual filter segments without splitting on a ``|`` that appears inside
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -242,6 +242,26 @@ def _interpolate_expressions(template: str, namespace: dict[str, Any]) -> str:
     return "".join(out)
 
 
+def _split_top_level(text: str, sep: str) -> list[str]:
+    """Split *text* on each occurrence of *sep* that lies outside any quoted
+    string or nested bracket.
+
+    Used to break a filter chain (``a | map('x') | join(',')``) into its
+    individual filter segments without splitting on a ``|`` that appears inside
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

mnriem (approved):

mnriem:
Thank you!

PR #3341, opened 2026-07-05
fix(toml): escape control characters so generated command files parse

### LINKED ISSUE #3340: generated .toml command files fail to parse when a description or body contains a control character
both toml renderers write control characters into the output raw, which produces a .toml file that a parser rejects.

toml forbids literal control characters (U+0000-U+001F except tab and newline, plus U+007F) in every string form, and a bare carriage return that is not part of a CRLF pair.

two separate code paths hit this:

- `TomlIntegration._render_toml_string` in `src/specify_cli/integrations/base.py`, used by the gemini and tabnine integrations for the `description` and `prompt` fields.
- `CommandRegistrar.render_toml_command` / `_render_basic_toml_string` in `src/specify_cli/agents.py`, used for extension/preset command files.

the multiline `"""` and `'''` forms and the basic-string fallback all emit the control char verbatim, so the resulting file cannot be loaded back.

repro (either renderer):

```python
import tomllib
from specify_cli.integrations.base import TomlIntegration
rendered = TomlIntegration._render_toml_string("ab")
tomllib.loads(f"prompt = {rendered}")  # tomllib.TOMLDecodeError: Illegal character
```

Quratulain-bilal:
fixes #3340

both toml renderers wrote control characters into their output raw, so a description or command body containing one produced a `.toml` file that fails to parse. toml forbids literal control characters (U+0000-U+001F except tab and newline, plus U+007F) in every string form, and a bare CR that is not part of a CRLF pair.

two code paths were affected:

- `TomlIntegration._render_toml_string` in `src/specify_cli/integrations/base.py` (gemini, tabnine)
- `CommandRegistrar.render_toml_command` / `_render_basic_toml_string` in `src/specify_cli/agents.py` (extension/preset commands)

the fix adds a small guard: any value that contains an illegal control char is routed to a fully-escaped basic string that emits the leftover control chars as \uXXXX, so the output always parses. the normal triple-quote multiline paths are unchanged for values without control chars.

both changes have a regression test that round-trips the rendered output through tomllib (test_toml_string_escapes_control_characters in the base toml suite, test_render_toml_command_escapes_control_characters in test_extensions). i confirmed each test fails on the pre-fix code by stashing the source and re-running.

### DIFF
--- a/src/specify_cli/_toml_string.py
+++ b/src/specify_cli/_toml_string.py
@@ -1,0 +1,60 @@
+"""Shared TOML string-escaping helpers.
+
+Both TOML command renderers — ``TomlIntegration`` (gemini, tabnine) in
+``specify_cli.integrations.base`` and ``CommandRegistrar.render_toml_command``
+(extension/preset commands) in ``specify_cli.agents`` — need the same rules for
+detecting characters TOML forbids literally and for emitting a fully-escaped
+basic string. Keeping one implementation here avoids the two drifting apart if
+the escaping rules change again.
+"""
+from __future__ import annotations
+
+
+def has_illegal_toml_control(value: str) -> bool:
+    """True when *value* contains a character TOML forbids literally.
+
+    TOML basic/literal strings (single- or multi-line) allow tab and, in the
+    multiline forms, newlines — but every other control character
+    (``U+0000``–``U+001F`` and ``U+007F``) must be ``\\u``-escaped, which only a
+    basic string can do. A bare carriage return counts too: a multiline basic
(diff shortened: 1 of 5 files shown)


mnriem (changes requested):

Quratulain-bilal:
pushed 0bf4bdc which moves both helpers into a single specify_cli/_toml_string.py and has both renderers delegate to it, so there's only one implementation to keep in sync now. no behavior change, all toml suites still pass.

mnriem (approved):

mnriem:
Thank you!

PR #3343, opened 2026-07-05
fix(yaml): pin goose recipe prompt block-scalar indentation

### LINKED ISSUE #3342: goose recipe with an indented first prompt line generates unparseable yaml
the goose recipe renderer (`YamlIntegration._render_yaml` in `src/specify_cli/integrations/base.py`) writes the prompt body under a bare `|` block scalar. yaml infers a plain block scalar's indentation from its first non-empty line, so when a command body's first line is itself indented (for example a markdown code block or a nested list continuation), the parser expects that deeper indentation for the whole block and rejects the later, shallower lines. the generated `.goose` recipe then fails to parse.

repro:

```python
import yaml
from specify_cli.integrations.base import YamlIntegration
body = "    indented first line\nback to normal"
out = YamlIntegration._render_yaml("Title", "Desc", body, "src")
yaml.safe_load(out)  # yaml.parser.ParserError: expected <block end>, but found '<scalar>'
```

it also reproduces through the public `CommandRegistrar().render_yaml_command({"description": "d"}, body, "src")` path.

a plain body (first line not indented) is unaffected, which is why the existing fixtures don't catch it.

Quratulain-bilal:
fixes #3342

the goose recipe renderer wrote the prompt body under a bare `|` block scalar. yaml infers a plain block scalar's indentation from its first non-empty line, so a command body whose first line is itself indented (a markdown code block, a nested list item) made the parser expect that deeper indent for the whole block and reject the later, shallower lines. the generated `.goose` recipe then failed to parse. a plain body was unaffected, which is why the existing fixtures didn't catch it.

the fix uses an explicit `|2` indentation indicator so the block is always read at 2 spaces regardless of what the body looks like. the body is already emitted at a fixed 2-space indent, so this just tells the parser what the renderer already does.

added a regression test (`test_yaml_prompt_with_indented_first_line_stays_valid`) that round-trips an indented-first-line body through the yaml parser. i confirmed it fails on the pre-fix code by stashing the source and re-running, and the full goose/yaml suites pass with the fix.

### DIFF
--- a/src/specify_cli/integrations/base.py
+++ b/src/specify_cli/integrations/base.py
@@ -1178,12 +1178,18 @@ class YamlIntegration(IntegrationBase):
             default_flow_style=False,
         ).strip()
 
-        # Indent the body for YAML block scalar
+        # Indent the body for YAML block scalar. Use an explicit indentation
+        # indicator ("|2") rather than a bare "|": YAML infers a plain block
+        # scalar's indentation from its first non-empty line, so a body whose
+        # first line is itself indented (e.g. a markdown code block or a nested
+        # list item) would make the parser expect that deeper indent for the
+        # whole block and reject the later, less-indented lines. Pinning the
+        # indent to 2 keeps the recipe parseable whatever the body looks like.
         indented = "\n".join(f"  {line}" for line in body.split("\n"))
 
         lines = [
             header_yaml,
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3345, opened 2026-07-05
fix(extensions): coerce non-mapping YAML config roots to {} in ConfigManager

jawwad-ali:
## Description

`ConfigManager._load_yaml_config` guards only *falsy* YAML roots:

```python
return yaml.safe_load(file_path.read_text(encoding="utf-8")) or {}
```

A **truthy non-mapping** root — a YAML list (`- foo`) or a scalar (`just a string`) — passes straight through, then `_merge_configs` does `.items()` on it:

```text
```

reproduced on main @ `bba473c` via `ConfigManager(root, 'jira').get_config()` with a list-root `jira-config.yml`. `has_value`/`get_value` crash the same way, and — worse — `HookExecutor.should_execute_hook` wraps evaluation in a blanket `except Exception: return False`, so a malformed config file **silently disables every `config.* ` hook condition** for that extension instead of surfacing an error.

### Fix

Coerce a non-`dict` root (list/scalar, or `None` for an empty file) to `{}`, mirroring the existing non-dict-root guard already present in `get_project_config()`. One place hardens all three callers (`_get_extension_defaults`, `_get_project_config`, `_get_local_config`).

## Testing

New `TestConfigManagerNonMappingYaml` (5 tests, all **fail-before / pass-after**, verified by source-stash):
- list-root and scalar-root `get_config()` return `{}` instead of raising;
- `has_value`/`get_value` don't raise;
- a **valid** `local-config.yml` still layers over a malformed (list-root) project config;
- `HookExecutor._evaluate_condition('config.x is set', 'jira')` returns `False` cleanly (asserted directly, since `should_execute_hook` masks the exception).

`uvx ruff check` clean.

## AI Disclosure

Found and fixed with Claude Code (Claude Fable 5) under my direction. AI traced the crash from `_load_yaml_config` through `_merge_configs` into the swallowed hook exception; I reproduced it, verified fail-before/pass-after, and reviewed the diff before submitting.

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -2688,7 +2688,12 @@ class ConfigManager:
             return {}
 
         try:
-            return yaml.safe_load(file_path.read_text(encoding="utf-8")) or {}
+            data = yaml.safe_load(file_path.read_text(encoding="utf-8"))
+            # Coerce a non-mapping root (list/scalar, or None for an empty
+            # file) to {} so callers that iterate/merge the result — e.g.
+            # _merge_configs' .items() — never crash. Mirrors the same
+            # non-dict-root guard in get_project_config().
+            return data if isinstance(data, dict) else {}
         except (yaml.YAMLError, OSError, UnicodeError):
             return {}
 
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3346, opened 2026-07-05
fix(integrations): hermes honors SPECKIT_INTEGRATION_HERMES_EXTRA_ARGS

jawwad-ali:
## Description

`HermesIntegration.build_exec_args` routes argv[0] through `_resolve_executable()` (so the `_EXECUTABLE` override works) but never calls `_apply_extra_args_env_var()`:

```python
args = [self._resolve_executable(), "chat", "-Q"]
if model:
    args.extend(["-m", model])
...
```

So `SPECKIT_INTEGRATION_HERMES_EXTRA_ARGS` is **silently dropped** — the same class of bug fixed for `cursor-agent` in #3265. Sibling CLI-dispatch integrations (`codex`, `devin`, `opencode`) all call the hook.

### Fix

Insert `self._apply_extra_args_env_var(args)` after the base `chat -Q` command and **before** Spec Kit's canonical `-m`/`--json`/`-s`/`-q` flags, mirroring `opencode`'s placement — so operator-supplied flags can't displace or clobber the canonical ones.

## Testing

New `TestHermesBuildExecArgs` (hermes had **zero** `build_exec_args` coverage):
- default argv shape pinned;
- `test_build_exec_args_honors_extra_args`: with the env var set, `--temperature 0.2` appears **before** `--json`/`-s` (asserts both membership and insertion order). **Fails before** (tokens absent — verified by source-stash), **passes after**;
- executable-override test.

`uvx ruff check` clean.

## AI Disclosure

Found and fixed with Claude Code (Claude Fable 5) under my direction. AI flagged hermes as another integration skipping the extra-args hook; I confirmed it, pinned insertion order in the test (per the #3265 review), verified fail-before/pass-after, and reviewed the diff.

### DIFF
--- a/src/specify_cli/integrations/hermes/__init__.py
+++ b/src/specify_cli/integrations/hermes/__init__.py
@@ -253,6 +253,11 @@ class HermesIntegration(SkillsIntegration):
         """
         args = [self._resolve_executable(), "chat", "-Q"]
 
+        # Operator-supplied SPECKIT_INTEGRATION_HERMES_EXTRA_ARGS go here —
+        # after the base command but before Spec Kit's canonical -m/--json/-s/-q
+        # flags — so they can't displace or clobber them (mirrors opencode).
+        self._apply_extra_args_env_var(args)
+
         if model:
             args.extend(["-m", model])
         if output_json:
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3347, opened 2026-07-05
fix(integrations): agy honors SPECKIT_INTEGRATION_AGY_EXTRA_ARGS

jawwad-ali:
## Description

`AgyIntegration.build_exec_args` returned its argv in one line without calling `_apply_extra_args_env_var()`:

```python
# agy does not support --model or JSON output; both params are ignored
return [self._resolve_executable(), "--print", prompt]
```

So `SPECKIT_INTEGRATION_AGY_EXTRA_ARGS` was **silently dropped** — the same class of bug fixed for `cursor-agent` in #3265. (argv[0] already went through `_resolve_executable()`, so the `_EXECUTABLE` override worked; only the extra-args hook was missing.)

### Fix

Append `self._apply_extra_args_env_var(args)` after the positional prompt, matching the `devin` integration's shape (`[exe, "-p", prompt]` then hook). agy still ignores `model`/`output_json` as before.

## Testing

Added to the existing `TestAgyBuildExecArgs`:
- `test_build_exec_args_honors_extra_args`: with the env var set, argv == `['agy','--print','my prompt','--verbose']`. **Fails before** (`--verbose` absent — verified by source-stash), **passes after**;
- executable-override test.

The 3 existing tests pin the env-unset shape and stay green. `uvx ruff check` clean.

## AI Disclosure

Found and fixed with Claude Code (Claude Fable 5) under my direction. AI flagged agy as another integration skipping the extra-args hook; I confirmed it, verified fail-before/pass-after, and reviewed the diff.

### DIFF
--- a/src/specify_cli/integrations/agy/__init__.py
+++ b/src/specify_cli/integrations/agy/__init__.py
@@ -89,7 +89,11 @@ class AgyIntegration(SkillsIntegration):
         output_json: bool = True,
     ) -> list[str] | None:
         # agy does not support --model or JSON output; both params are ignored
-        return [self._resolve_executable(), "--print", prompt]
+        args = [self._resolve_executable(), "--print", prompt]
+        # Honor SPECKIT_INTEGRATION_AGY_EXTRA_ARGS (operator-supplied flags),
+        # appended after the positional prompt like the devin integration.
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3352, opened 2026-07-05
fix(workflows): route run/resume errors to stderr under --json

jawwad-ali:
## Description

`specify workflow run`/`resume --json` is contracted to emit a single JSON object on **stdout** (the module already goes to great lengths for this — see `_emit_workflow_json` and `_stdout_to_stderr_when`, which `dup2`-redirects step output to stderr during execution). But the command's own **error paths** use `console.print`, which writes to stdout even under `--json`:

```python
except FileNotFoundError:
    console.print(f"[red]Error:[/red] Workflow not found: {source}")   # -> stdout
...
console.print("[red]Workflow validation failed:[/red]")               # -> stdout
```

So a not-found / invalid / validation-failed / failed run — and the shared `_parse_input_values` invalid-input error — all **corrupt the JSON stream** a caller is trying to parse.

### Fix

Route those error messages through `err_console` when `--json` is set (a no-op in normal text mode, where they stay on the regular console). This mirrors the stderr-only error routing already used by `specify bundle` (`_fail`) and by `err_console` elsewhere in this same module. Covers both `workflow run` and `workflow resume` plus the shared input parser.

## Testing

New `TestWorkflowRunJsonErrorStream` (CliRunner separates stdout/stderr in this repo's Click):
- validation-failure under `--json`: `"validation failed"` is **not** on stdout and **is** on stderr;
- invalid `--input` under `--json`: `"Invalid input format"` is **not** on stdout and **is** on stderr.

Both **fail-before / pass-after** (verified by source-stash). `uvx ruff check` clean.

## AI Disclosure

Found and fixed with Claude Code (Claude Fable 5) under my direction. AI spotted that the error paths bypassed the module's own stdout-cleanliness machinery; I confirmed the stream corruption, verified fail-before/pass-after, and reviewed the diff.

### DIFF
--- a/src/specify_cli/workflows/_commands.py
+++ b/src/specify_cli/workflows/_commands.py
@@ -50,7 +50,17 @@ workflow_step_catalog_app = typer.Typer(
 workflow_step_app.add_typer(workflow_step_catalog_app, name="catalog")
 
 
-def _parse_input_values(input_values: list[str] | None) -> dict[str, Any]:
+def _error_console(json_output: bool):
+    """Console for error text: stderr under ``--json`` so the JSON stdout
+    stream stays parseable, the normal console otherwise. Mirrors the
+    stderr-only error routing already used by ``specify bundle``.
+    """
+    return err_console if json_output else console
+
+
+def _parse_input_values(
+    input_values: list[str] | None, *, json_output: bool = False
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3376, opened 2026-07-06
fix(integrations): guard _sha256 against unreadable managed files

jawwad-ali:
## Description

`manifest.py::_sha256` does an unguarded `open(path, "rb")`. Two callers invoke it on a **regular file that exists but can't be opened** (e.g. permission denied) without catching `OSError`:

- `check_modified()` (line ~312) — only symlinks/non-regular files are guarded above; a readable-looking-but-unopenable regular file lets the `OSError` escape.
- `uninstall()` (line ~361) — the sibling `path.unlink()` right below **is** wrapped in `except OSError`, but the hash read is not (asymmetric).

So `specify integration upgrade` / `uninstall` / `switch` dump a raw `PermissionError` traceback instead of handling it. Reproduced on main @ `92b7cf7`.

### Fix (single file — `manifest.py`)

- `check_modified()`: catch `OSError` and treat an unreadable file as **modified**, consistent with how symlinks / non-regular files are already appended just above.
- `uninstall()`: catch `OSError` around the hash read and treat it as **skipped/preserved** (can't verify ownership → don't delete), mirroring the existing `unlink()` `OSError` guard. The `force` short-circuit is unchanged.

## Testing

New `TestManifestUnreadableFile` — deterministic and **cross-platform** via `monkeypatch` of `_sha256` to raise `PermissionError` (no real `chmod`/symlink needed):
- `check_modified()` returns the file as modified (was: raised);
- `uninstall(force=False)` returns it in `skipped`, removes nothing, leaves it on disk (was: raised).

Both **fail-before / pass-after** (verified by source-stash). Note: 6 pre-existing symlink tests in this file fail locally on Windows (symlink privilege) — identical on clean main, unrelated to this change, and green on Linux CI.

## AI Disclosure

Found and fixed with Claude Code (Claude Fable 5) under my direction. AI flagged the unguarded `_sha256` against the guarded `unlink()` sibling; I reproduced the traceback path, wrote cross-platform monkeypatch tests, verified fail-before/pass-after, and confirmed the 6 remaining failures are pre-existing Windows-symlink environment issues.

### DIFF
--- a/src/specify_cli/integrations/manifest.py
+++ b/src/specify_cli/integrations/manifest.py
@@ -309,7 +309,14 @@ class IntegrationManifest:
             if abs_path.is_symlink() or not abs_path.is_file():
                 modified.append(rel)
                 continue
-            if _sha256(abs_path) != expected_hash:
+            try:
+                changed = _sha256(abs_path) != expected_hash
+            except OSError:
+                # Unreadable regular file (e.g. permission denied): treat as
+                # modified, consistent with the symlink / non-regular-file
+                # handling above, rather than letting the OSError escape.
+                changed = True
+            if changed:
                 modified.append(rel)
         return modified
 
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3379, opened 2026-07-06
fix(github-http): return None on malformed GHES port instead of raising

jawwad-ali:
## Description

`resolve_github_release_asset_api_url` is contracted to either resolve a browser release-download URL to its REST API asset URL or **return `None`** — every unresolvable case (unknown host, not-an-asset path, network error, missing asset) returns `None`. But the GHES branch builds the authority via `parsed.port`:

```python
elif is_ghes:
    authority = hostname if parsed.port is None else f"{hostname}:{parsed.port}"
```

`urllib`'s `.port` **raises `ValueError`** on a malformed port. So an allowlisted GHES host with a bad port (e.g. `https://ghes.example:notaport/...`) crashes the caller instead of returning `None`. Reproduced on main @ `92b7cf7`: `ValueError: Port could not be cast to integer value as 'notaport'`.

### Fix

Read `parsed.port` inside a `try/except ValueError: return None`, matching the function's existing graceful `return None` behavior.

## Testing

New `test_returns_none_on_malformed_ghes_port`: an allowlisted GHES host with `:notaport` returns `None` and induces **no** network call. **Fails before** (`ValueError` — verified by source-stash), **passes after**. The existing valid-port GHES test still passes (happy path unchanged). Full `test_github_http.py`: 27 passed. `uvx ruff check` clean.

## AI Disclosure

Found and fixed with Claude Code (Claude Fable 5) under my direction. AI spotted the unguarded `parsed.port` against the function's return-None contract; I reproduced the ValueError, verified fail-before/pass-after, and reviewed the diff.

### DIFF
--- a/src/specify_cli/_github_http.py
+++ b/src/specify_cli/_github_http.py
@@ -127,7 +127,14 @@ def resolve_github_release_asset_api_url(
     if hostname == "github.com":
         api_base = "https://api.github.com"
     elif is_ghes:
-        authority = hostname if parsed.port is None else f"{hostname}:{parsed.port}"
+        # ``parsed.port`` raises ValueError on a malformed port (e.g.
+        # ``host:notaport``); the function's contract is to return None for
+        # anything it can't resolve, not to raise.
+        try:
+            port = parsed.port
+        except ValueError:
+            return None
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3392, opened 2026-07-07
test(integrations): guard multiline/control-char SKILL.md frontmatter escaping

Quratulain-bilal:
Resolved the conflicts by rebasing onto `main`.

When this PR was opened it added a helper to escape multiline / control characters in generated SKILL.md frontmatter. In the meantime #3399 landed `yaml_quote()` in `integrations/base.py` and routes every SKILL.md frontmatter field through it, which already provides that escaping. So the production change here was fully redundant and I dropped it during the rebase (including the now-unused `_yaml_string.py`).

What remains is the **regression coverage**, which `main` did not have: two tests on the shared `SkillsIntegrationTests` mixin that round-trip a **multiline** description and a **control-character** description back through the YAML parser, asserting the rendered SKILL.md frontmatter stays valid and lossless. They run across every skills integration (agy, codex, hermes, kimi, …) and lock in the `yaml_quote()` behavior so a future hand-rolled quote can't silently regress it.

Net diff is now test-only. Happy to close instead if you'd rather not carry the extra coverage, but it guards a real corruption path that had no direct test before.

### DIFF
--- a/tests/integrations/test_integration_base_skills.py
+++ b/tests/integrations/test_integration_base_skills.py
@@ -141,6 +141,91 @@ class SkillsIntegrationTests:
             assert isinstance(fm["description"], str)
             assert len(fm["description"]) > 0, f"{f} has empty description"
 
+    def test_skill_frontmatter_preserves_multiline_description(
+        self, tmp_path, monkeypatch
+    ):
+        """A multiline (block-scalar) description must round-trip exactly.
+
+        The hand-built SKILL.md frontmatter used to only escape backslash and
+        quote, so a block-scalar description was emitted with raw newlines inside
+        a double-quoted scalar and reparsed with those newlines collapsed to
+        spaces. The description must survive byte-for-byte."""
+        from pathlib import Path
+
+        i = get_integration(self.KEY)
+        # Hermes writes to ~/.hermes/skills/ — isolate Path.home() to prevent
+        # overwriting a developer's real global skill directory.
+        if self.KEY == "hermes":
+            home = tmp_path / "home"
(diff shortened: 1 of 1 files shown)


mnriem (changes requested):

Quratulain-bilal:
good call. added a second round-trip (fcdbbcd) that feeds a C0 control char (U+0008, via a yaml escape in the source template) through quote_yaml_double and asserts it survives byte for byte. confirmed it fails on the pre-fix naive quote (raw control char makes the SKILL.md unparseable, yaml ReaderError) and passes with the fix.

mnriem:
Please resolve conflicts

Quratulain-bilal:
Closing this as superseded rather than resolving the conflicts, since the conflict is with the fix that replaced it.

This PR added a `_yaml_string.py` helper to escape multiline / control characters in SKILL.md frontmatter. That exact need is now covered on `main` by #3399, which introduced `yaml_quote()` in `integrations/base.py` and routes every generated SKILL.md frontmatter field through it (`name`, `description`, `compatibility`, `metadata.author`, `metadata.source`). `yaml_quote()` uses the YAML emitter to produce the escapes, so newlines and control characters are handled — the same guarantee this PR aimed for.

Rebasing would just re-introduce a second, parallel helper for behavior that already ships, so closing is cleaner. Happy to reopen if there's an edge case `yaml_quote()` misses that this approach caught.

mnriem:

mnriem (approved):

mnriem:
Thank you!

PR #3399, opened 2026-07-07
fix(integrations): escape control characters in SKILL.md frontmatter

### LINKED ISSUE #3391: SKILL.md frontmatter corrupts multiline descriptions and breaks on control characters
### description

the skills setup path builds `SKILL.md` frontmatter by hand (to match the release packaging output) instead of using yaml.safe_dump. the `_quote` helper wraps each value in a double-quoted yaml scalar but only escapes backslash and the quote:

```python
def _quote(v):
    escaped = v.replace("\\", "\\\\").replace('"', '\\"')
    return f'"{escaped}"'
```

a double-quoted yaml scalar cannot carry a raw newline or a control character. so:

- a multiline (block-scalar) `description:` in the command template is emitted with raw newlines inside the quoted scalar, and reparses with the newlines collapsed to spaces (silent corruption).
- a control character (U+0000–U+001F except tab/newline, or U+007F) makes the generated SKILL.md fail to load at all (`ReaderError: unacceptable character`).

the description comes straight from the template frontmatter (`frontmatter.get("description")`), so any command whose description is a block scalar hits the multiline case.

this is the SKILL.md sibling of the toml escaping fixed in #3341 and the goose yaml issue #3382. affects `TomlIntegration`/skills path in `src/specify_cli/integrations/base.py` and the identical `_quote` in `src/specify_cli/integrations/hermes/__init__.py`.

repro (via any skills integration, e.g. hermes): a template with

```
---
description: |
  first line
  second line
---
```

produces a SKILL.md whose parsed description is `"first line second line "` instead of `"first line\nsecond line\n"`.

marcelsafin:
## Summary

Fixes #3391

`SkillsIntegration.setup()` and the hermes skill builder quoted SKILL.md frontmatter values with a hand-rolled helper that only escaped `\` and `"`. A template description containing a newline or a control character (for example a YAML escape like `\a` in the source frontmatter, which parses to a real control byte) produced a broken or lossy SKILL.md: the newline split the quoted scalar across lines and the control byte was emitted raw, which is invalid in double-quoted YAML.

## Changes

- New module-level `yaml_quote()` in `integrations/base.py`: delegates to `yaml.safe_dump(default_style='"')`, which handles every escape case per spec. Output is byte-identical to the old helper for plain strings.
- Both copies of the inline `_quote` helper (base.py skill builder and hermes) now use it, removing the duplicated logic.
- New regression tests: unit tests for `yaml_quote` plus integration tests that render SKILL.md from hostile templates and re-parse the frontmatter with `yaml.safe_load`.

## Testing

- `uv run pytest tests/integrations/test_skill_frontmatter_quoting.py` — 6 new tests
- Full suite: 3808 passed, 105 skipped
- `uvx ruff check` — clean

## Checklist

- [x] Tests pass locally
- [x] Lint passes
- [x] I did use AI assistance for this contribution

### DIFF
--- a/src/specify_cli/integrations/base.py
+++ b/src/specify_cli/integrations/base.py
@@ -51,6 +51,18 @@ _CORE_COMMAND_TEMPLATE_RANK = {
 }
 
 
+def yaml_quote(value: str) -> str:
+    """Emit *value* as a double-quoted YAML scalar on a single line.
+
+    A hand-rolled quote cannot carry raw newlines (YAML folds them to
+    spaces) or control characters (the reader rejects them), so let the
+    YAML emitter produce the escapes.
+    """
+    return yaml.safe_dump(
+        str(value), default_style='"', allow_unicode=True, width=sys.maxsize
+    ).strip()
+
+
 # ---------------------------------------------------------------------------
 # IntegrationOption
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3413, opened 2026-07-08
feat(git-extension): add configurable Conventional Commit support (#3390)

### LINKED ISSUE #3390: [Feature]: Add configurable Conventional Commit support to the bundled Git extension
### Problem Statement

The bundled `git` extension currently supports auto-commit hooks using fixed/static commit messages from `git-config.yml`.

This works for the default Spec Kit workflow, but it does not fit teams that use Conventional Commits for changelogs, release automation, semantic versioning, or CI policy checks.

For example, today the extension can produce messages like `[Spec Kit] Add specification` or `[Spec Kit] Add implementation plan`, but some projects need messages like `feat: add OAuth specification` or `docs: update implementation plan`.

There does not appear to be a built-in way to keep using the Git extension’s auto-commit hooks while asking the agent to generate Conventional Commit messages from the actual diff.

### Proposed Solution

Add a new configuration option to the bundled `git` extension, for example:

`commit_style: fixed`

Supported values:

- `fixed` - default; preserve the current behavior and use configured/static messages.
- `conventional` - ask the agent to inspect the relevant changes and provide a Conventional Commit message.

In `conventional` mode:

- `auto_commit` should still control whether commits happen for each hook/event.
- The configured `auto_commit.<event>.message` values should not be used as the primary commit message source.
- The agent-facing `speckit.git.commit` command should instruct the agent to inspect the diff and generate an appropriate Conventional Commit message.
- The shell/PowerShell scripts should receive the generated message explicitly, for example via an argument or environment variable such as `SPECKIT_COMMIT_MESSAGE`.
- If no generated message is provided, the command should fail or skip with a clear warning instead of silently falling back to the fixed message format.

This would keep semantic commit-message generation in the agent layer while keeping the Git scripts simple and deterministic.

### Alternatives Considered

Possible alternatives:

- Manually edit the generated commit messages after each auto-commit. This defeats much of the value of auto-commit hooks.
- Disable the Git extension’s auto-commit hooks and commit manually. This works, but loses the integrated Spec Kit lifecycle behavior.
- Add custom local scripts outside the bundled extension. This is flexible, but each project has to reinvent the same behavior.
- Try to generate Conventional Commit messages inside the shell/PowerShell scripts. This seems less desirable because the scripts should not be responsible for semantic diff analysis.

### Component

Other

### AI Agent (if applicable)

None

### Use Cases

1. Teams using Conventional Commits with automated changelog generation.
2. Projects using semantic-release or similar release automation.
3. Repositories with CI checks that enforce Conventional Commit format.
4. Teams that want Spec Kit lifecycle auto-commits while preserving project commit conventions.
5. Multi-agent workflows where commit message style should be configured once in `git-config.yml`.

### Acceptance Criteria

- [ ] `extensions/git/config-template.yml` includes a documented `commit_style` option.
- [ ] Existing projects without `commit_style` continue to behave exactly as they do today.
- [ ] `commit_style: fixed` preserves current fixed/static commit message behavior.
- [ ] `commit_style: conventional` causes `speckit.git.commit` to ask the agent to generate a Conventional Commit message from the diff.
- [ ] Bash and PowerShell auto-commit scripts can accept an explicit generated commit message.
- [ ] In conventional mode, missing generated commit messages produce a clear warning/failure instead of silently using fixed messages.
- [ ] Tests cover default behavior, fixed mode, conventional mode, and missing-message behavior.
- [ ] Documentation is updated for the Git extension configuration.

### Additional Context

The current Git extension already centralizes commit hook behavior in:

- `extensions/git/config-template.yml`
- `extensions/git/git-config.yml`
- `extensions/git/commands/speckit.git.commit.md`
- `extensions/git/scripts/bash/auto-commit.sh`
- `extensions/git/scripts/powershell/auto-commit.ps1`

This seems like a natural place to add a small configuration option without changing the broader Spec Kit workflow.

dhruv-15-03:
## Summary

Implements issue #3390: adds a configurable `commit_style` option (`fixed` | `conventional`) to the bundled **git extension**'s auto-commit feature.

- `commit_style: fixed` (default) — unchanged, static configured messages.
- `commit_style: conventional` — the `speckit.git.commit` hook now instructs the agent to generate a Conventional Commit message from the diff and pass it explicitly to `auto-commit.sh` / `auto-commit.ps1` as a second argument.
- If `commit_style: conventional` is set but no generated message is supplied, the scripts fail loudly (stderr message + exit 1) instead of silently committing with the fixed message — while still short-circuiting cleanly when there are no changes to commit.

## Changes

- `extensions/git/config-template.yml`, `extensions/git/git-config.yml`: new `commit_style` option with docs.
- `extensions/git/scripts/bash/auto-commit.sh`: optional `[generated_message]` arg, `commit_style` parsing, conventional-mode enforcement (ordered after the "no changes" check).
- `extensions/git/scripts/powershell/auto-commit.ps1`: mirrored PowerShell implementation.
- `extensions/git/commands/speckit.git.commit.md`: documents the new "Commit Message Styles" section and updated execution/config/graceful-degradation guidance.
- `extensions/git/README.md`: documents the new option in the configuration example.
- `tests/extensions/git/test_git_extension.py`: new `TestAutoCommitBashCommitStyle` / `TestAutoCommitPowerShellCommitStyle` regression tests covering fixed-default, conventional success, conventional missing-message failure, and no-changes short-circuit.

## Testing

- `pytest tests/extensions/git/test_git_extension.py -v` — all bash/PowerShell tests requiring bash/pwsh skip consistently in this environment (matching existing `TestAutoCommitBash`/`TestAutoCommitPowerShell` behavior via `@requires_bash` / `HAS_PWSH`).
- Manually validated all four scenarios (fixed default, conventional success, conventional missing-message failure, no-changes short-circuit) end-to-end against `auto-commit.ps1` using Windows PowerShell, since pwsh (PowerShell 7) isn't installed in this environment.
- `pytest tests/` (full suite): 3614 passed, 295 skipped, 39 failed — all 39 failures are pre-existing, environment-specific (Windows symlink-permission tests and pwsh-only tests), unrelated to this change.
- `ruff check extensions/git tests/extensions/git`: all checks passed.

Fixes #3390

---


dhruv-15-03:
Fixed the flaky macOS CI failure in `pytest (macos-latest, 3.13)`: `test_conventional_without_generated_message_fails` asserted on `result.stderr` only, but pwsh's `Write-Warning` stream placement (stdout vs. stderr) isn't deterministic across platforms/pwsh versions. Updated the assertion to check the combined stdout+stderr, matching the pattern already used elsewhere in this file (e.g. `test_not_a_repo_still_detected_with_autocrlf`). Pushed as 7a6951c.

mnriem:

grafvonb:
@dhruv-15-03 could you follow up soon? Speckit is a moving target. Thx.

dhruv-15-03:
Addressed the outstanding Copilot review feedback in c1d245d:

- **bash** (`extensions/git/scripts/bash/auto-commit.sh`): fixed `commit_style` parsing to correctly strip trailing YAML inline comments (e.g. `commit_style: conventional  # team standard`). The previous regex only matched a single space before `#`, so quoted values with two+ spaces before the comment (e.g. `commit_style: "conventional"  # x`) left a stray trailing `"` after quote-stripping. Now uses a `{1,}` interval to consume all whitespace + the comment together.
- **powershell**: already handled this correctly (`\s+#.*$` + `.Trim()`), verified with manual regex tests - no change needed.
- **tests**: added regression coverage for `commit_style` values with trailing inline comments (bash + pwsh), plus the pwsh no-changes short-circuit regression test suggested in the review.

All existing + new tests pass locally (ruff clean, targeted commit_style test classes green).

mnriem (changes requested):

dhruv-15-03:
Addressed the remaining Copilot review feedback in 5f20fd5:

- Validated `commit_style` against the documented `fixed`/`conventional` values in both bash and pwsh; an unrecognized value (e.g. a typo like `conventonal`) now emits a warning and falls back to `fixed` instead of silently mis-parsing.
- Reworded the conventional-mode-without-generated-message message from "skipped auto-commit" to "aborting auto-commit" (the script exits 1, which is a failure, not a skip), and added the concrete remediation (pass the generated message as arg 2, or set `commit_style: fixed`).
- Added bash + pwsh regression tests covering the unknown `commit_style` fallback.

Left the test-duplication suggestion (parametrizing the bash/pwsh commit-style test classes) unaddressed for now since it's a structural refactor rather than a behavior bug; happy to do it in a follow-up if preferred.

All changes verified locally (bash syntax check, PowerShell AST parse, manual script runs for both the unknown-style-fallback and abort-message paths).

mnriem:

dhruv-15-03:
Addressed the latest Copilot review feedback in 70024d4: the bash `commit_style` parser now uses `grep -m1` to limit matching to the first `commit_style:` line in the config, preventing value concatenation if a config accidentally contains duplicate keys (e.g. from a bad merge). Added a regression test (`test_duplicate_commit_style_lines_use_first_match`) covering this case. The PowerShell version already limited to the first match via its `break` statement, so no change was needed there.

mnriem:

dhruv-15-03:
- **Whitespace bug**: removed `tr -d '[:space:]'` from the bash `commit_style` parsing pipeline. It was stripping *all* whitespace from the value (not just surrounding it), so `commit_style: con ventional` was silently normalized to `conventional` instead of being rejected as unknown - inconsistent with the PowerShell version, which already rejected it correctly. Both now behave the same.
- **Command injection risk**: added a file-based message channel - `--message-file <path>` (bash) / `-MessageFile <path>` (PowerShell) - so an agent-generated commit message (which may contain quotes, `$(...)`, or backticks) is read from a file instead of being interpolated into a shell command string. Updated `speckit.git.commit.md` to instruct writing the message to a temp file via the file-editing tool and passing the path, explicitly warning against shell inlining. The old positional-argument form is kept for backward compatibility.
- **Test coverage**: added explicit `commit_style: fixed` tests (previously only the absent-key default was covered), plus success/missing-file tests for the new message-file flag, including an injection-shaped message (`feat: add $(dangerous) `injection` test`) verified to land in the commit message verbatim with no shell expansion - for both bash and PowerShell suites.

All three fixes were manually smoke-tested locally (bash and PowerShell) confirming correct behavior before pushing.

mnriem:

dhruv-15-03:
Addressed the latest Copilot review feedback in e16db03: the message file passed via `--message-file`/`-MessageFile` was read but left in the worktree. If written inside the project (e.g. by an agent's file-editing tool), it was staged into the commit by `git add .`, and its mere presence as an untracked file could also defeat the no-changes short-circuit, producing a spurious commit containing only that file. Both scripts now remove the transport file immediately after its content is captured, before the change-detection check and before staging. Added bash + pwsh regression tests covering: (1) the message file is excluded from the resulting commit and removed from disk, and (2) when it is the only 'change' present, auto-commit correctly reports no changes rather than committing.

Verified locally (bash syntax check, PowerShell AST parse, and manual runs of both new scenarios against the real test fixtures).

mnriem (approved):

mnriem:
Thank you!

PR #3417, opened 2026-07-08
fix(templates): remove self-referencing path in plan-template.md note

### LINKED ISSUE #1148: Wrong directory path in the "Note" section of plan-template.md
https://github.com/github/spec-kit/blob/e6d6f3cdee99752baee578896797400a72430ec0/templates/plan-template.md?plain=1#L6

It seems that this should be amended by the CLI based on the agent used. For example, if we use cursor-agent, it's not there. It's in `.cursor/commands/speckit.plan.md`. 

Or am I wrong, and this note won't confuse the agent, and it's for a reference in the spec-kit repo?

marcelsafin:
Fixes #1148

## Root cause

The note at the top of `templates/plan-template.md` says:

> See `.specify/templates/plan-template.md` for the execution workflow.

That path is the file itself — a self-reference. The execution workflow actually lives in the plan command's definition, which the same sentence already names via the `__SPECKIT_COMMAND_PLAN__` placeholder (rendered per-agent, e.g. `/speckit.plan`).

An earlier attempt (#1342) went stale and was auto-closed; this picks the issue back up with the same intent — make the note point at the right thing without hardcoding any agent-specific path.

## Change

One line: the note now reads "This template is filled in by the `__SPECKIT_COMMAND_PLAN__` command; its definition describes the execution workflow."

## Testing

No tests pin this template's prose; the placeholder continues to render through the existing `resolve_command_refs` pipeline (shared_infra.py).

### DIFF
--- a/templates/plan-template.md
+++ b/templates/plan-template.md
@@ -4,7 +4,7 @@
 
 **Input**: Feature specification from `/specs/[###-feature-name]/spec.md`
 
-**Note**: This template is filled in by the `__SPECKIT_COMMAND_PLAN__` command. See `.specify/templates/plan-template.md` for the execution workflow.
+**Note**: This template is filled in by the `__SPECKIT_COMMAND_PLAN__` command; its definition describes the execution workflow.
 
 ## Summary
 


mnriem:
Thank you!

PR #3433, opened 2026-07-09
fix(bundler): raise BundlerError, not raw ValueError, on a malformed catalog URL

### LINKED ISSUE #3432: bundler catalog validator leaks raw ValueError on a malformed URL
### description

`_validate_remote_url` in `src/specify_cli/bundler/services/adapters.py` reads `parsed.hostname` on the user-supplied catalog url. urlparse/hostname raise `ValueError` on a malformed authority (e.g. an unclosed ipv6 bracket `https://[::1`), so instead of the clean `BundlerError` every other reject path in that function produces, a raw `ValueError: Invalid IPv6 URL` escapes and crashes the catalog fetch.

this is the bundler sibling of #3369, which fixed the cli `extension`/`preset`/`workflow add` entry points but not this validator (nor the `bundle catalog add` path that reaches it via #3367).

repro:

```python
from specify_cli.bundler.services import adapters
adapters._validate_remote_url("team", "https://[::1")  # ValueError: Invalid IPv6 URL
```

expected: a `BundlerError` like the host-less and non-https cases already raise.

Quratulain-bilal:
fixes #3432

`_validate_remote_url` read `parsed.hostname`, which raises `ValueError` on a malformed authority (e.g. an unclosed ipv6 bracket `https://[::1`). the function's contract is to raise `BundlerError` for any bad url - every other reject path does - so the raw ValueError leaked to the caller and crashed the fetch instead of failing cleanly.

bundler sibling of #3369, which fixed the cli extension/preset/workflow add paths but not this validator (reached via `bundle catalog add`, #3367).

**fix:** wrap the parse + hostname access and convert `ValueError` to `BundlerError`, then reuse the parsed hostname for the existing host check.

added a regression test (`test_validate_remote_url_rejects_malformed_url_cleanly`) covering an unclosed ipv6 bracket and a non-ip bracket host. i confirmed it fails on the pre-fix code by stashing the source and re-running; the adapters suite passes with the fix.

### DIFF
--- a/src/specify_cli/bundler/services/adapters.py
+++ b/src/specify_cli/bundler/services/adapters.py
@@ -68,8 +68,18 @@ def _validate_remote_url(source_id: str, url: str) -> None:
     Mirrors ``specify_cli.catalogs`` URL validation to avoid MITM/downgrade
     issues before any network call.
     """
-    parsed = urlparse(url)
-    is_localhost = parsed.hostname in ("localhost", "127.0.0.1", "::1")
+    # A malformed authority (e.g. an unclosed IPv6 bracket ``https://[::1``)
+    # makes urlparse / hostname access raise ValueError. This function's
+    # contract is to raise BundlerError for a bad URL, so surface that as a
+    # clean error rather than leaking a raw ValueError to the caller.
+    try:
+        parsed = urlparse(url)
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3435, opened 2026-07-09
fix(catalogs): raise catalog error, not raw ValueError, on a malformed URL

### LINKED ISSUE #3434: catalog url validator leaks raw ValueError on a malformed URL
### description

`CatalogStackBase._validate_catalog_url` in `src/specify_cli/catalogs.py` reads `parsed.hostname` on the catalog url. urlparse/hostname raise `ValueError` on a malformed authority (e.g. an unclosed ipv6 bracket `https://[::1`, or a bracketed non-ip host `https://[not-an-ip]`), so instead of the clean catalog error every other reject path in that method produces, a raw `ValueError` escapes.

this reaches user input via `specify integration catalog add <url>`.

repro:

```python
from specify_cli.catalogs import IntegrationCatalog
IntegrationCatalog._validate_catalog_url("https://[::1")  # ValueError: Invalid IPv6 URL
```

expected: the normal catalog error, like the non-https and host-less cases already raise.

twin of the bundler validator issue #3432.

Quratulain-bilal:
fixes #3434

`CatalogStackBase._validate_catalog_url` read `parsed.hostname`, which raises `ValueError` on a malformed authority (unclosed ipv6 bracket `https://[::1`, bracketed non-ip host `https://[not-an-ip]`). the method's other reject paths all raise the class catalog error, so the raw ValueError leaked instead of failing cleanly. reachable via `specify integration catalog add <url>`.

twin of the bundler validator fix (#3432 / #3433).

**fix:** wrap the parse + hostname access and convert `ValueError` to the normal error via `cls._error`, then reuse the parsed hostname for the existing scheme/host checks.

added a regression test (`test_malformed_url_rejected_cleanly`) covering an unclosed ipv6 bracket and a bracketed non-ip host. i confirmed it fails on the pre-fix code by stashing the source and re-running (raw ValueError leaks); the url-validation suite passes with the fix.

### DIFF
--- a/src/specify_cli/catalogs.py
+++ b/src/specify_cli/catalogs.py
@@ -71,8 +71,12 @@ class CatalogStackBase:
         """Validate that a catalog URL uses HTTPS, except localhost HTTP."""
         from urllib.parse import urlparse
 
-        parsed = urlparse(url)
-        is_localhost = parsed.hostname in ("localhost", "127.0.0.1", "::1")
+        try:
+            parsed = urlparse(url)
+            hostname = parsed.hostname
+        except ValueError:
+            raise cls._error(f"Catalog URL is malformed: {url}") from None
+        is_localhost = hostname in ("localhost", "127.0.0.1", "::1")
         if parsed.scheme != "https" and not (parsed.scheme == "http" and is_localhost):
             raise cls._error(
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3437, opened 2026-07-09
fix(auth): return no matches, not raw ValueError, for a malformed URL

### LINKED ISSUE #3436: auth find_entries_for_url leaks raw ValueError on a malformed URL
### description

`find_entries_for_url` in `src/specify_cli/authentication/config.py` does `(urlparse(url).hostname or "").lower()` unguarded. a malformed authority (e.g. an unterminated ipv6 bracket `https://[::1`, or a bracketed non-ip host `https://[not-an-ip]`) makes urlparse/hostname raise `ValueError`, so instead of the empty list the function already returns for a host-less url, a raw `ValueError` escapes.

this function is the first step of the shared http client - `build_request` and `open_url` both call it before any url validation - so the raw error can leak out of catalog/extension/preset fetches.

repro:

```python
from specify_cli.authentication.config import find_entries_for_url
find_entries_for_url("https://[::1", [])  # ValueError: Invalid IPv6 URL
```

expected: no matches (`[]`), same as the host-less url case just above it.

same bug class as #3210 / the recent bundler + catalog validator fixes.

Quratulain-bilal:
fixes #3436

`find_entries_for_url` did `(urlparse(url).hostname or "").lower()` unguarded. a malformed authority (unterminated ipv6 bracket `https://[::1`, bracketed non-ip host `https://[not-an-ip]`) makes urlparse/hostname raise `ValueError`, so a raw ValueError leaked instead of the empty list the function already returns for a host-less url. this is the first step of the shared http client (`build_request` / `open_url` call it before any url validation), so it can surface out of catalog/extension/preset fetches.

**fix:** wrap the parse + hostname access; on `ValueError` return no matches, exactly like the host-less case just below it (no auth entry can match a url with no usable host anyway).

added a regression test (`test_malformed_url_returns_empty`) over an unterminated bracket and a bracketed non-ip host. i confirmed it fails on the pre-fix code by stashing the source and re-running (raw ValueError leaks); the auth suite passes with the fix.

same bug class as #3210 and the recent bundler/catalog validator fixes.

### DIFF
--- a/src/specify_cli/authentication/config.py
+++ b/src/specify_cli/authentication/config.py
@@ -196,7 +196,15 @@ def find_entries_for_url(
     url: str, entries: list[AuthConfigEntry]
 ) -> list[AuthConfigEntry]:
     """Return entries whose ``hosts`` match the hostname of *url*."""
-    hostname = (urlparse(url).hostname or "").lower()
+    # A malformed authority (e.g. an unterminated IPv6 bracket "https://[::1")
+    # makes urlparse/hostname raise ValueError. Treat that the same as a
+    # host-less URL: no entry can match, so return no matches rather than
+    # leaking a raw ValueError out of the shared HTTP client (build_request /
+    # open_url call this before any URL validation).
+    try:
+        hostname = (urlparse(url).hostname or "").lower()
+    except ValueError:
+        return []
     if not hostname:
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

Issue #3457, opened 2026-07-10
--integration-options with an unbalanced quote crashes with a raw ValueError


Quratulain-bilal:
### description

`_parse_integration_options` in `src/specify_cli/integrations/_helpers.py` calls `shlex.split(raw_options)` unguarded. an unbalanced quote in the flag value (e.g. `--integration-options='--commands-dir "foo'`) makes shlex raise `ValueError: No closing quotation`, so a raw traceback escapes instead of the clean `typer.Exit(1)` error every other bad-input path in this function produces (unknown option, missing value, unexpected value all print a message and exit 1).

reachable from `specify init --integration-options=...` and every `specify integration install/switch/upgrade/migrate --integration-options=...`.

repro:

```
specify integration install generic --integration-options='--commands-dir "foo'
# ValueError: No closing quotation  (raw traceback)
```

expected: a clean one-line error and exit 1, like a mistyped option name already gives.

note: i used an ai assistant to help investigate and write this up.

PR #3458, opened 2026-07-10
fix(integrations): escape Rich markup in --integration-options error messages

Quratulain-bilal:
Follow-up hardening to #3457 (the unbalanced-quote crash, since fixed on `main`).

`_parse_integration_options` interpolated user-controlled option tokens straight into `console.print` in two branches:

- the **unexpected value** branch — a token that does not start with `-`;
- the **unknown option** branch — a token that starts with `-` but is not declared.

Because `console.print` parses Rich markup, a token carrying an unbalanced tag (e.g. `[/red]foo` or `--[/red]bad`) raises `rich.errors.MarkupError` and leaks a traceback instead of the clean `typer.Exit(1)` every other bad-input path produces.

**fix:** wrap both token values in `rich.markup.escape(...)` before printing, so malformed markup is shown literally and the command still exits 1.

**scope note:** the `shlex.split` `ValueError` conversion for #3457 is already on `main`; this PR does not re-introduce it. The only production change here is the markup escaping of the two token prints.

Added `test_bad_option_token_with_rich_markup_exits_cleanly`, which asserts a clean `typer.Exit` for both the unexpected-value (`[/red]foo`) and unknown-option (`--[/red]bad`) branches; it fails on the pre-escape code (raw `MarkupError`) and passes with the fix. The existing `test_unbalanced_quote_exits_cleanly` continues to cover the shlex path.

### DIFF
--- a/src/specify_cli/integrations/_helpers.py
+++ b/src/specify_cli/integrations/_helpers.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from typing import Any, Callable
 
 import typer
+from rich.markup import escape
 
 from .._agent_config import SCRIPT_TYPE_CHOICES
 from .._console import console
@@ -206,7 +207,7 @@ def _parse_integration_options(integration: Any, raw_options: str) -> dict[str, 
     while i < len(tokens):
         token = tokens[i]
         if not token.startswith("-"):
-            console.print(f"[red]Error:[/red] Unexpected integration option value '{token}'.")
+            console.print(f"[red]Error:[/red] Unexpected integration option value '{escape(token)}'.")
             if allowed:
                 console.print(f"Allowed options: {allowed}")
             raise typer.Exit(1)
@@ -217,7 +218,7 @@ def _parse_integration_options(integration: Any, raw_options: str) -> dict[str, 
             name, value = name.split("=", 1)
         opt = declared.get(name)
         if not opt:
-            console.print(f"[red]Error:[/red] Unknown integration option '{token}'.")
+            console.print(f"[red]Error:[/red] Unknown integration option '{escape(token)}'.")
             if allowed:
                 console.print(f"Allowed options: {allowed}")
             raise typer.Exit(1)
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

Quratulain-bilal:
escaped the value before interpolating it in aa810ba — you're right that `--commands-dir "[/red]foo` tripped the intended ValueError but then the error print itself raised MarkupError on the unbalanced tag and leaked a traceback anyway.

swept the two sibling branches in the same function too (the 'unexpected option value' and 'unknown option' prints both interpolate the raw token), since a bare `[/red]foo` token would hit the same wall without an unbalanced quote.

regression covers both: an unbalanced-quote value carrying markup, and a plain markup token. confirmed both raise MarkupError on the pre-fix code and exit cleanly with the escape.

mnriem:
Please resolve conflicts

mnriem:

Quratulain-bilal:
Addressed the Copilot feedback. The flagged test (`test_malformed_quoting_with_rich_markup_exits_cleanly`) had an inaccurate docstring — the shlex failure branch prints only `{exc}`, which never carries the caller's markup, so that path was already clean via the existing `test_unbalanced_quote_exits_cleanly`. Its `[/red]foo` assertion also duplicated `test_bad_option_token_with_rich_markup_exits_cleanly`. Removed the redundant test; the substantive change remains escaping the two user-controlled token prints (the unexpected-value and unknown-option branches) so an unbalanced Rich tag exits cleanly instead of raising MarkupError.

Quratulain-bilal:
Updated the PR title and description to match the actual scope. You're right that the `shlex.split` `ValueError` conversion for #3457 is already on `main` and unchanged in this diff — the only production behavior here is escaping the two user-controlled token prints (unexpected-value and unknown-option branches) so an unbalanced Rich tag exits cleanly instead of raising `MarkupError`. The title now says "escape Rich markup in --integration-options error messages" and the body frames this as follow-up hardening to the already-fixed #3457 rather than re-adding that fix.

mnriem (approved):

mnriem:
Thank you!

PR #3463, opened 2026-07-10
docs: clarify that release tags keep the leading v prefix

emrdgrmnci:
## Summary
- Clarify install docs so readers keep the leading `v` when substituting release tags (e.g. `v0.12.11`, not `0.12.11`).
- Update the same guidance in README, the installation guide, and the pipx/uvx install pages to avoid the common `couldn't find remote ref` failure.

## Test plan
- [ ] Confirm [Releases](https://github.com/github/spec-kit/releases) tags are named `vX.Y.Z`.
- [ ] Skim README + docs/installation.md wording for clarity.
- [ ] Optionally verify `uv tool install specify-cli --from git+https://github.com/github/spec-kit.git@v0.12.11` succeeds and `@0.12.11` fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Spec-Driven Development **flips the script** on traditional software development
 
 ### 1. Install Specify CLI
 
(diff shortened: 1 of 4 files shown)


mnriem (approved):

mnriem:
Thank you!

PR #3468, opened 2026-07-11
fix(workflows): evaluate 'in'/'not in' safely on a non-iterable right operand (#3447)

### LINKED ISSUE #3447: workflow 'in' operator crashes with raw TypeError on a non-iterable right operand
### description

the `in` / `not in` operators in `_evaluate_simple_expression` (src/specify_cli/workflows/expressions.py) only guard `right is not None`:

```python
if op == " in ":
    return left in right if right is not None else False
```

but `left in right` also raises `TypeError` for any other non-iterable right operand (int, bool, float). so a workflow condition like `{{ inputs.tag in inputs.count }}` where `count` is a number leaks a raw `TypeError: argument of type 'int' is not iterable` and crashes the whole run, instead of evaluating like the None case does.

this is asymmetric with `_safe_compare`, which already swallows `TypeError` and returns False for the ordering operators (`<`, `>`, etc.).

repro:

```python
from specify_cli.workflows.expressions import evaluate_condition
evaluate_condition("{{ inputs.tag in inputs.count }}", StepContext(inputs={"tag":"x","count":5}))
# TypeError: argument of type 'int' is not iterable
```

expected: `False` (nothing is contained in a non-iterable), same as the `right is None` branch right beside it.

Noor-ul-ain001:
## Summary

Closes #3447. The `in` / `not in` operators in `_evaluate_simple_expression` (`src/specify_cli/workflows/expressions.py`) only guarded `right is not None`:

```python
if op == " in ":
    return left in right if right is not None else False
```

But `left in right` also raises `TypeError` for any other non-iterable right operand (int, bool, float). So a workflow condition like `{{ inputs.tag in inputs.count }}` where `count` is a number leaked a raw `TypeError: argument of type 'int' is not iterable` and crashed the whole run, instead of evaluating like the None case right beside it.

This is asymmetric with `_safe_compare`, which already swallows `TypeError` and returns `False` for the ordering operators (`<`, `>`, etc.).

**Before:**
```python
evaluate_condition("{{ inputs.tag in inputs.count }}", StepContext(inputs={"tag":"x","count":5}))
# TypeError: argument of type 'int' is not iterable  (crashes the run)
```

**After:** evaluates to `False` (nothing is contained in a non-container), same as the `right is None` branch.

## Changes

- Add a `_safe_contains(left, right)` helper mirroring `_safe_compare`: a `None` or non-container `right` means "nothing is contained", so `in` → `False` and `not in` → `True` rather than a raw `TypeError`.
- Route both membership branches through it (`not in` negates the result), so the two operators can't drift.

## Testing

- Added `test_in_operator_non_iterable_right_operand`: asserts `in`/`not in` against int/bool/float/None right operands don't crash and return the None-branch result, and that genuine containment against iterables still works.
- Test-the-test: confirmed the new test fails without the source change (raw `TypeError` propagates).
- `pytest tests/test_workflows.py` → 395 passed, 1 skipped. (The 11 `TestWorkflow*SymlinkGuard` failures are pre-existing on Windows without elevation and unrelated to this change.)


mnriem (changes requested):

Noor-ul-ain001:
Thanks for the review. Addressed the Copilot feedback:

**On the redundant refactor (main finding):** You are right — #3447 was fixed independently by #3448, which merged first and added the same `_safe_membership` helper this branch introduced. I have reverted the `_safe_contains` rename so `expressions.py` now matches `main` exactly (empty diff). The membership guard that fixes #3447 already lives on `main`.

**On the duplicate test:** Removed `test_in_operator_non_iterable_right_operand` and folded its only genuinely-new coverage into the existing `test_membership_against_non_iterable_is_false_not_error` — the base test only exercised `not in` for the `int` case, so I extended it to loop `not in` over the `float`/`bool`/`None` right operands too. No second regression test.

**On rebase:** Merged the latest `upstream/main` into the branch.

The remaining diff is now just that small test-hardening addition on top of the already-merged fix. The earlier docstring/indentation comments were resolved in prior commits. `pytest tests/test_workflows.py::TestExpressions` → 42 passed.

mnriem (approved):

mnriem:
Thank you!

PR #3470, opened 2026-07-11
fix(workflows): reject bool max_iterations in engine re-eval guard

Noor-ul-ain001:
## Summary

The `while` / `do-while` step validators already reject `max_iterations: true` as a type error — `bool` is a subclass of `int`, so a boolean passes a bare `isinstance(..., int)` check, and since `True - 1 == 0` it would silently cap the loop at a **single iteration** (0 re-iterations).

The engine's re-evaluation guard in `WorkflowEngine`, however, still used the older shape:

```python
if not isinstance(max_iters, int) or max_iters < 1:
    max_iters = 10
```

On an **unvalidated** run (executing a definition without first calling the validators), a boolean `max_iterations` slipped through this guard and ran the loop body just once instead of falling back to the safe default of 10.

## Fix

Add the explicit `isinstance(max_iters, bool)` rejection so the engine guard matches `WhileStep.validate()` / `DoWhileStep.validate()`. An unvalidated run now degrades to the default of 10 rather than silently under-running the loop.

## Testing

- Added `test_while_loop_bool_max_iterations_falls_back_to_default`, which runs a `while` step with `max_iterations: true` through the engine and asserts the body executes 10 times (not 1).
- Verified the test **fails without** the engine change and **passes with** it (test-the-test via `git stash`).
- Full while/do-while/loop suite: 16 passed.


mnriem (changes requested):
Please resolve conflicts

Noor-ul-ain001:
Conflicts resolved — merged the latest `upstream/main` in; the PR now shows mergeable.

While resolving I found the conflict was substantive, not just textual: **`main` already contains this exact fix.** Another change landed the identical engine guard —

```python
if (
    isinstance(max_iters, bool)
    or not isinstance(max_iters, int)
    or max_iters < 1
):
    max_iters = 10
```

— and an equivalent engine-level regression test (`test_loop_with_bool_max_iterations_uses_default_cap`) with the same `while` + `max_iterations: true` workflow asserting the body runs 10 times. My branch had the same code plus a near-duplicate test under a different name (`test_while_loop_bool_max_iterations_falls_back_to_default`); the merge conflict was only in the comment wording and the test name.

I resolved both conflicts by taking `main`, so this branch now has an **empty diff** against `main`. `pytest ...::test_loop_with_bool_max_iterations_uses_default_cap` passes on the merged tree.

Since the fix is already upstream, this PR no longer adds anything and is a reasonable candidate to **close as superseded** — happy to do that, or leave it open if you would prefer. Just flagging so the empty diff is not a surprise.

mnriem:
Closing as superseded as indicated by the comment above

PR #3474, opened 2026-07-11
fix: rewrite extension skill command references

### LINKED ISSUE #3451: [Bug]: Extension command references not rewritten for skills-based integrations (Codex), breaking cross-command invocation
### Bug Description

Extension command markdown is copied **verbatim** into generated agent skills for skills-based integrations (e.g. Codex CLI), so in-body command references keep their slash-command spelling (`/speckit.foo.bar`). In Codex skills mode there is no such dispatchable slash command, so the agent fails or falls back to a markdown-only path.

Reported in discussion #3422 with the Codex CLI + `spec-kit-memory-hub` extension. Example failure the user saw:

```text
Attempted the required memory prep command path (/speckit.memory-md.prepare-context),
but no runnable command was available in this environment, so I followed the
markdown-first fallback (...) before editing.
```

**Root cause (code-confirmed):**

- The skills rendering path `ExtensionManager._register_extension_skills` (`src/specify_cli/extensions/__init__.py`) copies each command file body into `SKILL.md` after only:
  - `CommandRegistrar.resolve_skill_placeholders` — resolves `{SCRIPT}`, `{ARGS}`, `__AGENT__`, project-relative paths only.
  - `integration.post_process_skill_content` — for `SkillsIntegration` this only injects a *note* about hyphenating hook command names; it does not rewrite in-body references.
- The helper that would fix this, `CommandRegistrar._hyphenate_body_refs` (`src/specify_cli/agents.py:94`), already exists but is applied **only** on the Cline extension path (`is_cline_ext`, `agents.py:628-629`). The generic skills path never calls it.

Net effect for Codex: the skill *directory name* and *frontmatter hook refs* get hyphenated (`speckit-memory-md-prepare-context`), but body text keeps `/speckit.memory-md.prepare-context` — wrong separator (`.`) and a `/` prefix that maps to nothing in skills mode.

There is also **no extension-authoring guidance** on command-reference portability: `extensions/EXTENSION-USER-GUIDE.md` itself uniformly uses `/speckit.foo.bar` notation and never describes skills-mode rendering.

### Steps to Reproduce

1. `specify init myproj --integration codex` (skills-based integration).
2. Install an extension whose command bodies reference other commands in slash-dot form, e.g. `spec-kit-memory-hub`:
   `specify extension add https://github.com/DyanGalih/spec-kit-memory-hub`
3. Inspect a generated skill: `cat .agents/skills/speckit-memory-md-prepare-context/SKILL.md`
4. Observe the body still contains `/speckit.memory-md.prepare-context`.
5. Ask Codex to run that skill; it reports the command path is unavailable and falls back.

### Expected Behavior

When rendering extension commands into skills, spec-kit should normalize in-body command references to the target integration's invocation form (hyphenated skill name, no bare slash-dot spelling) — the same way it already does for Cline and for frontmatter hook refs. Existing extensions written in slash-dot notation should "just work" for Codex without per-agent templates.

### Actual Behavior

Body command references are copied verbatim (`/speckit.foo.bar`), which is not dispatchable in skills mode, so the agent errors or silently falls back to a markdown path.

### Specify CLI Version

Reproduced against current `main` (source inspection); reporter on discussion #3422 did not pin a version.

### AI Agent

Codex CLI

### Operating System

Any (logic is OS-independent)

### Python Version

Any supported (logic is OS/Python-independent)

### Error Logs

```text
Attempted the required memory prep command path (/speckit.memory-md.prepare-context),
but no runnable command was available in this environment, so I followed the
markdown-first fallback (./docs/memory/INDEX.md + feature memory-synthesis.md) before editing.
```

### Additional Context

From discussion #3422 (https://github.com/github/spec-kit/discussions/3422).

Proposed fix (two parts, ideally both):

1. **Code:** Apply `_hyphenate_body_refs` in the skills rendering path (`_register_extension_skills`), and normalize `/speckit.x` → the integration's invocation form (reuse `SkillsIntegration.build_command_invocation`). This makes existing extensions work for Codex without author changes.
2. **Docs:** Add an extension-authoring portability note — prefer the `__SPECKIT_COMMAND_*__` token / neutral wording over hard-coded `/speckit.…` references so commands render correctly across slash-command and skills-based integrations.

NgoQuocViet2001:
## Description

- rewrite standalone portable slash-dot references in extension skill bodies using the active agent's native invocation syntax
- share the invocation renderer with hook execution so Codex, Kimi, Cline, and slash-based integrations stay consistent
- preserve URL, filesystem-path, Markdown-link, HTML-attribute, and query-path references
- document the portable extension command-reference behavior

Closes #3451

## Testing

Additional validation:
- `uv run pytest tests/test_extension_skills.py tests/test_extensions.py tests/integrations/test_base.py tests/integrations/test_integration_kimi.py tests/integrations/test_integration_zcode.py -q` (527 passed, 14 skipped)
- `uvx ruff check src/specify_cli/extensions/__init__.py tests/test_extension_skills.py`
- `git diff --check`

The sample-project coverage installs an extension into temporary Codex, Kimi, and Claude projects and verifies both rewritten invocations and preserved path/link references.

## AI Disclosure

Codex (GPT-5, autonomous) analyzed the issue, implemented and iteratively tested the change on behalf of @NgoQuocViet2001. The operator requested this contribution but did not manually author or line-by-line review the patch.

### DIFF
--- a/extensions/EXTENSION-USER-GUIDE.md
+++ b/extensions/EXTENSION-USER-GUIDE.md
@@ -202,6 +202,8 @@ Jira Integration (v1.0.0)
 
 When an extension is removed, its corresponding skills are also cleaned up automatically. Pre-existing skills that were manually customized are never overwritten.
 
+Extension command bodies can keep portable slash-dot references such as `/speckit.jira.sync-status`. During skill registration, spec-kit rewrites standalone command references to the active agent's invocation style—for example, `$speckit-jira-sync-status` for Codex, `/skill:speckit-jira-sync-status` for Kimi, or `/speckit-jira-sync-status` for slash-based skills integrations. URL and path-like references are left unchanged.
+
 ---
 
 ## Using Extensions
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -49,12 +49,74 @@ _FALLBACK_CORE_COMMAND_NAMES = frozenset(
     }
 )
 EXTENSION_COMMAND_NAME_PATTERN = re.compile(r"^speckit\.([a-z0-9-]+)\.([a-z0-9-]+)$")
+_SPECKIT_SLASH_REF_PATTERN = re.compile(
+    r"(?<!\]\()(?<![.\w:/\\])"
+    r"/(speckit\.[A-Za-z0-9_-]+(?:\.[A-Za-z0-9_-]+)*)\b"
+    r"(?![\w/\\?#-]|\.[A-Za-z0-9_-])"
+)
+_MARKDOWN_LINK_DESTINATION_PREFIX_PATTERN = re.compile(r"\]\(\s*$")
+_HTML_LINK_ATTRIBUTE_PREFIX_PATTERN = re.compile(
+    r"\b(?:href|src)\s*=\s*['\"]\s*$", re.IGNORECASE
+)
 
 VALID_EFFECTS = frozenset({"read-only", "read-write"})
 
 DEFAULT_HOOK_PRIORITY = 10
 
 REINSTALL_COMMAND = "uv tool install specify-cli --force --from git+https://github.com/github/spec-kit.git"
+
+
+def _skill_name_from_command(command: Any) -> str:
+    """Map a command id like ``speckit.plan`` to its installed skill name."""
+    if not isinstance(command, str):
+        return ""
+    command_id = command.strip()
+    if not command_id.startswith("speckit."):
+        return ""
+    return f"speckit-{command_id[len('speckit.') :].replace('.', '-')}"
+
+
+def _render_agent_command_invocation(
+    command: Any, selected_ai: Any, ai_skills_enabled: bool
+) -> str:
+    """Render a command using the active agent's invocation syntax."""
+    if not isinstance(command, str):
+        return ""
+
+    command_id = command.strip()
+    if not command_id:
+        return ""
+
+    skill_name = _skill_name_from_command(command_id)
+    if is_dollar_skills_agent(selected_ai, ai_skills_enabled) and skill_name:
+        return f"${skill_name}"
+    if selected_ai == "kimi" and skill_name:
+        return f"/skill:{skill_name}"
+    if selected_ai == "cline":
+        from ..integrations.cline import format_cline_command_name
+
+        return f"/{format_cline_command_name(command_id)}"
+    if is_slash_skills_agent(selected_ai, ai_skills_enabled) and skill_name:
+        return f"/{skill_name}"
+    return f"/{command_id}"
+
+
+def _rewrite_skill_body_command_refs(
+    body: str, selected_ai: str, ai_skills_enabled: bool
+) -> str:
+    """Rewrite slash-dot command references for an agent skill body."""
+    def replace(match: re.Match[str]) -> str:
+        line_start = body.rfind("\n", 0, match.start()) + 1
+        line_prefix = body[line_start : match.start()]
+        if _MARKDOWN_LINK_DESTINATION_PREFIX_PATTERN.search(
+            line_prefix
+        ) or _HTML_LINK_ATTRIBUTE_PREFIX_PATTERN.search(line_prefix):
+            return match.group(0)
+        return _render_agent_command_invocation(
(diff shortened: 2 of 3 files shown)


mnriem:
Thanks for the detailed writeup and the PR — the diagnosis of where skills rendering diverges is accurate and genuinely useful. I'd like to reframe the fix, though, because I think the root cause is different from "command markdown is copied verbatim."

Spec Kit already ships a first-class, agent-neutral way to reference sibling commands from a command body: the `__SPECKIT_COMMAND_<NAME>__` token, resolved by `IntegrationBase.resolve_command_refs()`. Our own first-party bundled extensions use it exclusively — see `extensions/bug/` and `extensions/git/`, which reference each other via `__SPECKIT_COMMAND_BUG_FIX__`, `__SPECKIT_COMMAND_SPECIFY__`, etc. A command file is a *template* that gets rendered per agent; the neutral dialect (`{SCRIPT}`, `{ARGS}`, `__AGENT__`, `__SPECKIT_COMMAND_*__`) is how authors stay portable, and rendering it into each agent's native invocation is Spec Kit's job.

`spec-kit-memory-hub` hard-coded `/speckit.memory-md.prepare-context` — a literal in one agent's surface syntax — instead of the provided token. So I don't think the right fix is to teach the renderer to detect and rewrite arbitrary literal slash-dot strings. That requires heuristically telling a command reference apart from a URL, filesystem path, Markdown link, HTML attribute, and query path — which is exactly what the PR's negative-lookahead regex tries to do, and it still mis-handles inline code spans, fenced code blocks, autolinks (`</speckit.foo.bar>`), and reference-style link definitions. The token has none of this ambiguity because intent is explicit.

That said, this surfaced two legitimate defects on our side, plus a docs gap. I'd like to fix all three, phased:

**1. Discoverability (docs only).** `EXTENSION-USER-GUIDE.md` never documents `__SPECKIT_COMMAND_*__` and its own examples use `/speckit.foo.bar` throughout — so an author has no signal the token exists. We'll document it as the required way to write cross-command references and fix the guide's examples to model it. *(Independent, shippable on its own.)*

**2. Enable the render path and define the default skill render.** `_register_extension_skills` only calls `resolve_skill_placeholders` (handles `{SCRIPT}`/`{ARGS}`/`__AGENT__`/paths); it never resolves `__SPECKIT_COMMAND_*__`. This work wires token resolution into that path and establishes the *default* invocation for a skills agent — the generic slash-skill form `/speckit-<name>` — as the baseline every skills integration inherits. This is independently correct: it immediately fixes every default slash-skills agent (claude, copilot, cursor, …) and doesn't regress Codex.

**3. Codex-specific render override.** On top of that default, some agents override the invocation form: Codex renders `$speckit-<name>` (Kimi `/skill:...`, Cline its own form) rather than the default `/speckit-<name>`. The invocation logic already exists in `_render_hook_invocation` / `_invocation_style.py` — this hooks that per-agent override into the same renderer so the default from #2 resolves to Codex's dispatchable form.

Plan: **three tracked issues, three PRs.** #1 (docs) is fully independent. #2 wires token resolution into the skill path with the default `/speckit-<name>` render — that's independently correct and immediately fixes every default slash-skills agent without regressing Codex. #3 layers the Codex-specific `$speckit-<name>` override (and Kimi/Cline) on top of #2's default. #3 sequences after #2, but each PR lands a self-contained, correct increment, so they don't need to be bundled. This supersedes the regex-rewrite approach here, so I'd propose closing this PR in favor of that work.

You clearly did the hard part already — tracing the divergence and the render paths — and I'd love to see you address these if you have the bandwidth. If you're up for it, any of the three (starting with the docs fix, then #2, then #3) would be a great contribution, and I'm happy to review and help along the way. No pressure if the timing doesn't work — just let me know either way.

The outcome is the same thing you're after — existing-style extensions render correctly for Codex — but achieved by making Spec Kit's own portability mechanism work end-to-end rather than by guessing intent from un-templated literals.

NgoQuocViet2001:
The token-based portability model is the correct boundary here. I’m closing this PR rather than extending the heuristic rewrite.

I’m available to take the three follow-ups in the proposed order—documentation, default skill token rendering, then agent-specific overrides—and will keep each change independently shippable. I’ll wait for the tracking issues, or can open them if you prefer.

AI disclosure: this response was drafted with Codex assistance.

PR #3477, opened 2026-07-11
fix(integrations): declare kiro-cli multi-install safe

### LINKED ISSUE #3471: [Bug]: kiro-cli integration should be declared multi_install_safe
### Bug Description

  The `kiro-cli` integration uses fully isolated paths (`.kiro/prompts/`) and meets all documented criteria for multi-install safety, but is not declared
  `multi_install_safe = True` on the `KiroCliIntegration` class. Co-installing `kiro-cli` alongside any other integration (e.g. `claude`) produces a permanent ERROR
  from `specify integration status` with no way to acknowledge or resolve it.

### Steps to Reproduce

  1. Initialize a Spec Kit project: `specify init`
  2. Install claude integration: `specify integration install claude`
  3. Install kiro-cli alongside it: `specify integration install kiro-cli --force`
  4. Check status: `specify integration status`

### Expected Behavior

  `specify integration status` should report OK with `Multi-install safe: yes`, since kiro-cli uses isolated paths (`.kiro/prompts/`) that do not collide with any
  other multi-install safe integration.

### Actual Behavior

  `specify integration status` reports ERROR with finding:
  `unsafe-multi-install: Installed integrations are not all declared multi-install safe: kiro-cli`
  
  The `--force` flag on install bypasses the install-time gate but does not resolve the status error. There is no flag or config to suppress it — the error is
  permanent as long as both integrations are installed.

### Specify CLI Version

0.12.11

### AI Agent

Kiro CLI

### Operating System

macOS arm64 (Darwin Kernel Version 25.5.0)

### Python Version

3.12.13

### Error Logs

```shell
Integration status: ERROR
  Default integration: kiro-cli
  Installed integrations: kiro-cli, claude
  Multi-install safe: no
  Shared templates target alignment: kiro-cli
  Modified managed files: 0
  Missing managed files: 0
  Invalid manifest paths: 0
  Unchecked manifests: 0

  Findings:
  - error unsafe-multi-install: Installed integrations are not all declared multi-install safe: kiro-cli
    Suggestion: Use `specify integration use <key>` to change defaults, or `specify integration switch <key>` only when replacing integrations.
```

### Additional Context

Per the [integrations reference](https://github.github.com/spec-kit/reference/integrations.html#which-integrations-are-multi-install-safe), an integration qualifies
  as multi-install safe when it uses:
  
  1. **Isolated agent directory** — ✅ `.kiro/prompts/`
  2. **Dedicated context file that does not collide with another safe integration** — ✅ No overlap with `.claude/`, `.cursor/`, `.augment/`, etc.
  3. **Stable command invocation settings** — ✅ Uses `.` separator, no dynamic paths
  4. **Separate install manifest** — ✅ `.specify/integrations/kiro-cli.manifest.json`
  
  The install manifests confirm zero file overlap between `kiro-cli` (`.kiro/prompts/speckit.*.md`) and `claude` (`.claude/skills/speckit-*/SKILL.md`).
  
  **Proposed fix:** Add `multi_install_safe = True` to `KiroCliIntegration` in `src/specify_cli/integrations/kiro_cli/__init__.py`.
  
  **Workaround:** Patch `multi_install_safe = True` directly in the installed package at
  `~/.local/share/uv/tools/specify-cli/lib/python3.12/site-packages/specify_cli/integrations/kiro_cli/__init__.py`. This workaround is lost on every `specify self
  upgrade`.

Quratulain-bilal:
fixes #3471

kiro-cli uses a fully isolated `.kiro/` root and `.kiro/prompts` command dir, a stable `.` separator with no dynamic paths, and a dedicated install manifest — it meets every documented multi-install-safe criterion. but `multi_install_safe` was never set on `KiroCliIntegration`, so co-installing it (e.g. alongside claude) left `specify integration status` permanently in ERROR (`unsafe-multi-install`) with no way to resolve it. `--force` bypasses the install-time gate but not the status finding.

**fix:** set `multi_install_safe = True` on `KiroCliIntegration`.

i verified the isolation rather than trusting the flag: the registry contract tests in `tests/integrations/test_registry.py` auto-parametrize over every declared-safe integration and check pairwise-disjoint agent roots, command dirs, and (via a real init+install) install manifests. with kiro-cli added they still all pass (365 tests), which empirically proves `.kiro/` doesn't collide with any other safe integration.

also added an explicit `multi_install_safe is True` assertion in the kiro-cli test file (fails on the pre-fix code) and listed kiro-cli in the multi-install-safe docs table.

### DIFF
--- a/docs/reference/integrations.md
+++ b/docs/reference/integrations.md
@@ -271,6 +271,7 @@ The currently declared multi-install safe integrations are:
 | `grok` | `.grok/skills` |
 | `junie` | `.junie/commands`, `.junie/AGENTS.md` |
 | `kilocode` | `.kilocode/workflows`, `.kilocode/rules/specify-rules.md` |
+| `kiro-cli` | `.kiro/prompts` |
 | `qodercli` | `.qoder/commands`, `QODER.md` |
 | `qwen` | `.qwen/commands`, `QWEN.md` |
 | `shai` | `.shai/commands`, `SHAI.md` |
--- a/docs/reference/integrations.md
+++ b/docs/reference/integrations.md
@@ -250,34 +250,32 @@ Spec Kit tracks one default integration in `.specify/integration.json` with `def
 
 ### Which integrations are multi-install safe?
 
-An integration is multi-install safe when it uses a static, unique agent root and command directory, stable command invocation settings, and a separate install manifest whose managed files do not overlap another safe integration. Registry tests enforce those path and manifest invariants. Shared Spec Kit templates remain aligned to the single default integration.
+An integration is multi-install safe when it uses a static, unique command directory, stable command invocation settings, and a separate install manifest whose managed files do not overlap another safe integration. Registry tests enforce those path and manifest invariants. Shared Spec Kit templates remain aligned to the single default integration.
 
-The Isolation column below lists paths Spec Kit manages for that integration (skills/commands roots and any integration-owned rule files). It is not a full inventory of every file an agent may read.
-
-**Agent-context defaults are separate.** The optional agent-context extension maps each integration to a default context file in `extensions/agent-context/agent-context-defaults.json`. Those defaults are independent of multi-install safety: several agents may share a root file such as `AGENTS.md` when the extension is enabled. Multi-install safety does not require a unique context file per safe integration.
+The Command directory column below lists the directory each integration installs its commands or skills into. The optional agent-context extension is a separate concern — it seeds a per-agent context file (for example `AGENTS.md` or `CLAUDE.md`) and is *not* multi-install safe, since several agents can map to the same context file; see the agent-context extension for details.
 
 The currently declared multi-install safe integrations are:
 
-| Key | Isolation |
-| --- | --------- |
-| `auggie` | `.augment/commands`, `.augment/rules/specify-rules.md` |
-| `claude` | `.claude/skills`, `CLAUDE.md` |
-| `cline` | `.clinerules/workflows`, `.clinerules/specify-rules.md` |
-| `codebuddy` | `.codebuddy/commands`, `CODEBUDDY.md` |
-| `codex` | `.agents/skills`, `AGENTS.md` |
-| `cursor-agent` | `.cursor/skills`, `.cursor/rules/specify-rules.mdc` |
-| `firebender` | `.firebender/commands`, `.firebender/rules/specify-rules.mdc` |
-| `gemini` | `.gemini/commands`, `GEMINI.md` |
+| Key | Command directory |
+| --- | ----------------- |
+| `auggie` | `.augment/commands` |
+| `claude` | `.claude/skills` |
+| `cline` | `.clinerules/workflows` |
+| `codebuddy` | `.codebuddy/commands` |
+| `codex` | `.agents/skills` |
+| `cursor-agent` | `.cursor/skills` |
+| `firebender` | `.firebender/commands` |
+| `gemini` | `.gemini/commands` |
 | `grok` | `.grok/skills` |
-| `junie` | `.junie/commands`, `.junie/AGENTS.md` |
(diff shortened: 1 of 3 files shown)


mnriem (changes requested):

Quratulain-bilal:
thanks, you're right that the docs read like a contradiction — fixed the wording in 050446f.

digging in though, it's not a real collision. two separate layers were getting conflated:

- the `multi_install_safe` flag governs only the install footprint: agent root, command dir, and manifest. kiro-cli's are `.kiro/` + `.kiro/prompts` + its own manifest, disjoint from codex's `.agents/` + `.agents/skills`. the registry contract tests (`test_registry.py`) enforce this pairwise across every safe key — adding the flag pulled kiro-cli in and all 389 still pass.
- neither codex nor kiro-cli writes `AGENTS.md` as part of install. that mapping lives in `agent-context-defaults.json` and is owned by the optional agent-context extension.

and that extension can't actually make two safe integrations write the same file: in `update-agent-context.sh` it reads a single key — `opts.get('integration') or opts.get('ai')` from init-options.json — and seeds the context file for that one default integration only, never for every installed integration at once. so codex + kiro-cli co-installed still only seeds whichever one is the default.

so the flag is correct as-is; the docs just described a constraint (`dedicated context file`) that the flag doesn't govern. dropped that clause from the definition and added a note explaining what the Isolation column and the flag actually cover.

mnriem:
We need to resolve by remove the agent file mention altogether. By default integrations are now multi-install safe. The agent-context extension is not. Maybe we should call that out in its README. Can you make that change?

mnriem:

Quratulain-bilal:
done in 62faa8b.

- dropped the agent-file entries from the table and renamed the column to "Command directory" — it now lists only what each integration actually manages (its command/skills dir), not the agent-context file. that removes the kiro-cli/codex both-map-to-AGENTS.md contradiction Copilot flagged.
- reworded the definition to say the agent-context extension is a separate concern and is not multi-install safe (several agents can map to the same context file).
- removed the duplicate test — test_declares_multi_install_safe already covers it.

on calling it out in the agent-context extension's README: happy to do that as a small follow-up PR so this one stays focused. want me to?

mnriem:

Quratulain-bilal:
Addressed both remaining Copilot threads in 6eaf46a:

- **Agent-root requirement (L253):** restored "static, unique agent root" alongside the command directory in the definition. You're right that both are part of the contract — `IntegrationBase` documents it (`base.py:133-139`) and `test_registry.py` enforces it via `test_safe_integrations_have_distinct_agent_roots` **and** `test_safe_integrations_have_distinct_command_dirs`. Dropping the root half would have wrongly classified separate subdirs under a shared root as safe.

- **Agent-context framing (L255):** reworded so context-file targeting is described as independent of integration multi-install safety, rather than calling the extension itself "not multi-install safe." `multi_install_safe` is an integration-level declaration about command/skill paths; the agent-context extension is opt-in and can even synchronize multiple anchors via `context_files` (`extensions/agent-context/README.md`), so several agents sharing one context file is expected and orthogonal to the flag.

mnriem:
Please address Copilot feedback. If not applicable, please explain why. If Copilot is going in circles please also let me know ;)

mnriem (approved):

Quratulain-bilal:
Addressed in be4b48e. On this one Copilot was reading a partial view: the `multi_install_safe = True` declaration was **not** removed — the diff only dropped a duplicate near the top of the class, and the flag was still declared once further down. So `get_integration('kiro-cli').multi_install_safe` already resolved to `True` and the `unsafe-multi-install` path was not triggered.

To make that unambiguous in review, I've hoisted the single declaration to the top of `KiroCliIntegration` (right after `key`), where the diff touched, so the opt-in is visible at exactly the spot Copilot flagged. Verified the integration still resolves `multi_install_safe` as `True`; the 24 kiro-cli tests pass.

PR #3482, opened 2026-07-12
fix(workflows): fail fan-in step on non-list wait_for instead of crashing

Noor-ul-ain001:
## Summary

`FanInStep.validate()` and the engine's fan-in wiring checks both reject a non-list `wait_for`, but the engine's `execute()` path does **not** auto-validate — `WorkflowEngine.load_workflow`'s docstring notes the returned definition is "not yet validated; call `validate_workflow()` or `engine.validate()` separately."

On such an unvalidated run, `FanInStep.execute` iterated the raw value (`for step_id in wait_for`), with two bad outcomes depending on the type:

- **Scalar** (`wait_for: 5`, `wait_for: null`) → `TypeError`, which took down the whole run, since the engine calls `step_impl.execute()` with no surrounding try/except.
- **String** (`wait_for: stepA`) → silently iterated the string's *characters*, returning a join of empty results with a `COMPLETED` status:

```python
>>> FanInStep().execute({"id": "f", "wait_for": "stepA"}, ctx).output["results"]
[{}, {}, {}, {}, {}]   # one bogus empty result per character, COMPLETED
```

That silent-wrong case is exactly the "silent empty result + COMPLETED" wiring bug the engine's own fan-in validation comment (`engine.py`) warns against.

## Fix

Guard `execute` to return a `FAILED` `StepResult` naming the type error — mirroring the fan-out step's non-list `items` handling (`test_execute_non_list_items_fails_loudly`) and the switch/`max_iterations`/`timeout`/fan-in-`output` guards. A **missing** `wait_for` key still defaults to an empty list (`COMPLETED`), unchanged; the guard fires only on an explicit non-list value.

Same bug class as #3481 and prior fixes: a validator catches the bad value, but the runtime path crashes (or silently misbehaves) on it when validation is skipped.

## Testing

- Added `TestFanInStep::test_execute_non_list_wait_for_fails_loudly` (parametrized over `str`, `int`, `None`, `dict`).
- Full `tests/test_workflows.py` passes except the 11 pre-existing Windows symlink-guard tests (fail identically on a clean base; require elevation).

### DIFF
--- a/src/specify_cli/workflows/steps/fan_in/__init__.py
+++ b/src/specify_cli/workflows/steps/fan_in/__init__.py
@@ -24,6 +24,24 @@ class FanInStep(StepBase):
         if not isinstance(output_config, dict):
             output_config = {}
 
+        # The engine does not auto-validate step config, so an unvalidated run
+        # with a non-list ``wait_for`` reaches here raw. Iterating it then
+        # either crashes the whole run (a scalar like an int or None raises
+        # TypeError) or, worse, silently iterates a string's characters and
+        # yields a bogus join of empty results with a COMPLETED status — the
+        # exact "silent empty result + COMPLETED" wiring bug the engine's
+        # fan-in validation guards against. Fail this step loudly instead,
+        # mirroring the fan-out step's non-list ``items`` handling.
+        if not isinstance(wait_for, list):
+            return StepResult(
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3484, opened 2026-07-12
fix(workflows): raise catalog error, not raw ValueError, on a malformed catalog URL

Noor-ul-ain001:
## Summary

The four catalog URL validators in `src/specify_cli/workflows/catalog.py` accessed `urlparse(url).hostname` unguarded:

- `WorkflowCatalog._validate_catalog_url` (raises `WorkflowValidationError`)
- `StepCatalog._validate_catalog_url` (raises `StepValidationError`)
- the nested fetch-path validators (raise `WorkflowCatalogError` / `StepCatalogError`)

A malformed authority — an unterminated IPv6 bracket `https://[::1` or a bracketed non-IP host `https://[not-an-ip]` — makes `urlparse` / `.hostname` raise `ValueError`.

Each validator's contract is to raise its **domain** error, and the CLI handlers catch only those. So a bad URL leaked a raw `ValueError` traceback instead of the clean message + exit 1 a bad URL should produce:

```
$ specify workflow catalog add "https://[::1"
# before: Traceback ... ValueError: Invalid IPv6 URL
# after:  Error: Catalog URL is malformed: https://[::1   (exit 1)
```

The two **fetch-path** validators also run on the post-redirect `resp.geturl()`, so a hostile/broken redirect target could crash the fetch the same way.

## Fix

Guard each `urlparse` / `.hostname` access with `try/except ValueError -> domain error`, and read `hostname` once and reuse it for the host check — mirroring the fixes already applied to `specify_cli.catalogs` (#3435) and the bundler adapters (#3433). This is the direct `workflows/catalog.py` twin of those; the same bug class as the auth-config fix (#3437).

## Testing

- Added `test_validate_url_malformed_raises_validation_error` to both the `WorkflowCatalog` and `StepCatalog` test classes (parametrized: unterminated IPv6 bracket, bracketed non-IP host).
- Verified end-to-end: `specify workflow catalog add "https://[::1"` now exits 1 with `Error: Catalog URL is malformed: ...` and no traceback.
- Full `tests/test_workflows.py` passes except the 11 pre-existing Windows symlink-guard tests (fail identically on a clean base; require elevation).


mnriem (changes requested):

Noor-ul-ain001:
Thanks for the review. Addressed both Copilot comments — they asked for regression coverage of the two post-redirect validators (`_validate_catalog_url(resp.geturl())` / `_validate_url(resp.geturl())`), which the original tests only exercised at config time.

**Added** `test_fetch_malformed_redirect_target_raises_catalog_error` to both `TestWorkflowCatalog` and `TestStepCatalog`:
- Stub `open_url` to return a response whose `geturl()` is a malformed authority (`https://[::1` for the workflow catalog, `https://[not-an-ip]/x` for the step catalog), while `entry.url` is a valid host — so validation only trips on the redirect target, isolating the post-redirect branch.
- Assert `_fetch_single_catalog(...)` raises `WorkflowCatalogError` / `StepCatalogError` matching `malformed`. Uses `force_refresh=True` on a fresh `project_dir` so no cache masks the error.

**Test-the-test:** both fail on the pre-fix source — the raw `ValueError` there is re-wrapped by the broad `except` as `Failed to fetch catalog ...: Invalid IPv6 URL`, which does not match `malformed` — and pass once the guard converts it to a clean malformed-URL refusal. So they genuinely cover the redirect path, not just the code shape.

Also merged the latest `upstream/main`. `pytest tests/test_workflows.py::TestWorkflowCatalog tests/test_workflows.py::TestStepCatalog` → 40 passed; `ruff check` clean.

mnriem (approved):

PR #3496, opened 2026-07-13
fix(workflows): validate non-mapping `options` in command step

### LINKED ISSUE #3492: fix(workflows): command step crashes with AttributeError on `input: null`
## Summary

The `command` step in the workflow engine accepts an `input:` field that is expected to be a mapping. However, a workflow that sets `input: null` (or omits the value entirely — YAML shorthand for `null`) crashes at execution with a cryptic `AttributeError: 'NoneType' object has no attribute 'items'`, and validation does not catch it.

This is the same class of type-validation gap that has been getting cleaned up in recent commits — e.g. #3327 (shell timeout), #3270 (loop cap ignores `bool max_iterations`), #3448 (membership test on non-iterable). Reporting so the `command` step gets the same treatment.

## Reproduction

Minimal repro against `main` (`0.12.13.dev0`):

```python
from specify_cli.workflows.steps.command import CommandStep

step = CommandStep()
ctx = StepContext(project_root="/tmp")
step.execute(
    {"id": "test", "command": "/speckit.specify", "input": None},
    ctx,
)
```

Output:

```
AttributeError: 'NoneType' object has no attribute 'items'
```

The offending line is `src/specify_cli/workflows/steps/command/__init__.py:37`:

```python
input_data = config.get("input", {})
# ...
for key, value in input_data.items():   # <-- crashes when input is null / list / str
    resolved_input[key] = evaluate_expression(value, context)
```

`CommandStep.validate()` (same file, line ~152) only checks that `command` is present — nothing for `input`. So `WorkflowEngine.validate()` returns no errors and the crash escapes at run time instead of surfacing as a clean validation error at load time.

The natural YAML that triggers this looks harmless:

```yaml
- id: draft
  type: command
  command: /speckit.specify
  input:            # <-- YAML parses this as null
```

Also crashes on `input: "some string"` and `input: [1, 2]` for the same reason (`.items()` on a non-mapping).

## Suggested fix

Match the pattern already applied in the shell / while-loop / do-while validators:

1. In `validate()`: reject non-mapping `input` (allow absence) with a clear error, e.g. `input must be a mapping, got <type>`.
2. In `execute()`: defensively coerce `input_data` to `{}` when it is not a mapping, so a caller that bypasses `WorkflowEngine.validate()` still gets a clean skip rather than a crash.

Happy to open a PR — I have a fix + tests ready.

---

*Disclosure: this issue and the accompanying PR were prepared with Claude (Anthropic) assistance. The reproducer was executed against a fresh clone of `main` and the fix is covered by new unit tests in `tests/test_workflows.py`.*

### LINKED ISSUE #3493: fix(workflows): command step crashes with TypeError on non-mapping `options`
## Summary

The `command` step accepts an `options:` field that is expected to be a mapping of step-level option overrides. If the workflow author sets `options:` to a non-mapping (e.g. a list, a string, or `null`-with-a-truthy-branch-taken), execution crashes with a cryptic `TypeError` or `ValueError` from `dict.update`, and `validate()` does not catch it.

Sibling report to #3492 (same file, related pattern). Also fits the recent `fix(workflows)` cleanup thread — #3327, #3270, #3448.

## Reproduction

Against `main` (`0.12.13.dev0`):

```python
from specify_cli.workflows.steps.command import CommandStep

step = CommandStep()
ctx = StepContext(project_root="/tmp")

# options: [1, 2]
step.execute({"id": "t", "command": "/x", "options": [1, 2]}, ctx)
# -> TypeError: cannot convert dictionary update sequence element #0 to a sequence

# options: "foo"
step.execute({"id": "t", "command": "/x", "options": "foo"}, ctx)
# -> ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

Offending code in `src/specify_cli/workflows/steps/command/__init__.py:50-54`:

```python
options = dict(context.default_options)
step_options = config.get("options", {})
if step_options:                # truthy for non-empty lists / strings
    options.update(step_options)  # crashes when step_options is not a mapping
```

The same pattern applies to `input` (see #3492) — same file, same class of gap.

Natural YAML that triggers this:

```yaml
- id: draft
  type: command
  command: /speckit.plan
  options:
    - approve
    - reject      # authored options as a list; step-level overrides expect a mapping
```

## Suggested fix

Mirror the treatment already given to shell / while-loop / do-while validators:

1. In `validate()`: reject non-mapping `options` (allow absence) with a clear message pointing at the field.
2. In `execute()`: guard `step_options` with `isinstance(step_options, dict)` before the `.update()` so a caller that bypasses `WorkflowEngine.validate()` still fails cleanly rather than raising an internal `dict.update` traceback.

Happy to open a PR — I have a fix + tests ready.

---

*Disclosure: this issue and the accompanying PR were prepared with Claude (Anthropic) assistance. The reproducer was executed against a fresh clone of `main` and the fix is covered by new unit tests in `tests/test_workflows.py`.*

thejesh23:
Fixes #3493.

## What

`CommandStep.execute()` merges step-level `options` into the workflow-level defaults via `dict.update()`, but `CommandStep.validate()` did not check the type of `options`, and the merge itself was gated only on truthiness (`if step_options:`). Any truthy non-mapping value therefore passed validation and crashed at run time:

```
options: [1, 2]   -> TypeError: cannot convert dictionary update sequence element #0 to a sequence
options: "foo"    -> ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

Sibling of #3492 (same file, same class of type-validation gap in the command step).

## Fix

Same shape as the fix already applied in the shell / while-loop / do-while validators and in #3492:

- `validate()` rejects a non-mapping `options` with a message that names the field.
- `execute()` guards the `.update()` with `isinstance(step_options, dict)` so a caller that bypasses `WorkflowEngine.validate()` (e.g. ad-hoc `step.execute(...)`) still fails cleanly and the workflow-level defaults survive.

## Tests

Two new regression tests in `TestCommandStep`:
- `test_validate_rejects_non_mapping_options` — a list and a string both report a clear `'options' must be a mapping` error.
- `test_execute_non_mapping_options_does_not_crash` — the validate-bypass path returns `StepStatus.FAILED` and the workflow-level defaults are preserved unchanged (a malformed step-level override is discarded, not partially applied).

Targeted run:

```
$ .venv/bin/python -m pytest tests/test_workflows.py::TestCommandStep -q
============================== 12 passed in 0.35s ==============================
```

## Manual test results

**Agent**: N/A — workflow-engine-only fix, no slash-command surface changed.  |  **OS/Shell**: macOS 15 / zsh

| Command tested | Notes |
|----------------|-------|
| `/speckit.plan` | Not affected — CLI dispatch surface unchanged. `TestCommandStep::test_options_merge` still passes (well-formed step options are still merged over defaults). |

Only touches `src/specify_cli/workflows/steps/command/__init__.py`; no `templates/commands/*`, no `scripts/bash|powershell/*`, no `src/specify_cli/*` CLI entry point. Per CONTRIBUTING.md's mapping rules, no slash command's behavior changes.

## Relation to #3495

#3495 (input validation, closes #3492) is an independent sibling fix in the same file. Both share the pattern the recent `fix(workflows)` commits have been applying elsewhere. They land cleanly on top of each other; happy to rebase either way.

## AI disclosure

Per `CONTRIBUTING.md#ai-contributions-in-spec-kit`: this PR (bug identification, patch, and tests) was prepared with **Claude (Anthropic) assistance** in an autonomous agent session. Reproducer executed against a fresh clone of `main`; the fix follows the existing type-validation pattern in sibling step validators; the workflow test suite was run green before submission. The change is minimal — one `isinstance` guard in `execute()`, one check in `validate()`, two focused tests.

### DIFF
--- a/src/specify_cli/workflows/steps/command/__init__.py
+++ b/src/specify_cli/workflows/steps/command/__init__.py
@@ -61,15 +61,25 @@ class CommandStep(StepBase):
         if model and isinstance(model, str) and "{{" in model:
             model = evaluate_expression(model, context)
 
-        # Merge options (workflow defaults ← step overrides)
+        # Merge options (workflow defaults ← step overrides). Same rationale as
+        # 'input': a malformed options fails the step *before dispatch* rather
+        # than being silently ignored — a silent skip would let an installed
+        # CLI report ``COMPLETED`` for a malformed step, and ``dict.update``
+        # would still crash on a list.
         options = dict(context.default_options)
         step_options = config.get("options", {})
-        # Same rationale as 'input': a malformed options fails the step rather
-        # than being silently ignored (which would let an invalid step run and
-        # apparently complete).
         if not isinstance(step_options, dict):
             return StepResult(
                 status=StepStatus.FAILED,
+                output={
+                    "command": command,
+                    "integration": integration,
+                    "model": model,
+                    "options": options,
+                    "input": resolved_input,
+                    "dispatched": False,
+                    "exit_code": 1,
+                },
                 error=(
                     f"Command step {config.get('id', '?')!r}: 'options' must be a "
                     f"mapping, got {type(step_options).__name__}."
--- a/tests/test_workflows.py
+++ b/tests/test_workflows.py
@@ -960,11 +960,13 @@ class TestCommandStep:
         step = CommandStep()
         # execute() does input.items() / options.update(); a non-mapping must be
         # reported by validate(), not crash at run time (like switch 'cases').
+        # ``None`` is included because a bare YAML ``options:`` parses as ``None``.
         for bad in (None, "args", ["a", "b"], 5):
             errs = step.validate({"id": "c", "command": "/x", "input": bad})
             assert any("'input' must be a mapping" in e for e in errs), bad
-        errs = step.validate({"id": "c", "command": "/x", "options": 42})
-        assert any("'options' must be a mapping" in e for e in errs)
+        for bad in (None, "foo", [1, 2], 42):
+            errs = step.validate({"id": "c", "command": "/x", "options": bad})
+            assert any("'options' must be a mapping" in e for e in errs), bad
         # a valid mapping config is still accepted
         assert step.validate({"id": "c", "command": "/x", "input": {"args": "y"}, "options": {"k": 1}}) == []
         # execute() has no auto-validation guarantee (the engine may skip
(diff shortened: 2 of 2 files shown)


mnriem (changes requested):
Please resolve conflicts

thejesh23:
Superseded by #3262, which landed the same `validate()`/`execute()` guards for non-mapping `input`/`options` on `main` (commit 52c1acf) — including the identical error message and the "FAILED before dispatch" behavior in `execute()`. The remaining delta on this branch is a minor output-payload shape change on the FAILED `StepResult`, which doesn't warrant a standalone PR. Closing as superseded; #3493 is fixed.

PR #3503, opened 2026-07-13
docs: document __SPECKIT_COMMAND_ token for portable cross-command references

Quratulain-bilal:
phase 1 of the plan @mnriem laid out in #3474: document the `__SPECKIT_COMMAND_<NAME>__` token so extension authors have a signal it exists.

## why

the development guide's "Body (Markdown)" section listed `$ARGUMENTS` and `{SCRIPT}` but never mentioned `__SPECKIT_COMMAND_<NAME>__` — the agent-neutral token that `resolve_command_refs()` (integrations/base.py) renders into each agent's own invocation syntax. with no mention of it, an author naturally hard-codes a literal like `/speckit.my-ext.prepare`, which is correct for one agent and breaks on the rest. that literal-instead-of-token mistake is the root cause behind #3451.

## what

docs-only, in EXTENSION-DEVELOPMENT-GUIDE.md:
- added the token to the placeholder list in "Body (Markdown)"
- added a "Referencing other commands" subsection: why a hard-coded literal isn't portable, the command-name -> token encoding (upper-case, `speckit.` prefix dropped, dotted segments -> underscores), and a worked example
- noted the one limitation i verified: the resolver maps each `_` back to the agent separator, so a token can't carry a hyphen inside a name segment

## verification

every rendered example in the docs was checked against `resolve_command_refs` directly — e.g. `__SPECKIT_COMMAND_BUG_FIX__` -> `/speckit.bug.fix` (slash) and `/speckit-bug-fix` (skills). the token names match the first-party `bug` and `git` extensions, which use this token exclusively.

scope: this is the independent, shippable docs increment. phases 2 (wire token resolution into the skill render path) and 3 (codex `$speckit-` override) from the #3474 plan are separate.

### DIFF
--- a/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
+++ b/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
@@ -252,6 +252,7 @@ Use standard Markdown with special placeholders:
 
 - `$ARGUMENTS`: User-provided arguments
 - `{SCRIPT}`: Replaced with script path during registration
+- `__SPECKIT_COMMAND_<NAME>__`: Replaced with the invocation of another command, rendered in the active agent's syntax (see [Referencing other commands](#referencing-other-commands))
 
 **Example**:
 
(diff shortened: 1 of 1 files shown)


mnriem:
Can you please address Copilot feedback. We will should tackle the documentation change for skill rendering once that lands. Thanks!

Quratulain-bilal:
addressed in 8fb401a. you're right  resolve_command_refs only emits /speckit<separator>... based on the active integration's invoke_separator; the $ and /skill: prefixes come from later skills-output post-processing, not the token resolver itself.

reworded the placeholder-list entry and both paragraphs to describe separator-based rendering, and moved the skills-mode prefix detail (Kimi /skill:, Codex/ZCode $) into a parenthetical example rather than stating it as the token's guaranteed output.

@mnriem agreed on holding the fuller skill-rendering docs until that work lands  happy to follow up with it then.

mnriem (changes requested):

Quratulain-bilal:
Addressed in fc72805. You're right — I verified that `_register_extension_skills` (`src/specify_cli/extensions/__init__.py:978`) resolves placeholders and calls `post_process_skill_content`, but never `resolve_command_refs`, so the `__SPECKIT_COMMAND_<NAME>__` token reaches Codex/ZCode/Kimi verbatim when an extension is registered as skills. Token resolution currently only runs in the command-file rendering path (`CommandRegistrar`, `agents.py:719`).

Rather than overstate portability, I added an explicit **"Current limitation — skills mode"** note right after the example: the token is safe for command-file extensions today, and authors should avoid relying on it inside skill bodies destined for skills-based agents until that rendering step lands. Happy to swap the note for the resolver wiring instead if you'd prefer to close the gap in this PR.

mnriem (approved):

PR #3515, opened 2026-07-14
fix(workflows): fail if/switch steps on non-list branch instead of crashing

Noor-ul-ain001:
## Summary

`IfThenStep.validate()` and `SwitchStep.validate()` already reject a non-list branch (`then`/`else`, and `case`/`default`), but the engine's `execute()` path does not auto-validate — see `WorkflowEngine.load_workflow`, whose docstring notes the definition is "not yet validated". On an unvalidated run the selected branch is fed straight into `next_steps`, which `_execute_steps` iterates as step mappings. A non-list branch (a single mapping or scalar authoring mistake) was iterated element-wise — a dict yields its string keys, a str its characters — and raised `AttributeError` on `.get()`, taking down the whole run; the engine invokes `step_impl.execute()` with no surrounding try/except.

This guards both `execute` paths to return a FAILED `StepResult` naming the type error instead, mirroring the switch non-mapping `cases` guard (#3481) and the fan-out non-list `items` handling.

## Changes

- **`if_then`** — guard both `then` and `else` branches at execute time.
- **`switch`** — guard `case` and `default` branches via a shared `_non_list_branch_failure` helper.
- A missing `else`/`default` still defaults to an empty list (COMPLETED), unchanged; the guard fires only on an explicit non-list value.
- The condition/expression is still evaluated first, so its result is surfaced in the step output for downstream context.

## Tests

Added parametrized tests for non-list `then` and `else` (dict / str / int). Full workflow logic suite passes locally.

Follows the same pattern as #3481 (switch non-mapping cases) and #3482 (fan-in non-list wait_for).

### DIFF
--- a/src/specify_cli/workflows/steps/if_then/__init__.py
+++ b/src/specify_cli/workflows/steps/if_then/__init__.py
@@ -37,7 +37,7 @@ class IfThenStep(StepBase):
         # ``.get()``. ``validate`` already rejects a non-list branch; fail this
         # step loudly on an unvalidated run instead, mirroring the switch/fan-out
         # steps. A missing ``else`` defaults to ``[]`` and stays valid.
-if branch is None and branch_name == "else":
+        if branch is None and branch_name == "else":
             branch = []
         elif not isinstance(branch, list):
             return StepResult(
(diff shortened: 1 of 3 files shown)


mnriem (changes requested):

mnriem:
Please address Copilot feedback and resolve test & lint errors

mnriem:

mnriem (approved):

mnriem:
Thank you!

PR #3519, opened 2026-07-14
fix(workflows): fail while/do-while steps on non-list steps instead of crashing

Noor-ul-ain001:
## Summary

`WhileStep.validate()` and `DoWhileStep.validate()` already reject a non-list `steps` body, but the engine's `execute()` path does not auto-validate — see `WorkflowEngine.load_workflow`, whose docstring notes the definition is "not yet validated". On an unvalidated run the body is returned as `next_steps`, and the engine feeds it straight into `_execute_steps`, which iterates it as step mappings. A non-list `steps` (a single mapping or scalar authoring mistake) was iterated element-wise — a dict yields its string keys, a str its characters — and raised `AttributeError` on `.get()`, taking down the whole run; the engine invokes `step_impl.execute()` with no surrounding try/except.

This guards both `execute` paths to return a FAILED `StepResult` naming the type error instead, mirroring the if/switch non-list-branch guards and the fan-out non-list `items` handling.

## Changes

- **`while`** — guard fires only when the condition is truthy (the only case where the body is dispatched). A false condition leaves a non-list `steps` benign and the step completes, unchanged.
- **`do-while`** — body always dispatches on the first call, so the guard is unconditional.
- The condition/expression is still evaluated first, so its result is surfaced in the step output for downstream context.

## Tests

Added parametrized tests (dict / str / int) for:
- `while` — non-list `steps` fails loudly when condition is true; stays benign/COMPLETED when condition is false.
- `do-while` — non-list `steps` fails loudly (always dispatched).

All while/do-while step tests pass locally.

Follows the same pattern as the merged switch (non-mapping cases), fan-in (non-list `wait_for`), and if/switch non-list-branch crash guards.

### DIFF
--- a/src/specify_cli/workflows/steps/do_while/__init__.py
+++ b/src/specify_cli/workflows/steps/do_while/__init__.py
@@ -27,6 +27,30 @@ class DoWhileStep(StepBase):
         nested_steps = config.get("steps", [])
         condition = config.get("condition", "false")
 
+        # The engine does not auto-validate step config (see
+        # ``WorkflowEngine.load_workflow``) and feeds ``next_steps`` straight
+        # into ``_execute_steps``, which iterates them as step mappings. A
+        # non-list ``steps`` (a single mapping or scalar authoring mistake)
+        # would otherwise be iterated element-wise — a dict yields its string
+        # keys, a str its characters — and crash the whole run with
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3521, opened 2026-07-14
fix(workflows): fan-out max_concurrency .inf falls back to sequential instead of crashing

jawwad-ali:
## What

A fan-out step with `max_concurrency: .inf` crashes the entire workflow run:

```
OverflowError: cannot convert float infinity to integer
```

`WorkflowEngine._run_fan_out` coerces the user-controlled `max_concurrency` with `int()` inside `except (TypeError, ValueError)`:

```python
try:
    workers = max(1, int(max_concurrency))
except (TypeError, ValueError):
    workers = 1
```

But `int(float("inf"))` raises **`OverflowError`**, which is not in that tuple — so `.inf` (a valid YAML float) escapes the guard and takes down the run, contradicting the method's own docstring ("a value that cannot be coerced … runs sequentially").

## Fix

Add `OverflowError` to the except tuple so `.inf` falls back to sequential like any other uncoercible value. (`.nan` already coerced via `ValueError`.) One-line change; direction matches the documented intent and the sibling numeric-coercion guards in this codebase (e.g. `_coerce_input`'s OverflowError handling).

## Test

Extends the existing `test_invalid_max_concurrency_coerces_to_sequential` parametrization with `float("inf")`/`float("nan")` — fails before on `inf` (OverflowError), passes after (sequential, all items in order).

---

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -1197,9 +1197,9 @@ class WorkflowEngine:
         already flipped), so the prefix never drops the actual halting item.
 
         ``max_concurrency`` is coerced with ``int()``; a value that cannot be
-        coerced (``None``, a non-numeric string, …) or that coerces to <= 1 runs
-        sequentially, while a numeric string like ``"4"`` or a float like ``4.0``
-        is honored.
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3538, opened 2026-07-15
fix(extensions/git): reject negative -Number in create-new-feature-branch.ps1

Noor-ul-ain001:
## Summary

The `create-new-feature-branch` script has three twins (bash, PowerShell, Python). The bash and Python twins validate `--number` against `^[0-9]+$` and reject a negative value with:

```
Error: --number must be a non-negative integer
```

The **PowerShell** twin declares the parameter as `[long]$Number`, so PowerShell binds `-5` as the integer `-5` instead of rejecting it. That value then formats via `'{0:000}'` to `-005` and produces a branch name starting with a dash — e.g. `-005-my-feature` — which git refuses (refs cannot begin with `-`). The user gets a confusing late `git` failure instead of the twins' clear, early error.

## Fix

- Guard for `$Number -lt 0` up front (before the empty-description check, matching the bash twin's parse-time validation order) and emit the identical error message.
- An explicit `-Number 0` is still honored (`-lt 0`, not `-le 0`), preserving the #3412 fix.

## Tests

Added matching negative-number parity tests to both the bash and PowerShell create-feature suites, mirroring the existing `test_explicit_number_zero_is_honored` pair.

Verified manually on Windows PowerShell 5.1:
- `-Number -5` → exit 1, `Error: --number must be a non-negative integer`
- `-Number 5` → `005-...` (regression: valid numbers still work)
- `-Number 0` → `000-...` (regression: #3412 behavior preserved)

Same PowerShell-parity bug class as #3412 (`-Number 0`).

### DIFF
--- a/extensions/git/scripts/powershell/create-new-feature-branch.ps1
+++ b/extensions/git/scripts/powershell/create-new-feature-branch.ps1
@@ -41,6 +41,16 @@ if ($Help) {
     exit 0
 }
 
+# -Number is [long], so PowerShell binds "-5" as -5 rather than rejecting it
+# the way the bash/Python twins do (`^[0-9]+$`). A negative value would format
+# via '{0:000}' to e.g. "-005" and produce a branch name starting with "-",
+# which git refuses (refs cannot begin with a dash). Reject it here, before the
+# description check, matching the bash twin's parse-time validation order.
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3552, opened 2026-07-15
fix(workflows): reject non-list input 'enum' instead of crashing

Noor-ul-ain001:
## Problem

A workflow input `enum` is never type-checked. A scalar value crashes the validator and the runtime input resolver with a raw `TypeError` instead of producing a clean error:

```yaml
inputs:
  scope:
    type: string
    enum: 5        # or `true`, `3.14`, ...
```

```
TypeError: argument of type 'int' is not a container or iterable
```

The crash comes from the `value not in enum_values` membership test in `WorkflowEngine._coerce_input`. Two paths are affected:

1. **Authoring time** — `validate_workflow()` documents a "return errors, never raise" contract, but the `TypeError` escapes its `except ValueError` and propagates.
2. **Run time** — `execute()` does not auto-validate the definition (see `load_workflow`'s docstring), so `_resolve_inputs` hits the same crash when a provided value is coerced.

A bare-string `enum` (`enum: "abc"`) is silently worse: `value in "abc"` becomes a substring/character test rather than enum membership, so it passes validation while doing the wrong thing at runtime.

## Fix

Require `enum` to be a list in both places:

- `_coerce_input` (the single runtime choke point) raises a clean `ValueError` on a non-list `enum`.
- The authoring-time inputs loop in `validate_workflow` reports it as a validation error — needed independently because `_coerce_input` is only reached from validation when a `default` is present, and the `integration: auto` case strips `enum` before coercing.

The redundant "invalid default" report is suppressed when `enum` is already flagged, so the operator sees one clear error rather than the same problem re-framed twice.

This is the same unvalidated-config crash class as the existing switch / fan-in / fan-out non-list guards.

## Tests

Added to `tests/test_workflows.py`:
- scalar `enum` (no default) → clean validation error, no crash
- string `enum` → rejected + reported exactly once (no duplicate)
- scalar `enum` at run time via `_resolve_inputs` → clean `ValueError`, not `TypeError`
- valid list `enum` → still validates and still checks the default against membership

Full `test_workflows.py` passes except the pre-existing Windows-without-elevation symlink-guard tests (unrelated to this change).


mnriem (changes requested):
Address Copilot feedback

Noor-ul-ain001:
@
Addressed both Copilot comments in a6ed066.

**1. `integration: auto` runtime path bypassed the guard** (comment on `engine.py:1455`)

`_resolve_inputs` stripped `enum` before coercion whenever `"enum" in input_def`, so a malformed `enum: 5` on the auto-integration input was removed and never reached the shape check — it resolved successfully instead of raising the clean `ValueError`. Now only a *list* `enum` is stripped (`isinstance(input_def.get("enum"), list)`); a scalar/string `enum` stays in so `_coerce_input` rejects it.

**2. Skipping default coercion hid independent default errors** (comment on `engine.py:224`)

When `enum` was malformed, validation skipped the default check entirely, so `type: string` + `default: 5` + `enum: 5` reported only the enum error and silently accepted the wrong-typed default. Now the malformed `enum` is removed only from `validation_input_def` (suppressing the duplicate membership error) while `_coerce_input` still runs, so the defaults type is validated and both errors surface.

**Tests**
- `test_validate_workflow_scalar_enum_still_reports_bad_default_type` — asserts the enum error appears exactly once *and* the wrong-typed default is still reported.
- `test_resolve_inputs_auto_integration_rejects_scalar_enum` — asserts the `auto` sentinel with `enum: 5` raises a clean `ValueError`, not a silent pass.

Both fail on the pre-fix code (test-the-test verified). Full `test_workflows.py` passes except the pre-existing Windows symlink-guard and cross-project-registry tests, which fail on the clean tree too and are unrelated to this change.
@

mnriem:

Noor-ul-ain001:
@
Addressed the remaining Copilot feedback and resolved the merge conflicts.

**Copilot feedback (the `enum: null` null-hole).** Both enum-shape guards resolved the value with `input_def.get("enum")`, which returns `None` for *both* an omitted key and a present `enum:` / `enum: null`. A declared null enum is a non-list too, but the `None`-based check silently accepted it — `validate_workflow` passed and, at run time, `_coerce_input` treated it as "no restriction".

Fixed by distinguishing an omitted key from a present null via `"enum" in input_def` in both places:

```python
enum_present = "enum" in input_def
enum_values = input_def.get("enum")
enum_is_valid = not enum_present or isinstance(enum_values, list)
```

so every declared non-list enum — scalar, string, or null — raises the clean `"enum must be a list"` error, while an omitted enum stays valid. This mirrors the existing present-but-null `requires:` handling in the same function.

**Tests.** Added `test_validate_workflow_rejects_null_enum` (authoring-time), `test_resolve_inputs_rejects_null_enum_at_runtime` (run time, since `execute()` skips validation), and `test_validate_workflow_omitted_enum_is_valid` (guards against over-reach). Test-the-test: with the fix stashed the two null tests fail; the omitted-enum test stays green.

**Conflicts.** Merged `upstream/main` — the base enum-guard commits had already landed there, so the conflicts were in `engine.py` between that older `is not None` form and this null-aware superset; kept the null-aware version and dropped the now-redundant `enum_is_valid` recomputation. `tests/test_workflows.py` auto-merged.

Full `tests/test_workflows.py`: 666 passed, 7 skipped, 20 failed — the 20 are all the pre-existing Windows symlink-guard / cross-project tests, identical on a clean tree and unrelated to this change.
@

Noor-ul-ain001:
Superseded by #3601 (merged), which took the same approach with expanded test coverage.

PR #3577, opened 2026-07-17
fix(extensions,presets): surface clean error on malformed download URL

Noor-ul-ain001:
## Summary

`ExtensionCatalog.download_extension` and `PresetCatalog.download_pack` read a `download_url` from catalog payload data and pass it through `urlparse(...).hostname` during their HTTPS validation. A malformed authority — e.g. an unterminated IPv6 bracket like `https://[::1`, or `https://[not-an-ip]/x` — makes `urlparse`/`.hostname` raise a raw `ValueError`.

That `ValueError` escapes the command handlers, which only catch `ExtensionError` / `PresetError`, so instead of a clean CLI message the user gets an uncaught traceback.

This wraps the parse in a `try/except ValueError` and re-raises as the module's domain error (`ExtensionError` / `PresetError`) with a clear "download URL is malformed" message.

## Bug class

Same shape as the already-fixed unguarded-`urlparse` leaks in:
- catalogs (#3435)
- `workflows/catalog.py` (#3484)

This closes the two remaining sibling code paths (extensions + presets) that had the same latent crash.

## Changes

- `src/specify_cli/extensions/__init__.py` — guard the `download_url` parse in `download_extension`.
- `src/specify_cli/presets/__init__.py` — guard the `download_url` parse in `download_pack`.
- `tests/test_extensions.py` — regression test asserting a malformed URL raises `ExtensionError` (not `ValueError`).
- `tests/test_presets.py` — matching regression test for `download_pack`.

## Testing

- New tests fail on the pre-fix source (raw `ValueError`) and pass with the guard ("test-the-test" verified).
- Full `tests/test_extensions.py` and `tests/test_presets.py` pass (717 passed, 5 skipped).

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -2631,8 +2631,20 @@ class ExtensionCatalog(CatalogStackBase):
         # Validate download URL requires HTTPS (prevent man-in-the-middle attacks)
         from urllib.parse import urlparse
 
-        parsed = urlparse(download_url)
-        is_localhost = parsed.hostname in ("localhost", "127.0.0.1", "::1")
+        # A malformed authority (e.g. an unterminated IPv6 bracket
+        # "https://[::1") makes urlparse / hostname access raise ValueError.
+        # The download_url comes from catalog payload data, so surface a clean
+        # ExtensionError rather than leaking a raw ValueError past the command
+        # handler (which only catches ExtensionError). Mirrors catalogs (#3435)
+        # and workflows/catalog.py (#3484).
+        try:
+            parsed = urlparse(download_url)
+            hostname = parsed.hostname
+        except ValueError:
+            raise ExtensionError(
+                f"Extension download URL is malformed: {download_url}"
+            ) from None
+        is_localhost = hostname in ("localhost", "127.0.0.1", "::1")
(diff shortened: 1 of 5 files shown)


mnriem (changes requested):

Noor-ul-ain001:
@
Addressed the Copilot feedback in 07b5611.

The malformed-URL fix stopped short: `download_pack` now raises a clean `PresetError`, but the `preset_add` handler rendered `{e}` unescaped, so a catalog `download_url` like `https://[not-an-ip]/x` was passed through to Rich as markup and could crash the renderer instead of exiting cleanly.

Changes:
- `presets/_commands.py` — escape `str(e)` in the preset command error handlers (matching the extension handler at `extensions/_commands.py:657`) and hoist the `rich.markup` import to module scope.
- `tests/test_presets.py` — added CLI-level regression tests: a bracketed-host `download_url` exits cleanly, and the compatibility/validation/error handlers escape markup-bearing messages. Both fail on the pre-fix handler (test-the-test verified).

Full `tests/test_presets.py` (367 passed, 2 skipped) and `tests/test_extensions.py` (354 passed, 3 skipped) still pass.
@

mnriem (approved):

mnriem:
Thank you!

PR #3592, opened 2026-07-19
fix(copilot): honor preset command template overrides

### LINKED ISSUE #2761: Copilot speckit.specify does not honor preset spec-template override

faqehanoor:
**## Summary**

**Fixes #2761 by updating the Copilot integration to resolve command templates through the existing preset-aware `PresetResolver`**.

Previously, Copilot always used the bundled shared command templates, causing preset-provided command template overrides (such as `specify.md`) to be ignored. With this change, Copilot now honors preset overrides while preserving the existing fallback behavior when no override is available.

## What changed

- Updated the Copilot setup path to resolve command templates using `PresetResolver.resolve(..., template_type="command")`
- Preserved the existing fallback behavior when no preset override exists
- Added regression tests for:
  - Default fallback behavior
  - Preset override behavior

## Testing

```bash
ruff check src/specify_cli/integrations/copilot/init.py tests/integrations/test_integration_copilot.py

pytest -q tests/integrations/test_integration_copilot.py

### DIFF
--- a/src/specify_cli/integrations/copilot/__init__.py
+++ b/src/specify_cli/integrations/copilot/__init__.py
@@ -350,6 +350,10 @@ class CopilotIntegration(IntegrationBase):
         if not templates:
             return []
 
+        from ...presets import PresetResolver
+
+        preset_resolver = PresetResolver(project_root_resolved)
+
         dest = self.commands_dest(project_root)
         dest_resolved = dest.resolve()
         try:
@@ -367,7 +371,11 @@ class CopilotIntegration(IntegrationBase):
 
         # 1. Process and write command files as .agent.md
         for src_file in templates:
-            raw = src_file.read_text(encoding="utf-8")
+            resolved_template = preset_resolver.resolve(
+                src_file.stem, template_type="command"
+            )
+            source_path = resolved_template or src_file
+            raw = source_path.read_text(encoding="utf-8")
             processed = self.process_template(
                 raw, self.key, script_type, arg_placeholder,
                 project_root=project_root,
--- a/src/specify_cli/integrations/copilot/__init__.py
+++ b/src/specify_cli/integrations/copilot/__init__.py
@@ -372,7 +372,7 @@ class CopilotIntegration(IntegrationBase):
         # 1. Process and write command files as .agent.md
         for src_file in templates:
             resolved_template = preset_resolver.resolve(
-                src_file.stem, template_type="command"
+                f"speckit.{src_file.stem}", template_type="command"
             )
             source_path = resolved_template or src_file
             raw = source_path.read_text(encoding="utf-8")
--- a/tests/integrations/test_integration_copilot.py
+++ b/tests/integrations/test_integration_copilot.py
@@ -188,6 +188,43 @@ class TestCopilotIntegration:
         assert "Copy `.specify/templates/spec-template.md`" not in content
         assert "Load `.specify/templates/spec-template.md`" not in content
 
+    def test_setup_falls_back_to_bundled_command_template_without_preset_override(self, tmp_path):
+        """Copilot should keep using the bundled specify command template when no preset override exists."""
+        from specify_cli.integrations.copilot import CopilotIntegration
+        copilot = CopilotIntegration()
+        m = IntegrationManifest("copilot", tmp_path)
+
+        copilot.setup(tmp_path, m)
+
+        specify_file = tmp_path / ".github" / "agents" / "speckit.specify.agent.md"
+        content = specify_file.read_text(encoding="utf-8")
+        assert "Create or update the feature specification" in content
+        assert "preset override content" not in content
+
+    def test_setup_uses_preset_command_override_when_present(self, tmp_path):
+        """Copilot should prefer a preset-provided command template over the bundled one."""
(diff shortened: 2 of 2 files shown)


mnriem (changes requested):

faqehanoor456:
Thanks for catching this. You're right, the resolver should use the canonical speckit.<command> key so it matches bundled presets and manifest-defined overrides. I'll update the resolver accordingly and adjust the regression test to use the canonical command name instead of the nonstandard commands/specify.md.

faqehanoor:
Thanks for the review! I've addressed the requested change by updating the resolver to use the canonical `speckit.<command>` key and adjusted the regression test to use the canonical command filename (`commands/speckit.specify.md`). I also verified that the existing fallback behavior is preserved. The updated changes have been pushed in commit `0ef16bf`. When you have a chance, I'd appreciate another look. Thanks!


> **## Summary**
> 
> **Fixes #2761 by updating the Copilot integration to resolve command templates through the existing preset-aware `PresetResolver`**.
> 
> Previously, Copilot always used the bundled shared command templates, causing preset-provided command template overrides (such as `specify.md`) to be ignored. With this change, Copilot now honors preset overrides while preserving the existing fallback behavior when no override is available.
> 
> ## What changed
> * Updated the Copilot setup path to resolve command templates using `PresetResolver.resolve(..., template_type="command")`
> * Preserved the existing fallback behavior when no preset override exists
> * Added regression tests for:
>   
>   * Default fallback behavior
>   * Preset override behavior
> 
> ## Testing
> ```shell
> ruff check src/specify_cli/integrations/copilot/init.py tests/integrations/test_integration_copilot.py
> 
> pytest -q tests/integrations/test_integration_copilot.py
> ```

faqehanoor:
Thanks for the review! I've addressed the requested change by updating the resolver to use the canonical `speckit.<command>` key and adjusted the regression test to use the canonical command filename (`commands/speckit.specify.md`). I also verified that the existing fallback behavior is preserved. The updated changes have been pushed in commit `0ef16bf`. When you have a chance, I'd appreciate another look. Thanks!

mnriem (approved):

mnriem:
Thank you!

PR #3595, opened 2026-07-20
fix(workflows): fail gate step loudly on a malformed 'options'

Noor-ul-ain001:
## Summary

`GateStep.validate` rejects a non-list (or empty) `options` and requires every option to be a string, but the workflow engine does **not** auto-validate a step before calling `execute`. On an unvalidated run a malformed `options` slipped straight through to the interactive prompt and took down the whole workflow:

- **scalar / dict / `None` `options`** → raw `TypeError` (`enumerate`/`len` on a non-iterable) or `KeyError` (indexing a dict) inside `_prompt`
- **empty list `[]`** → `_prompt`'s `while True:` input loop can never satisfy `1 <= n <= 0`, so it spins forever
- **non-string element** (e.g. `[123, 456]`) → passes the shape check but crashes the reject test at `choice.lower()` with `AttributeError`

This mirrors an established bug class in this repo: `validate()` correctly rejects bad input, but `execute()` (reachable on an unvalidated run) crashes on it instead of failing the step cleanly — same shape as the switch `cases` (#3481) and command `input` guards.

## Fix

Guard `execute` to return `StepStatus.FAILED` with a clear error when `options` is not a non-empty list of strings. The check runs **before** the non-TTY `PAUSED` short-circuit, so a misconfigured gate surfaces as FAILED in CI rather than pausing and only crashing later when an operator resumes on a real terminal.

## Tests

Added regression tests under `TestGateStep`:
- `test_execute_non_list_options_fails_cleanly` (parametrized over `5`, `{...}`, `None`, `[]`, `"approve"`)
- `test_execute_non_string_options_element_fails_cleanly`
- `test_execute_non_list_options_fails_in_non_tty_too`

Verified test-the-test: with the source fix stashed, the scalar cases crash and the empty-list case hangs `_prompt` forever; with the fix all 21 `TestGateStep` tests pass.

### DIFF
--- a/src/specify_cli/workflows/steps/gate/__init__.py
+++ b/src/specify_cli/workflows/steps/gate/__init__.py
@@ -43,6 +43,35 @@ class GateStep(StepBase):
         options = config.get("options", ["approve", "reject"])
         on_reject = config.get("on_reject", "abort")
 
+        # ``validate`` rejects a non-list (or empty) ``options``, and requires
+        # every option to be a string, but the engine does not auto-validate
+        # before ``execute``. An unvalidated run with a scalar/dict/None
+        # ``options`` would otherwise reach ``_prompt`` and crash the whole run
+        # with a raw ``TypeError`` (``enumerate``/``len`` on a non-iterable) or
+        # ``KeyError`` (indexing a dict); a non-string option would crash at the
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3596, opened 2026-07-20
fix(workflows): reject a non-string 'command' in command-step

Noor-ul-ain001:
## Summary

`CommandStep.validate` only checked that a `command` field is **present**, never its type. But the workflow engine does not auto-validate a step before calling `execute`, so on an unvalidated run a non-string `command` — `null`, a list, an int — was passed straight through `_try_dispatch` to the integration's `build_command_invocation`, which does:

```python
stem = command_name
if stem.startswith("speckit."):   # AttributeError if command_name isn't a str
```

Once a resolvable integration with an installed CLI is found, this crashes the entire workflow run with a raw `AttributeError` instead of failing just the step.

Confirmed directly:
```
['a', 'b'] -> AttributeError 'list' object has no attribute 'startswith'
None       -> AttributeError 'NoneType' object has no attribute 'startswith'
42         -> AttributeError 'int' object has no attribute 'startswith'
```

This is the same class of gap that #3582 closed for the prompt-step `prompt` field — `validate()` checked presence only while `execute()` fed the value to the CLI layer.

## Fix

Guard both paths, matching the sibling steps already in this file:

- **`validate()`** rejects a non-string `command`, mirroring prompt-step `prompt` (#3582) and shell-step `run`.
- **`execute()`** fails the step cleanly with the same contract error *before* dispatch, mirroring the existing `input`/`options` guards in `CommandStep.execute` — so an unvalidated run FAILs the step (honouring `continue_on_error`) instead of crashing the run.

An expression like `{{ inputs.cmd }}` is still a `str`, so it stays valid.

## Tests

Added to `TestCommandStep`:
- `test_validate_rejects_non_string_command` (null / list / int / dict, plus accepts a plain string and an expression)
- `test_execute_non_string_command_fails_cleanly` (patches `shutil.which` so dispatch would actually be attempted absent the guard — proving the guard prevents the crash)

Verified test-the-test: with the source fix stashed, `validate()` misses the bad value and `execute()` crashes at `build_command_invocation` with the raw `AttributeError`; with the fix, all 13 `TestCommandStep` tests pass. The 20 unrelated `test_workflows.py` failures are the pre-existing Windows symlink-guard tests that require elevation.

### DIFF
--- a/src/specify_cli/workflows/steps/command/__init__.py
+++ b/src/specify_cli/workflows/steps/command/__init__.py
@@ -30,6 +30,21 @@ class CommandStep(StepBase):
 
     def execute(self, config: dict[str, Any], context: StepContext) -> StepResult:
         command = config.get("command", "")
+        # validate() rejects a non-string 'command', but the engine does not
+        # auto-validate before execute(); an unvalidated run would pass the value
+        # to build_command_invocation() (via _try_dispatch) and crash there with a
+        # raw AttributeError (command_name.startswith(...) on a list/int/None).
+        # Fail the step with the same contract error instead, mirroring the
+        # 'input'/'options' guards below.
+        if not isinstance(command, str):
+            return StepResult(
+                status=StepStatus.FAILED,
+                error=(
+                    f"Command step {config.get('id', '?')!r}: 'command' must be a "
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3597, opened 2026-07-20
fix(workflows): reject a non-string 'integration'/'model' in command & prompt steps

Noor-ul-ain001:
## Problem

A command or prompt step whose `integration:` (or `model:`) is a non-string crashes the run with a raw `TypeError` instead of a clean validation/step error.

In `execute()`, the resolved `integration` value is passed to `get_integration()`, which does `INTEGRATION_REGISTRY.get(key)` — using the value as a **dict key**. An unhashable value (list/dict) raises `TypeError: unhashable type` right there:

```python
>>> from specify_cli.workflows.steps.command import CommandStep
>>> from specify_cli.workflows.base import StepContext
>>> CommandStep().execute({"id": "x", "command": "speckit.specify", "integration": ["claude"]}, StepContext())
TypeError: cannot use 'list' as a dict key (unhashable type: 'list')
```

Crucially this is **not** limited to unvalidated runs: neither the step's `validate()` nor `validate_workflow` checked the type of `integration`/`model`, so `validate()` returns `[]` for `integration: [claude]` and the crash survives validation. A non-string `model` similarly reaches `build_exec_args()` and would be fed into the CLI argv.

This is the same recurring shape as the existing `command`/`prompt`/`input`/`options` guards — a string-typed dispatch field that `execute()` passes downstream without a type check.

## Fix

Both `command` and `prompt` steps:

- **`validate()`** — reject a literal non-string `integration`/`model`, mirroring the existing sibling type checks. An explicit YAML-null (`integration:` → inherit the workflow default) and a `"{{ ... }}"` expression both stay valid.
- **`execute()`** — guard the *resolved* value (after the `or`-fallback to the workflow default and expression evaluation) and return a clean `FAILED` `StepResult` rather than letting the run crash. This also catches an expression that resolves to a non-string and an unvalidated workflow-level default.

## Tests

Added parametrized `validate()` rejection tests, "accepts None / expression" tests, and `execute()`-fails-loudly tests to both `TestCommandStep` and `TestPromptStep`. Test-the-test: with the fix stashed, the new tests fail — including the raw `TypeError: unhashable type` traceback at `get_integration`. Full `tests/test_workflows.py` shows no new failures (the 20 pre-existing failures are all Windows symlink-guard tests, identical on a clean tree).


mnriem:

Noor-ul-ain001:
@
Addressed the Copilot feedback and resolved the merge conflicts.

**Copilot feedback (falsey non-string bypass).** Both flagged spots resolved the value with `config.get("integration") or context.default_integration` (and the model equivalent), which coerces a *falsey* non-string — `[]`, `{}`, `0`, `False` — into the workflow default **before** the new type guard runs. Since `execute()` is explicitly supported without prior validation, such a value was silently accepted and, with a configured default, could dispatch using the wrong integration/model instead of returning the contract error.

Fixed in both `command` and `prompt` steps by falling back to the workflow default only for a genuinely-unset value (missing / YAML-null / empty string):

```python
integration = config.get("integration")
if integration is None or integration == "":
    integration = context.default_integration
```

so every non-string now reaches the guard.

**Tests.** Added parametrized `execute()` cases over `[], {}, 0, False` to both `TestCommandStep` and `TestPromptStep`, each with `default_integration`/`default_model` set so a regression would dispatch-with-the-default rather than fail-not-possible. Test-the-test: with the source fix stashed, all 8 fail; with the fix, they pass.

**Conflicts.** Merged `upstream/main`; the only conflict was in `tests/test_workflows.py` where both sides added new test methods to `TestCommandStep` (the incoming non-string `command` guard tests and these integration/model tests) — kept both.

Full `tests/test_workflows.py`: 677 passed, 7 skipped, 20 failed — the 20 are all the pre-existing Windows symlink-guard tests, identical on a clean tree and unrelated to this change.
@

mnriem:
Thank you!

PR #3599, opened 2026-07-20
fix: render extension skills for target agent, not just active agent …

### LINKED ISSUE #2948: Extension commands aren't rendered as skills for non-active skills-mode agents on add/install/upgrade
## Summary

Extension commands are not rendered as **skills** for an explicitly installed skills-mode agent unless that agent is the *active* one. They are rendered as command files instead. This affects `extension add`, `integration install`, and `integration upgrade`. Only `integration switch` renders them as skills because it makes the target the active agent first.

This is about parity for integrations that are already being registered in skills mode. It is not asking to change multi-install safety rules or make unsafe multi-install automatic.

## Root cause

`init-options.json` tracks a single active agent (`ai`) and one `ai_skills` flag. There is no per-agent skills tracking. Extension skill rendering is scoped to that active agent:

- `register_enabled_extensions_for_agent` (`src/specify_cli/extensions.py`) computes `skills_mode_active` from `active_agent == agent_name and is_ai_skills_enabled(init_options)`, so a non-active target never takes the skills path.
- `_register_extension_skills` / `_get_skills_dir` resolve the skills dir and frontmatter from `init_options["ai"]` (the active agent), not the target agent.
- `register_commands_for_all_agents` (used by `extension add`) has the same active-agent scoping.

So for any skills-mode agent that is not active, extensions fall back to command files.

## Steps to reproduce

```bash
specify init --here --integration claude
specify extension add git
specify integration install copilot --integration-options "--skills" --force
```

`--force` is intentional here because this reproduction uses an integration that is not declared multi-install-safe today; this issue is about the files produced after the user explicitly opts into that multi-install path.

Inspect `.specify/extensions/.registry` and the filesystem:

- `extensions.git.registered_commands` contains `copilot` (registered as commands).
- `.github/agents/speckit.git.feature.agent.md` exists.
- `.github/skills/speckit-git-feature/SKILL.md` does **not** exist.

Compare with `specify integration switch copilot --integration-options "--skills"`, which *does* produce `.github/skills/speckit-git-feature/SKILL.md` because it activates the target agent before registering.

## Expected

A skills-mode agent should receive extension **skills** matching how its own core commands are scaffolded, when that agent is being explicitly registered or refreshed. This should provide parity with `switch` for `add` / `install` / `upgrade`, without changing which integrations are considered multi-install-safe.

## Proposed direction

Make extension skill rendering agent-aware: derive the skills directory and frontmatter from the target agent's config rather than the active-agent init-options, so `register_commands_for_all_agents` and `register_enabled_extensions_for_agent` can render skills for any skills-mode agent that is already in the registration path.

This is separate from #2886, which only wires `install` / `upgrade` to call `register_enabled_extensions_for_agent` for parity with `switch` in the common active-agent case.

## Scope / acceptance criteria

Keep the fix narrowly scoped to parity with `switch`:

- A target agent installed or upgraded in skills mode receives extension commands as `SKILL.md` files even when another agent remains active. For integrations that are not multi-install-safe, this applies only after the user explicitly opts into that path with `--force`.
- `extension add` applies the same target-agent-aware behavior when registering commands for all detected agents.
- Skills are rendered from the target agent integration config, including skills directory, frontmatter, argument handling, and post-processing, not from `init-options["ai"]`.
- Default command-file registration for non-skills-mode agents remains unchanged.
- Do not change any integration's multi-install safety status or make unsafe multi-install automatic.
- Avoid introducing a broad per-agent state model unless it is required for the narrow parity fix.

For the reproduction above, the target agent should end up with `.github/skills/speckit-git-feature/SKILL.md`, and should not receive `.github/agents/speckit.git.feature.agent.md` for that extension command.

Follow-up to #2886.

Nimraakram22:
Fixes #2948.

## Problem
Extension skill rendering was scoped to the active agent (init-options.json's `ai` field) even when registering extensions for a different, non-active agent — e.g. after `specify integration install copilot --integration-options "--skills"` while Claude remained active, or `specify integration upgrade copilot`. The target agent received command files instead of SKILL.md files.

## Root cause
`ExtensionManager._get_skills_dir()` and `_register_extension_skills()` always resolved the skills directory/settings from the *active* agent in `init-options.json`, ignoring the `agent_name` argument passed to `register_enabled_extensions_for_agent()`.

## Fix
- Added `is_agent_skills_enabled()` to resolve per-agent skills-mode from `.specify/integration.json`'s `integration_settings[agent].parsed_options.skills`, falling back to the legacy global `ai_skills` flag for the active agent.
- Threaded `agent_name` through `_get_skills_dir()` and `_register_extension_skills()`.
- `register_enabled_extensions_for_agent()` now computes `skills_mode_active` per the *target* agent, and always attempts skill registration for that agent (matching prior best-effort error handling).

## Testing
- Manually verified: `specify integration upgrade copilot --force` with a non-active, skills-mode Copilot integration now renders `.github/skills/speckit-git-*/SKILL.md` files instead of command files, with no regression to the active agent's own skills.
- `pytest tests/ -k extension`: 616/616 pass.
- Full suite: 44 pre-existing failures unrelated to this change (Windows symlink-privilege and PowerShell path-resolution issues — confirmed present on `main` without this change).

### DIFF
--- a/src/specify_cli/_init_options.py
+++ b/src/specify_cli/_init_options.py
@@ -34,3 +34,28 @@ def load_init_options(project_path: Path) -> dict[str, Any]:
 def is_ai_skills_enabled(opts: Mapping[str, Any] | None) -> bool:
     """Return True only when init options explicitly enable AI skills."""
     return isinstance(opts, Mapping) and opts.get("ai_skills") is True
+
+
+def is_agent_skills_enabled(project_path: Path, agent_name: str, opts: Mapping[str, Any] | None) -> bool:
+    """Return True when *agent_name* should render extension skills.
+
+    Prefers the per-agent ``skills`` flag recorded in
+    ``.specify/integration.json`` (``integration_settings[agent_name].parsed_options.skills``),
+    which reflects the ``--skills`` option passed to that specific agent's
+    install/upgrade/switch. Falls back to the legacy global
+    ``init-options.json`` ``ai_skills`` flag only when *agent_name* is the
+    active agent recorded there (pre-multi-install behaviour).
+    """
+    from .integration_state import try_read_integration_json, integration_setting
+
+    state, _error = try_read_integration_json(project_path)
+    if state:
+        setting = integration_setting(state, agent_name)
+        parsed = setting.get("parsed_options")
+        if isinstance(parsed, Mapping) and "skills" in parsed:
+            return parsed.get("skills") is True
+
+    if isinstance(opts, Mapping) and opts.get("ai") == agent_name:
+        return is_ai_skills_enabled(opts)
+
+    return False
--- a/src/specify_cli/_init_options.py
+++ b/src/specify_cli/_init_options.py
@@ -46,7 +46,7 @@ def is_agent_skills_enabled(project_path: Path, agent_name: str, opts: Mapping[s
     ``init-options.json`` ``ai_skills`` flag only when *agent_name* is the
     active agent recorded there (pre-multi-install behaviour).
     """
-    from .integration_state import try_read_integration_json, integration_setting
+    from .integration_state import integration_setting, try_read_integration_json
 
     state, _error = try_read_integration_json(project_path)
     if state:
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -1299,19 +1299,19 @@ class ExtensionManager:
         Called during extension removal to clean up skill files that
         were created by ``_register_extension_skills()``.
 
-        If *skills_dir* is not provided and ``_get_skills_dir()`` returns
-        ``None`` (e.g. the user removed init-options.json or toggled
-        ai_skills after installation), we fall back to scanning all known
-        agent skills directories so that orphaned skill directories are
-        still cleaned up.  In that case each candidate directory is
-        verified against the SKILL.md ``metadata.source`` field before
-        removal to avoid accidentally deleting user-created skills with
-        the same name.
+        Skills may have been rendered for more than one agent (#2948),
+        so this always scans every known agent skills directory (plus
+        *skills_dir* or the resolved active-agent directory, if any)
+        rather than only a single resolved directory, to avoid orphaning
+        skills registered for a non-active agent. Each candidate
+        directory is verified against the SKILL.md ``metadata.source``
+        field before removal to avoid accidentally deleting user-created
+        skills with the same name.
 
         Args:
             skill_names: List of skill names to remove.
             extension_id: Extension ID used to verify ownership during
-                fallback candidate scanning.
+                candidate scanning.
             skills_dir: Optional explicit skills directory to use instead
                 of resolving via ``_get_skills_dir()``.  Useful when the
                 caller needs to target a specific agent's skills directory
@@ -1365,8 +1365,13 @@ class ExtensionManager:
                 except (OSError, UnicodeDecodeError, Exception):
                     continue
                 shutil.rmtree(skill_subdir)
-        else:
-            # Fallback: scan all possible agent skills directories
+        # Additionally always scan every other known agent skills
+        # directory: skills may have been rendered for more than one
+        # agent (#2948), so limiting cleanup to only the directory above
+        # would orphan skills registered for a non-active agent.
+        # Directories already handled above are silently skipped since
+        # their matching skill subdirectories no longer exist.
+        if True:
             from .. import AGENT_CONFIG, DEFAULT_SKILLS_DIR
 
             candidate_dirs: set[Path] = set()
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -910,8 +910,15 @@ class ExtensionManager:
 
         return _ignore
 
-    def _get_skills_dir(self) -> Optional[Path]:
-        """Return the active skills directory for extension skill registration.
+    def _get_skills_dir(self, agent_name: Optional[str] = None) -> Optional[Path]:
+        """Return the skills directory for extension skill registration.
+
+        When *agent_name* is given, resolves the skills directory and
+        skills-enabled state for that specific agent (using per-agent
+        settings in ``.specify/integration.json``), so skill rendering
+        works correctly for non-active agents (#2948). When omitted,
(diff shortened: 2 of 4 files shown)


mnriem (changes requested):
Please address Copilot feedback and address test & lint errors

Nimraakram22:
@mnriem Thanks for the review! Pushed a follow-up commit addressing both points:

- **Copilot's `extension add` gap**: `install_from_directory()` now uses a new `_register_extension_skills_for_installed_agents()` helper that renders skills for the active agent *and* every agent in `.specify/integration.json`'s installed integrations that has skills mode enabled — not just the active one. Verified manually: `specify extension add git` with Claude active + Copilot installed in `--skills` mode now reports "5 agent skill(s) auto-registered" and creates SKILL.md files under both `.claude/skills/` and `.github/skills/`.
- **Test coverage**: added `TestNonActiveAgentSkillRegistration` (3 tests) covering the non-active-agent `upgrade` path, isolation from the active agent's own skills, and the `extension add` multi-agent path.

Also had to fix a regression this exposed: my agent-specific `_get_skills_dir()` branch was resolving the naive per-agent directory without the same symlink/containment/is-a-directory safety checks that `resolve_active_skills_dir()` applies for the active agent this broke the Hermes marker-file detection test, now fixed to mirror that logic exactly.

`pytest tests/ -k extension`: 619/619 pass. `git diff --check`: clean.

mnriem:

Nimraakram22:
@mnriem Pushed another commit addressing Copilot's second point (skill cleanup across agents):

- **Orphaned skills on remove**: `_unregister_extension_skills()` previously only scanned every known agent directory as a fallback when the active agent's own skills directory couldn't be resolved at all. Now it always scans every known agent directory (in addition to any explicitly resolved/passed one), so skills rendered for a non-active agent aren't left behind after `extension remove`. Kept `registered_skills` as a flat, deduped list rather than switching to a per-agent dict the existing SKILL.md `metadata.source` ownership check already makes a full-scan-based cleanup safe, and a schema change would be a much larger, riskier diff. Added `test_remove_cleans_up_skills_for_non_active_agent` to cover this.

On the **first point** (making plain `integration install` immediately register extension commands/skills for the newly-installed agent): I implemented this, but it directly contradicts the existing, explicitly-tested design from #2886 see `test_install_defers_extension_commands_until_use` and `test_install_skills_mode_secondary_agent_defers_extension_artifacts` in `tests/integrations/test_integration_subcommand.py`, whose docstrings say *"Maintainer-requested behavior for #2886: extension command back-fill is limited to `integration use`/`switch`/`upgrade`. Plain `install` only adds the integration..."*

Since that looks like a deliberate, previously-approved decision, I didn't want to silently reverse it based on an automated review comment. @mnriem should `install` now also immediately back-fill extension artifacts for a skills-mode secondary agent (superseding that part of #2886), or should the existing deferred-registration behavior stay as-is and #2948 be considered fully addressed by `use`/`switch`/`upgrade`/`extension add` (all already fixed in this PR)? Happy to implement whichever you prefer.

`pytest tests/ -k extension`: 620/620 pass. `git diff --check`: clean.

mnriem:
> @mnriem Pushed another commit addressing Copilot's second point (skill cleanup across agents):
> 
> * **Orphaned skills on remove**: `_unregister_extension_skills()` previously only scanned every known agent directory as a fallback when the active agent's own skills directory couldn't be resolved at all. Now it always scans every known agent directory (in addition to any explicitly resolved/passed one), so skills rendered for a non-active agent aren't left behind after `extension remove`. Kept `registered_skills` as a flat, deduped list rather than switching to a per-agent dict the existing SKILL.md `metadata.source` ownership check already makes a full-scan-based cleanup safe, and a schema change would be a much larger, riskier diff. Added `test_remove_cleans_up_skills_for_non_active_agent` to cover this.
> 
> On the **first point** (making plain `integration install` immediately register extension commands/skills for the newly-installed agent): I implemented this, but it directly contradicts the existing, explicitly-tested design from #2886 see `test_install_defers_extension_commands_until_use` and `test_install_skills_mode_secondary_agent_defers_extension_artifacts` in `tests/integrations/test_integration_subcommand.py`, whose docstrings say _"Maintainer-requested behavior for #2886: extension command back-fill is limited to `integration use`/`switch`/`upgrade`. Plain `install` only adds the integration..."_

The existing behavior should be retained

> Since that looks like a deliberate, previously-approved decision, I didn't want to silently reverse it based on an automated review comment. @mnriem should `install` now also immediately back-fill extension artifacts for a skills-mode secondary agent (superseding that part of #2886), or should the existing deferred-registration behavior stay as-is and #2948 be considered fully addressed by `use`/`switch`/`upgrade`/`extension add` (all already fixed in this PR)? Happy to implement whichever you prefer.
> 
> `pytest tests/ -k extension`: 620/620 pass. `git diff --check`: clean.

mnriem:
Address Copilot feedback and fix test & lint errors. Note that cleanup should never touch any files it does not own. So if a specific integration is uninstalled it should only cleanup those files

Nimraakram22:
@mnriem Pushed a commit addressing this round:

1. **Scoped the skill cleanup fix**: `_unregister_extension_skills()` now takes a `scan_all_agents` parameter (default `False`). Single-agent-scoped callers (`unregister_agent_artifacts`, used when switching away from an agent) keep the exact original behavior — only touch the resolved/passed directory, falling back to a full scan only when no directory resolves at all. Only `remove()` (full extension uninstall) passes `scan_all_agents=True`, since that's the only path that needs to clean up skills across every agent (#2948). Added `test_unregister_agent_artifacts_does_not_touch_other_agents_skills` to lock this in — confirmed switching away from Copilot no longer touches Claude's skills.

2. **`install` deferred-registration behavior retained** as requested — no changes to `_install_commands.py` in this PR.

3. **Fixed the ruff failures**: diffed `ruff check` output between `main` and this branch to isolate exactly what this PR introduced (an unused variable, a couple of unsorted import blocks, and a few `List`/`Optional` type hints in new code that should be `list`/`X | None`). Fixed all of those; this file now has *fewer* ruff errors than `main` (118 vs 119 baseline), not more.

`pytest tests/ -k extension`: 621/621 pass. `git diff --check`: clean.

mnriem:

PR #3601, opened 2026-07-21
fix(workflows): reject non-list input 'enum' instead of crashing

Noor-ul-ain001:
## Problem

A workflow input whose `enum` is not a list slips past validation and crashes at run time:

```yaml
inputs:
  scope:
    type: string
    enum: 5          # scalar instead of a list
```

- **Runtime crash:** `_resolve_inputs` → `_coerce_input` runs `value not in enum_values`, and `x not in 5` raises a raw `TypeError: argument of type 'int' is not iterable`.
- **Broken validator contract:** the same `TypeError` escapes `validate_workflow`'s `except ValueError`, breaking its documented *"return a list of errors, never raise"* contract — so authoring-time validation crashes instead of reporting an error.
- **Silently-wrong string case:** a bare string `enum: "abc"` doesn't crash but turns enum membership into a substring test (`value in "abc"`), silently accepting `"a"`, `"b"`, `"ab"`, etc.

This is the same **unvalidated-`execute()` crash class** as #3482 (fan-in `wait_for`) and #3537 (fan-out step template): `validate()` should reject the value, but `execute()` accepts unvalidated definitions, so the bad value still reaches the engine.

## Fix

- `_coerce_input` now requires `enum` to be a `list` (or `None`), raising a clean `ValueError` with a clear message for any other shape. Both the validation path and runtime resolution now fail fast.
- `validate_workflow` checks the `enum` shape **directly**, not only via the default-coercion path — that path only runs when a `default` is present, so a defaultless bad `enum` would otherwise slip through. A malformed `enum` is stripped before coercing the default so the wrong-typed-default error isn't duplicated as an enum-shape error.
- The `integration: auto` sentinel only strips a **list** `enum` before coercion (enum membership is a runtime concern for `auto`). A non-list `enum` stays in the definition so it's rejected rather than silently exempted.

## Behavior

Before → `TypeError` escapes / silent substring match.
After → `Input 'scope' has invalid 'enum': must be a list, got int.` at both authoring and run time.

## Tests

New coverage across all three layers:
- `_coerce_input` directly (scalar, bool, string, dict; valid list and `None` still work)
- `validate_workflow` with **no default present** (the path the default-coercion check misses)
- runtime `_resolve_inputs`
- the `integration: auto` interaction (non-list `enum` still rejected)

Full `tests/test_workflows.py` passes locally except the pre-existing Windows symlink-guard tests, which require elevation and are unrelated to this change.

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -193,6 +193,20 @@ def validate_workflow(definition: WorkflowDefinition) -> list[str]:
                     f"Must be 'string', 'number', or 'boolean'."
                 )
 
+            # ``enum`` must be a list. Checked here — not only via the
+            # ``_coerce_input`` call below — because that call is reached only
+            # when a ``default`` is present, and the ``integration: auto`` case
+            # strips ``enum`` before coercing; a scalar/string ``enum`` on an
+            # input with no default (or the auto-integration default) would
+            # otherwise slip through here and then crash ``_resolve_inputs`` with
+            # a raw ``TypeError`` at run time. ``None`` means "no enum".
+            enum_values = input_def.get("enum")
+            if enum_values is not None and not isinstance(enum_values, list):
+                errors.append(
+                    f"Input {input_name!r} has invalid 'enum': must be a list, "
+                    f"got {type(enum_values).__name__}."
+                )
+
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3623, opened 2026-07-22
fix(bundler): reject non-list 'catalogs' in bundle-catalogs.yml with a clean error

jawwad-ali:
## What

`_merge_config` (the source-stack loader for `bundle-catalogs.yml`) guards only the falsy case:

```python
catalogs = data.get("catalogs") if isinstance(data, dict) else None
if not catalogs:
    return
for raw in catalogs:   # <-- raw TypeError if `catalogs` is a non-empty scalar
```

A non-empty scalar like `catalogs: 5` (or `catalogs: true`) passes the falsy check and then raises a raw **`TypeError: 'int' object is not iterable`** from the loop — escaping the module's `BundlerError` error contract.

The **sibling reader of the same file** — `commands_impl/catalog_config.py` (used by `specify bundle catalog list`) — already raises an actionable `BundlerError("Malformed catalog config … 'catalogs' must be a list …")` for the identical mis-shape. So the two readers of `bundle-catalogs.yml` diverge.

## Fix

Add the same `isinstance(catalogs, list)` guard, raising the same-shaped `BundlerError`. One file, ~3 lines (`BundlerError` already imported). Matches the shape-guard vein used across `catalogs.py`, `workflows/catalog.py`, presets, and extensions.

## Test

`catalogs: 5` now raises `BundlerError(…must be a list…)` — fails before (raw `TypeError`), passes after. Full `test_catalog_schema.py` (15) passes.

---

### DIFF
--- a/src/specify_cli/bundler/models/catalog.py
+++ b/src/specify_cli/bundler/models/catalog.py
@@ -251,14 +251,18 @@ def _merge_config(by_id: dict[str, CatalogSource], config_path: Path, scope: Sco
         return
     data = load_yaml(config_path)
     catalogs = data.get("catalogs") if isinstance(data, dict) else None
-    if not catalogs:
+    if catalogs is None:
         return
     if not isinstance(catalogs, list):
-        # A non-empty scalar (``catalogs: 5``) passes the falsy check above and
-        # would raise a raw ``TypeError: 'int' object is not iterable`` from the
-        # loop below. Report the same actionable BundlerError the sibling reader
-        # of this file raises (commands_impl/catalog_config.py) so both readers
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

jawwad-ali:
Fixed in 094f1bf: changed the early return to `if catalogs is None` so only an absent value is treated as 'nothing to merge'. Every other non-list value — including the falsy ones (`false`, `0`, `''`, `{}`) — now raises the `BundlerError`, matching the sibling reader. Added parametrized tests for the falsy non-list cases and for the absent/empty-list no-op.

mnriem (approved):

PR #3624, opened 2026-07-22
fix(workflows): gate prompt uses isdecimal() so a superscript digit can't crash it

jawwad-ali:
## What

The interactive `gate` step prompt parses a numeric choice like this:

```python
if raw.isdigit() and 1 <= int(raw) <= len(options):
```

`str.isdigit()` returns `True` for characters `int()` **rejects** — e.g. the superscript `²` (`'²'.isdigit()` is `True`, but `int('²')` raises `ValueError`). So entering `²` at the prompt passes the guard and then crashes the input loop with an uncaught `ValueError`.

## Fix

Use `raw.isdecimal()` — exactly the decimal-digit set `int()` accepts (`Numeric_Type=Decimal`), which excludes superscripts/subscripts while preserving every real decimal digit across scripts. Such input is now treated as an invalid choice and re-prompted. No behavior change for valid input.

## Test

`test_interactive_prompt_rejects_non_decimal_digit`: input `²` then `1` returns the first option — fails before (`ValueError`), passes after.

---

### DIFF
--- a/src/specify_cli/workflows/steps/gate/__init__.py
+++ b/src/specify_cli/workflows/steps/gate/__init__.py
@@ -168,7 +168,11 @@ class GateStep(StepBase):
             except (EOFError, KeyboardInterrupt):
                 print()
                 return options[-1]  # default to last (usually reject)
-            if raw.isdigit() and 1 <= int(raw) <= len(options):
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3625, opened 2026-07-22
docs(core): document the 'py' (Python) --script type

jawwad-ali:
## What

`specify init --script` accepts three values — `sh`, `ps`, **and `py`** — per `init.py` (`--script` help: "sh, ps, or py") and `SCRIPT_TYPE_CHOICES` in `_agent_config.py` (`{"sh", "ps", "py"}`). `specify init … --script py` scaffolds Python scripts.

But the **Core Commands** reference (`docs/reference/core.md`) — the canonical option table — listed only `--script sh|ps`, so a user following the docs never discovers `--script py`.

## Fix

Update the one option row to `--script sh|ps|py` with the Python description. Single-file, one-line docs change, kept to `core.md` to stay single-concern.

---
🤖 Written with the assistance of Claude Code (AI). Gap self-found (flag present in code + `SCRIPT_TYPE_CHOICES`, absent from docs).


mnriem:
Thank you!

PR #3626, opened 2026-07-22
fix(workflows): command/prompt steps fail cleanly on a non-string integration

jawwad-ali:
## What

Both the `command` and `prompt` steps dispatch via `_try_dispatch`, which guarded only the falsy case before the registry lookup:

```python
if not integration_key:
    return None
...
impl = get_integration(integration_key)   # dict lookup
```

A **non-string** `integration` — a list/dict, or an expression like `integration: "{{ steps.pick.output.agents }}"` that resolves to a list — passes the falsy guard and hits `get_integration`'s dict lookup, raising **`TypeError: unhashable type: 'list'`** and aborting the entire workflow run.

## Fix

Widen the guard to also require a `str`:

```python
if not integration_key or not isinstance(integration_key, str):
    return None
```

`None` routes execution to the already-present **FAILED** `StepResult` ("Cannot dispatch … integration … not found"), so the step fails cleanly instead of crashing the run. Unconfigured `integration=None` still returns `None` as before. Applied to both `command` and `prompt` steps.

## Test

`test_execute_non_string_integration_fails_cleanly` (command + prompt): a list integration yields a `FAILED` result — fails before (`TypeError: unhashable type: 'list'`), passes after.

---

### DIFF
--- a/src/specify_cli/workflows/steps/command/__init__.py
+++ b/src/specify_cli/workflows/steps/command/__init__.py
@@ -153,7 +153,11 @@ class CommandStep(StepBase):
         not possible (integration not found, CLI not installed, or
         dispatch not supported).
         """
-        if not integration_key:
+        if not integration_key or not isinstance(integration_key, str):
+            # A non-string integration (a list/dict/expression that resolved to
+            # one) would raise TypeError: unhashable type from get_integration's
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3627, opened 2026-07-22
fix(integrations): validate cached catalog shape before returning it

jawwad-ali:
## What

`IntegrationCatalog`'s fetch validates a **fresh** payload's shape (root must be a JSON object, `integrations` must be a mapping) before use — but the **cache-read** branch returned the cached object directly:

```python
if age < self.CACHE_DURATION:
    return json.loads(cache_file.read_text(encoding="utf-8"))   # no shape check
```

So a poisoned or older-format cache (e.g. `{"integrations": []}`) is returned as-is and later crashes with **`AttributeError: 'list' object has no attribute 'items'`** when the caller iterates `integrations`.

## Fix

Validate the cached object with the same key checks as the fresh-fetch path before returning it. The raised `ValueError` is already caught by the existing surrounding `except (… ValueError …)`, which deletes the corrupt cache files and refetches from source — the same recovery a poisoned network payload already gets.

## Test

`test_poisoned_cache_shape_is_dropped_and_refetched`: a fresh-but-mis-shaped cache (`integrations` as a list) is dropped and the valid source refetched — fails before (`AttributeError`), passes after.

---

### DIFF
--- a/src/specify_cli/integrations/catalog.py
+++ b/src/specify_cli/integrations/catalog.py
@@ -40,6 +40,25 @@ class IntegrationDescriptorError(Exception):
     """Raised when an integration.yml descriptor is invalid."""
 
 
+def _catalog_shape_error(payload: Any) -> Optional[str]:
+    """Return a human-readable reason if *payload* is not a valid integration
+    catalog document, else ``None``.
+
+    Shared by the fresh-fetch and cache-read paths so both enforce the same
+    format contract: a JSON object carrying ``schema_version`` and a mapping
+    ``integrations``. Keeping a single validator prevents the two paths from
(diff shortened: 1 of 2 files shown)


mnriem:

jawwad-ali:
Fixed in 0f362d3: extracted a shared `_catalog_shape_error()` validator and use it in **both** the fresh-fetch and cache-read paths, so they now enforce the same contract (dict + `schema_version` + dict `integrations`). A cache like `{"integrations": {}}` (missing `schema_version`) is now dropped and refetched instead of bypassing validation. Replaced the integration-level test (it was masked by multi-source merging) with a direct unit test of the validator.

mnriem:
Thank you!

PR #3630, opened 2026-07-22
fix(workflows): StepRegistry.add tolerates a corrupted non-dict existing entry

jawwad-ali:
## What

`StepRegistry.add` reads the existing entry then calls `.get()` on it:

```python
existing = self.data["steps"].get(step_id, {})
metadata_to_store["installed_at"] = existing.get("installed_at", ...)   # AttributeError if existing is a str/list/int
```

A corrupted-but-parseable registry — e.g. `{"steps": {"foo": "corrupted"}}`, which `_load()` accepts because it validates only the top-level dict and that `steps` is a dict — makes `add()` raise `AttributeError: 'str' object has no attribute 'get'`.

Its sibling `WorkflowRegistry.add` was hardened for exactly this in #3419 (`existing = raw_existing if isinstance(raw_existing, dict) else {}`); `StepRegistry.add` was left unguarded — a same-file parity gap.

## Fix

Mirror the WorkflowRegistry guard so a non-dict existing entry is treated as absent.

## Test

`test_step_registry_add_survives_non_dict_existing_entry` (a 1:1 copy of the WorkflowRegistry sibling test): a `"corrupted"` string entry no longer crashes `add()` — fails before (`AttributeError`), passes after.

---

### DIFF
--- a/src/specify_cli/workflows/catalog.py
+++ b/src/specify_cli/workflows/catalog.py
@@ -882,7 +882,11 @@ class StepRegistry:
         import copy
         from datetime import datetime, timezone
 
-        existing = self.data["steps"].get(step_id, {})
+        raw_existing = self.data["steps"].get(step_id)
+        # Corrupted-but-parseable registries may hold non-dict entries; treat
+        # them as absent rather than crashing on existing.get() (mirrors
+        # WorkflowRegistry.add).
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3632, opened 2026-07-22
fix(scripts): git-ext PowerShell emits the '# To persist' SPECIFY_FEATURE hint (parity)

jawwad-ali:
## What

The Git extension's `create-new-feature-branch.ps1` prints a non-JSON hint that diverges from **every** twin:

```
SPECIFY_FEATURE environment variable set to: 001-add-user-auth
```

Its bash twin (`create-new-feature-branch.sh`), its python twin (`create_new_feature_branch.py`), **and** the core `create-new-feature.ps1` all emit:

```
# To persist in your shell: $env:SPECIFY_FEATURE = '001-add-user-auth'
```

The old wording is also misleading: `$env:SPECIFY_FEATURE` is set only in this child process and never propagates to the agent's shell, so the actionable output is precisely the persist hint the siblings print.

## Fix

Mirror the core PowerShell twin: build `$featureAssignment = '$env:SPECIFY_FEATURE = ' + $quotedBranchName` (single-quoted, `'`-escaped) and change the output line to `# To persist in your shell: $featureAssignment`. `BRANCH_NAME`/`FEATURE_NUM` lines are untouched.

## Test

`test_persist_hint_matches_twins` (pwsh CI): the non-JSON output uses `# To persist in your shell: $env:SPECIFY_FEATURE = '001-persist'` — fails before (old wording). Verified end-to-end via `powershell.exe`:

```
BRANCH_NAME: 001-persist
FEATURE_NUM: 001
# To persist in your shell: $env:SPECIFY_FEATURE = '001-persist'
```

Merge precedent for this exact bash-parity vein on these files: #3195, #3129/#3130.

---
🤖 Written with the assistance of Claude Code (AI). Bug self-found; fix/tests verified locally (fail-before / pass-after), ruff & ASCII-only checks clean.

### DIFF
--- a/extensions/git/scripts/powershell/create-new-feature-branch.ps1
+++ b/extensions/git/scripts/powershell/create-new-feature-branch.ps1
@@ -565,6 +565,12 @@ if (-not $DryRun) {
     $env:SPECIFY_FEATURE = $branchName
 }
 
+# Build the PowerShell-idiomatic persist hint, mirroring the core
+# create-new-feature.ps1 twin (and the bash/python twins of this script), which
+# all emit "# To persist in your shell: ...".
+$quotedBranchName = "'" + $branchName.Replace("'", "''") + "'"
+$featureAssignment = '$env:SPECIFY_FEATURE = ' + $quotedBranchName
+
 if ($Json) {
     $obj = [PSCustomObject]@{
         BRANCH_NAME = $branchName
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

jawwad-ali:
Fixed in 9d2fbd4: removed the doubled trailing apostrophe in the test docstring so the documented form reads `$env:SPECIFY_FEATURE = '<name>'`. Docstring-only — the assertions were already correct.

mnriem (approved):

mnriem:
Thank you!

PR #3633, opened 2026-07-22
fix(workflows): workflow add detects local YAML files case-insensitively

jawwad-ali:
## What

`specify workflow add` gated its two local-file branches on a **case-sensitive** extension check — `.suffix in (".yml", ".yaml")` — at both the `--dev` branch (`src/specify_cli/workflows/_commands.py`) and the plain local-path branch. Every other YAML-file detector in the CLI normalizes case first:

- `workflow run` → `source_path.suffix.lower() in (".yml", ".yaml")`
- `WorkflowEngine.load_workflow` → `path.suffix.lower() in (".yml", ".yaml")`

## Why it matters

An add/run inconsistency for uppercase extensions (e.g. `Sample.YAML`, `wf.YML`):

- `specify workflow run Sample.YAML` **loads** the file (case-insensitive), but
- `specify workflow add Sample.YAML` does **not** recognize it as a local workflow:
  - the `--dev` branch rejects it with *"--dev source must be a workflow YAML file or a directory containing workflow.yml"*
  - the plain path falls through to a catalog lookup and fails with *"not found in catalog"*

`Path("x.YAML").suffix` preserves case regardless of filesystem, so the mismatch isn't masked on case-insensitive systems.

## Fix

Add `.lower()` to both suffix reads so `workflow add` matches `workflow run` and the engine loader. The lowercase happy path is unchanged (2-line source change).

## Tests

`tests/test_workflows.py::TestWorkflowAddCaseInsensitiveSuffix`:
- `test_plain_path_accepts_uppercase_extension` — `workflow add Sample.YAML` installs (fails before the fix with "not found in catalog")
- `test_dev_path_accepts_uppercase_extension` — `workflow add --dev Sample.YAML` installs (fails before with the "--dev source must be…" error)
- `test_lowercase_extension_still_installs` — happy-path regression guard

Verified fail-before / pass-after (reverting only the source hunk makes the two uppercase tests fail with the exact errors above). `ruff check` clean.

---
*AI-assisted: authored with Claude Code. I reviewed the change, verified it against the `workflow run` / `load_workflow` siblings, and confirmed the fail-before/pass-after behavior.*

### DIFF
--- a/src/specify_cli/workflows/_commands.py
+++ b/src/specify_cli/workflows/_commands.py
@@ -1555,7 +1555,7 @@ def workflow_add(
     # precedence over --from so a URL that would be ignored is never fetched.
     if dev:
         dev_path = Path(source).expanduser()
-        if dev_path.is_file() and dev_path.suffix in (".yml", ".yaml"):
+        if dev_path.is_file() and dev_path.suffix.lower() in (".yml", ".yaml"):
             _validate_and_install_local(dev_path, str(dev_path))
             return
         if dev_path.is_dir():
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3634, opened 2026-07-22
fix(extensions): parse SKILL.md on the --- delimiter line during removal

Quratulain-bilal:
`ExtensionManager._unregister_extension_skills` verifies an installed skill before deleting it by reading `metadata.source` back from its `SKILL.md`. It located the closing frontmatter delimiter with a raw `split("---", 2)`.

That substring split stops at the first `---` *anywhere* after the opening delimiter — including one embedded in a command description (e.g. `"Separate sections with --- markers"`). The frontmatter is then truncated mid-value, `metadata.source` parses empty, the skill looks unrelated to the extension, and its directory is **left orphaned on uninstall**.

This is the same class of bug fixed in #3590 for `CommandRegistrar.parse_frontmatter` (`content.find("---", 3)`), which now scans on line boundaries. The two removal paths here were the remaining substring-split sites reading generated `SKILL.md` files.

## Fix

Parse on the `---` delimiter *line* by reusing the line-anchored `CommandRegistrar.parse_frontmatter` (the #3590 parser) in both removal paths:
- the fast, registry-driven path, and
- the fallback directory-scan path.

## Repro

```python
raw = '''---
name: speckit-myext-report
metadata:
  source: extension:myext
---
Body here.
'''
# old: raw.split("---", 2)[1] -> "\nname: ...\ndescription: Separate sections with "
#      -> metadata.source == ""  -> dir treated as unrelated, orphaned
```

## Test

Added `test_skills_removed_when_description_contains_dashes`: installs an extension whose command description contains `---`, removes it, and asserts the skill directory is gone. It **fails before the fix** (`assert not skill_dir.exists()` → dir orphaned) and passes after. Full `tests/test_extension_skills.py`, `test_extensions.py`, and `test_extension_registration.py` suites pass (467 tests).

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -1330,19 +1330,20 @@ class ExtensionManager:
                 if not skill_md.is_file():
                     continue
                 try:
-                    import yaml as _yaml
+                    from ..agents import CommandRegistrar as _Registrar
 
                     raw = skill_md.read_text(encoding="utf-8")
-                    source = ""
-                    if raw.startswith("---"):
-                        parts = raw.split("---", 2)
-                        if len(parts) >= 3:
-                            fm = _yaml.safe_load(parts[1]) or {}
-                            source = (
-                                fm.get("metadata", {}).get("source", "")
-                                if isinstance(fm, dict)
-                                else ""
-                            )
+                    # Parse on the ``---`` delimiter *line*, not any ``---``
+                    # substring: a description containing ``---`` would trip a
+                    # raw ``split("---", 2)`` and hide metadata.source, so this
+                    # extension's own skill would look unrelated and be left
+                    # orphaned. Mirrors the #3590 parse_frontmatter fix.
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

Quratulain-bilal:
Good catch — addressed in 2b44ad2. You're right that `skills_project` keeps `ai_skills` enabled, so `remove()` resolved the skills dir and only the fast branch ran; a substring-split regression in the fallback would have stayed green.

Added `test_skills_removed_with_dashes_via_fallback_scan`, which deletes `init-options.json` after install so `_get_skills_dir()` returns `None` and removal takes the fallback directory-scan branch. Confirmed it protects that independently duplicated parser: reverting **only** the fallback site back to the old `split("---", 2)` fails the new test (skill dir orphaned) while the fast-path test still passes.

mnriem (approved):

mnriem:
Thank you!

PR #3637, opened 2026-07-22
fix(workflows): validate every redirect hop when fetching workflow/step catalogs

Quratulain-bilal:
`WorkflowCatalog._fetch_single_catalog` and `StepCatalog._fetch_single_catalog` opened the catalog URL with `open_url(entry.url, timeout=30)` and validated only the final `resp.geturl()`.

`open_url` follows redirects, so an `https://` catalog entry that 30x-redirects through a non-HTTPS host **mid-chain** could let a network attacker rewrite the next hop and slip a payload past the terminal-URL-only check. That payload then drives the workflow/step catalog data used by discovery and install.

This is the same security-parity gap already closed for the presets (#3523) and extensions (#3524) catalog fetchers — both now validate **every** redirect hop, not just the final URL. The two workflow catalog loaders were missed.

## Fix

Pass a `redirect_validator` that runs the existing HTTPS/hostname check (`_validate_catalog_url` / `_validate_url`) before every redirect hop, in both loaders. The final `geturl()` check is kept as a defense-in-depth backstop.

## Tests

- Add `test_fetch_validates_every_redirect_hop` for both `TestWorkflowCatalog` and `TestStepCatalog`: a non-HTTPS **intermediate** hop is rejected with an HTTPS refusal. Both **fail before the fix** (`'NoneType' object is not callable` — no validator was passed, so the redirect is followed unchecked) and pass after.
- Update the two existing `test_fetch_malformed_redirect_target_raises_catalog_error` stubs whose `open_url` lambda lacked the `redirect_validator` kwarg.

All catalog-related workflow tests pass (78 passed, 1 skipped).

### DIFF
--- a/src/specify_cli/workflows/catalog.py
+++ b/src/specify_cli/workflows/catalog.py
@@ -523,8 +523,20 @@ class WorkflowCatalog:
 
         _validate_catalog_url(entry.url)
 
+        # Validate EVERY redirect hop, not just the final URL: _open_url follows
+        # redirects, so an https:// entry that 30x-redirects through http:// (or
+        # to a non-HTTPS host mid-chain) could otherwise let a network attacker
+        # rewrite the next hop and slip a payload past a final-URL-only check.
+        # redirect_validator runs before each hop; the geturl() check below is
+        # retained as a defense-in-depth backstop. Mirrors the presets/extensions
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3642, opened 2026-07-22
fix(init): show hyphenated /speckit-<name> in Next Steps for Forge projects

jawwad-ali:
## What

The post-init **Next Steps** panel renders the recommended slash commands through the nested `_display_cmd()` in `src/specify_cli/commands/init.py`. It special-cased dollar-skills agents, `kimi`, slash-skills agents, and `cline` — but not Forge. For a Forge project `_display_cmd` fell through to:

```python
return f"/speckit.{name}"
```

printing `/speckit.constitution`, `/speckit.specify`, `/speckit.plan`, etc.

Forge only registers the **hyphenated** form `/speckit-<name>` (via `format_forge_command_name` / `ForgeIntegration.build_command_invocation`; the generated Forge command-file tests in `test_integration_forge.py` already assert dot-notation must not appear). So the panel instructed Forge users to run commands that don't exist under the dotted name.

## Fix

Add `forge_skill_mode = selected_ai == "forge"` alongside the existing `cline_skill_mode`, and include it in the hyphenated-slash condition — mirroring exactly how Cline (also a non-skills markdown agent with hyphenated commands) is handled. Other agents are unaffected.

## Tests

`tests/integrations/test_integration_forge.py::TestForgeInitNextSteps` — runs `specify init --here --integration forge` and asserts the output contains `/speckit-plan` and **not** `/speckit.plan`, mirroring the existing Copilot/Grok next-steps tests.

Verified fail-before / pass-after (reverting only the source hunk makes the test fail — the panel renders the dotted `/speckit.checklist` / `/speckit.plan`). `ruff check` clean.

---
*AI-assisted: authored with Claude Code. I confirmed Forge's hyphenated registration, that Cline is the exact sibling precedent in this same closure, and verified fail-before/pass-after.*

### DIFF
--- a/src/specify_cli/commands/init.py
+++ b/src/specify_cli/commands/init.py
@@ -700,6 +700,7 @@ def register(app: typer.Typer) -> None:
         zed_skill_mode = selected_ai == "zed" and _is_skills_integration
         grok_skill_mode = selected_ai == "grok" and _is_skills_integration
         cline_skill_mode = selected_ai == "cline"
+        forge_skill_mode = selected_ai == "forge"
         bob_skill_mode = selected_ai == "bob" and _is_skills_integration
         native_skill_mode = (
             codex_skill_mode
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3651, opened 2026-07-22
fix(cli): guard lazy .hostname ValueError in extension/preset add --from

Noor-ul-ain001:
## What

Two CLI install paths validated a user-supplied `--from` URL by reading `parsed.hostname` **outside** their `try/except ValueError` guard:

- `extension add --from <url>` — `extensions/_commands.py:434`
- `preset add --from <url>` — `presets/_commands.py` (`_is_allowed_download_url`)

A bracketed-but-invalid IPv6 authority (e.g. `https://[not-an-ip]/x.zip`) parses cleanly under `urlparse()` on Python < 3.14 and only raises `ValueError` **lazily** on the first `.hostname` access. On the interpreters spec-kit supports (`>=3.11`), that raw `ValueError` leaked past the CLI, printing an uncaught traceback instead of the clean `Invalid URL` error. (The raise moved eager into `urlparse()` only in 3.14, so the pre-fix code happened to be safe there — but not on 3.11–3.13, i.e. CI.)

Same bug class as the maintainer-fixed `#3433` / `#3435` / `#3437` / `#3577` (unguarded `urlparse().hostname` leaking a raw `ValueError`).

## Fix

- **extensions**: read `parsed.hostname` inside the existing `try` and reuse it for the localhost check.
- **presets**: guard the up-front `urlparse(from_url).hostname` read (preserving the existing `Invalid URL` message), and harden the nested `_is_allowed_download_url` to accept a URL **string** and parse + read `.hostname` inside its own `try/except` → returns `False` on malformed input. This also covers the redirect-validator and post-redirect final-URL checks, where the URL is **server-controlled** (a malicious redirect target).

## Tests

For each command:
- a bracketed-non-IP `--from` URL exits cleanly (exit 1, no traceback);
- a monkeypatched lazy-`.hostname` raiser reproducing the pre-3.14 shape **independently of the running interpreter**. Test-the-test: both fail with a raw `ValueError` before the fix and pass after.

Full `tests/test_extensions.py` + `tests/test_presets.py` pass locally (751 passed, 7 pre-existing version/pwsh skips).


mnriem:

Noor-ul-ain001:
Addressed the Copilot feedback in fec5a68 (rebased on top of the Copilot Autofix commit).

**1. Inaccurate CPython version history in the guard comments.** Copilot was right. The eager bracketed-host check landed in `urlsplit` via [gh-103848](https://github.com/python/cpython/issues/103848) (CVE-2024-11168), which was backported to the 3.11 branch (shipped in 3.11.4) and to 3.10/3.9. So on every interpreter spec-kit supports (`>=3.11`), `https://[not-an-ip]/…` is rejected **eagerly at `urlparse()`** — the "parses cleanly / raises lazily on `.hostname` / only eager on 3.14+" framing was wrong. All three source comments (`extensions/_commands.py`, and both spots in `presets/_commands.py`) now state the guard as a defensive policy — parsing *or* the `.hostname` read can raise `ValueError`, so both stay inside the `try` — without asserting any version history.

**2. Monkeypatch tests mischaracterized as the exact production path.** Relabeled both `*_lazy_hostname_valueerror_exits_cleanly` tests as **synthetic defensive coverage** that does not reproduce any specific CPython behavior — they just exercise the case where the `ValueError` surfaces on the `.hostname` read rather than at parse time. Also dropped the version-history claims from the two `bracketed_non_ip` docstrings.

**3. `_is_allowed_download_url(final_url)` redirect refactor.** Already applied in the original commit — the final-URL and redirect-validator calls both pass the URL string and parse internally.

These are comment/docstring changes only; behavior is unchanged and the URL-guard tests still pass.

mnriem:
Please resolve conflicts

Noor-ul-ain001:
Merged latest `main` and resolved the conflict (branch is now mergeable).

The conflict was in `presets/_commands.py`, and the resolution is worth flagging because `main` moved underneath this PR:

- **Presets — superseded, took upstream.** `main` refactored preset URL validation into the shared `src/specify_cli/_download_security.py` module. Its `_parse_url()` already reads `.hostname`/`.port` inside a `try/except (TypeError, ValueError)`, and both `is_https_or_localhost_http()` and `is_safe_download_redirect()` go through it. That centralizes exactly the guard this PR was adding locally, so I dropped the PR's presets-side changes (the local `_is_allowed_download_url` + up-front `.hostname` read) in favor of upstream's version.
- **Extensions — still needed, kept.** `main`'s `extensions/_commands.py` still reads `parsed.hostname` **outside** the `try`, so `extension add --from` can still leak a raw `ValueError`. The guard fix auto-merged cleanly and remains.
- **Tests.** Extensions regression tests are unchanged. On the presets side, the bracketed-non-IP test still passes against the shared helper; I retargeted the synthetic lazy-`.hostname` monkeypatch test to an **out-of-range port** (`https://example.com:99999/…`), which raises `ValueError` lazily on the `.port` access that upstream's up-front guard now performs — a realistic case that maps to actual code rather than a synthetic mock.

`tests/test_presets.py` + `tests/test_extensions.py` + `tests/test_download_security.py`: 828 passed, 7 skipped (pre-existing version/pwsh skips).

mnriem:
Thank you!

PR #3656, opened 2026-07-22
docs(workflows): gate step docstring lists the 'retry' on_reject behaviour

jawwad-ali:
## What

The `GateStep` class docstring said `on_reject` *"controls abort / skip behaviour"* — listing only two behaviours. The step actually supports three:

- `validate()` rejects any `on_reject` not in `('abort', 'skip', 'retry')`
- `execute()` has a dedicated `retry` branch (returns `PAUSED` so the next `resume` re-executes the gate), distinct from `abort` (`FAILED`) and `skip` (`COMPLETED`)

The docstring was the only place omitting `retry`, contradicting the same file's `validate()` and `execute()`.

## Fix

Docstring-only: `controls abort / skip behaviour.` → `controls abort / skip / retry behaviour.`

## Tests

`tests/test_workflows.py::TestGateStep::test_docstring_lists_every_on_reject_behaviour` — asserts all three behaviours appear in `GateStep.__doc__` (fails before the fix). `ruff` clean.

---
*AI-assisted: authored with Claude Code. Verified against the step's own `validate()` accepted set and `execute()` branches.*

### DIFF
--- a/src/specify_cli/workflows/steps/gate/__init__.py
+++ b/src/specify_cli/workflows/steps/gate/__init__.py
@@ -26,7 +26,7 @@ class GateStep(StepBase):
     later with ``specify workflow resume``.
 
     The user's choice is stored in ``output.choice``.  ``on_reject``
-    controls abort / skip behaviour.
+    controls abort / skip / retry behaviour.
     """
 
     type_key = "gate"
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3657, opened 2026-07-22
fix(integrations): Cline overrides post_process_command_content (correct hook name)

jawwad-ali:
## What

`ClineIntegration` defined its command-content transform as `post_process_content`, but the overridable base hook is `IntegrationBase.post_process_command_content` — the method `CommandRegistrar.register_commands()` dispatches to (`agents.py:813` and `:877`) for every non-skills integration.

Because the names differed, Cline's method never overrode the base hook. For **extension/preset** command files registered via `CommandRegistrar`, the registrar therefore ran the base **no-op**, so those files never received Cline's dot-to-hyphen **hook-command note** (`_inject_hook_command_note`) — the note instructing the agent to replace dots with hyphens when invoking hook commands.

(Cline's own `setup()` still works because it calls the method directly. And handoff references are already hyphenated independently by the registrar's `_hyphenate_body_refs`, so that transform was unaffected — renaming just makes Cline's own copy run too, harmlessly, since both are idempotent.)

## Fix

Rename `post_process_content` → `post_process_command_content` (matching the base hook and the `post_process_skill_content` exact-name-override convention used by claude/copilot/agy/kimi/droid/vibe), and update the single internal caller in `setup()`. No test referenced the old name.

## Tests

`tests/test_post_process.py::TestClineRealPostProcess::test_cline_transforms_applied_via_registrar` — registers an extension command through `CommandRegistrar` and asserts Cline's **real** transform ran (the hook-command note's "replace dots" text is present). Fails before the fix (note absent — the base no-op ran); passes after. `ruff` clean; all 11 post-process tests pass.

---
*AI-assisted: authored with Claude Code. I traced the registrar dispatch (`post_process_command_content` at agents.py:813/877), confirmed the exact-name-override convention across sibling integrations, and verified via fail-before that the hook-command note is missing without the rename.*

### DIFF
--- a/src/specify_cli/integrations/cline/__init__.py
+++ b/src/specify_cli/integrations/cline/__init__.py
@@ -138,8 +138,14 @@ class ClineIntegration(MarkdownIntegration):
             content,
         )
 
-    def post_process_content(self, content: str) -> str:
-        """Apply Cline-specific transformations to command content."""
+    def post_process_command_content(self, content: str) -> str:
+        """Apply Cline-specific transformations to command content.
+
+        Overrides the ``IntegrationBase`` hook of the same name so that
+        ``CommandRegistrar.register_commands()`` (which dispatches to
+        ``post_process_command_content``) applies these transforms to
+        extension/preset command files too, not just core commands.
+        """
         updated = self._inject_hook_command_note(content)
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3658, opened 2026-07-22
fix(bundler): order bundle members by canonical POSIX arcname (reproducible builds)

jawwad-ali:
## What

`_collect_files` returned `sorted(collected)` — i.e. `pathlib.Path` comparison order. That ordering is **platform-dependent**: on Windows `PurePath` compares case-folded with backslash separators, while on Linux/macOS it compares the case-preserving forward-slash string.

But the zip member **names** are the canonical POSIX arcnames (`build_bundle`: `file_path.relative_to(bundle_dir).as_posix()`), which are platform-independent. So member *order* was derived from a platform-dependent key while member *names* were canonical — the same bundle built on Windows vs Linux/macOS produced archives whose members were laid out in different order, i.e. **not byte-for-byte identical across build hosts**. That contradicts the packager's stated reproducible-build guarantee (fixed timestamps + canonical 0644/0755 modes; `test_build_is_deterministic` asserts "byte-for-byte reproducible").

## Fix

Order by the **same canonical POSIX-arcname key** used to name the members:

```python
return sorted(collected, key=lambda p: p.relative_to(bundle_dir).as_posix())
```

## Tests

`tests/unit/test_bundler_packager.py::test_member_order_is_platform_independent` — builds a bundle with mixed-case sibling files and asserts `archive.namelist() == sorted(archive.namelist())`. Fails before the fix on Windows (Path order groups case differently: `apple.txt` sorts before `Foo.txt`); passes after. CI's `windows-latest` matrix exercises this. All existing packager tests (incl. `test_build_is_deterministic`) still pass. `ruff` clean.

---
*AI-assisted: authored with Claude Code. Verified the arcname key matches `build_bundle`'s naming, and confirmed fail-before on Windows / pass-after.*

### DIFF
--- a/src/specify_cli/bundler/services/packager.py
+++ b/src/specify_cli/bundler/services/packager.py
@@ -142,4 +142,10 @@ def _collect_files(
                 # Skip symlinked files to avoid escaping the bundle directory.
                 continue
             collected.append(path)
-    return sorted(collected)
+    # Order by the canonical POSIX arcname (the same key build_bundle uses to
+    # NAME each member), not by pathlib.Path comparison. Path ordering is
+    # platform-dependent (Windows folds case and uses backslash separators),
+    # which would lay out zip members differently across build hosts and break
+    # the byte-for-byte reproducible-build guarantee even though the member
+    # names are identical.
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3661, opened 2026-07-22
fix(bundler): reject falsy non-mapping requires/provides in manifest from_dict

jawwad-ali:
## What

`BundleManifest.from_dict` used `data.get("requires") or {}` and `data.get("provides") or {}`. The `or {}` coerces a **falsy** non-mapping value (`[]`, `''`, `0`, `false`) to `{}` **before** the `isinstance` guard runs — so a malformed manifest silently passed validation as one that requires/provides nothing. Only a *truthy* non-mapping (e.g. `provides: "extensions"`) reached the guard and raised.

The sibling `integration` guard (added in #3629) already does this correctly with an explicit `is not None` check; `requires`/`provides` were the remaining holes.

## Fix

Handle `None` explicitly (default to `{}`) and reject every other non-mapping:

```python
requires_raw = data.get("requires")
if requires_raw is None:
    requires_raw = {}
elif not isinstance(requires_raw, dict):
    raise BundlerError("'requires' must be a mapping when present.")
```

…and the same for `provides`. Absent fields still parse to the empty default.

## Tests

`tests/contract/test_manifest_schema.py`:
- `test_non_mapping_provides_rejected_including_falsy` and `test_non_mapping_requires_rejected_including_falsy` — parametrized over `[]`, `''`, `0`, `False`, and a truthy string; all raise after the fix (the falsy ones passed silently before)
- `test_absent_provides_and_requires_do_not_raise_mapping_error` — absent (None) optional mappings still parse without tripping the guard

`ruff` clean; all 28 contract tests pass.

---
*AI-assisted: authored with Claude Code. Verified the falsy-coercion hole against the merged `integration` guard (#3629) and confirmed fail-before/pass-after.*

### DIFF
--- a/src/specify_cli/bundler/models/manifest.py
+++ b/src/specify_cli/bundler/models/manifest.py
@@ -111,8 +111,10 @@ class BundleManifest:
             license=str(bundle_raw.get("license", "")).strip(),
         )
 
-        requires_raw = data.get("requires") or {}
-        if not isinstance(requires_raw, dict):
+        requires_raw = data.get("requires")
+        if requires_raw is None:
+            requires_raw = {}
+        elif not isinstance(requires_raw, dict):
             raise BundlerError("'requires' must be a mapping when present.")
         requires = Requires(
             speckit_version=str(requires_raw.get("speckit_version", "")).strip(),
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3664, opened 2026-07-22
fix(integrations): recompute invoke_separator from retained parsed_options

jawwad-ali:
## What

`with_integration_setting` (`integration_runtime.py`) recomputed `invoke_separator` from the **raw `parsed_options` argument**:

```python
current["invoke_separator"] = integration.effective_invoke_separator(parsed_options, project_root)
```

When a caller updates **only** `script_type` (`parsed_options=None`, `raw_options=None`), the previously-stored `parsed_options` are **retained** on the setting (the code only pops them when `raw_options is not None`). But the separator was still derived from the `None` argument — so an options-dependent separator (e.g. Copilot in `--skills` mode → `"-"`) silently reverted to the default `"."`, desynchronizing `invoke_separator` from the `parsed_options` actually stored.

## Fix

Derive the separator from `current.get("parsed_options")` — the options actually stored on the setting after the update — so it stays consistent across all branches (new options provided, options cleared via `raw_options`, or only `script_type` changed).

## Tests

`tests/integrations/test_integration_state.py::test_with_integration_setting_recomputes_separator_from_retained_options` — stores Copilot `{"skills": True}` (separator `"-"`), then updates only `script_type`; asserts `parsed_options` is retained and `invoke_separator` stays `"-"`. Fails before the fix (recomputed to `"."`); passes after. `ruff` clean.

---
*AI-assisted: authored with Claude Code. Verified via Copilot's options-dependent `effective_invoke_separator` and confirmed fail-before/pass-after.*

### DIFF
--- a/src/specify_cli/integration_runtime.py
+++ b/src/specify_cli/integration_runtime.py
@@ -64,8 +64,15 @@ def with_integration_setting(
     elif raw_options is not None:
         current.pop("parsed_options", None)
 
+    # Recompute the separator from the options actually STORED on ``current``
+    # after the update, not the raw ``parsed_options`` argument. When only
+    # ``script_type`` changes (``parsed_options`` and ``raw_options`` both
+    # None), the previously-stored ``parsed_options`` are retained above, so
+    # deriving the separator from the argument (None) would drop an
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3666, opened 2026-07-22
fix(bundler): reject falsy non-list bundles/contributed_components in records

jawwad-ali:
## What

Both readers of `.specify/bundle-records.json` defaulted their list fields with `data.get(...) or []` **before** the `isinstance(list)` guard, making the guard dead code for **falsy** non-list values:

- `load_records` (L124): `bundles = data.get("bundles") or []`
- `InstalledBundleRecord.from_dict` (L58): `components_raw = data.get("contributed_components") or []`

A corrupt `bundles: 0` / `bundles: {}` (or the same for `contributed_components`) was coerced to `[]` and silently read as "no bundles"/"no components" instead of raising `Corrupt records file … must be a list`. Only a *truthy* non-list (e.g. `"x"`) reached the guard.

## Fix

Handle `None` explicitly (default to `[]`) and reject every other non-list, mirroring the merged `requires`/`provides`/`integration` manifest guards (#3629, #3661) and the `catalog_config._read` sibling reader.

## Tests

`tests/unit/test_bundler_records.py`:
- `test_load_records_rejects_falsy_non_list_bundles` — parametrized `0`, `False`, `''`, `{}`
- `test_from_dict_rejects_falsy_non_list_contributed_components` — same set

All raise after the fix; all passed silently before. `ruff` clean; the full records suite passes (the one Windows-only `test_save_records_refuses_symlinked_specify_escape` failure is a pre-existing symlink-privilege limitation of the local box, unrelated to this change, and passes on CI/Linux).

---
*AI-assisted: authored with Claude Code. Verified against the sibling guards and confirmed fail-before/pass-after.*

### DIFF
--- a/src/specify_cli/bundler/models/records.py
+++ b/src/specify_cli/bundler/models/records.py
@@ -55,8 +55,13 @@ class InstalledBundleRecord:
     def from_dict(cls, data: Any) -> "InstalledBundleRecord":
         if not isinstance(data, dict):
             raise BundlerError("Each installed-bundle record must be a mapping.")
-        components_raw = data.get("contributed_components") or []
-        if not isinstance(components_raw, list):
+        components_raw = data.get("contributed_components")
+        if components_raw is None:
+            components_raw = []
+        elif not isinstance(components_raw, list):
+            # `or []` would coerce a FALSY non-list (0, '', False, {}) to []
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3669, opened 2026-07-22
fix(extensions): hyphenate command names in the Forge post-install listing

jawwad-ali:
## What

After `specify extension add`, the **"Provided commands"** summary hyphenated command names only for Cline:

```python
is_cline = load_init_options(project_root).get("ai") == "cline"
...
if is_cline:
    cmd_name = format_cline_command_name(cmd_name)
```

For a **Forge** project the names were printed in dotted form (e.g. `speckit.test-ext.hello`), but Forge registers its commands hyphenated (`speckit-test-ext-hello`, via `format_forge_command_name`). So the post-install summary told Forge users to run commands that don't exist under the dotted name.

## Fix

Extend the existing Cline handling to Forge via `format_forge_command_name`. This completes the Forge command-name parity already fixed for hook invocations (#3641) and the init next-steps panel (#3642) — the extension-install listing was the third and last site.

## Tests

`tests/test_extensions.py::test_forge_extension_install_listing_hyphenates_command_names` — installs an extension (`extension add --dev`) in a Forge project and asserts the listing shows `speckit-test-ext-hello`, not the dotted form. Fails before the fix; passes after. `ruff` clean.

---
*AI-assisted: authored with Claude Code. Completes the Forge parity series (#3641/#3642) and verified fail-before/pass-after.*

### DIFF
--- a/src/specify_cli/extensions/_commands.py
+++ b/src/specify_cli/extensions/_commands.py
@@ -623,16 +623,22 @@ def extension_add(
         for warning in manifest.warnings:
             console.print(f"\n[yellow]⚠  Compatibility warning:[/yellow] {_escape_markup(str(warning))}")
 
-        is_cline = load_init_options(project_root).get("ai") == "cline"
+        selected_ai = load_init_options(project_root).get("ai")
+        is_cline = selected_ai == "cline"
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3671, opened 2026-07-22
fix: harden redirect validation and bounded reads

PascalThuet:
## Summary

This is a follow-up to #3140, prepared after that PR merged while a final security and robustness audit was still running.

- reject remote redirects to local targets across canonical, legacy numeric, percent-encoded, IDNA, scoped IPv6, mapped IPv4, and unspecified-address spellings
- keep HTTP available only for unambiguous loopback URLs used by local development, while preserving normal HTTPS/CDN redirects
- make bounded response reads memory-stable under adversarial short reads by using 64 KiB reads and a single `BytesIO` buffer
- validate malformed redirect ports and unexpected Azure CLI/Azure AD token payload shapes without leaking raw parser exceptions
- centralize the shared download URL policy used by authentication, presets, and workflows
- ensure self-upgrade tests exercise the isolated opener used by production code

## Root cause and impact

The merged redirect check handled the obvious remote-to-HTTP-loopback case, but validation and connection behavior could still disagree for alternative hostname spellings. It also allowed a remote HTTPS URL to enter a local HTTPS target. In environments that consume attacker-controlled URLs, that gap could enable a conditional SSRF-style read from a service on the local machine or CI runner.

The bounded reader assumed `read(size)` would normally return large chunks. A response producing one-byte short reads retained every bytes object and then duplicated the payload during `join`; the 1 MiB reproduction increased RSS by about 94.5 MiB. The new implementation keeps the same byte limit while reducing that reproduction to about 1.6 MiB.

These findings are defense-in-depth rather than a critical vulnerability. Normal remote HTTPS redirects remain supported, and authorization headers are still stripped when a redirect leaves configured trusted hosts.

## Compatibility and limits

The redirect policy is deliberately lexical and performs no DNS lookup. DNS or hosts-file rebinding would require connection-level address pinning and is outside this focused follow-up. Raw IPv6 zone syntax is rejected; the RFC `%25zone` form remains supported.

## Validation

- `.venv/bin/python -m pytest`: 5,095 passed, 155 skipped
- `uvx ruff check src tests`: passed
- Bandit on the changed source files at medium/high severity: no findings
- `git diff --check`: passed

Prepared on behalf of @PascalThuet by OpenAI Codex (model: GPT-5, autonomous).

### DIFF
--- a/src/specify_cli/_download_security.py
+++ b/src/specify_cli/_download_security.py
@@ -2,14 +2,17 @@
 
 from __future__ import annotations
 
+import io
+import socket
+from ipaddress import IPv4Address, IPv6Address, ip_address
 from typing import NoReturn, TypeVar
-from urllib.parse import urlparse
+from urllib.parse import ParseResult, urlparse
 
 
 ErrorT = TypeVar("ErrorT", bound=Exception)
 
 MAX_DOWNLOAD_BYTES = 50 * 1024 * 1024
-READ_CHUNK_SIZE = 1024 * 1024
+READ_CHUNK_SIZE = 64 * 1024
 
 # Tighter ceiling for responses that are read fully into memory and parsed as
 # JSON. The 50 MiB MAX_DOWNLOAD_BYTES default is sized for archive/payload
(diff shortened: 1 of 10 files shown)


mnriem:
Thank you!

PR #3673, opened 2026-07-23
fix(git-extension): trim trailing whitespace before stripping commit-message quotes

Quratulain-bilal:
The auto-commit **bash** and **Python** twins read the configured `message:` value and strip one leading + trailing quote with an end-of-string-anchored strip (`sed 's/["\']*$//'` / `re.sub(r'["\']*$', ...)`).

When the YAML value has trailing whitespace **after** the closing quote — e.g. `message: "Done"  ` — the close-quote strip never matches, because the quote is no longer at end-of-string (spaces follow it). The resulting commit message keeps a **dangling quote and trailing spaces**: `Done"  `.

The **PowerShell** twin already `.Trim()`s before stripping, so it produced the clean `Done`. So the three script variants were out of parity, and two of them wrote a malformed commit subject.

## Fix

Trim the value before stripping quotes in the bash and Python twins, matching the PowerShell twin.

## Verification

Exact-code level (the actual sed pipeline / regex from the scripts):

```
input:    message: "spec done"
OLD bash: [spec done"  ]      # dangling quote + spaces
NEW bash: [spec done]         # clean
```

Python `_strip_quotes` matches after the fix (`'spec done'` for trailing-space, single-quote, and bare cases).

Added a parity regression test (`test_custom_message_with_trailing_whitespace_after_quote`) asserting bash and Python produce the same clean `spec done`. It runs under CI where Git-for-Windows bash / native bash is resolvable; the existing bash-parity tests already skip on environments without a working native bash.

### DIFF
--- a/extensions/git/scripts/bash/auto-commit.sh
+++ b/extensions/git/scripts/bash/auto-commit.sh
@@ -94,7 +94,12 @@ if [ -f "$_config_file" ]; then
                     [ "$_val" = "false" ] && _enabled=false
                 fi
                 if echo "$_line" | grep -Eq '[[:space:]]+message:'; then
-                    _commit_msg=$(echo "$_line" | sed 's/^[^:]*:[[:space:]]*//' | sed 's/^["'\'']//' | sed 's/["'\'']*$//')
+                    # Trim trailing whitespace before stripping the closing quote:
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3587, opened 2026-07-17
feat: add Factory Droid CLI integration (#822)

NurfitraPujo:
Adds a skills-based integration for the Factory Droid CLI alongside the existing Claude/Codex skills agents. The integration scaffolds `.factory/skills/speckit-*` directories, injects canonical `--model`/`--output-format` flags ahead of any operator-supplied extra args, and documents the install step in the devcontainer post-create script (mirroring the Kiro install layout, with a chained EXIT trap so both installer tmpfiles are released on script exit).

Includes:
- `src/specify_cli/integrations/droid/__init__.py` (subpackage)
- `tests/integrations/test_integration_droid.py` (46 tests, including regression coverage for the no-trailing-newline frontmatter fusion bug, idempotent skill injection, and env-var path resolution)
- `integrations/catalog.json` entry + `updated_at` bump
- Alphabetical registration in `src/specify_cli/integrations/__init__.py` and `tests/integrations/test_registry.py`
- Devcontainer Droid install block

Closes #822

Assisted-by: Droid (oracle-reviewer)

## Description

This PR add support for factory droid cli integration (https://factory.ai/product/cli)

## Testing

since this is a new isolated module for adding new integration, I already added new tests cases to make sure the integration running correctly.

- [x] Tested locally with `uv run specify --help`
- [x] Ran existing tests with `uv sync && uv run pytest`
- [x] Tested with a sample project (if applicable)

## AI Disclosure

- [ ] I **did not** use AI assistance for this contribution
- [x] I **did** use AI assistance (describe below)

I use AI for research, planning, implementing, testing and reviewing the changes. I did review the research and plan results, but I don't do much on reviewing the implementation, it's majorly reviewed by AI.

PR #2780, opened 2026-05-31
Add Zed integration

arrrrny:
## Summary
- add a built-in `zed` integration that installs Spec Kit commands as Zed skills under `.agents/skills`
- render Zed skill invocations correctly in init next-steps output and hook messages
- add integration coverage for Zed plus registry/list assertions and docs updates

## Testing
- `uv run --extra test python3 -m pytest tests/integrations/test_registry.py tests/integrations/test_integration_subcommand.py::TestIntegrationList::test_list_shows_available_integrations tests/integrations/test_integration_zed.py -q`

Closes #2779

PR #3688, opened 2026-07-23
fix(integrations): render hyphenated /speckit-<name> for Droid (always-slash agent)

jawwad-ali:
## What

`DroidIntegration` is an always-skills agent — it installs commands as `.factory/skills/speckit-<name>/SKILL.md` and its `build_command_invocation` returns the hyphenated `/speckit-<name>`. But `"droid"` was missing from every set in `_invocation_style.py` (`DOLLAR_SKILLS_AGENTS`, `ALWAYS_SLASH_AGENTS`, `CONDITIONAL_SLASH_AGENTS`), so `is_slash_skills_agent("droid", True)` returned `False`.

As a result, both `HookExecutor._render_hook_invocation` and `specify init`'s next-steps `_display_cmd` fell through to the **dotted** `/speckit.<name>` form — a command Droid never registers. Hook dispatch and the printed getting-started commands pointed at nonexistent commands for Droid users.

## Fix

Add `"droid"` to `ALWAYS_SLASH_AGENTS`, matching its always-skills siblings `grok`/`trae`/`zed`/`devin` (each added to that set by its own integration PR — e.g. grok in #3535, zed in #2780; Droid's #3587 omitted it).

## Tests

`tests/integrations/test_integration_droid.py::TestDroidIntegration::test_is_slash_skills_agent` — asserts `is_slash_skills_agent("droid", True) is True` (fails before the fix). Full droid suite (46 tests) passes; `ruff` clean.

---
*AI-assisted: authored with Claude Code. I confirmed Droid is an always-skills `SkillsIntegration` (hyphenated `build_command_invocation`, no `is_skills_mode` override) and that grok/trae/zed/devin establish the `ALWAYS_SLASH_AGENTS` precedent.*

### DIFF
--- a/src/specify_cli/_invocation_style.py
+++ b/src/specify_cli/_invocation_style.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 DOLLAR_SKILLS_AGENTS: frozenset[str] = frozenset({"codex", "zcode"})
 
 # Agents that always render /speckit-<name>, regardless of ai_skills.
-ALWAYS_SLASH_AGENTS: frozenset[str] = frozenset({"devin", "grok", "trae", "zed"})
+ALWAYS_SLASH_AGENTS: frozenset[str] = frozenset({"devin", "droid", "grok", "trae", "zed"})
 
 # Agents that render /speckit-<name> only when ai_skills is enabled.
 CONDITIONAL_SLASH_AGENTS: frozenset[str] = frozenset(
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

jawwad-ali:
Thanks! The Copilot feedback is addressed — I added the disabled-case assertion so the test now covers both states:

```python
assert is_slash_skills_agent("droid", True) is True
assert is_slash_skills_agent("droid", False) is True
```

The Copilot thread is resolved and the change is pushed. Ready for another look when you have a moment.

mnriem (approved):

PR #3690, opened 2026-07-23
fix(workflows): escape step-graph brackets in `workflow info` so the step type shows

jawwad-ali:
## What

`specify workflow info <id>` renders each step's line as `→ <id> [<type>]` — with the step type meant to appear inside **literal** square brackets. But `console.print` has Rich markup enabled, so `[<type>]` is parsed as an opening style tag named after the step type (`command`, `gate`, `prompt`, …). Those aren't valid Rich styles and there's no closing tag, so Rich **silently swallows the whole `[<type>]` token** — every step prints as `→ <id> ` with the type gone.

## Fix

Escape the literal bracket with `\[` so Rich renders `[<type>]` verbatim, and escape the id/type through `_escape_markup` (as the sibling `workflow_list` already does for displayed fields). Mirrors the in-file `\[disabled]` precedent in `workflow_list`.

## Tests

`tests/test_workflows.py::TestWorkflowInfoStepGraph::test_step_type_rendered_in_literal_brackets` — monkeypatches `WorkflowEngine.load_workflow` to return a definition with a `gate` step and asserts `[gate]` appears in `workflow info` output. Fails before the fix (output is `→ step-one ` — the `[gate]` swallowed). `ruff` clean.

---
*AI-assisted: authored with Claude Code. Reproduced the Rich markup-swallowing behavior and verified fail-before/pass-after.*

### DIFF
--- a/src/specify_cli/workflows/_commands.py
+++ b/src/specify_cli/workflows/_commands.py
@@ -2376,7 +2376,15 @@ def workflow_info(
             console.print(f"\n  [bold]Steps ({len(definition.steps)}):[/bold]")
             for step in definition.steps:
                 stype = step.get("type", "command")
-                console.print(f"    → {step.get('id', '?')} [{stype}]")
+                # Escape the literal bracket (\[) so Rich renders `[<type>]`
+                # instead of parsing it as a style tag named after the step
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3691, opened 2026-07-23
fix(auth): normalize whitespace in auth-config env-var/id references at store time

jawwad-ali:
## What

In `load_auth_config`, the fields `token_env`, `client_secret_env`, `tenant_id`, and `client_id` are **validated** on their `.strip()`ed form (proving they're non-empty ignoring surrounding whitespace) but then **stored raw** on the `AuthConfigEntry`.

So an accidentally padded value (e.g. `"token_env": " GH_TOKEN "`) passes validation, yet downstream:
- `base.py` → `os.environ.get(entry.token_env)`
- `azure_devops.py` → `os.environ.get(entry.client_secret_env, "")`, and `tenant_id`/`client_id` used verbatim in the OAuth URL/body

…all use the padded name literally, so the env var is never found. `load_auth_config` succeeds with **no error**, but `resolve_token` returns `None` and the request silently downgrades to unauthenticated (401/403 on private resources) with no diagnostic — the classic validate-normalized-but-store-original bug.

The `hosts` field in the same function already does it right (`hosts = [h.strip().lower() for h in hosts]`); these references were just never given the same normalization.

## Fix

A module-level `_norm` helper strips whitespace-insignificant string references at store time (non-strings like `None` pass through). Applied to `token_env`, `tenant_id`, `client_id`, `client_secret_env`. `token` is left unchanged (already stripped at resolve time in `base.py`).

## Tests

`tests/test_authentication.py::TestLoadAuthConfig::test_padded_token_env_is_normalized_and_resolves` — a `" GH_TOKEN "` config now stores `"GH_TOKEN"` and `GitHubAuth().resolve_token(...)` finds the token. Fails before the fix (`resolve_token` returns `None`). Full auth suite (140 tests) passes; `ruff` clean.

---
*AI-assisted: authored with Claude Code. Verified the validate-vs-store asymmetry against the `hosts` sibling and confirmed fail-before/pass-after end-to-end through `resolve_token`.*

### DIFF
--- a/src/specify_cli/authentication/config.py
+++ b/src/specify_cli/authentication/config.py
@@ -13,6 +13,7 @@ import stat
 from dataclasses import dataclass
 from fnmatch import fnmatch
 from pathlib import Path
+from typing import Any
 from urllib.parse import urlparse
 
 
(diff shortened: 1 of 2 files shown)


mnriem:

mnriem:
Thank you!

PR #3692, opened 2026-07-23
fix(bundler): InstallResult.changed counts uninstalled as a change

jawwad-ali:
## What

`InstallResult.changed` only considered `installed` and `refreshed`, omitting `uninstalled`:

```python
return bool(self.installed or self.refreshed)
```

A `bundle update` whose new manifest **drops** components (e.g. the new version ships a reduced/empty `provides` set) removes them via the refresh path — populating `result.uninstalled` — while installing/refreshing nothing. That yields `installed=[]`, `refreshed=[]`, `uninstalled=[dropped set]`, so `changed` returned `False`, **misreporting a mutating update as a no-op**.

## Fix

Include `uninstalled` in the disjunction. It's the third mutating outcome list on the same dataclass (populated by the update-drops-components path) and the sole output of the `remove_bundle` path, so it's a real change signal.

## Tests

`tests/integration/test_bundler_install_flow.py::test_install_result_changed_reports_uninstalled` — an `InstallResult` with only `uninstalled` populated now reports `changed is True` (empty still `False`). Fails before the fix. `ruff` clean.

---
*AI-assisted: authored with Claude Code. Verified fail-before/pass-after against the dataclass's mutating-outcome semantics.*

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
           python-version: "3.14"
 
       - name: Run ruff check
-        run: uvx ruff check src tests
+        run: uvx ruff@0.15.0 check src tests
 
   pytest:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3694, opened 2026-07-23
fix(workflows): guard non-mapping 'workflow:' block in WorkflowDefinition

jawwad-ali:
## What

A present-but-non-mapping top-level `workflow:` block crashed `WorkflowDefinition.__init__` with `AttributeError`:

- bare `workflow:` → YAML `null`
- `workflow: some-string` → `str`
- `workflow: [a, b]` → `list`

`workflow = data.get("workflow", {})` only substitutes `{}` when the key is **absent**, so a present non-dict value reached `workflow.get("id", "")` → `AttributeError: 'NoneType'/'str'/'list' object has no attribute 'get'`.

This fires inside `from_yaml`/`from_string` — **before** `validate_workflow` can report the malformed shape — and in the CLI (`workflow run`/`resume`, whose `load_workflow` is wrapped to catch only `FileNotFoundError`/`ValueError`) it escapes as a raw traceback instead of a clean "Invalid workflow" message.

## Fix

Normalize the local `workflow` to `{}` when it isn't a mapping (`self.data` keeps the raw value so `validate_workflow` still reports it), mirroring the adjacent `default_options` guard (`if not isinstance(...): = {}`).

## Tests

`tests/test_workflows.py::TestWorkflowDefinition::test_non_mapping_workflow_block_parses_then_validates` (parametrized null/str/list) — construction no longer raises, and `validate_workflow` reports the missing `workflow.id`. Fails before the fix (all three crash). `ruff` clean.

---
*AI-assisted: authored with Claude Code. Verified against the `default_options` sibling guard and confirmed fail-before/pass-after, incl. `self.data` round-trip.*

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -42,6 +42,17 @@ class WorkflowDefinition:
         self.source_path = source_path
 
         workflow = data.get("workflow", {})
+        # A present-but-non-mapping ``workflow:`` block (bare ``workflow:`` ->
+        # None, or ``workflow: <str/list>``) would crash the following
+        # ``workflow.get(...)`` calls with AttributeError, so construction fails
+        # before any validation can run. Normalize the local to {} instead: the
+        # header fields fall back to their defaults and ``validate_workflow``
+        # (which reads those parsed attributes) reports the missing
+        # ``workflow.id``/``workflow.name``. ``self.data`` is deliberately left
+        # holding the raw value, since it is what gets written back out when a
+        # definition is serialized. Mirrors the default_options guard below.
+        if not isinstance(workflow, dict):
+            workflow = {}
         self.id: str = workflow.get("id", "")
         self.name: str = workflow.get("name", "")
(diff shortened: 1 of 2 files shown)


mnriem:

mnriem (changes requested):

jawwad-ali:
Correct — fixed in 4b26886. I verified `validate_workflow` never reads `.data` (it reads the parsed attributes `definition.id`/`.name`/`.version`/`.schema_version`); the only reader of `.data` is the `yaml.safe_dump` serialization path. The comment no longer claims a validation dependency:

> Normalize the local to {} instead: the header fields fall back to their defaults and `validate_workflow` (which reads those parsed attributes) reports the missing `workflow.id`/`workflow.name`. `self.data` is deliberately left holding the raw value, since it is what gets written back out when a definition is serialized.

Also rebased onto latest `main` to clear the conflict with the now-merged #3696 (both tests coexist; 5 pass).

mnriem (approved):

mnriem:
Thank you!

PR #3696, opened 2026-07-23
fix(workflows): guard non-mapping 'inputs:' block in engine._resolve_inputs

jawwad-ali:
## What

`WorkflowEngine.execute()` / `resume()` run **unvalidated** definitions (`load_workflow`'s docstring: *"not yet validated; call validate_workflow() separately"*), and `WorkflowDefinition` stores `inputs` raw. So a non-mapping `inputs:` block — bare `inputs:` → `None`, or `inputs: []` → `list` — reached:

```python
for name, input_def in definition.inputs.items():
```

…and crashed the whole run with `AttributeError: 'NoneType'/'list' object has no attribute 'items'`. The steps layer is explicitly hardened for exactly this "engine does not auto-validate" case; the `inputs` container was the gap (the existing per-input `isinstance` skip is only reached *after* the crashing iteration).

## Fix

Return `{}` when `definition.inputs` is not a mapping, mirroring `validate_workflow`'s own `isinstance(definition.inputs, dict)` check. Protects both call sites (`execute` and `resume`); normal dict resolution (defaults + provided overrides) is unchanged, and `validate_workflow` still reports the malformed shape.

## Tests

`tests/test_workflows.py::TestWorkflowDefinition::test_resolve_inputs_tolerates_non_mapping_inputs` (parametrized `inputs: []` and bare `inputs:`) — `_resolve_inputs` returns `{}` instead of raising. Fails before the fix (`AttributeError`). `ruff check` reports no new issues (the pre-existing engine.py `I001` is unrelated and present on main).

---
*AI-assisted: authored with Claude Code. Verified against the `validate_workflow` sibling check and confirmed fail-before/pass-after.*

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -1400,6 +1400,14 @@ class WorkflowEngine:
     ) -> dict[str, Any]:
         """Resolve workflow inputs against definitions and provided values."""
         resolved: dict[str, Any] = {}
+        # execute()/resume() accept UNVALIDATED definitions (load_workflow does
+        # not validate). A non-mapping ``inputs:`` block (bare ``inputs:`` ->
+        # None, or ``inputs: []``) is stored raw, so iterating ``.items()`` here
+        # would crash the run with AttributeError. Treat a non-mapping inputs
+        # block as "no inputs"; validate_workflow reports the malformed shape
+        # via its own isinstance check.
+        if not isinstance(definition.inputs, dict):
+            return {}
         for name, input_def in definition.inputs.items():
             if not isinstance(input_def, dict):
                 continue
--- a/tests/test_workflows.py
+++ b/tests/test_workflows.py
@@ -3612,6 +3612,23 @@ class TestWorkflowDefinition:
         assert definition.id == "test-workflow"
         assert len(definition.inputs) == 2
 
+    @pytest.mark.parametrize(
+        "block",
+        [
+            "workflow:\n  id: w\n  name: W\nsteps: []\ninputs: []\n",   # list
+            "workflow:\n  id: w\n  name: W\nsteps: []\ninputs:\n",       # null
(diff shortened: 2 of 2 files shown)


mnriem (changes requested):
Please pull in upstream/main

jawwad-ali:
Heads-up: the failing `ruff` check here is **not** from this change. `uvx ruff check src tests` on a clean `main` currently reports **1476 errors** — dominated by rules the repo does not select in `[tool.ruff.lint]` (which only `extend-select`s S602/604/605): 733× `I001`, 161× `UP006`, 105× `RUF012`, 92× `PLW1510`, 49× `BLE001`, etc. A recent `ruff` release expanded its default rule set, so the unpinned `uvx ruff` in CI now lights up long-standing code repo-wide.

This affects every open PR, not just this one (the `SUCCESS` you see on older PRs is cached from before the ruff release). Pinning ruff in `.github/workflows/test.yml` (e.g. `uvx ruff@<last-green-version> check src tests`) would restore green CI across the board. This PR's own change touches only `engine._resolve_inputs` (a 2-line guard) and introduces no new lint findings. Happy to send a separate ruff-pin PR if that'd help.

jawwad-ali:
Rebased onto the latest `upstream/main` (now on top of `4d3a428`, the 0.14.2.dev0 release commit). 👍

One note on the red `ruff` check: it is **not** caused by this PR's diff — it's a repo-wide regression from `uvx ruff` (unpinned) resolving a newer Ruff whose expanded default rule set now flags ~1476 pre-existing findings across `src`/`tests` on a clean `main`. `pytest` is green on all six OS/Python combinations here. I opened #3705 to lock the Ruff rule set explicitly so CI goes green repo-wide again; once that lands (or on the next rebase after it) this check will pass.

mnriem (approved):

Issue #23, opened 2025-09-04
Proposal: Integrate VeriTrail for hallucination detection & provenance in SpecKit


richgo:
Disclosure: I work for a bank, where is a lot of distrust for generated content. Given the early nature of both SpecKit and VeriTrail (which I hope will be open sourced) this may well be a consideration for later iterations of SpecKit.

Summary
It would be valuable to explore integration between SpecKit and Microsoft Research’s VeriTrail, a framework for hallucination detection and provenance tracing in multi-step AI workflows. https://www.microsoft.com/en-us/research/blog/veritrail-detecting-hallucination-and-tracing-provenance-in-multi-step-ai-workflows/

Motivation
- Multi-step AI workflows in spec generation risk compounding hallucinations; catching errors only at the final step is insufficient.
- VeriTrail’s DAG-based approach can trace generated specs back to their sources and highlight exactly where unsupported content was introduced.
- Improved trust and adoption: Developers would gain confidence in AI-assisted spec generation when outputs are verifiable and auditable.

Benefits
- Detect hallucinations earlier and with greater precision
- Provide provenance mapping from final spec outputs to intermediate steps and source content
- Enhance transparency, auditability, and debugging in SpecKit workflows

tiagostutz:
I like the idea, but I'd propose support for something a bit more generic like linters (maybe called `evaluators`?), being VeriTrail one of those possible linters, but then something that can be registered/extended, so people could even use their private/paid evaluators if wanted and community could also build new ones.

At the end, I think we are looking for a way to register "linters" that would run after each phase (spec, plan, task) and that would trigger iteration of those in case of linting error. We just need a way to be able to register those linters. Btw, that linter capability is what made JS/TS grow as it grown, so this can be an inspiration here.

Thoughts?

localden:
Thank you for the suggestion. We flagged this for future investigation, but it won't be immediately part of Spec Kit.

PR #22, opened 2025-09-04
feat: add Codex CLI support

danizord:

PR #21, opened 2025-09-04
fix minor typo in constitution's Article I

thegovind:
This pull request updates the documentation to reflect the correct product name in the constitution's first article. The change ensures consistency in naming conventions throughout the documentation.

* Documentation update: Changed the product name from `Specify2` to `Specify` in Article I to maintain accurate and consistent terminology.

Issue #17, opened 2025-09-04
Create extensions folder for Gemini CLI


nalbion:
https://github.com/google-gemini/gemini-cli/blob/main/docs/extension.md

Commands should be in `<workspace>/.gemini/extensions/spec-kit/` with `gemini-extension.json` and `GEMINI.md` for spec-kit.

Are we supposed to create our own `GEMINI.md` file in the project root, or is spec-kit helping to create that?

PR #3704, opened 2026-07-23
feat: first-class agent-native runtime hooks for integrations

kanfil:
### 🎯 Problem & Motivation

While Spec Kit excels at scaffolding commands and prompt templates during `specify init`, there was previously no first-class mechanism for **event-driven execution** natively tied to AI coding agents in upstream.

This PR introduces an **agent-native runtime events layer** that bridges Spec Kit directly to agent lifecycle events (canonical snake_case names: `session_start`, `pre_tool_use`, `post_tool_use`, `session_end`, `user_prompt_submit`, `stop`).

---

### ✨ Key Features Added

1. **The Event Dispatcher (`.specify/events.py`)**:
   A lightweight, zero-dependency Python script scaffolded in projects during `specify init`. It maps registered Spec Kit commands to their underlying scripts and executes them on demand, passing any payload through `stdin`.

2. **Agent Event Adapters** (registry-driven, one per integration):
   Target-aware adapters that generate/merge native event configurations seamlessly during `setup()` and `teardown()`:
   - **Claude Code**: `.claude/settings.json` (nested JSON, single `command` string)
   - **Cursor**: `.cursor/hooks.json` (flat, camelCase structure)
   - **Codex**: `.codex/config.toml` (TOML format)
   - **Copilot CLI**: `.github/hooks/speckit.json` (dedicated file, bash/powershell variants)
   - **opencode**: `.opencode/plugin/speckit-events.ts` + `opencode.json` (TypeScript plugin)
   - **Gemini, Qwen, Devin, Tabnine**: Native settings merge

3. **Layered Event Resolution**:
   Events are resolved cleanly using a 4-tier priority stack:
   - **Layer 1 (CLI)**: Passing `--events false` during init/upgrade disables events entirely (and strips prior hooks).
   - **Layer 2 (User Override)**: Users can create `.specify/integration-events.yml` to override event behaviors.
   - **Layer 3 (Extension-Declared)**: Extensions declare `events:` in their `extension.yml` (completely decoupling core code from extension details). Multiple extensions declaring the same event all run.
   - **Layer 4 (Built-in Defaults)**: Native integration class defaults.

   Adapters translate canonical snake_case event names to each agent's native casing (Claude `SessionStart`, Copilot/Cursor `sessionStart`, Gemini `BeforeTool`, …), so extension authors never learn agent-specific names.

---

### 🛠️ Example Usage

An extension can declare events in `extension.yml`:

```yaml
extension:
  id: "team-ai-directives"
  name: "Team AI Directives"
  version: "1.0.0"

events:
  session_start:
    command: "speckit.team-directives.boot"
    description: "Load team rules and constitution on session startup"
```

When the user runs `specify init --integration claude`, the Claude adapter automatically detects this, generates the Event Dispatcher, and injects the hook into `.claude/settings.json`. When Claude Code launches, it natively triggers the dispatcher, booting the team context seamlessly. Installing/enabling/disabling an extension refreshes each installed integration's native event config automatically.

---

### 🧪 Verification & Dependencies

- **Dependencies**: Uses `pyyaml` (already a spec-kit dependency) to parse extension-declared events. No new external dependencies.
- **Teardown**: Run `specify integration uninstall` to cleanly strip all native event config entries/plugins, leaving the user's files exactly as they were. The shared `.specify/events.py` dispatcher is reference-counted so it is only deleted when no installed event-capable integration still references it.
- **Unit Tests**: Includes a suite of unit tests verifying layered event resolution, override parsing/validation, extension collection (with registry `enabled` honoring), multi-handler-per-event accumulation, adapter configurations, idempotent merge/teardown, JSONC preservation, and safe-write containment.


mnriem:
Thanks for this — the agent-runtime wiring is a real gap and the canonical→native event mapping, marker-based surgical teardown, and layered resolution are the right instincts. Sharing some architectural feedback before this goes further; happy to pair on any of it.

**1. Call the feature "events", and use snake_case names consistent with the existing event vocabulary.** `hooks` is already load-bearing in Spec Kit: extensions declare `hooks:` in `extension.yml`, there's a documented Hook System + `HookExecutor`, and those fire on *SDD-workflow* points (`before_specify`, `after_tasks`) as soft/in-prompt prompts. This PR fires on *agent-runtime* points (`SessionStart`, `PreToolUse`) wired into native agent config — genuinely different, and reusing "hooks" will confuse users and docs. Suggest:
- extension field `events:` (not `runtime_hooks`),
- CLI flag `--events`,
- canonical event names in the same `xx_yyy` style as the existing workflow events: `session_start`, `pre_tool_use`, `post_tool_use`, `session_end`, `user_prompt_submit`, `stop`,
- dispatcher `events.py`, marker `__speckit_event__`.

Adapters translate those canonical snake_case names to each agent's native casing (Claude `SessionStart`, Copilot/Cursor `sessionStart`, Gemini `BeforeTool`, …), so extension authors never learn agent-specific names.

**2. Drive adapters from the integration registry, not a parallel `_ADAPTERS` dict.** AGENTS.md makes the integration registry the single source of truth; the hardcoded `{"claude": ..., "cursor-agent": ...}` map means new agents must be registered twice and keys can drift. Fold each adapter into its integration class as `CANONICAL_TO_NATIVE` + `events_config_file` + `emit_events()`/`remove_events()` the registry drives — the same pattern Copilot already uses to own `.vscode/settings.json`. This also collapses the 4 duplicated insertions in `base.py` into one path.

**3. Split sourcing from writing so `integrations/` never imports `extensions/`.** Right now `_hooks.py` both *writes* native config and *sources* declarations by scanning `.specify/extensions/*/extension.yml` (`from ..extensions import ValidationError`). Those are two responsibilities:
- **Writing** native config is legitimately integration-domain (like Copilot's settings.json) → keep it in the integration class.
- **Sourcing** the event→command map (built-in defaults + extension `events:` + user override) should move out: have the init orchestration resolve it and **inject** it via `setup(..., events=map)`. The integration owns the write but doesn't reach up into the extension system.

**4. The dispatcher is agent-agnostic → make it core.** One `bridge.py` serves every agent and re-implements command→script resolution that already lives in `_commands.py`/the registrar. Lift it to a core `.specify/events.py` that reuses the existing resolver instead of duplicating it.

**5. Two capabilities the PR description advertises aren't actually wired yet — make sure the reworked versions are real and covered by tests.** As filed: the disable flag reads `parsed_options.get("hooks")`, but no integration declares that option via `options()` and there's no top-level flag, so it can't be set the way the description implies; and `validate_runtime_hooks()` documents itself as "called from `ExtensionManifest._validate()`", but the manifest isn't touched in this PR (only 3 files changed), so malformed declarations are never validated at load. After the rework that means: `--events` should be a real, declared gate, and event validation should actually be invoked from the manifest validator — both with end-to-end tests.

**6. Missing Copilot CLI adapter.** Copilot CLI has first-class hooks ([docs](https://docs.github.com/en/copilot/reference/hooks-reference)): native events (`sessionStart`, `preToolUse`, `postToolUse`, `sessionEnd`, `userPromptSubmitted`), per-entry `bash`/`powershell` variants, and a dedicated `.github/hooks/*.json` location. Writing our own `.github/hooks/speckit.json` (rather than merging a shared settings file) makes teardown a plain file-delete with no marker scanning, and the `bash`/`powershell` split maps cleanly onto Spec Kit's existing `scripts: {sh, ps}` frontmatter. Note Copilot also reads cross-tool `.claude/settings.json`, so a project with both integrations needs a de-dupe guard in the resolver.

**Smaller correctness notes:** the Codex adapter embeds `$(git rev-parse --show-toplevel)` inside a TOML command string (relies on the agent shell-expanding it); several native event names (Cursor `beforeSubmitPrompt`, Gemini `BeforeTool`/`AfterAgent`, Tabnine) read as best-guesses worth verifying per-agent or marking experimental; and silent writes into `.claude/settings.json` etc. during `init` should be surfaced in output.

Net: the writing mechanism belongs in the integration layer as you have it — the main structural change is **splitting the one module in two**: native-config writing stays per-integration (registry-driven), while event *sourcing* moves to an injected core resolver, with the dispatcher lifted to core. Add the terminology change and the Copilot adapter and I think this is in great shape. Glad to help implement or review.

---
*Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8).*

mnriem (changes requested):
Please address my feedback above. Great proposal!

mnriem:

kanfil:
Posted on behalf of @kanfil by opencode (model: glm-5.2, autonomous).

Addressed all 23 Copilot inline review findings plus the ruff lint errors. Pushed as four commits on top of `da6c20d9` (head now `2623ddbd`):

- `6ec91bb2` — lint: E402 noqa on the event-command import, removed unused `console`/`patch`/`yaml`/`Path`/integration imports, fixed F541 f-strings. `uvx ruff@0.15.0 check src tests` is clean.
- `61f0a45e` — native-config generation: resolved events map now carries an ordered list of handlers per event (#2, so two extensions declaring the same event both run); each adapter renders a single complete shell `command` string (#6, Claude/Gemini/etc. reject `command`+`args`) using a portably-resolved Python interpreter instead of hard-coded `python3` (#16); Gemini timeouts converted to ms (#7); the core runner resolves the project's sh/ps/py variant and splits the script command into argv instead of treating `scripts:` strings as bare paths (#4), with the bundled-template fallback using the canonical `_assets` resolvers (#5); OpenCode TS plugin failures throw instead of `process.exit(2)` (#13); user YAML override validated before returning (#21).
- `0d546a2e` — merge/teardown safety: marker detection recurses into nested `hooks` (#9); merge strips all Specify-marked entries before adding the new set so a changed override doesn't leave stale events (#11); empty resolved map (`--events false`) now strips prior hooks instead of early-returning (#3); Spec-Kit-created configs that become empty are deleted, not left as `{}` (#14); the shared `.specify/events.py` dispatcher is reference-counted across installed event-capable integrations (#10); Copilot's `.github/hooks/speckit.json` merges owned entries and teardown removes only owned entries (#8); JSONC/unparseable native configs are preserved instead of reset to `{}` (#22, #23); write destinations are validated for symlinks/containment before any bytes are written (#12).
- `2623ddbd` — lifecycle/validation/contract: `collect_extension_events` honors the registry's `enabled` flag and a new `refresh_integration_events()` is wired into extension add/remove/enable/disable so native event config regenerates on extension state changes (#1); `validate_events` requires `command` to be a non-empty string, not just truthy (#17); PR description updated to the implemented `events`/`--events`/`integration-events.yml`/`events.py` terminology (#15).

Tests: `tests/integrations/test_events.py` grew from 23 to 46 tests (added regressions for each finding). 189 adapter tests and 14 extension tests still pass. Version bumped `0.14.2.dev0` → `0.14.2.dev2` with CHANGELOG entries for the `__init__.py` touches (AGENTS.md convention).

One implementation note for #2: I moved it into the native-config tier (rather than the lifecycle tier in the original plan) because it is inseparable from the adapter command-string rendering — both rewrite the same formatters against the events-map shape, so doing them together avoided a throwaway intermediate rewrite.

mnriem:
Please do not update CHANGELOG.md nor pyproject.toml as those are done by release process

mnriem:

kanfil:
Posted on behalf of @kanfil by opencode (model: glm-5.2, autonomous).

Addressed all 11 actionable findings from Copilot review 4790195897 (the 4 high-confidence inline + 7 low-confidence suppressed). Pushed as five commits on top of `2623ddbd` (head now `be43bc4c`):

- `c5b9a93f` — per maintainer request, reverted the CHANGELOG.md/pyproject.toml version bumps from the prior events commits. Both files now match their pre-PR (`da6c20d9`) state, so the PR makes no net change to them. (Bonus: this also clears the pending merge conflicts with upstream/main's 0.14.2 release — our side no longer diverges on those two files.)
- `aaab4d23` — #8/#9: `DevinIntegration.options()` and `CopilotIntegration.options()` now compose with `super().options()` so `--events` is declared alongside `--skills`; previously `--integration-options "--events false"` was rejected as unknown for both adapters.
- `524c2107` — #7: Cursor's `.cursor/hooks.json` now gets top-level `version: 1` (added a `version` kwarg to `_merge_json_fragment`, preserving a user's value). S3: json-nested groups handlers by distinct matcher (one matcher-group per matcher) instead of placing all under the first handler's matcher. S4: Copilot's `bash`/`powershell` fields get independent OS-targeted interpreters (`python3` / `python`) so a config generated on one OS works on the other.
- `ceee5071` — S2: the `py` variant now anchors `scripts/...` under `.specify/` (shared with sh/ps) before prepending the interpreter. S6: the `ps` variant prefixes argv with `pwsh -File` (fallback `powershell -File`) so `subprocess.run(shell=False)` can execute `.ps1` on Windows.
- `be43bc4c` — S5: merge helpers return bool; install branches skip `manifest.record_existing()` when a merge is skipped on parse failure, so an untouched JSONC/malformed config isn't tracked and `uninstall()` can't later delete it. S1: `remove_integration_events` drops this integration's manifest claim on the shared dispatcher even when the file is retained, so the subsequent `manifest.uninstall()` in `teardown()` doesn't delete a file another integration references (test now exercises full `teardown()`, not just `remove_integration_events`). S7: `refresh_integration_events` passes each integration's stored `parsed_options` to `resolve_events`, so a persisted `--events false` survives extension add/enable/disable. #10: an invalid override entry abandons the whole override and keeps accumulated built-in+extension layers (instead of resetting to `{}` and disabling all hooks on a typo); only a fully-valid override, including explicit `events: {}`, replaces.

Tests: `tests/integrations/test_events.py` grew 56→61 tests; added regressions for each finding (options composition for all event-capable adapters, Cursor version + preservation, matcher grouping per-distinct/shared, Copilot bash≠powershell, py anchoring, ps launcher prefix, override preserve-layers, skipped-merge not tracked, full-teardown dispatcher retention, refresh honors `--events false`). 312 adapter tests still pass. `uvx ruff@0.15.0 check src tests` clean.

Note on stale inline comments: GitHub re-anchored the six comments from the original review (`original_commit_id: da6c20d9`) onto the new HEAD — they describe pre-fix code and correspond to findings already addressed in the prior round (#3 empty-map removal, #7 Gemini ms, #9 marker recursion, #12 safe-write, #13 opencode throw, #15 PR contract), so they're not actionable here.

mnriem:

kanfil:
Posted on behalf of @kanfil by opencode (model: glm-5.2, autonomous).

Fixed the CI-blocking stale test and addressed all 7 findings from Copilot review 4791088500 (5 inline + 2 suppressed). Pushed as three commits on top of `be43bc4c` (head now `6761c948`); CHANGELOG.md/pyproject.toml remain untouched per the earlier instruction.

- `c74e5a44` — CI: `tests/test_extensions.py::test_no_commands_no_hooks` matched the old validation string; updated the regex to the current `"must provide at least one command, hook, or event"` message (this was the sole failing test).
- `123e10a3` — data-loss & resolution:
  - S9: `_remove_native_event_hooks` now unconditionally drops this integration's manifest claim on the native config (not only when the file is deleted). Previously a config whose owned entries were cleaned but user content retained stayed tracked, so `teardown(force=True)` → `manifest.uninstall(force=True)` deleted the entire user-owned settings file. Config-file mirror of the earlier shared-dispatcher fix.
  - S8: `_find_command_template` resolved extension event commands via a broken registry lookup (the registry stores per-agent `registered_commands` name-lists, not a `{name, file}` map) and a file-stem scan that only matched when the `.md` stem equaled the command name. A manifest mapping `speckit.selftest.extension → commands/selftest.md` resolved as missing. It now enumerates installed extensions via `ExtensionManager.get_extension()` and matches `provides.commands[].name → file`, with the directory scan and core-template lookups kept as fallbacks.
  - R3: `_remove_toml_entries` now validates the destination with `_ensure_safe_destination` before read/write (the merge path already did), so a symlink swap of `.codex/config.toml` after install can't make teardown overwrite a file outside the project.
- `6761c948` — native-config hardening:
  - R1: the generated dispatcher and `resolve_and_run_event_command` now run their subprocesses with `cwd` = the project root. Previously `specify event run` (and the resolved script) inherited the agent's CWD while `event_run` resolves the project via `Path.cwd()`, so a hook fired from a subdirectory targeted the wrong project.
  - R2: `_dispatcher_command` shell-quotes each component for the target shell (POSIX `shlex.quote`; PowerShell single-quoted literals), so an interpreter path with spaces or an override `command` with shell metacharacters is passed as one argument. Claude's `${CLAUDE_PROJECT_DIR}` prefix is left unquoted so the shell still expands it.
  - R4: the Codex TOML `matcher` is now rendered through the shared TOML escaper like `command`, so a matcher with a quote/backslash/newline/control char no longer produces malformed `config.toml`.
  - R5: Tabnine declares `events_timeout_unit="ms"` (its hook schema mirrors Gemini's `BeforeTool`/`AfterTool`), so the 60s default becomes 60000ms instead of `timeout: 60` (60 ms).

Tests: `tests/integrations/test_events.py` grew 64→70 tests; added regressions for forced-teardown user-config preservation, manifest-driven command resolution (stem ≠ name), TOML teardown symlink rejection, cwd-forced execution from a subdirectory, POSIX/PowerShell quoting of space/metacharacter components, TOML matcher escaping, and Tabnine ms conversion. Full events suite + 235 adapter tests pass; `uvx ruff@0.15.0 check src tests` clean.

Note on R5: Tabnine's ms timeout unit is inferred from its Gemini-parity native mapping (`BeforeTool`/`AfterTool`) since the repo doesn't pin Tabnine's hook schema; happy to revert if Tabnine actually uses seconds.

kanfil:
Posted on behalf of @kanfil by opencode (model: glm-5.2, autonomous).

Fixed the three Windows CI failures and addressed all 15 findings from the latest Copilot review (5 inline + 2 suppressed high-confidence, plus 8 lower-confidence generated comments). Pushed as six commits on top of `6761c948` (head now `9b3f2496`); CHANGELOG.md/pyproject.toml remain untouched per the standing instruction.

**Windows CI (3 failures):**
- `ea333696` — `EVENTS_DISPATCHER_REL` was `str(Path(".specify")/"events.py")`, which yields `.specify\events.py` on Windows; manifest keys are POSIX-form, so the shared-dispatcher manifest-claim drop was skipped and `uninstall(force=True)` deleted the dispatcher another integration depended on. Now a POSIX constant (`.as_posix()`). Also made the py/ps argv assertions platform-agnostic (POSIX-form comparison, case-insensitive launcher stem) so they hold on Windows.

**Copilot review — 15 findings:**
- `f7715a5c` — C4: a malformed override handler (`stop: []`) now abandons the whole override (keeps prior layers) instead of disabling all hooks. C6: a non-mapping integration entry (`claude: bad`) is ignored as malformed, not treated as a valid disable. C10: `matcher` validated as a string in both manifest and override paths (a `matcher: []` previously crashed `by_matcher.setdefault`). C11: event `command` refs are now canonicalized via `rename_map` in `ExtensionManifest._validate`, mirroring hook refs.
- `a51ff920` — C3: shared dispatcher added to `events_stale_exclusions` so an `--events false` upgrade can't delete it via the generic stale pass and break other integrations. C5: a Spec-Kit-created Cursor file retaining only `{"version": 1}` is now deleted on teardown (not left as a stub). C12: `refresh_integration_events` resolves first and installs once; a mid-refresh failure no longer destroys the working config (the destructive pre-strip is gone — `install_integration_events` already handles cleanup).
- `cb66271e` — C1: `target_os="host"` always uses POSIX `shlex.quote` (safe tokens bare on every platform); the previous Windows-PS `'python'` form wasn't invocable without `&`. PowerShell quoting is now only for the explicit `windows` target (Copilot's powershell field), prefixed with `& `. C2: Claude's `${CLAUDE_PROJECT_DIR}` dispatcher is double-quoted so the variable still expands but a spaced project path doesn't word-split.
- `e113956a` — opencode TS plugin rewrite: C8 dispatcher+interpreter resolved from the `directory` OpenCode passes (not `process.cwd()`); C9 launched via `execFileSync` with an argv array (no shell, no injection); C7 both `input` and `output` forwarded to `runEvent` so pre/post_tool_use can inspect args/results.
- `9b3f2496` — U1: Qwen `events_timeout_unit="ms"` (Qwen Code command hooks use ms per its docs). U2: Devin `json-root-nested` format — its `hooks.v1.json` is a root event map with no `hooks` wrapper (per Devin docs); added `_merge_json_root`/`_remove_json_root_entries`. U3: Copilot `"stop": "agentStop"`.

**Verification:** `tests/integrations/test_events.py` 70→85 tests (all pass); the previously-failing `test_extensions.py::test_no_commands_no_hooks` passes; 312 adapter tests pass; `uvx ruff@0.15.0 check src tests` clean. CHANGELOG.md/pyproject.toml net-diff vs the pre-PR base is empty.

Note on U1/U2/U3: these were verified against each agent's published hook documentation (Qwen Code command-hook timeout = ms; Devin `hooks.v1.json` = root event map, "the hooks object is the entire file"; Copilot `agentStop` native name). Happy to revert any if a maintainer confirms otherwise.

mnriem:

kanfil:
Posted on behalf of @kanfil by opencode (model: glm-5.2, autonomous).

Addressed all 10 findings from the latest Copilot review (3 inline + 7 suppressed). Pushed as six commits on top of `9b3f2496` (head now `dbf1fcb5`); CHANGELOG.md/pyproject.toml remain untouched per the standing instruction.

- `7b12ebef` — R1: `collect_extension_events` now reads events from a validated `ExtensionManifest` (whose command refs were canonicalized at install validation, C11) instead of the raw `extension.yml` YAML; an obsolete ref like `my-ext.boot` is now resolved to `speckit.my-ext.boot` so refresh doesn't emit a hook that silently no-ops. R3: `refresh_integration_events` accumulates per-integration failures and raises `EventRefreshError` so the extension lifecycle commands (add/remove/enable/disable) can't claim full deactivation while a stale native hook may still be active; a `_refresh_events_and_warn` helper surfaces the aggregated failures at each call site.
- `78ffdf94` — R2: the dispatcher's `_find_specify` and the opencode TS resolver now probe a project venv python for `specify_cli` importability before selecting it, so an unrelated project venv no longer shadows the PATH `specify` fallback and fails every event. S2: the opencode TS PATH fallback is now `python` on Windows (where `python3` is commonly absent), `python3` on POSIX.
- `44bebf88` — S1: `_build_opencode_plugin` serializes every interpolated value (command, event, native hook key, matcher tool names) as a JSON string literal, so a quote/backslash/backtick in a command or matcher can't break the generated TypeScript or inject code.
- `ba3f1aed` — S4: the per-handler timeout now flows through the chain (native command → dispatcher 4th arg → `event run` → inner runner) instead of a fixed 120s cap that killed handlers configured for longer. S5: the `sh` script variant prefixes a bash/sh launcher on Windows so `subprocess.run(shell=False)` can execute `.sh` files (mirroring the `ps` branch's `pwsh -File`).
- `9f721ed6` — S3: the empty-resolved-map install path (`--events false` upgrade) now runs the shared-dispatcher refcount cleanup, so the last event integration disabling events deletes `.specify/events.py` instead of orphaning it permanently (the new manifest wouldn't claim it and stale cleanup excludes it).
- `dbf1fcb5` — S6: Gemini `user_prompt_submit` → `BeforeAgent` (verified against Gemini CLI's hooks docs). S7: Tabnine `user_prompt_submit` → `BeforeAgent`, `stop` → `AfterAgent` (Tabnine's Gemini-compatible schema).

**Verification:** `tests/integrations/test_events.py` 86→94 tests (all pass); 312 adapter tests pass; `uvx ruff@0.15.0 check src tests` clean. CHANGELOG.md/pyproject.toml net-diff vs the pre-PR base is empty.

Note on S6: Gemini's `BeforeAgent` mapping was verified against the Gemini CLI hooks documentation (BeforeAgent fires after the user submits a prompt, before planning). S7 (Tabnine) is inferred from Tabnine's deliberate Gemini-schema parity; happy to revert if Tabnine's docs say otherwise.

mnriem:

kanfil:
Posted on behalf of @kanfil by opencode (model: glm-5.2, autonomous).

Addressed all 6 findings from the latest Copilot review (2 inline + 4 suppressed). Pushed as five commits on top of `dbf1fcb5` (head now `0a5b1ddc`); CHANGELOG.md/pyproject.toml remain untouched per the standing instruction.

- `3f676c2e` — R2: `_dispatcher_command` was passing `_native_timeout(timeout_seconds)` as the dispatcher's 4th argument, but the dispatcher interprets it as seconds — so for Gemini/Qwen/Tabnine (ms adapters) 60s became 60000s (~16h). It now passes raw seconds; the +5s buffer moves to the native hook timeout field (`_native_timeout(seconds + EVENT_TIMEOUT_BUFFER)`) so the agent's outer cap fires after the dispatcher's inner subprocess, killing the child cleanly instead of orphaning it. S3: the opencode TS `runEvent` now accepts a per-handler `timeoutSec` — `execFileSync` uses `(timeoutSec + buffer) * 1000` ms and appends the seconds to the dispatcher argv, so handlers configured above 60s aren't killed early.
- `0991d874` — S1: `_find_command_template` now filters `enabled: false` (shared `_disabled_extension_ids` helper, reused by `collect_extension_events`) in both the manifest-resolution loop and the on-disk fallback scan, so a stale hook can't keep executing a disabled extension.
- `81551645` — S2: `_cleanup_shared_dispatcher` now deletes the shared dispatcher based on the refcount alone, independent of whether the fresh upgrade manifest currently claims it. Previously `integration upgrade --events false` passed a fresh manifest that never recorded the dispatcher, so the condition was false and the old on-disk dispatcher was orphaned (stale cleanup excludes it).
- `56b311bb` — S4: `extension_update` now refreshes native event config after a successful update (it replaces `extension.yml` but previously never refreshed, leaving event declarations stale until a manual integration upgrade).
- `0a5b1ddc` — R1: the generated `.specify/events.py` dispatcher is now self-contained. It prefers `import specify_cli` (durable pip/pipx/uv-tool install) and delegates to `resolve_and_run_event_command`; when the package isn't importable (e.g. a one-time `uvx` init whose environment is discarded) it falls back to an inline stdlib-only resolver that finds the command template, parses its `scripts:` frontmatter, resolves the project's script variant, and runs the script with the correct launcher — so hooks work without a persistent `specify` executable on PATH. Both paths verified end-to-end.

**Verification:** `tests/integrations/test_events.py` grew 94→98 tests (all pass); 312 adapter tests pass; `uvx ruff@0.15.0 check src tests` clean. CHANGELOG.md/pyproject.toml net-diff vs the pre-PR base is empty.

mnriem:

kanfil:
Posted on behalf of @kanfil by opencode (model: glm-5.2, autonomous).

Addressed feedback from Copilot review 4801895963. Pushed as three commits on top of `0a5b1ddc` (head now `16860e30`); CHANGELOG.md/pyproject.toml remain untouched per the standing instruction.

- `570cdbb5` — Safe destination validation on removals: added `_ensure_safe_destination` guards to all removers (`_remove_json_entries`, `_remove_copilot_entries`, `_remove_json_root_entries`, `_remove_opencode_entries`), `_cleanup_shared_dispatcher`, and `remove_integration_events` unlinks. Prevents teardown from modifying or deleting external files if a config directory or file is replaced with a symlink.
- `a2e08b3b` — Manifest-driven resolution in dispatcher template: updated `_find_command_template` in `_EVENTS_DISPATCHER_TEMPLATE` to filter disabled extensions (`enabled: false`) from `.registry` and parse `provides.commands` in `extension.yml` so commands whose file stem differs from the command name (e.g. `speckit.selftest.extension` -> `commands/selftest.md`) resolve correctly when `specify_cli` is unavailable (`uvx` one-time installs).
- `16860e30` — Timeout validation & OpenCode error aggregation: added positive non-boolean integer validation for `timeout` in `validate_events` and `_validate_resolved_event`. Updated `_build_opencode_plugin` so multi-handler event functions wrap each handler call in try/catch, execute all handlers for the event, and throw an aggregate error at the end if any handler failed.

**Verification:** `tests/integrations/test_events.py` (103 tests pass); `tests/test_extensions.py::test_no_commands_no_hooks` passes; 312 adapter tests pass; `uvx ruff@0.15.0 check src tests` clean. CHANGELOG.md/pyproject.toml net-diff vs the pre-PR base is empty.

**Note on Qwen timeout unit (Suppressed #5):** Qwen Code's official documentation (`QwenLM/qwen-code/docs/users/features/hooks.md`) specifies milliseconds for Command hooks (`timeout` default 60000ms). We intentionally kept `events_timeout_unit = "ms"` for Qwen.

mnriem (approved):

mnriem:
Thank you! Looking forward to see what everyone is going to do with your contribution :)

kanfil:
Thank youm you can inject the constitution enforcement using an event.

PR #3706, opened 2026-07-24
fix(workflows): reject non-string/non-boolean 'condition' in if/while/do-while steps

jawwad-ali:
## Problem

`IfThenStep`, `WhileStep`, and `DoWhileStep` `validate()` confirm that `condition` is **present** but never check its type.

At run time `execute()` feeds `condition` to `evaluate_condition()`, which first delegates to `evaluate_expression()` — that returns a non-string **unchanged** — and then coerces the result with `bool()`:

```python
def evaluate_expression(template, context):
    if not isinstance(template, str):
        return template   # non-string returned as-is
    ...

def evaluate_condition(condition, context):
    result = evaluate_expression(condition, context)
    ...
    return bool(result)
```

So a **list/dict/number** condition silently resolves to its truthiness instead of erroring on the authoring mistake:

- `condition: [1, 2]` → `bool([1, 2])` → **always True** → the if-branch is always taken / the while-loop spins to `max_iterations`
- `condition: {}` → always False, etc.

No error is reported, so the mistake is invisible.

## Fix

Reject a present-but-non-`str`/non-`bool` `condition` in each of the three steps' `validate()`, mirroring the existing `prompt`/`shell`/`command` type checks.

### Why booleans are accepted (contract note)

An earlier revision of this PR rejected **every** non-string, including `bool`. That was wrong — it would have been a **breaking change**, so the validator now accepts `(str, bool)`:

- An unquoted `condition: false` is idiomatic YAML, and it already resolves **correctly** today: `evaluate_expression` passes the bool through and `evaluate_condition`'s `bool()` is a no-op. Verified on `main`: `evaluate_condition(False)` → `False`, `evaluate_condition(True)` → `True`, and end-to-end `condition: false` takes the else-branch while `condition: true` dispatches `then`.
- Two of the three steps **default** `condition` to the bool `False` themselves (`config.get("condition", False)`), so `bool` is the field's own natural type — not an authoring mistake.

A bool is therefore not part of the silent-coercion bug this PR fixes; only `list`/`dict`/`int`/`float` are. `"true"`/`"false"` and expressions like `"{{ inputs.flag }}"` are strings and stay valid, as before.

## Verification

- New parametrized tests (`[list, dict, int, float]`) across all three step classes: **fail before** the guard, **pass after**.
- `test_validate_accepts_string_or_bool_condition` covers `"true"`/`"false"`/`"{{ ... }}"` **and** literal `True`/`False` in each step — the regression guard for the contract above.
- Full `TestIfThenStep`/`TestWhileStep`/`TestDoWhileStep` green; `ruff@0.15.0` clean.

---
*AI-assisted: authored with Claude Code. I traced the runtime path (`evaluate_expression` → `evaluate_condition` → `bool()`), confirmed the silent coercion on `main`, and confirmed that literal booleans already behave correctly there — which is why they are accepted rather than rejected.*

### DIFF
--- a/src/specify_cli/workflows/steps/do_while/__init__.py
+++ b/src/specify_cli/workflows/steps/do_while/__init__.py
@@ -72,12 +72,14 @@ class DoWhileStep(StepBase):
             )
         elif not isinstance(config["condition"], str):
             # The engine re-evaluates 'condition' via evaluate_condition() after
-            # each iteration; it returns a non-string as-is and takes bool() of
-            # it -- so a list/dict/number condition silently resolves to a
-            # truthiness (e.g. condition: [1, 2] is always truthy, looping to
-            # max_iterations) with no error. Reject non-strings at validation,
-            # mirroring the prompt/shell/command 'must be a string' checks.
-            # "true"/"false" and an expression like "{{ ... }}" stay valid.
+            # each iteration. That call first delegates to
+            # evaluate_expression() -- which returns a non-string unchanged --
+            # and then coerces the result with bool(). So a list/dict/number
+            # condition silently resolves to its truthiness (e.g.
+            # condition: [1, 2] is always truthy, looping to max_iterations)
+            # with no error. Reject non-strings at validation, mirroring the
+            # prompt/shell/command 'must be a string' checks. "true"/"false"
+            # and an expression like "{{ ... }}" stay valid.
             errors.append(
                 f"Do-while step {config.get('id', '?')!r}: 'condition' must be a "
                 f"string, got {type(config['condition']).__name__}."
--- a/src/specify_cli/workflows/steps/do_while/__init__.py
+++ b/src/specify_cli/workflows/steps/do_while/__init__.py
@@ -70,6 +70,18 @@ class DoWhileStep(StepBase):
                 f"Do-while step {config.get('id', '?')!r} is missing "
                 f"'condition' field."
             )
+        elif not isinstance(config["condition"], str):
+            # The engine re-evaluates 'condition' via evaluate_condition() after
+            # each iteration; it returns a non-string as-is and takes bool() of
+            # it -- so a list/dict/number condition silently resolves to a
+            # truthiness (e.g. condition: [1, 2] is always truthy, looping to
+            # max_iterations) with no error. Reject non-strings at validation,
+            # mirroring the prompt/shell/command 'must be a string' checks.
+            # "true"/"false" and an expression like "{{ ... }}" stay valid.
+            errors.append(
+                f"Do-while step {config.get('id', '?')!r}: 'condition' must be a "
+                f"string, got {type(config['condition']).__name__}."
+            )
         max_iter = config.get("max_iterations")
         if max_iter is not None:
             # bool is a subclass of int, so isinstance(True, int) is True and
--- a/src/specify_cli/workflows/steps/do_while/__init__.py
+++ b/src/specify_cli/workflows/steps/do_while/__init__.py
@@ -70,19 +70,23 @@ class DoWhileStep(StepBase):
                 f"Do-while step {config.get('id', '?')!r} is missing "
                 f"'condition' field."
             )
-        elif not isinstance(config["condition"], str):
+        elif not isinstance(config["condition"], (str, bool)):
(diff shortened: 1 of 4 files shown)


mnriem (changes requested):

jawwad-ali:
Good catch — fixed in 6a78c02. You're right that `evaluate_condition()` always returns a `bool`; the pass-through happens one level down in `evaluate_expression()`, which `evaluate_condition()` calls before applying `bool()`. I reworded all four sites (the if/while/do-while guards and the mirror test comment) to name the two stages explicitly:

> execute() feeds 'condition' to evaluate_condition(), which first delegates to evaluate_expression() -- that returns a non-string unchanged -- and then coerces the result with bool().

Comments only; no behaviour change. `ruff@0.15.0` clean and the condition tests still pass (28).

mnriem:

jawwad-ali:
Thanks — I took the second option you offered ("update the PR/API contract if boolean literals are intentionally supported"), because rejecting `bool` would be a **breaking change** rather than a fix.

Accepting booleans is deliberate. Verified against `upstream/main` (pre-PR):

```
condition=False -> completed | condition_result: False | next: []
condition=True  -> completed | condition_result: True  | next: ['i']
```

So `condition: false` already resolves **correctly** today: `evaluate_expression` passes the bool through and `evaluate_condition`'s `bool()` is a no-op. It is not part of the silent-coercion bug — unlike `[1, 2]`, which becomes `True` and always takes the branch. On top of that, two of the three steps **default** `condition` to the bool `False` themselves (`config.get("condition", False)`), so `bool` is the field's own natural type. An unquoted `condition: false` is also idiomatic YAML, so rejecting it would fail workflows that work on `main` today.

I've updated the **title and description** to state the contract accurately ("non-string/non-boolean") and documented why booleans are accepted, and the bad-value matrix is now `[list, dict, int, float]` with `True`/`False` moved to the positive test in all three step classes. Happy to flip to string-only if you'd prefer that contract — but it would need a deprecation note, since it drops support for input that currently works.

mnriem (approved):

jawwad-ali:
Addressed the latest Copilot review (the three suppressed low-confidence notes on `tests/test_workflows.py`) in e3a42d1.

**It was right, and it was my leftover.** When I updated the contract to accept booleans I renamed the acceptance test to `test_validate_accepts_string_or_bool_condition` but left its sibling called `test_validate_rejects_non_string_condition` — and a bool *is* a non-string, so the two names contradicted each other in the test report.

Renamed in all three step classes (if/while/do-while):

```
test_validate_rejects_non_string_condition
  -> test_validate_rejects_non_string_non_bool_condition
```

That now matches the validator's own message, `'condition' must be a string or boolean, got <type>`.

Test names only — no behaviour change, and the parametrized values are untouched (`[["a","b"], {"k":"v"}, 5, 1.5]` rejected; `["true", "false", "{{ inputs.flag }}", True, False]` accepted).

Verification: 38 condition tests pass · `uvx ruff@0.15.0 check src tests` clean · full `tests/test_workflows.py` failure set is byte-identical before and after the rename (20 pre-existing Windows symlink-privilege failures, 743 passed).

@mnriem this should clear the outstanding Copilot feedback on this PR. To restate the one earlier point I *didn't* action, so it's not lost: Copilot asked for `bool` to be rejected outright, and I took the alternative its own comment offered ("update the PR/API contract if boolean literals are intentionally supported") — because `condition: false` already resolves correctly today and two of the three steps default `condition` to the bool `False` themselves, so rejecting it would be a breaking change rather than a fix. Happy to flip it behind a deprecation if you'd rather have the string-only contract.

jawwad-ali:
Correction to my comment above: I posted it without realising this PR had already been merged (a9c9905, ~3 minutes earlier), so the commit I referenced never became part of it. Thanks for the merge, @mnriem.

The Copilot note it addressed is still valid against `main` — `test_validate_rejects_non_string_condition` sits next to `test_validate_accepts_string_or_bool_condition`, and a bool is a non-string, so the two names disagree. I've opened it as a separate follow-up instead: **#3808** (3 insertions / 3 deletions, test names only, no behaviour change).

Everything else in my comment above still stands — in particular the reason booleans stay valid rather than being rejected.

PR #3707, opened 2026-07-24
fix(workflows): reject falsy non-mapping workflow-catalogs.yml top level

jawwad-ali:
## Problem

`WorkflowCatalog._load_catalog_config` (in `src/specify_cli/workflows/catalog.py`) parses the project/user `workflow-catalogs.yml` like this:

```python
data = yaml.safe_load(config_path.read_text(encoding="utf-8")) or {}
...
if not isinstance(data, dict):
    raise WorkflowValidationError(f"Invalid catalog config: expected a mapping, got ...")
```

The `or {}` coerces a **falsy** non-mapping top level to `{}` **before** the `isinstance` guard runs, so those are silently swallowed instead of raising — while a **truthy** non-mapping raises as intended. Reproduced on `main`:

| top-level value | current behaviour |
| --- | --- |
| `[]`, `false`, `0`, `''` (falsy non-mappings) | **silently** returns `None` → built-in defaults, no error |
| `5`, a bare list (truthy non-mappings) | raises `expected a mapping` ✅ |
| empty document / `null` | returns `None` (valid no-op) ✅ |

So a malformed config like a top-level `[]` (probably meant `catalogs: [...]`) is ignored with no diagnostic. This is the same silent-swallow the bundler catalog reader already guards against for its own config.

## Fix

Drop the `or {}` and branch on `None` explicitly:

```python
data = yaml.safe_load(...)
if data is None:        # empty document / explicit null -> valid no-op
    return None
if not isinstance(data, dict):
    raise WorkflowValidationError(...)   # now catches [], false, 0, '' too
```

`None` (empty file / explicit `null`) stays a valid no-op; every non-mapping — falsy or truthy — now raises the same actionable error. **Correct configs are unaffected** (no behaviour change).

## Verification

- New `test_falsy_non_mapping_config_rejected` (`[]`/`false`/`0`/`''`): **fails before** (silently returns `None`), **passes after** (raises).
- New `test_empty_or_null_config_is_noop` (empty/comment-only/`null`/`~`): passes before and after — regression guard that the no-op path is preserved.
- Full `TestWorkflowCatalog`: **32 passed**. `ruff` clean.

---
*AI-assisted: authored with Claude Code. I reproduced the falsy/truthy split on `main` with a small script before writing the fix, and confirmed the empty/null no-op is preserved.*

### DIFF
--- a/src/specify_cli/workflows/catalog.py
+++ b/src/specify_cli/workflows/catalog.py
@@ -355,16 +355,25 @@ class WorkflowCatalog:
                 f"Invalid catalog config: expected a mapping, "
                 f"got {type(data).__name__}"
             )
-        catalogs_data = data.get("catalogs", [])
-        if not catalogs_data:
-            # Empty catalogs list (e.g. after removing last entry)
-            # is valid — fall back to built-in defaults.
+        # Same asymmetry as the top level above, one nesting level down: the
+        # shape check has to run BEFORE the emptiness check, or a FALSY non-list
+        # (``catalogs: {}``/``''``/``0``/``false``) is silently swallowed as
+        # "no catalogs" while a TRUTHY non-list (``catalogs: 5``) correctly
+        # raises. An absent key or an explicit ``catalogs:`` null means "nothing
+        # configured here" (matching the bundler's reader), and an empty list
+        # stays valid too.
+        catalogs_data = data.get("catalogs")
+        if catalogs_data is None:
             return None
         if not isinstance(catalogs_data, list):
             raise WorkflowValidationError(
                 f"Invalid catalog config: 'catalogs' must be a list, "
                 f"got {type(catalogs_data).__name__}"
             )
+        if not catalogs_data:
+            # Empty catalogs list (e.g. after removing last entry)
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

jawwad-ali:
All three addressed in c1dfa8b:

1. **Moved** the StepCatalog tests into `TestStepCatalog` — you're right that a targeted `pytest ...::TestStepCatalog` run skipped them where they were.
2. **Added the nested coverage**: parametrized falsy `catalogs:` (`{}`/`''`/`0`/`false`) plus the absent/null/empty no-op cases. Verified against `upstream/main`'s `catalog.py`: **8 fail there, 8 pass here**, so the duplicated loader can't regress independently now.
3. **Removed the parity parenthetical** — you're correct that `src/specify_cli/catalogs.py` is not the behavior this matches: it coerces a null document to `{}` and *raises* for missing/empty `catalogs` rather than treating them as a no-op. The comment now just states what changed (only the misreported shapes) without claiming parity.

mnriem (approved):

PR #3709, opened 2026-07-24
fix(auth): resolve az via shutil.which so azure-cli token works on Windows

jawwad-ali:
## Problem

`AzureDevOpsAuth._acquire_via_az_cli()` runs:

```python
result = subprocess.run(["az", "account", "get-access-token", ...], ...)
```

with a bare `"az"`. On **Windows** the Azure CLI is installed as `az.cmd`, and `subprocess.run` calls `CreateProcess`, which — unlike a shell — does **not** consult `PATHEXT`. So a bare `"az"` fails with `WinError 2 (FileNotFoundError)` even when `az` works fine in the user's terminal after `az login`. The `except OSError` swallows it, so `auth: azure-cli` silently returns **no token** on Windows.

(Verified on a Windows box: `shutil.which("az")` → `...\wbin\az.CMD`, i.e. the runnable target is the `.CMD` shim, which the bare name never reaches.)

## Fix

Resolve the executable with `shutil.which("az")` (which honors `PATHEXT`) before invoking it — exactly the fix the maintainer already applied in `integrations/base.py` for the same CreateProcess/`.cmd` problem:

```python
az = shutil.which("az") or "az"
result = subprocess.run([az, "account", "get-access-token", ...], ...)
```

`or "az"` preserves the previous behaviour (and the existing not-installed `OSError` path) when `az` is absent. On POSIX `shutil.which` returns the same executable, so behaviour is unchanged there.

## Verification

- New `test_resolve_token_azure_cli_resolves_executable`: patches `shutil.which` → a fake `az.CMD` path and asserts the resolved path is `argv[0]`. **Fails before** (no `shutil` import to resolve with), **passes after**.
- New `test_resolve_token_azure_cli_falls_back_to_bare_az`: `shutil.which` → `None` still yields `argv[0] == "az"` (fallback preserved).
- Existing azure-cli success/failure/not-installed tests unchanged: **11 passed**. `ruff` clean.

---
*AI-assisted: authored with Claude Code. I confirmed on Windows that `shutil.which("az")` resolves to `az.CMD` and mirrored the maintainer's existing `integrations/base.py` resolution.*

### DIFF
--- a/src/specify_cli/authentication/azure_devops.py
+++ b/src/specify_cli/authentication/azure_devops.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import json as _json
 import os
+import shutil
 import subprocess
 from typing import TYPE_CHECKING
 
@@ -71,9 +72,17 @@ class AzureDevOpsAuth(AuthProvider):
     def _acquire_via_az_cli() -> str | None:
         """Run ``az account get-access-token`` and return the access token."""
         try:
+            # Windows: ``subprocess.run`` calls ``CreateProcess``, which does
+            # not consult ``PATHEXT``, so a bare ``"az"`` (installed as
+            # ``az.cmd``) fails with ``WinError 2`` even after ``az login``.
+            # Resolve via ``shutil.which`` (which honors ``PATHEXT``) so the
+            # ``.cmd`` shim works. On POSIX this is a harmless lookup that
+            # returns the same executable; ``or "az"`` preserves the prior
+            # behavior (and the existing OSError path) when ``az`` is absent.
+            az = shutil.which("az") or "az"
             result = subprocess.run(  # noqa: S603, S607
                 [
-                    "az",
+                    az,
                     "account",
                     "get-access-token",
                     "--resource",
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
Please address Copilot feedback and resolve test & lint errors

jawwad-ali:
Good catch — that was the macOS CI failure, fixed in 0dd35a7. The hardcoded `C:\Program Files\...` reads as **relative** under `posixpath`, so on Linux/macOS the production `os.path.isabs()` sent it down the fallback branch and `argv[0]` was `"az"`.

The path is now built with the host's rules — `os.path.join(os.path.abspath(os.sep), "opt", "az", "az.CMD")` — with an `assert os.path.isabs(...)` so the intended branch is guaranteed to be the one under test (`/opt/az/az.CMD` on POSIX, `D:\opt\az\az.CMD` on Windows). The fallback test's inputs (`.\az.CMD`, `az.cmd`, `./az`) are relative under both `ntpath` and `posixpath`, so those were already portable.

mnriem (approved):

PR #3710, opened 2026-07-24
fix(extensions): guard non-numeric catalog downloads in search/info rendering

jawwad-ali:
## Problem

`specify extension search` and `specify extension info <id>` render a catalog entry's `downloads` field with the `:,` thousands separator:

```python
stats.append(f"Downloads: {ext['downloads']:,}")   # extension_search
...
stats.append(f"Downloads: {ext_info['downloads']:,}")  # _print_extension_info
```

guarded only by `is not None`. But `:,` is only valid for `int`/`float` — on a string it raises `ValueError: Cannot specify ',' with 's'`.

Catalog payloads are only **shape**-validated (`_validate_catalog_payload` checks that `extensions` is a dict); individual entry fields are never type-checked, and `_get_merged_extensions` returns the raw catalog dicts. So a catalog entry with `"downloads": "1500"` (a quoted number — realistic from a community catalog, a `SPECKIT_CATALOG_URL` catalog, a project/user catalog, or a poisoned cache) makes the **whole command** abort with an uncaught traceback. Reproduced end-to-end: with such a cache, both `extension search` and `extension info` exit non-zero with `ValueError("Cannot specify ',' with 's'.")`.

## Fix

Group-format `downloads` only when it is actually numeric; otherwise render it as-is:

```python
downloads = ext.get('downloads')
if downloads is not None:
    stats.append(
        f"Downloads: {downloads:,}"
        if isinstance(downloads, (int, float))
        else f"Downloads: {downloads}"
    )
```

(and the equivalent in `_print_extension_info`). Numeric values (`int`/`float`, including `bool`) take the identical `:,` branch, so **correct catalogs render byte-for-byte the same** — only the previously-crashing case changes to print the value. This mirrors the module's existing defensive treatment: every other field in these two renderers is already `str()`-wrapped.

## Verification

- `test_info_renders_non_numeric_downloads` (direct) and `test_search_survives_non_numeric_downloads` (end-to-end via `CliRunner`): both **fail before** with the exact `ValueError`, **pass after** (`exit_code == 0`, output shows `Downloads: 1500`).
- `TestExtensionCatalog`: **64 passed**. `ruff` clean.

---
*AI-assisted: authored with Claude Code. I reproduced the crash on `main` through the real CLI (catalog cache with a string `downloads`) before writing the guard, and confirmed numeric output is unchanged.*

### DIFF
--- a/src/specify_cli/extensions/_commands.py
+++ b/src/specify_cli/extensions/_commands.py
@@ -790,8 +790,16 @@ def extension_search(
 
             # Stats
             stats = []
-            if ext.get('downloads') is not None:
-                stats.append(f"Downloads: {ext['downloads']:,}")
+            downloads = ext.get('downloads')
+            if downloads is not None:
+                # Catalog fields are untrusted; a non-numeric ``downloads``
+                # (e.g. the JSON string "1500") would crash the ``:,`` format
+                # with "Cannot specify ',' with 's'". Only group-format numbers.
+                stats.append(
+                    f"Downloads: {downloads:,}"
+                    if isinstance(downloads, (int, float))
+                    else f"Downloads: {downloads}"
+                )
             if ext.get('stars') is not None:
                 stats.append(f"Stars: {ext['stars']}")
             if stats:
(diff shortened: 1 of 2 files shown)


jawwad-ali:
You're right — fixed in 135444f. Guarding the `:,` ValueError just traded it for a `MarkupError`: I confirmed `Console().print("Downloads: [/red]foo")` raises `closing tag '[/red]' doesn't match any open tag`, so `extension search`/`info` still aborted on a hostile catalog value.

Both sites now escape the fallback — `f"Downloads: {_escape_markup(str(downloads))}"` — matching how every other catalog field in these renderers is already handled. The numeric branch keeps the identical `:,` format, so correct catalogs render byte-for-byte the same.

Tests parametrized over `"1500"`, `"[/red]foo"` (unbalanced → MarkupError) and `"[bold]x[/bold]"` (balanced → would silently restyle); the two markup cases **fail before** this commit and pass after.

mnriem:
Thank you!

PR #3711, opened 2026-07-24
fix(bundler): reject unsupported schema_version in _merge_config (align readers)

jawwad-ali:
## Problem

`.specify/bundle-catalogs.yml` has **two** readers that are meant to agree:

- `commands_impl/catalog_config._read` — used by `specify bundle catalog list/add/remove`
- `models/catalog._merge_config` — used by `load_source_stack`, which feeds `bundle search` / `info` / `install` via `CatalogStack`

`_read` rejects an unsupported **major** `schema_version` (its lines 54–64). `_merge_config` validates the top-level mapping and the `catalogs` list shape but **never checks `schema_version`**. So a project/user config written by a future incompatible Spec Kit (e.g. `schema_version: '2.0'`) is **rejected** by `_read` but **silently parsed under v1 assumptions** by `_merge_config` — on the exact path (`search`/`install`) where a source's `install_policy` governs whether bundles may be installed from it. The two readers of the same file disagree.

`#3623` (non-list `catalogs`) and `#3659` (top-level non-mapping) already aligned these two readers guard-by-guard — the in-code comments say so. This is the last unaligned guard.

## Fix

- Add the same forward-compatible major-version check to `_merge_config`, immediately after its top-level-mapping guard (identical wording/logic to `_read`).
- Promote `CONFIG_SCHEMA_VERSION = "1.0"` to `models/catalog.py` as the **single source of truth** and import it in `catalog_config.py` (it was a local duplicate), so the two readers cannot drift apart.

An **absent** `schema_version` stays valid (backward compatible with configs that omit it); a **matching major** (`1.x`) stays valid. Only a different major now raises — consistently across both readers.

## Verification

- `test_load_source_stack_rejects_unknown_schema_version` (`2.0`): **fails before** (`DID NOT RAISE`), **passes after**.
- `test_load_source_stack_accepts_matching_or_absent_schema_version` (`1.5` and no key): passes — backward-compat regression guard.
- Both readers' suites (`test_catalog_schema.py` + `test_bundler_catalog_config.py`): all pass except one **pre-existing** Windows symlink-privilege test (`WinError 1314`, environmental — passes on CI/Linux). `ruff` clean.

---
*AI-assisted: authored with Claude Code. I reproduced the reader disagreement on `main` (v2.0 accepted by `_merge_config`, rejected by `_read`) before adding the aligned guard.*

### DIFF
--- a/src/specify_cli/bundler/commands_impl/catalog_config.py
+++ b/src/specify_cli/bundler/commands_impl/catalog_config.py
@@ -14,13 +14,12 @@ from .. import BundlerError
 from ..lib.yamlio import dump_yaml, ensure_within, load_yaml
 from ..models.catalog import (
     CONFIG_FILENAME,
+    CONFIG_SCHEMA_VERSION,
     BUILTIN_DEFAULT_STACK,
     CatalogSource,
     InstallPolicy,
     Scope,
 )
-
-CONFIG_SCHEMA_VERSION = "1.0"
 
 _BUILTIN_IDS = {raw["id"] for raw in BUILTIN_DEFAULT_STACK}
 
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3712, opened 2026-07-24
fix(presets): guard non-list/non-mapping provides.templates in PresetManifest

jawwad-ali:
## Problem

`PresetManifest._validate` iterates `provides["templates"]` with **no shape guards**, unlike the direct sibling `ExtensionManifest._validate` (which guards both the container and each entry). A malformed third-party `preset.yml` therefore crashes with a **raw `TypeError`** rather than a clean validation error:

```
templates: 5        →  TypeError: 'int' object is not iterable
templates: [null]   →  TypeError: argument of type 'NoneType' is not iterable
templates: [5]      →  TypeError: argument of type 'int' is not iterable
```

The install handler (`presets/_commands.py`) only catches `PresetCompatibilityError` / `PresetValidationError` / `PresetError`, so the `TypeError` escapes all three and dumps an **unhandled Python traceback** to the user on `specify preset install`. (A string/list entry is arguably worse — it slips past into the misleading `"Template missing 'type', 'name', or 'file'"`.)

Reproduced on `main` (`4d3a428`): the same three malformed shapes fed to the sibling `ExtensionManifest` return clean `ValidationError`s — confirming the parity gap.

## Fix

Add the two guards `ExtensionManifest` already has, raising `PresetValidationError`:

```python
templates = provides["templates"]
if not isinstance(templates, list):
    raise PresetValidationError("Invalid provides.templates: expected a list")
for tmpl in templates:
    if not isinstance(tmpl, dict):
        raise PresetValidationError("Each template entry in 'provides.templates' must be a mapping")
    ...
```

Valid manifests (a list of mappings) pass both guards unchanged — **no behaviour change for correct input**.

## Verification

- New `test_non_list_templates_raises_validation_error` + parametrized `test_non_mapping_template_entry_raises_validation_error` (`None`/`5`/`"oops"`/`["nested"]`): **fail before** (raw `TypeError` or the misleading "Template missing…" message), **pass after** (clean `PresetValidationError`). Mirrors the existing `test_non_mapping_yaml_raises_validation_error` precedent.
- Full `TestPresetManifest`: **24 passed**. `ruff` clean.

---
*AI-assisted: authored with Claude Code. I reproduced the raw `TypeError`s on `main` and confirmed the sibling `ExtensionManifest` already returns clean errors for the identical shapes.*

### DIFF
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -322,7 +322,7 @@ class PresetManifest:
 
         # Validate provides section
         provides = self.data["provides"]
-        if "templates" not in provides or not provides["templates"]:
+        if "templates" not in provides:
             raise PresetValidationError(
                 "Preset must provide at least one template"
             )
@@ -333,10 +333,20 @@ class PresetManifest:
         # install handler already catches, instead of a raw TypeError
         # ('int'/'NoneType' object is not iterable) that escapes to an
         # unhandled traceback. Mirrors the sibling ExtensionManifest guards.
+        #
+        # Order matters: the container's TYPE is checked before its emptiness,
+        # so a FALSY non-list (``templates: 0``/``false``/``null``/``''``/``{}``)
+        # reports the accurate type error rather than the misleading "must
+        # provide at least one template". An empty list still reports the
+        # latter, since that genuinely is a list with no templates.
         templates = provides["templates"]
         if not isinstance(templates, list):
             raise PresetValidationError(
                 "Invalid provides.templates: expected a list"
+            )
+        if not templates:
+            raise PresetValidationError(
+                "Preset must provide at least one template"
             )
         for tmpl in templates:
             if not isinstance(tmpl, dict):
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -327,8 +327,22 @@ class PresetManifest:
                 "Preset must provide at least one template"
             )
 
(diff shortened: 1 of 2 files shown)


jawwad-ali:
Correct, and a good catch — fixed in 7612936. My original test used `templates: 5` (truthy), which is exactly why it passed while every falsy non-list still hit the misleading message.

Reordered to presence → container type → emptiness:
```python
if "templates" not in provides: raise ...("must provide at least one template")
templates = provides["templates"]
if not isinstance(templates, list): raise ...("Invalid provides.templates: expected a list")
if not templates: raise ...("must provide at least one template")
```
An **empty list** deliberately keeps the "at least one template" message — it genuinely is a well-typed container with no templates — so the pre-existing `test_no_templates_provided` is unaffected. I added a regression guard for that, and parametrized the non-list test over truthy (`5`, `"oops"`, `{"a": 1}`) **and** falsy (`0`, `False`, `None`, `""`, `{}`) values — the 5 falsy cases fail before this commit.

mnriem (changes requested):

jawwad-ali:
Agreed — removed in 97fc23a. `test_no_templates_provided` already sets `provides.templates = []` and asserts the same message, so it covers the empty-list result of the type-before-emptiness ordering on its own. Left a short pointer comment where mine was so the coupling is obvious to the next reader.

mnriem (approved):

mnriem:
Thank you!

PR #3713, opened 2026-07-24
fix(bundle): wrap local .zip errors in BundlerError instead of raw crash

jawwad-ali:
## Problem

The local `.zip` branch of `_local_manifest_source` opens the archive and parses its embedded `bundle.yml` with **no error handling**:

```python
with zipfile.ZipFile(candidate) as archive:      # raw BadZipFile
    ...
data = _yaml.safe_load(io.BytesIO(raw))           # raw YAMLError
```

- A `.zip`-suffixed file that isn't a valid archive → `zipfile.BadZipFile`.
- A valid zip whose embedded `bundle.yml` is malformed → `yaml.YAMLError`.

Neither is a `BundlerError`, and `bundle_install` only does `except BundlerError`, so both escape to an **unhandled traceback** instead of the intended clean `Error: …` + exit 1. Reproduced both on `main` (`4d3a428`).

This is a **parity gap**: the directory and `bundle.yml` branches go through `BundleManifest.from_file` → `load_yaml`, which wraps failures in `BundlerError`; and the remote sibling `_download_remote_manifest` explicitly wraps its `_local_manifest_source` call in `try/except → BundlerError`. Only the direct local `.zip` branch is unguarded.

## Fix

Wrap both failure modes in `BundlerError` (hardening the source function, so every caller benefits):

```python
try:
    archive = zipfile.ZipFile(candidate)
except (zipfile.BadZipFile, OSError) as exc:
    raise BundlerError(f"Artifact '{candidate}' is not a valid .zip bundle: {exc}") from exc
with archive:
    ... # existing KeyError guard for missing bundle.yml
try:
    data = _yaml.safe_load(io.BytesIO(raw))
except _yaml.YAMLError as exc:
    raise BundlerError(f"Artifact '{candidate}' contains an invalid bundle.yml: {exc}") from exc
```

Valid archives are unaffected — no behaviour change for correct input.

## Verification

- New `test_local_source_corrupt_zip_raises_clean_error` and `test_local_source_zip_with_malformed_bundle_yml_raises_clean_error`: **fail before** (raw `BadZipFile` / `YAMLError`), **pass after** (clean `BundlerError`).
- Full `test_bundler_local_install.py`: **12 passed**. `ruff` clean.

---
*AI-assisted: authored with Claude Code. I reproduced both raw crashes on `main` via a direct `_local_manifest_source` call and mirrored the remote-download path's existing error wrapping.*

### DIFF
--- a/src/specify_cli/commands/bundle/__init__.py
+++ b/src/specify_cli/commands/bundle/__init__.py
@@ -715,14 +715,30 @@ def _local_manifest_source(arg: str):
 
         import yaml as _yaml
 
-        with zipfile.ZipFile(candidate) as archive:
+        try:
+            archive = zipfile.ZipFile(candidate)
+        except (zipfile.BadZipFile, OSError) as exc:
+            # A .zip-suffixed file that is not a valid archive would otherwise
+            # raise a raw BadZipFile that escapes bundle_install's `except
+            # BundlerError` and dumps a traceback. Report it cleanly, like the
+            # remote-download path already does for the same source.
+            raise BundlerError(
+                f"Artifact '{candidate}' is not a valid .zip bundle: {exc}"
+            ) from exc
+        with archive:
             try:
                 raw = archive.read("bundle.yml")
             except KeyError as exc:
                 raise BundlerError(
                     f"Artifact '{candidate}' does not contain a bundle.yml."
                 ) from exc
-        data = _yaml.safe_load(io.BytesIO(raw))
+        try:
+            data = _yaml.safe_load(io.BytesIO(raw))
+        except _yaml.YAMLError as exc:
+            # Malformed embedded bundle.yml — same rationale as above.
+            raise BundlerError(
+                f"Artifact '{candidate}' contains an invalid bundle.yml: {exc}"
+            ) from exc
         return BundleManifest.from_dict(data)
 
     if candidate.name == "bundle.yml" or candidate.suffix in (".yml", ".yaml"):
--- a/src/specify_cli/commands/bundle/__init__.py
+++ b/src/specify_cli/commands/bundle/__init__.py
@@ -741,19 +741,21 @@ def _local_manifest_source(arg: str):
         )
         _ZIP_READ_ERRORS = _ZIP_OPEN_ERRORS + (zlib.error, EOFError, RuntimeError)
 
-        # The messages below escape the interpolated EXCEPTION text: it reaches
-        # the user through _fail(), which renders Rich markup, and zipfile
-        # embeds raw bytes from the archive in some messages -- e.g.
-        # BadZipFile("File name in directory 'bundle.yml' and header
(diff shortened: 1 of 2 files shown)


jawwad-ali:
Confirmed and fixed in 6d5f8d2 — you're right that `ZipFile()` only validates the central directory.

I reproduced both corruption modes on a directory-intact archive:
- flipped byte in a **stored** member → `zipfile.BadZipFile: Bad CRC-32 for file 'bundle.yml'` (exactly your example)
- mangled **deflate** stream → `zlib.error: Error -3 while decompressing data`

Since neither is an `OSError` subclass, an `OSError`-only clause would still have leaked the second one, so the read boundary lists `(BadZipFile, OSError, zlib.error, EOFError)` explicitly. The `KeyError` → "does not contain a bundle.yml" case is unchanged, and the new test asserts both modes now surface as `BundlerError` (the CRC case fails before this commit with a raw traceback).

mnriem (changes requested):

jawwad-ali:
Right again — fixed in aedb9e2. Confirmed the vector: rendering `Artifact '/tmp/x/[/red].zip' is not a valid .zip bundle: ...` raises `MarkupError: closing tag '[/red]' ... doesn't match any open tag`, and escaping it renders fine. Reachable in practice via a directory named `[` containing `red].zip`.

The path is now escaped once into `safe_path` and used at all three sites (open, missing-bundle.yml, read). One note for the record: the vector is **POSIX-specific in practice** — on Windows `Path()` normalizes the forward slash to a backslash (`[\red]`), which is inert markup — so the new test builds the POSIX-style path explicitly rather than relying on the host.

jawwad-ali:
Closing as **superseded by main**. #3141 (`harden: secure extension and preset archive downloads`) introduced the shared `open_zip_bounded` / `read_zip_member_limited` helpers and applied them to this exact code path, which covers everything this PR set out to fix — and more.

I re-ran my three repro archives against current `main` (`655a3cb`):

```
encrypted member      -> BundlerError: Failed to read ZIP member 'bundle manifest': ...
unsupported method 99 -> BundlerError: Unsupported ZIP compression method 99; ...
bad CRC-32            -> BundlerError: Failed to read ZIP member 'bundle manifest': ...
```

All clean — no raw `BadZipFile`/`RuntimeError`/`NotImplementedError`/`zlib.error` escapes. `open_zip_bounded` also wraps the open in `except Exception`, which is broader than the tuple I proposed, so there is nothing left for this PR to add.

One small residual observation, offered as information rather than a request: `_download_security.py`'s `Invalid ZIP archive: {zip_path}` interpolates the path, and the bundle CLI renders errors through Rich (`_fail`), so a path containing a tag-shaped bracket run (e.g. a directory named `[`) would raise `MarkupError` instead of showing the message. That is a shared-module concern affecting several messages, so it belongs in its own change rather than being retrofitted here — happy to open one if you'd like it.

Thanks — good to see the hardening landed centrally.

PR #3714, opened 2026-07-24
fix(integrations): reject empty --commands-dir in generic raw_options

jawwad-ali:
## Problem

`GenericIntegration._resolve_commands_dir` has a parity gap between its two resolution branches:

```python
commands_dir = parsed_options.get("commands_dir")
if commands_dir:            # parsed branch: empty value correctly falls through
    return commands_dir
...
for i, token in enumerate(tokens):        # raw_options branch: NO emptiness check
    if token == "--commands-dir" and i + 1 < len(tokens):
        return tokens[i + 1]              # "" returned verbatim
    if token.startswith("--commands-dir="):
        return token.split("=", 1)[1]     # "" returned verbatim
```

So `specify init --integration generic --integration-options="--commands-dir="` (and `--commands-dir ""`) resolves to `""`. `init`'s own generic guard (`if selected_ai == "generic" and not integration_options`) is satisfied because `integration_options` is the non-empty string `"--commands-dir="`, so it doesn't catch it. `setup()` then computes `dest = (project_root / "").resolve() == project_root`, passes the containment check, and writes **every speckit command file** (`specify.md`, `plan.md`, `tasks.md`, …) **directly into the project root** — silently bypassing the documented *"--commands-dir is required"* contract and polluting the repo root.

Verified on `main` (`4d3a428`): `_resolve_commands_dir({}, {"raw_options": "--commands-dir="})` returns `""` (the parsed-options path with the same empty value correctly raises).

## Fix

Apply the same non-empty guard the parsed branch already uses to the raw_options branch, so an empty value falls through to the existing `raise ValueError("--commands-dir is required …")` on every input form (`--commands-dir=`, `--commands-dir ""`). Non-empty values resolve exactly as before — **no behaviour change for valid usage**.

## Verification

- New parametrized `test_resolve_commands_dir_rejects_empty_raw_value` (`--commands-dir=`, `--commands-dir ''`, `--commands-dir ""`): **fails before** (`DID NOT RAISE`), **passes after**.
- New `test_resolve_commands_dir_accepts_nonempty_raw_value`: non-empty raw values still resolve unchanged.
- Full `TestGenericIntegration`: **35 passed**. `ruff` clean.

---
*AI-assisted: authored with Claude Code. I reproduced the project-root resolution on `main` (empty raw value → `""`) and confirmed the parsed-options branch already guards it.*

### DIFF
--- a/src/specify_cli/integrations/generic/__init__.py
+++ b/src/specify_cli/integrations/generic/__init__.py
@@ -63,10 +63,19 @@ class GenericIntegration(MarkdownIntegration):
             import shlex
             tokens = shlex.split(raw)
             for i, token in enumerate(tokens):
+                # Only accept a NON-EMPTY value, matching the parsed-options
+                # branch above (`if commands_dir:`). An empty value
+                # (``--commands-dir=`` or ``--commands-dir ""``) must fall
+                # through to the "required" error below, not be returned
+                # verbatim — otherwise it resolves to the project root and
+                # command files get written there instead of a dedicated dir.
                 if token == "--commands-dir" and i + 1 < len(tokens):
-                    return tokens[i + 1]
+                    if tokens[i + 1]:
+                        return tokens[i + 1]
                 if token.startswith("--commands-dir="):
-                    return token.split("=", 1)[1]
+                    candidate = token.split("=", 1)[1]
+                    if candidate:
+                        return candidate
 
         raise ValueError(
             "--commands-dir is required for the generic integration"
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

jawwad-ali:
You're right, and thanks for catching it — that contradicted my own "no behaviour change for valid usage" claim. Fixed in 84fa06a.

`strip()` is now used **only** to decide blankness; the accepted value is returned unchanged, so a quoted `--commands-dir ' commands '` still targets the literal ` commands ` directory exactly as before. Both branches apply the same rule so they can't drift.

The padding test was rewritten to assert the value **round-trips verbatim** instead of being normalized. One implementation detail worth noting: in `raw_options` the padded value has to be quoted (`--commands-dir='  x  '`), since `shlex.split()` consumes unquoted whitespace before this code ever sees it.

mnriem (approved):

PR #3715, opened 2026-07-24
fix(github-http): return None on malformed host in resolve_github_release_asset_api_url

Noor-ul-ain001:
## Summary

`resolve_github_release_asset_api_url()` documents a resolve-or-return-`None` contract ("Returns ... `None` when the URL is not a resolvable GitHub release download or the lookup fails"), but reading the parsed authority could raise `ValueError` for a malformed host and leak a raw traceback past the caller.

The function already guards the twin case for `parsed.port` (malformed port -> return `None`). This applies the same guard to the host read: on Python 3.14 `urlparse()` itself raises `ValueError` for an invalid bracketed IPv6 host such as `https://[not-an-ip]/...`; on older versions the raise happens on `.hostname`. Wrapping both `urlparse` and `.hostname` in the `try` covers every version.

This matters because `download_url` is server-controlled — it comes from a catalog `download_url` payload — so a malformed value must not crash resolution.

## Test plan

- Added `test_returns_none_on_malformed_host`, mirroring the existing `test_returns_none_on_malformed_ghes_port`. It fails without the fix (raw `ValueError`) and passes with it.
- `pytest tests/test_github_http.py` -> 30 passed.

### DIFF
--- a/src/specify_cli/_github_http.py
+++ b/src/specify_cli/_github_http.py
@@ -102,8 +102,17 @@ def resolve_github_release_asset_api_url(
 
     from specify_cli._download_security import read_response_limited
 
-    parsed = urlparse(download_url)
-    hostname = (parsed.hostname or "").lower()
+    # Accessing ``.hostname`` (like ``.port`` below) raises ValueError on a
+    # malformed authority, e.g. an invalid bracketed IPv6 host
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3719, opened 2026-07-24
docs: clarify shell-step interpolation safety

mnriem:
## Summary

Workflow `shell` steps run their `run` field through the system shell, and `{{ ... }}` expressions are resolved by plain string substitution with no automatic quoting. That's easy to overlook when a `run` field interpolates values that aren't fully under the workflow author's control — `inputs.*` supplied at run time, or a prior step's output (for a `prompt` step, AI-generated text influenced by whatever the agent read).

This PR documents the safe patterns and makes the shipped examples model them. No behavior change.

## Changes

- **`docs/reference/workflows.md`** — new **"Interpolation and shell safety"** section explaining that interpolation is raw substitution, which sources are untrusted, and how to guard `run` fields (quote substitutions, constrain inputs with `enum`, validate free-form values, prefer env/files, gate sensitive commands).
- **`workflows/README.md`** — warning callout under the Shell Steps example linking to the new section; the `inputs.project_dir` example now quotes its substitution.
- **`workflows/PUBLISHING.md`** — strengthened the "prefer explicit commands over interpolation" guidance and called out prior-step/agent output as untrusted.

## Notes

Docs-only. The intentional "Risky vs. Safer" example pair in the reference doc is kept as teaching content; `context.run_id` examples are left unquoted since that value is engine-generated/constrained.

---

_Opened on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8)._


mnriem:
Addressed the three review points in 803a360eba4814a16e6b23705a483124728db353. All were valid — the earlier wording over-promised, so I reframed the guidance to make clear none of the mitigations *solve* interpolation risk:

- **Quoting is not a security boundary.** There is no shell-escaping filter, and a value containing the matching quote can break out. Quoting is now presented only as correctness handling (word-splitting) for already-constrained values.
- **Removed the "environment or files" transport.** `ShellStep` has no `env` mapping — it only copies the process environment and sets `SPECKIT_WORKFLOW_DIR` — so that mechanism doesn't exist. Also dropped the claim that routing through a `command`/`prompt` step validates or binds a value.
- **Corrected the gate guidance.** A `gate` renders only its own `message`/`show_file` and does not inspect or sanitise the following step; approval doesn't neutralise an injectable interpolation. Authors must surface the exact resolved command/data in the gate themselves.

The section now frames constraining values at the source (`enum`/allowlist), and keeping unconstrained values out of `run` entirely, as the only reliable control — and explicitly states these practices reduce exposure but do not eliminate the risk.

_Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8)._

mnriem:
Addressed both points in 5180ff0a657e5c0d75a8ff0607ceceb28085cb23:

- **README shell example** no longer interpolates an unconstrained path — it contradicted the warning beneath it. Since shell steps already run from the project root, the example is now a plain `run: "npm test"` with no interpolation.
- **Gate guidance** corrected: `GateStep` prints `message` verbatim with no control-character stripping (stripping applies only to `show_file` path and contents), so surfacing untrusted data in `message` could inject terminal escapes that alter or hide the prompt. The guidance now says to keep `message` to trusted, constrained text and surface untrusted material via `show_file`, whose contents are stripped.

_Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8)._

mnriem:
Addressed in 01a057100d07a1c5898d709f9bc98f587a5cc934: a `prompt` step stores agent output under `output.stdout`, not `output.value`, so the example now references `{{ steps.plan.output.stdout }}` to correctly demonstrate the untrusted agent-output flow.

_Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8)._

PR #3731, opened 2026-07-24
fix(workflows): escape remaining untrusted fields in `workflow info`

Noor-ul-ain001:
## Problem

Follow-up to #3690, which escaped only the step-graph brackets in `workflow info`. Every *other* metadata field the command prints is untrusted content (from `workflow.yml` or catalog JSON), and `console.print` has Rich markup enabled — so an unescaped `[...]` in any of them is parsed as a style tag and silently swallowed.

Affected fields:
- **definition path:** name, version, author, description, integration, and each input's name/type
- **catalog path:** name, version, description, tags, and the `'<id>' not found` error id

Concretely, a description of `Does [stuff] nicely` rendered as `Does  nicely`; an integration of `claude [code]` rendered as `claude `; an input type of `str [ing]` rendered as `str `.

## Fix

Route every field through `_escape_markup` (already imported and used by the sibling `workflow list` and catalog `search` commands), so bracketed text renders literally. This is the same defect and the same remedy as #3690, extended to the rest of the command.

## Tests

Two regression tests added to `TestWorkflowInfoStepGraph`, covering the definition path and the catalog path. Both **fail on the pre-fix source** (bracketed fields come back truncated) and pass with the fix. Verified via `git stash` test-the-test.

```
tests/test_workflows.py::TestWorkflowInfoStepGraph  3 passed
```

### DIFF
--- a/src/specify_cli/workflows/_commands.py
+++ b/src/specify_cli/workflows/_commands.py
@@ -2354,14 +2354,25 @@ def workflow_info(
         raise typer.Exit(1)
 
     if definition:
-        console.print(f"\n[bold cyan]{definition.name}[/bold cyan] ({definition.id})")
-        console.print(f"  Version:     {definition.version}")
+        # Escape every user-controlled field: workflow.yml values (name,
+        # version, author, description, integration, input names/types) are not
+        # trusted, and console.print has Rich markup enabled, so an unescaped
+        # `[...]` in any of them is parsed as a style tag and silently swallowed
+        # (same defect fixed for the step graph below; the sibling workflow_list
+        # already escapes all of these).
+        console.print(
+            f"\n[bold cyan]{_escape_markup(str(definition.name))}[/bold cyan] "
+            f"({_escape_markup(str(definition.id))})"
+        )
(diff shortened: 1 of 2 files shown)


mnriem:

Noor-ul-ain001:
Addressed both Copilot points in 0069726:

1. **Version field never asserted** (`test_definition_metadata_fields_escaped`, `test_catalog_metadata_fields_escaped`): both tests now use a bracketed version (`1.0.0 [beta]` / `2.0.0 [rc]`) and assert the exact value survives in the output, so the version escapes at `_commands.py` L2367/L2418 are now covered.

2. **Not-found identifier path uncovered**: added `test_not_found_id_escaped`, which makes local load raise `FileNotFoundError` and the catalog lookup return `None`, invokes `workflow info` with a bracketed ID (`ghost [wf]`), and asserts the literal ID is preserved in the "not found" error — covering the escape at L2427.

Verified each new assertion fails when its source escape is removed (test-the-test).

mnriem:
Thank you!

Issue #3024, opened 2026-06-17
[Bug]: agent-context extension ignores scoped/nested spec directories (SPECIFY_FEATURE_DIRECTORY), so plan auto-discovery finds nothing


athemelis:
### Bug Description

The agent-context extension's plan auto-discovery uses a one-level glob (specs/*/plan.md). Repos that organize specs in nested/scoped directories via SPECIFY_FEATURE_DIRECTORY (e.g. specs/<scope>/<feature>/plan.md) never match, so the SPECKIT block is refreshed without a plan path even when a plan exists. The command doc states the one-level assumption.

### Steps to Reproduce

1. Set SPECIFY_FEATURE_DIRECTORY=specs/<scope>/<feature> and run speckit.specify / speckit.plan so a plan lands at specs/<scope>/<feature>/plan.md.
2. Let the agent-context after_specify/after_plan hook fire (or run speckit.agent-context.update directly).
3. Inspect the updated SPECKIT block.

### Expected Behavior

The extension finds the most-recent plan anywhere under specs/ (including nested specs/<scope>/<feature>/plan.md) and includes its path.

### Actual Behavior

The one-level glob specs/*/plan.md matches nothing in scoped layouts; the block updates with no plan path.

### Specify CLI Version

0.11.0

### AI Agent

GitHub Copilot

### Operating System

macOS Tahoe 26.5.1

### Python Version

3.11.15 (uv-managed standalone CPython, cpython-3.11-macos-aarch64-none — uv installs `specify` under its own Python, not the system CLT 3.9.6 or Homebrew 3.13/3.14)

### Error Logs

```shell
none (silent; no error raised, plan path simply absent)
```

### Additional Context

Files: .specify/extensions/agent-context/scripts/bash/update-agent-context.sh (glob specs.glob("*/plan.md")), .../powershell/update-agent-context.ps1, .../commands/speckit.agent-context.update.md (docs). Suggested fix: recurse (specs/**/plan.md) and pick the most-recently-modified match. Related to two sibling issues filed alongside (feature.json write side-effect; empty CURRENT_BRANCH).

marcelsafin:
I'd like to fix this. Plan: recurse with `specs/**/plan.md` in the bash, powershell and python variants of update-agent-context, pick the most recently modified match, and update the one-level wording in the command doc. Parity tests cover nested and flat layouts.

PR #3734, opened 2026-07-25
fix(agent-context): discover nested plans in Python port mtime fallback

### LINKED ISSUE #3733: [Bug]: agent-context Python port misses nested plans (one-level glob), a parity regression of #3024
### What happened

The `agent-context` extension's **Python port** (`extensions/agent-context/scripts/python/update_agent_context.py`) uses a one-level scan in its plan-path mtime fallback:

```python
plans = sorted(
    (root / "specs").glob("*/plan.md"),   # line ~176
    key=lambda p: p.stat().st_mtime,
    reverse=True,
)
```

This does **not** match nested scoped layouts such as `specs/<scope>/<feature>/plan.md`. As a result, when no explicit `plan_path` is given and `feature.json` is absent, the managed context section is written **without** the plan link.

### Why this is a regression

The Bash and PowerShell ports already search recursively for exactly this case, explicitly referencing #3024:

- Bash: `specs.rglob("plan.md")`
- PowerShell: `Get-ChildItem ... -Recurse`

The command documentation (`speckit.agent-context.update.md`) also promises recursive discovery. Issue #3024 fixed Bash/PowerShell, but the later Python port (#3387) reintroduced the one-level glob, so the three ports are no longer in parity and the Python one violates the documented contract.

### Expected

The Python port discovers the most recently modified `plan.md` anywhere under `specs/`, matching the Bash/PowerShell ports and the documented behavior.

### Fix

`(root / "specs").rglob("plan.md")`. PR incoming with a parity regression test (fails on the one-level glob, passes with the recursive scan).

### Version

matecardoso:
## Description

Closes #3733.

The `agent-context` Python port reintroduced a one-level plan scan (`specs/*/plan.md`) in its mtime fallback, while the Bash (`specs.rglob("plan.md")`) and PowerShell (`-Recurse`) ports search recursively per the fix for #3024. The three ports were therefore out of parity: for nested scoped layouts such as `specs/<scope>/<feature>/plan.md`, the Python port found no plan and omitted the plan link from the managed context section.

This changes the fallback to `(root / "specs").rglob("plan.md")` and updates the module docstring to match the documented recursive-discovery contract.

## Testing

- [x] Ran existing tests with `pytest tests/extensions/` — **213 passed, 63 skipped** (PowerShell parity skipped: no `pwsh` locally).
- [x] Added `test_python_mtime_fallback_finds_nested_plan_matching_bash` to the parity suite. Verified it **fails** on the old one-level glob (Bash finds the nested plan, Python omits it → context files diverge) and **passes** with the recursive scan.

## AI Disclosure

The one-line fix, the docstring update, and the regression test were written by Claude (Anthropic) under my direction, and I reviewed them. The bug was discovered while enabling this extension in a downstream project; an automated code review flagged the port divergence, which I then traced to the #3024 fix not being carried into the later Python port (#3387).

### DIFF
--- a/extensions/agent-context/scripts/python/update_agent_context.py
+++ b/extensions/agent-context/scripts/python/update_agent_context.py
@@ -11,8 +11,9 @@ Usage: update_agent_context.py [plan_path]
 
 When ``plan_path`` is omitted, the script derives it from
 ``.specify/feature.json`` (written by /speckit-specify). Falls back to the most
-recently modified ``specs/*/plan.md`` only when feature.json is absent or its
-plan does not exist yet.
+recently modified ``plan.md`` anywhere under ``specs/`` (including nested scoped
+layouts such as ``specs/<scope>/<feature>/plan.md``) only when feature.json is
+absent or its plan does not exist yet.
 """
 
 from __future__ import annotations
@@ -173,7 +174,7 @@ def _resolve_plan_path(project_root: str) -> str:
     if not plan_path:
         root = Path(project_root).resolve()
         plans = sorted(
-            (root / "specs").glob("*/plan.md"),
+            (root / "specs").rglob("plan.md"),
             key=lambda p: p.stat().st_mtime,
             reverse=True,
         )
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3736, opened 2026-07-25
fix(agent-context): apply default markers when config markers are blank (bash)

### LINKED ISSUE #3735: [Bug]: agent-context Bash port aborts as "malformed" when config markers are blank, instead of using defaults
### What happened

In the `agent-context` extension's **Bash port** (`extensions/agent-context/scripts/bash/update-agent-context.sh`), when the config omits `context_markers` (or sets them blank) to rely on the built-in defaults, the script aborts:

```
agent-context: malformed config parser output; expected 3 lines (context_files, marker_start, marker_end), got 1; skipping update.
```

and the context file is never updated. The Python and PowerShell ports handle the same config correctly.

### Root cause

The config parser prints three lines — context_files JSON, `context_markers.start`, `context_markers.end` — captured via `_raw_opts="$(...)"`. Command substitution strips *trailing* newlines, so when the marker lines are empty (blank/omitted markers), the output collapses to fewer than three lines. The guard then fires:

```bash
if (( ${#_opts_lines[@]} < 3 )); then ... exit 0 ; fi
```

which makes the default-marker substitution below **unreachable** in exactly the case it exists for:

```bash
[[ -z "$MARKER_START" ]] && MARKER_START="$DEFAULT_START"
[[ -z "$MARKER_END"   ]] && MARKER_END="$DEFAULT_END"
```

The Python port applies `... or DEFAULT_START`, and PowerShell initializes to defaults, so both succeed — the three ports diverge across platforms.

### Expected

With blank/omitted markers, the Bash port uses `DEFAULT_START`/`DEFAULT_END` and updates the context file, matching the Python/PowerShell ports.

### Fix

Require only the mandatory context_files line and default the marker lines to empty (`${_opts_lines[1]:-}` / `${_opts_lines[2]:-}`) so the existing default substitution runs. PR incoming with a parity regression test.

### Version

matecardoso:
## Description

Closes #3735.

When the extension config omits `context_markers` (or sets them blank) to rely on the built-in defaults, the Bash port aborted with `malformed config parser output` and never updated the context file, while the Python (`... or DEFAULT_*`) and PowerShell (default-initialized) ports handled it correctly.

The config parser prints three lines (context_files JSON, `marker_start`, `marker_end`), captured via `_raw_opts="$(...)"`. Command substitution strips trailing newlines, so blank marker lines collapse the output to fewer than three, tripping the `(( ${#_opts_lines[@]} < 3 ))` guard and making the `DEFAULT_START`/`DEFAULT_END` substitution below unreachable — the exact case it was written for.

Fix: require only the mandatory context_files line, and default the marker lines to empty (`${_opts_lines[1]:-}` / `${_opts_lines[2]:-}`) so the existing default-marker substitution fills them in. Single source of truth for the defaults stays in the Bash script.

## Testing

- [x] Ran existing tests with `pytest tests/extensions/` — **213 passed, 63 skipped** (PowerShell parity skipped: no `pwsh` locally).
- [x] Added `test_python_blank_markers_use_defaults_matching_bash` to the parity suite. Verified it **fails** on the old guard (Bash exits without updating while Python writes the section → parity mismatch) and **passes** with the fix.

## AI Disclosure

The fix and the regression test were written by Claude (Anthropic) under my direction, and I reviewed them. The bug surfaced while enabling this extension in a downstream project; an automated code review flagged the cross-platform divergence, which I traced to the command-substitution newline-stripping interacting with the 3-line guard.

### DIFF
--- a/extensions/agent-context/scripts/bash/update-agent-context.sh
+++ b/extensions/agent-context/scripts/bash/update-agent-context.sh
@@ -176,13 +176,18 @@ _opts_lines=()
 while IFS= read -r _line || [[ -n "$_line" ]]; do
   _opts_lines+=("$_line")
 done < <(printf '%s\n' "$_raw_opts")
-if (( ${#_opts_lines[@]} < 3 )); then
-  echo "agent-context: malformed config parser output; expected 3 lines (context_files, marker_start, marker_end), got ${#_opts_lines[@]}; skipping update." >&2
+if (( ${#_opts_lines[@]} < 1 )); then
+  echo "agent-context: malformed config parser output; expected at least the context_files line, got ${#_opts_lines[@]}; skipping update." >&2
   exit 0
 fi
+# The marker lines may be absent: the $(...) capture above strips trailing
+# newlines, so blank markers (the config omitting context_markers and relying on
+# defaults) collapse the 3-line output to fewer lines. Default them to empty here
+# and let the DEFAULT_START/END substitution below fill them in, matching the
+# Python and PowerShell ports.
 CONTEXT_FILES_JSON="${_opts_lines[0]}"
-MARKER_START="${_opts_lines[1]}"
-MARKER_END="${_opts_lines[2]}"
+MARKER_START="${_opts_lines[1]:-}"
+MARKER_END="${_opts_lines[2]:-}"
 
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3738, opened 2026-07-25
fix: escape Rich markup in catalog list output

Noor-ul-ain001:
## Summary

The `catalog list` subcommands print user-editable catalog fields (`name`, `url`, `description` from the `*-catalogs.yml` config files) through `console.print`, which has Rich markup enabled. Any bracketed content is parsed as a style tag and **silently swallowed** — e.g. a catalog description `Does [stuff] nicely` renders as `Does  nicely` — and a malformed tag can raise while rendering.

This affects four commands whose sibling `extension catalog list` already escapes these fields:

- `specify workflow catalog list` (`workflows/_commands.py`)
- `specify workflow step catalog list` (`workflows/_commands.py`)
- `specify preset catalog list` (`presets/_commands.py`)
- `specify integration catalog list` (`integrations/_query_commands.py`)

## Fix

Route each untrusted field through the module's already-imported `escape` helper (`from rich.markup import escape`), matching the reference pattern in `extension catalog list`. No new imports.

## Tests

Adds a regression test per command that injects bracketed `name`/`url`/`description` and asserts the brackets survive verbatim in the output. Confirmed each test fails without the corresponding source fix (test-the-test via stash).

The 20 unrelated `test_workflows.py` symlink-guard failures in my local run are pre-existing Windows-without-elevation failures, not touched by this change.

### DIFF
--- a/src/specify_cli/integrations/_query_commands.py
+++ b/src/specify_cli/integrations/_query_commands.py
@@ -489,13 +489,14 @@ def integration_catalog_list():
         display_name = str(raw_name).strip() if raw_name is not None else ""
         if not display_name:
             display_name = f"catalog-{i + 1}"
+        safe_name = _rich_escape(display_name)
         if env_override or project_configs is None:
-            console.print(f"  - [bold]{display_name}[/bold] — {install_status}")
+            console.print(f"  - [bold]{safe_name}[/bold] — {install_status}")
         else:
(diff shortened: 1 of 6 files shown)


mnriem (approved):

mnriem (approved):

mnriem:
Thank you!

PR #3739, opened 2026-07-25
fix(skills): match closing frontmatter delimiter on its own line

Quratulain-bilal:
### Problem

`SkillsIntegration.setup` parses each command template's frontmatter with `raw.split("---", 2)`. A bare substring split stops at the **first** `---` *anywhere* — including one embedded inside a value. A template whose description contains `---`, e.g.

```yaml
---
name-marker: sentinel
---
Body of the command.
```

is mis-parsed three ways:

- the description is truncated to `Separate sections with` (or dropped entirely, falling back to the generic `Spec Kit: <name> workflow`);
- every frontmatter key after the embedded marker is silently lost;
- the leftover frontmatter spills into the installed skill body.

### Fix

Locate the closing `---` on its **own line** (line-anchored scan) for both the description parse and the body strip, mirroring the hardened `CommandRegistrar.parse_frontmatter` scan.

- The frontmatter block is parsed **unstripped**, so trailing newlines in literal (`|`) block scalars still survive.
- The body slice keeps the newline after the marker, so output stays **byte-for-byte identical** to the old `split("---", 2)[2]` for well-formed templates (verified with a differential check across normal / empty-body / trailing-space cases).

### Tests

Adds `TestSkillFrontmatterEmbeddedDashes` covering the dashed-description truncation and the frontmatter-spilled-into-body cases. Both **fail on the pre-fix code** and pass after. Full skills integration suite: **342 passed, 3 skipped, 0 failures**.

### DIFF
--- a/src/specify_cli/integrations/hermes/__init__.py
+++ b/src/specify_cli/integrations/hermes/__init__.py
@@ -121,13 +121,27 @@ class HermesIntegration(SkillsIntegration):
             command_name = src_file.stem  # e.g. "plan"
             skill_name = f"speckit-{command_name.replace('.', '-')}"
 
-            # Parse frontmatter for description
+            # Parse frontmatter for description. Locate the closing ``---`` on
+            # its own line rather than with ``raw.split("---", 2)`` — a bare
+            # substring split stops at the first ``---`` *anywhere*, including
+            # one inside a value such as ``description: Separate sections
+            # with ---``, which truncates the frontmatter and drops later keys.
+            # The block between the delimiters is parsed unstripped so trailing
+            # newlines in literal (``|``) block scalars survive.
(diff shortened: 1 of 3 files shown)


mnriem (changes requested):

mnriem:
Please pull in latest changes from upstream/main to resolve security audit

Quratulain-bilal:
Branch is already up to date with upstream/main. No conflicts found. Ready for review.

mnriem:
Dependency audit failed so please rebase on main

Quratulain-bilal:
done @mnriem .

mnriem (approved):

mnriem:
Thank you!

PR #3744, opened 2026-07-25
fix(extensions): hyphenate command names in 'extension info' listing

Quratulain-bilal:
### Problem

The `Commands:` section of `specify extension info` (for a locally-installed extension) prints each command in its manifest **dotted** form, e.g. `speckit.jira.sync`. Cline and Forge register **hyphenated** command names (`/speckit-jira-sync`), so on those projects the names shown by `info` don't match what the user actually invokes.

This is the same parity `extension add` already applies to its "Provided commands" listing (#3669), and the same command-name parity fixed for hook invocations (#3641) and the init next-steps panel (#3642) — the `info` listing was simply missed.

### Fix

Format each command name through the active agent's formatter (`format_cline_command_name` / `format_forge_command_name`), mirroring the exact approach in `extension_add`. Other agents are unaffected (dotted form retained).

### Tests

Adds `test_forge_extension_info_hyphenates_command_names`: on a Forge project it asserts the `info` output contains the hyphenated `speckit-test-ext-hello` and **not** the dotted `speckit.test-ext.hello`. Fails on pre-fix code (prints dotted), passes after. Full `tests/test_extensions.py`: **385 passed**.

### DIFF
--- a/src/specify_cli/extensions/_commands.py
+++ b/src/specify_cli/extensions/_commands.py
@@ -879,9 +879,30 @@ def extension_info(
             console.print()
 
             if ext_manifest.commands:
+                # Print each command the way the active agent registers it.
+                # Cline and Forge hyphenate command names (e.g. Forge invokes
+                # `/speckit-jira-sync`, not the manifest's dotted
+                # `speckit.jira.sync`), so mirror the same formatting used by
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3747, opened 2026-07-26
fix(extensions): tolerate non-string catalog name in display-name lookup

Noor-ul-ain001:
@
## Summary

`_resolve_catalog_extension()` (used by `extension info <name>` and `extension add <name>`) resolves an argument to a catalog entry by display name:

```python
search_results = catalog.search(query=argument)
name_matches = [ext for ext in search_results if ext["name"].lower() == argument.lower()]
```

Extension catalog JSON is user-editable, so `catalog.search()` can return an entry whose `name` is a non-string. A hand-authored `name: 123` crashes the filter:

```
AttributeError: int object has no attribute lower
```

which takes down `extension info <name>` / `extension add <name>` with a raw traceback. A missing `name` key would similarly raise `KeyError`.

Notably, the **ambiguous-match display block a few lines below already `str()`-coerces the name** (`_escape_markup(str(ext.get("name", "")))`) — so the code already acknowledges `name` may be non-string/absent; only the filter line missed it.

## Fix

Coerce defensively with `str(ext.get("name", "")).lower()`, matching that sibling block. A bad-named entry simply fails to match, producing a clean not-found error instead of a traceback. Same bug class as the presets fix in #3743.

## Testing

Adds `test_add_by_name_tolerates_non_string_catalog_name`, which invokes `extension info <name>` against a mocked catalog whose search result has `name: 123`. It fails pre-fix with `AttributeError` and passes after; the existing display-name resolution test stays green.

@

### DIFF
--- a/src/specify_cli/extensions/_commands.py
+++ b/src/specify_cli/extensions/_commands.py
@@ -166,9 +166,17 @@ def _resolve_catalog_extension(
         if ext_info:
             return (ext_info, None)
 
-        # Try by display name - search using argument as query, then filter for exact match
+        # Try by display name - search using argument as query, then filter for exact match.
+        # Coerce name defensively: catalog JSON is user-editable, so a hand-authored
+        # non-string/missing name must not crash the match (the ambiguous-match display
+        # below already str()-coerces name for the same reason).
         search_results = catalog.search(query=argument)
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

mnriem (approved):

PR #3753, opened 2026-07-26
fix(powershell): resolve SPECIFY_INIT_DIR without .NET Core-only API

### LINKED ISSUE #3749: common.ps1 uses .NET Core-only TrimEndingDirectorySeparator, breaking SPECIFY_INIT_DIR on Windows PowerShell 5.1
### Summary

`scripts/powershell/common.ps1` resolves the project root with a .NET Core-only API, which is unavailable on Windows PowerShell 5.1 (.NET Framework). When `SPECIFY_INIT_DIR` is set, root resolution throws before any Spec Kit command can run.

The same file already documents this exact incompatibility — and then avoids it correctly 150 lines later.

### The inconsistency

`scripts/powershell/common.ps1:60`:

```powershell
$initRoot = [System.IO.Path]::TrimEndingDirectorySeparator($resolved.Path)
```

`scripts/powershell/common.ps1:212-214`:

```powershell
# TrimEnd (not [Path]::TrimEndingDirectorySeparator, which is .NET Core
# only) keeps this working on Windows PowerShell 5.1 / .NET Framework.
$featureDirTrimmed = $featureDir.TrimEnd('/', '\')
```

So the codebase already knows `TrimEndingDirectorySeparator` is .NET Core only, and line 60 uses it anyway.

### Impact

Windows PowerShell 5.1 users who set `SPECIFY_INIT_DIR` hit a `Method invocation failed` error during project-root resolution, before the requested command executes. Users on PowerShell 7+ are unaffected, which is likely why this hasn't surfaced widely.

### Suggested fix

Use the same normalization the file already endorses:

```powershell
$initRoot = $resolved.Path.TrimEnd('/', '\')
```

The stated no-op semantics still hold: `TrimEnd` is a no-op on a path with no trailing separator. Note it differs on a bare drive root (`C:\` → `C:`), which is the one case worth a guard if that path is reachable here.

### Version

Reproduced on `main` @ `c0fe0e4` (v0.14.2).

Found by an automated review pass over a vendored copy in a downstream project; verified against upstream `main` before filing.

Quratulain-bilal:
Fixes #3749

### Problem

`scripts/powershell/common.ps1` resolves the `SPECIFY_INIT_DIR` root with `[System.IO.Path]::TrimEndingDirectorySeparator(...)`, which only exists on .NET Core. On **Windows PowerShell 5.1** (.NET Framework) the call throws:

```
Method invocation failed because [System.IO.Path] does not contain a method named 'TrimEndingDirectorySeparator'.
```

so project-root resolution fails before any Spec Kit command can run whenever `SPECIFY_INIT_DIR` is set. PowerShell 7+ users are unaffected (the API exists there), which is why it went unnoticed. The same file already documents this incompatibility ~150 lines later and correctly uses `.TrimEnd('/','')`.

### Fix

Use the `.TrimEnd('/','')` approach the file already endorses, guarding the single case where the two differ — a bare drive root (`C:\`), where `TrimEnd` would strip the separator and yield an invalid `C:` — by preserving the original path there.

### Verification

Reproduced and verified on **Windows PowerShell 5.1** (5.1.26100, Desktop edition):

- Old line: `THROWS: Method invocation failed ... TrimEndingDirectorySeparator`
- New line: resolves correctly, `.specify/` detected.

The existing `@requires_pwsh` tests in `tests/test_init_dir.py` already exercise this path (trailing-slash, relative, precedence, compose). On PS 5.1 **6 of them fail on the current code and all pass with this fix**; they continue to pass on pwsh 7+. Edge cases confirmed: `C:ooar\ -> C:ooar`, `C:oo/ -> C:oo`, `C:\ -> C:\` (preserved), no-op when already clean.

### DIFF
--- a/scripts/powershell/common.ps1
+++ b/scripts/powershell/common.ps1
@@ -60,8 +60,12 @@ function Resolve-SpecifyInitDir {
     # matching note further below). TrimEnd is a no-op on a path with no trailing
     # separator; the one difference is a bare drive root (`C:\` -> `C:`), so keep
     # that separator to preserve a valid rooted path.
+    # A filesystem root is all separator: trimming it would leave an empty (or
+    # drive-only) string and break the Join-Path below. Covers the Windows
+    # drive root (`C:\` -> `C:`) and the POSIX root (`/` -> ``), which
+    # PowerShell on Linux/macOS can return.
     $initRoot = $resolved.Path.TrimEnd('/', '\')
-    if ($initRoot -match '^[A-Za-z]:$') {
+    if (-not $initRoot -or $initRoot -match '^[A-Za-z]:$') {
         $initRoot = $resolved.Path
     }
     if (-not (Test-Path -LiteralPath (Join-Path $initRoot '.specify') -PathType Container)) {
--- a/scripts/powershell/common.ps1
+++ b/scripts/powershell/common.ps1
@@ -55,9 +55,15 @@ function Resolve-SpecifyInitDir {
     }
     # Resolve-Path echoes back any trailing separator from the input; trim it so
     # the returned root matches the bash resolver, whose `cd && pwd` never yields
-    # one. TrimEndingDirectorySeparator is a no-op on a bare root and on a path
-    # that already has no trailing separator.
-    $initRoot = [System.IO.Path]::TrimEndingDirectorySeparator($resolved.Path)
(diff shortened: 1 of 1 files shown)


mnriem (changes requested):

mnriem:

PR #3757, opened 2026-07-27
fix(agent-context): recurse for nested plans in Python mtime fallback

### LINKED ISSUE #3733: [Bug]: agent-context Python port misses nested plans (one-level glob), a parity regression of #3024

Quratulain-bilal:
Fixes #3733

### Problem

The agent-context **Python** port's mtime fallback discovered plans with a one-level glob:

```python
(root / "specs").glob("*/plan.md")
```

So a scoped layout created via `SPECIFY_FEATURE_DIRECTORY` — `specs/<scope>/<feature>/plan.md`, which sits two levels below `specs/` — was missed whenever `feature.json` is absent. The fallback returned no plan, and the managed context section omitted its `at <plan>` line.

The **bash** and **PowerShell** twins were already fixed to recurse in #3024 (their comments explicitly call out "rather than the old one-level `specs/*/plan.md` scan"); the Python twin was left behind, so the three drifted out of parity.

### Fix

Switch to `specs.rglob("plan.md")` with the same symlink-safe containment check the bash twin uses: resolve each candidate and confirm it stays within the project root before ranking by mtime. This keeps a plan reached through a `specs/` symlink pointing outside the project from being selected (the lexical `relative_to` alone would accept it).

### Tests

Adds two parity regression tests covering a nested `specs/<scope>/<feature>/plan.md` with no `feature.json`:
- vs **bash** (`@requires_posix_bash`, runs in CI)
- vs **PowerShell** (runs on the Windows runner)

Both **fail on the pre-fix one-level glob** and pass after. Verified locally on Windows: the PowerShell-parity test fails without the fix (Python omits the nested plan, diverging from PowerShell) and passes with it.

### DIFF
--- a/extensions/agent-context/scripts/python/update_agent_context.py
+++ b/extensions/agent-context/scripts/python/update_agent_context.py
@@ -11,9 +11,8 @@ Usage: update_agent_context.py [plan_path]
 
 When ``plan_path`` is omitted, the script derives it from
 ``.specify/feature.json`` (written by /speckit-specify). Falls back to the most
-recently modified ``plan.md`` anywhere under ``specs/`` (including nested scoped
-layouts such as ``specs/<scope>/<feature>/plan.md``) only when feature.json is
-absent or its plan does not exist yet.
+recently modified ``specs/*/plan.md`` only when feature.json is absent or its
+plan does not exist yet.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

Quratulain-bilal:
@merium Conflict resolved and pushed. Branch is now rebased on latest main. Please review!

mnriem (approved):

mnriem:
Thank you!

PR #3765, opened 2026-07-27
fix: correct Optional type annotation for context_note parameter

Quratulain-bilal:
## Summary

The `context_note` parameter in `CommandRegistrar` methods was annotated as `str = None` - a type lie where the default is `None` but the type hint says `str`. Static type checkers (mypy/pyright) flag this as an error.

## Changes

- **`src/specify_cli/agents.py`** (lines 273, 600, 1019): Changed `context_note: str = None` to `context_note: Optional[str] = None`, consistent with how `extension_id` in the same class is already typed.

## Testing

This is a type-correctness fix - no runtime behavior change. Existing tests continue to pass.

### DIFF
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -270,7 +270,7 @@ class CommandRegistrar:
         return text
 
     def render_markdown_command(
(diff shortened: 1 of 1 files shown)


mnriem:
Thank you!

PR #3771, opened 2026-07-28
fix: correct nullable resolved directory annotation

marcelsafin:
## Description

`CommandRegistrar.register_commands()` accepts `None` for its internal
`_resolved_dir` override, but annotated the parameter as `Path`. This corrects
the type contract to `Optional[Path]`, matching the existing default and nearby
optional parameters. Runtime behavior is unchanged.

The regression test resolves the runtime type hints and verifies that
`_resolved_dir` accepts `None`.

## Testing

Full suite: 5422 passed, 172 skipped.

## AI Disclosure

GitHub Copilot (GPT-5.6 Sol) autonomously identified, implemented, tested, and
self-reviewed this change on behalf of @marcelsafin.

### DIFF
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -598,7 +598,7 @@ class CommandRegistrar:
         source_dir: Path,
         project_root: Path,
         context_note: Optional[str] = None,
-        _resolved_dir: Path = None,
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

Issue #3737, opened 2026-07-25
[Bug]: Constitution updates do not synchronize active preset-provided task templates


rcollette:
### Bug Description

# Synchronize Spec Kit constitution updates across built-in and preset-provided templates

## Problem

The `explicit-task-dependencies` preset supplies and activates its own `tasks-template.md`:

```text
.specify/presets/explicit-task-dependencies/templates/tasks-template.md
```

The preset declares this template as `replace`, so it can supersede the built-in task template. A constitution update
that synchronizes only built-in templates can therefore leave the active preset template stale. A Sync Impact Report can
incorrectly report the preset template as updated because no automated validation verifies the effective template stack.

The synchronization mechanism must preserve every template provider's specialized content while applying applicable
constitutional guidance.

## Scope

- Constitution updates must identify **all effective templates**: built-in, overrides, and enabled preset templates.
- A constitution amendment affecting task generation must synchronize both built-in and active preset task templates.
- The approach must generalize to templates supplied by any enabled preset, not hard-code this preset.
- Preserve each template provider's specialized content and associated generation behavior.

## Proposed Resolution

1. Add a reusable template-stack discovery mechanism for the constitution workflow. It must use the same precedence as
   task-template resolution: template overrides, enabled presets ordered by priority, extension-provided templates, and
   built-in templates.
2. Update the `speckit.constitution` agent/prompt checklist to enumerate every resolved template provider, including
   preset templates and preset command files that replace generation behavior.
3. Require constitution-driven task guidance to be propagated to every effective `tasks-template.md`.
4. Define a template-provider-agnostic synchronization contract that applies constitutional guidance without removing
   provider-specific structure, instructions, or generation behavior.
5. Make the Sync Impact Report derive its “updated” and “pending” template entries from the resolved template stack. It
   must not mark a template as updated unless that file was actually reviewed and changed or explicitly verified as
   already compliant.
6. Add a validation that fails or reports a pending follow-up when the active replacement template omits mandatory
   guidance present in the constitution-driven task-generation requirements.

## Acceptance Criteria

- [ ] Running a constitution update discovers enabled preset templates.
- [ ] The update workflow identifies `explicit-task-dependencies` as the active replacement for `tasks-template`.
- [ ] Both the built-in and active preset task templates receive applicable constitution-driven task guidance.
- [ ] Synchronization preserves the specialized content and generation behavior of each affected template provider.
- [ ] The preset template reflects applicable testing, implementation, and validation requirements from the
[...]

PR #3782, opened 2026-07-28
fix(templates): resolve dependent templates through the preset stack in constitution sync (#3737)

jawwad-ali:
Addresses #3737 (the reporter's items 1–3 and 5 — see *Scope* for what is deliberately left out).

## Problem

Step 4 of the constitution command hardcodes the three dependent template paths:

```
- Read `.specify/templates/plan-template.md`  ...
- Read `.specify/templates/spec-template.md`  ...
- Read `.specify/templates/tasks-template.md` ...
```

But a project override, an **enabled preset**, or an extension can supply the *active* template. When one does, the hardcoded path is **not** the file the other Spec Kit commands read, so the provider's copy is silently left stale — and the Sync Impact Report can list it as updated without it ever having been read.

Reproduced against `main` with the real resolver (throwaway tempdir; an enabled `replace`-strategy preset supplying `tasks-template`, exactly the pattern `presets/self-test/preset.yml` uses):

```
collect_all_layers('tasks-template'):
    explicit-task-dependencies v1.0.0 | replace | .specify/presets/.../tasks-template.md
    core                              | replace | .specify/templates/tasks-template.md

constitution.md step 4 reads : .specify/templates/tasks-template.md   -> "# CORE tasks template"
/speckit.tasks actually uses : .specify/presets/.../tasks-template.md -> "# PRESET tasks template (replace)"
MISMATCH: True
```

The same hardcoding applies to `plan-template` (`scripts/bash/setup-plan.sh` uses `resolve_template "plan-template"`) and `spec-template`, so all three bullets are affected — not just tasks.

## Fix

Prompt-level only. Instruct the agent to resolve each dependent template through the preset/template resolution stack before reading it, and to propagate the updated guidance to **every layer the stack reports** (active provider + built-in fallback), preserving each provider's specialized content.

This mirrors the **already-merged precedent** 14da893 — *"fix(copilot): resolve active spec template (#2765)"* — which fixed the identical class of hardcoding in `specify.md`, using the same wording and the same existing command:

```diff
-   - Copy `templates/spec-template.md` to ... as the starting point
+   - Resolve the active `spec-template` through the Spec Kit preset/template resolution stack (equivalent to `specify preset resolve spec-template`)
```

`specify preset resolve` already exists (`presets/_commands.py`) and already calls `collect_all_layers()`, so there is **no new API, no new `{SCRIPT}`, and no sh/ps/py parity work**. This is also what the issue's own bug assessment labels the *"Preferred"* remediation: `TemplateResolver.collect_all_layers()` already implements the precedence correctly — the prompt simply never used it.

Step 5's Sync Impact Report line is updated to require listing every reported layer and to mark a file ✅ updated only when it was actually read and changed.

## Verification

- New `test_constitution_agent_resolves_dependent_templates` asserts the generated agent file resolves all three templates and no longer carries the hardcoded reads. **Fails before this change**, passes after. Placed beside its precedent test, as 14da893 did.
- `tests/integrations/test_integration_copilot.py`: **53 passed**.
- `markdownlint-cli2` error count on this file is **identical before and after** (14 pre-existing, unchanged by this diff).

## Scope

Reporter items 4 and 6 — a provider-agnostic sync contract enforced in code, and a hard validation that *fails* when a `replace`-strategy preset omits mandated guidance — are intentionally **not** included: both need new CLI/validation surface and a product decision. Happy to follow up if you want that direction.

---
*AI-assisted: authored with Claude Code. I reproduced the MISMATCH on `main` with the real `TemplateResolver` before writing the fix, and mirrored the merged 14da893 precedent for both wording and test placement.*

### DIFF
--- a/templates/commands/constitution.md
+++ b/templates/commands/constitution.md
@@ -97,9 +97,10 @@ Follow this execution flow:
    - Ensure Governance section lists amendment procedure, versioning policy, and compliance review expectations.
 
 4. Consistency propagation checklist (convert prior checklist into active validations):
-   - Read `.specify/templates/plan-template.md` and ensure any "Constitution Check" or rules align with updated principles.
-   - Read `.specify/templates/spec-template.md` for scope/requirements alignment—update if constitution adds/removes mandatory sections or constraints.
-   - Read `.specify/templates/tasks-template.md` and ensure task categorization reflects new or removed principle-driven task types (e.g., observability, versioning, testing discipline).
+   - Resolve each dependent template through the Spec Kit preset/template resolution stack before reading it (equivalent to `specify preset resolve <template-name>`). A project override, an enabled preset, or an extension can supply the active `plan-template`, `spec-template`, or `tasks-template`; when one does, `.specify/templates/<template-name>.md` is NOT the file the other Spec Kit commands read. Propagate the updated guidance to every layer the stack reports — the active provider plus the built-in `.specify/templates/<template-name>.md` fallback — preserving each provider's specialized content and structure rather than overwriting it.
+   - Read the resolved `plan-template` and ensure any "Constitution Check" or rules align with updated principles.
+   - Read the resolved `spec-template` for scope/requirements alignment—update if constitution adds/removes mandatory sections or constraints.
+   - Read the resolved `tasks-template` and ensure task categorization reflects new or removed principle-driven task types (e.g., observability, versioning, testing discipline).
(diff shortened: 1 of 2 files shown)


rcollette:
The AI review made it pretty clear that cli assistance would make the fix more deterministic.  With AI authoring the code, it might as well go for the better long term solution.

mnriem:
Closing this as not planned as it was superseded by #3790

PR #3783, opened 2026-07-28
fix(workflows): escape the step-progress line so step ids render (and `/` stops failing the run)

jawwad-ali:
## Problem

`workflow run` and `workflow resume` install the same step-progress callback:

```python
engine.on_step_start = lambda sid, label: console.print(f"  ▸ [{sid}] {label} …")
```

Rich parses the bracketed step id as a **style tag**. Both arguments come from workflow YAML and neither is escaped, which produces three distinct failures on `main`:

**1. The step id is silently swallowed on every run** — the only identifying content on the line.

```
workflow with `id: greet`  ->  "  ▸  shell …"      ("[greet]" absent)
```

**2. An id that forms a closing tag fails the whole run.** `validate_workflow` places no charset restriction on step ids (it rejects only non-strings, `:`, and duplicates), so `id: "/"` is an accepted workflow. The callback then raises `MarkupError`, which propagates into `execute()`'s handler:

```
rich.errors.MarkupError: closing tag '[/]' at position 4 has nothing to close
-> run persisted as `failed`, step_results: {}   (the shell step never ran)
-> exit 1 with a Rich internals error
```

**3. An id that is a real style (`bold`, `red`, `dim`) is applied as formatting** to the rest of the line. The unescaped `label` (from `step_config["command"]`) compounds all of this.

## Fix

Escape the literal bracket with `\[` and escape both interpolated values — at **both** sites (`workflow_run` line 1051 and `workflow_resume` line 1173; the arrow appears exactly twice in `src/`).

This mirrors the `\[<type>]` step-graph precedent already in this file (`workflow_info`), including its comment rationale.

Escaping only the values is **not** sufficient — the f-string's own brackets are what Rich consumes. Verified:

| step id | before | escape values only | this fix |
| --- | --- | --- | --- |
| `greet` | `  ▸  shell …` | `  ▸  shell …` | `  ▸ [greet] shell …` |
| `bold` | `  ▸  shell …` | `  ▸  shell …` | `  ▸ [bold] shell …` |
| `/` | **MarkupError** | **MarkupError** | `  ▸ [/] shell …` |
| `a]b` | `  ▸ b] shell …` | `  ▸ b] shell …` | `  ▸ [a]b] shell …` |

## Verification

Driven through the real CLI (`CliRunner`, `workflow run`):

```
id='greet' exit=0 id_visible=True  line='  > [greet] shell ...'
id='bold'  exit=0 id_visible=True  line='  > [bold] shell ...'
id='/'     exit=0 id_visible=True  line='  > [/] shell ...'      <- was a failed run
id='a]b'   exit=0 id_visible=True  line='  > [a]b] shell ...'
```

- 5 new tests (3 parametrized ids, the `/` crash, plus a **resume-side** case that drives `run --json` → `resume <run_id>` so the twin is covered independently): **all fail before this change**, pass after.
- `TestWorkflowRunExitCodes` (the neighbouring pre-existing class) still green — **9 passed** together.
- Nothing depended on the broken output: no test or doc references the progress line (`docs/reference/workflows.md` documents only the `--json` payload).

---
*AI-assisted: authored with Claude Code. I reproduced all three failure modes on `main` through the real CLI, confirmed that escaping only the interpolated values leaves the bug, and matched the existing `\[` convention in this file.*

### DIFF
--- a/src/specify_cli/workflows/_commands.py
+++ b/src/specify_cli/workflows/_commands.py
@@ -1054,7 +1054,18 @@ def workflow_run(
     load_custom_steps(project_root)
     engine = WorkflowEngine(project_root)
     if not json_output:
-        engine.on_step_start = lambda sid, label: console.print(f"  \u25b8 [{sid}] {label} \u2026")
+        # Escape the literal bracket (\[) so Rich renders `[<step id>]` instead
+        # of parsing it as a style tag named after the step id -- which it
+        # silently swallows (losing the only identifying content on the line),
+        # applies as formatting when the id happens to be a real style such as
+        # `bold`, or raises MarkupError when the id forms a closing tag (`/`),
+        # failing the whole run. Escape the interpolated values too, since both
+        # come from workflow YAML. Mirrors the `\[<type>]` step-graph precedent
+        # in workflow_info below.
+        engine.on_step_start = lambda sid, label: console.print(
+            f"  \u25b8 \\[{_escape_markup(str(sid))}] "
+            f"{_escape_markup(str(label))} \u2026"
+        )
 
     err = _error_console(json_output)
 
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3784, opened 2026-07-28
fix(bundler): degrade non-UTF-8 config reads into BundlerError

jawwad-ali:
## Problem

`bundler/lib/yamlio.py` is the single chokepoint for every bundler read, and its module docstring states the contract explicitly:

> All reads/writes go through these functions so that **IO failures degrade into actionable `BundlerError`s rather than raw tracebacks**.

Both readers catch only `OSError`. But `Path.read_text(encoding="utf-8")` and `json.load()` raise `UnicodeDecodeError` on a non-UTF-8 file, and:

```python
issubclass(UnicodeDecodeError, OSError)  # False   (MRO: UnicodeError -> ValueError)
```

So the decode error escapes uncaught. Reproduced on `main` with a UTF-16 file:

```
load_yaml: LEAKED UnicodeDecodeError -> 'utf-8' codec can't decode byte 0xff in position 0
load_json: LEAKED UnicodeDecodeError -> 'utf-8' codec can't decode byte 0xff in position 0
```

In `load_json`, `json.JSONDecodeError` does **not** cover it — the two are *sibling* `ValueError` subclasses, neither subsumes the other.

This is realistic rather than theoretical: on Windows, PowerShell 5.1's `Out-File` and `>` default to **UTF-16**, so a hand-edited `.specify/bundle-catalogs.yml` or records file lands here.

## Fix

Widen both **read** clauses to `(OSError, UnicodeError)` — matching the sibling catalog readers (`catalogs.py:101`, `workflows/catalog.py:336`, which already use exactly this tuple).

`json.JSONDecodeError` deliberately stays **first** in `load_json`, so malformed-but-decodable JSON keeps its more specific `Invalid JSON` message while a decode failure falls through to the read-error clause. A regression test locks that ordering.

**Scope:** reads only. I verified the write paths do *not* leak `UnicodeEncodeError` — `dump_yaml`/`dump_json` both escape unencodable input (tested with a lone surrogate), so there is nothing to fix there.

## Verification

- `test_load_yaml_non_utf8_raises_bundler_error` and `test_load_json_non_utf8_raises_bundler_error` (UTF-16 input): **fail before** with the raw `UnicodeDecodeError`, pass after.
- `test_load_json_malformed_still_reports_invalid_json`: clause-order guard — decodable-but-malformed JSON still reports `Invalid JSON`.
- `tests/unit/test_bundler_yamlio.py`: **5 passed**.
- `tests/unit/` + `tests/contract/`: **252 passed** (2 pre-existing Windows symlink-privilege failures, unrelated — they need admin and pass on CI's Linux/macOS runners).

No behaviour change for UTF-8 input: those reads take the identical path and return identical values.

---
*AI-assisted: authored with Claude Code. I reproduced both leaks on `main`, checked `issubclass` rather than assuming, verified the write paths were safe before scoping the fix, and matched the existing sibling-reader tuple.*

### DIFF
--- a/src/specify_cli/bundler/lib/yamlio.py
+++ b/src/specify_cli/bundler/lib/yamlio.py
@@ -58,7 +58,12 @@ def load_yaml(path: Path) -> Any:
         raise BundlerError(f"File not found: {path}")
     try:
         text = path.read_text(encoding="utf-8")
-    except OSError as exc:
+    except (OSError, UnicodeError) as exc:
+        # A non-UTF-8 file raises UnicodeDecodeError, which is a ValueError --
+        # NOT an OSError -- so it escaped this module's "IO failures degrade
+        # into actionable BundlerError" contract as a raw traceback. Realistic
+        # on Windows, where PowerShell 5.1's `Out-File`/`>` default to UTF-16.
+        # Matches the sibling catalog readers (catalogs.py, workflows/catalog.py).
         raise BundlerError(f"Could not read {path}: {exc}") from exc
     try:
         has_node = yaml.compose(text) is not None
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3797, opened 2026-07-28
fix(extensions): guard the required manifest sections so one bad extension cannot break `extension list`

jawwad-ali:
## Problem

`ExtensionManifest.REQUIRED_FIELDS` only checks key **presence**:

```python
for field in self.REQUIRED_FIELDS:      # schema_version, extension, requires, provides
    if field not in self.data:
        raise ValidationError(...)
```

So a section that is *written but left empty* (`provides:` → `None`, which is just how YAML spells an empty field) or given the wrong shape passes that check and then fails on first use:

```
extension: null  -> TypeError: argument of type 'NoneType' is not iterable
requires:  null  -> TypeError: argument of type 'NoneType' is not iterable
provides:  null  -> AttributeError: 'NoneType' object has no attribute 'get'
provides:  []    -> AttributeError: 'list' object has no attribute 'get'
```

Neither is a `ValidationError`, so both escape the callers that **already handle** malformed manifests. `list_installed()` catches `ValidationError` only and has a deliberate `⚠️ Corrupted extension` fallback — so a single bad extension took down the entire command. Reproduced end-to-end with one broken and one good extension installed:

| | `specify extension list` |
| --- | --- |
| **before** | **exit 1**, raw `AttributeError`, no output at all |
| **after** | exit 0 — the good extension is listed, the bad one shows as `Corrupted extension` |

## Fix

Add an `isinstance` guard for each required section. This mirrors the guards **already in the same function** for the *nested optional* keys —

```python
if "commands" in provides and not isinstance(commands, list):
    raise ValidationError("Invalid provides.commands: expected a list")
if "hooks" in self.data and not isinstance(hooks, dict):
    raise ValidationError("Invalid hooks: expected a mapping")
```

— and `_load_yaml`'s document-root check. Only the three **required** sections lacked one; they were the odd ones out.

## Not a behaviour change for valid input

`provides: {}` is a well-*shaped* mapping, so the new check does not touch it:

- with hooks present → still validates (an extension may provide only hooks)
- with no hooks → still gets the pre-existing `must provide at least one command or hook`

Both are locked by regression tests. I got this wrong on the first attempt — my initial test asserted `provides: {}` must raise, which failed because the fixture defines hooks; the tests now assert the correct behaviour.

## Verification

- `test_required_section_not_mapping_rejected` — 3 sections × 3 bad shapes (`None`, `[]`, `"text"`): **9 fail before**, all pass after.
- 2 regression guards for the `provides: {}` cases above.
- `TestExtensionManifest`: **55 passed**.

---
*AI-assisted: authored with Claude Code. I reproduced the `extension list` exit-1 crash end-to-end (correct `.registry` layout, one broken + one good extension) and confirmed the graceful fallback now engages, rather than only testing the manifest in isolation.*

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -263,8 +263,25 @@ class ExtensionManifest:
                 f"(expected {self.SCHEMA_VERSION})"
             )
 
+        # The REQUIRED_FIELDS loop above only checks key PRESENCE, so a section
+        # that is written but left empty (``provides:`` -> None) or given the
+        # wrong shape (``provides: []``) passes it and then fails on first use:
+        # ``field not in None`` raises TypeError and ``None.get(...)`` raises
+        # AttributeError. Neither is a ValidationError, so both escape the
+        # callers that already handle malformed manifests -- list_installed()'s
+        # "Corrupted extension" fallback catches ValidationError only, so one bad
+        # extension made ``specify extension list`` exit 1 with a raw
+        # AttributeError instead of listing the rest. Guard each required
+        # section's shape, mirroring the nested guards below ("Invalid
+        # provides.commands: expected a list", "Invalid hooks: expected a
+        # mapping") and _load_yaml's document-root check.
+
         # Validate extension metadata
         ext = self.data["extension"]
+        if not isinstance(ext, dict):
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3798, opened 2026-07-28
fix(bundler): treat an explicit-null manifest field as missing, not the text "None"

jawwad-ali:
## Problem

`BundleManifest.from_dict` reads every required scalar as `str(raw.get(key, "")).strip()`. The `""` default only covers a **missing** key. A key that is *present but null* — exactly how YAML spells an empty field — yields `None`, and `str(None)` is the literal string `"None"`. That value is non-empty, so it sails straight past the `if not value` required-field checks in `structural_errors()`.

```yaml
bundle:
  id: demo
  name: Demo
  version: 1.0.0
  role: developer
  description:      # <- empty
  author:           # <- empty
  license:          # <- empty
```

Reproduced on `main`:

```
description='None'   author='None'   license='None'
structural_errors() == []
specify bundle validate -> exit 0: "demo is well-formed and valid."
```

So an empty required field is **silently accepted**, and the bundle ships the literal text `"None"` as its author/license/description — which is what `bundle info` and a catalog entry then display to users. A null `provides.<kind>[].id` likewise became a component literally named `"None"`.

## Fix

Add a `_text()` helper beside the existing `_parse_str_list` — matching this file's established "one coercion helper, applied at every site" shape — that maps an explicit null to `""`, and route the required scalars (`schema_version`, the six `bundle.*` fields, `requires.speckit_version`, component `id`) through it.

After:

```
description=''  author=''
structural_errors():
  - Missing required field: bundle.description.
  - Missing required field: bundle.author.
  - Missing required field: bundle.license.
```

This is the same silent-acceptance class as the already-merged guards in this exact function — #3629 (present-but-non-mapping `integration:` was silently dropped) and #3661 (falsy non-mapping `requires`/`provides`).

## Not a behaviour change for valid input

`_text` is equivalent to the code it replaces for every input except an explicit `None`: an absent key already produced `""`, and non-null scalars are still `str()`-coerced and stripped. Verified — a valid manifest still yields `structural_errors() == []` with `author='Ali'`.

## Verification

- `test_explicit_null_bundle_field_is_reported_as_missing` (5 parametrized fields) + null `speckit_version` + null component id: **7 fail before**, pass after.
- `tests/contract/test_manifest_schema.py`: **35 passed** (28 pre-existing untouched).
- `tests/contract/` + `tests/unit/`: **272 passed** — the 3 remaining failures are pre-existing Windows symlink-privilege tests (need admin; pass on CI's Linux/macOS runners).

---
*AI-assisted: authored with Claude Code. I reproduced the `"None"` values and the clean `bundle validate` on `main` first, then confirmed valid manifests are unaffected after the change.*

### DIFF
--- a/src/specify_cli/bundler/models/manifest.py
+++ b/src/specify_cli/bundler/models/manifest.py
@@ -96,19 +96,19 @@ class BundleManifest:
         if not isinstance(data, dict):
             raise BundlerError("Manifest must be a YAML mapping at the top level.")
 
-        schema_version = str(data.get("schema_version", "")).strip()
+        schema_version = _text(data.get("schema_version"))
 
         bundle_raw = data.get("bundle")
         if not isinstance(bundle_raw, dict):
             raise BundlerError("Manifest is missing the required 'bundle' mapping.")
         meta = BundleMeta(
-            id=str(bundle_raw.get("id", "")).strip(),
-            name=str(bundle_raw.get("name", "")).strip(),
-            version=str(bundle_raw.get("version", "")).strip(),
-            role=str(bundle_raw.get("role", "")).strip(),
-            description=str(bundle_raw.get("description", "")).strip(),
-            author=str(bundle_raw.get("author", "")).strip(),
-            license=str(bundle_raw.get("license", "")).strip(),
+            id=_text(bundle_raw.get("id")),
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3810, opened 2026-07-28
fix: add missing utf-8 encoding to registry file open calls

Quratulain-bilal:
Add encoding='utf-8' to 4 open() calls in extensions/__init__.py (lines 538, 554) and presets/__init__.py (lines 460, 481) that read/write registry JSON files. On Windows with non-UTF-8 locales, these would produce UnicodeDecodeError or corrupted data. The parallel code in workflows/catalog.py already specifies encoding='utf-8' correctly.


mnriem:
Thank you!

PR #3811, opened 2026-07-28
fix: eliminate TOCTOU race in file unlink calls

Quratulain-bilal:
Replace if path.exists(): path.unlink() with path.unlink(missing_ok=True) in integrations/_helpers.py and extensions/__init__.py to eliminate TOCTOU race conditions where the file could be removed between the check and the unlink.


mnriem:
Thank you!

PR #3816, opened 2026-07-28
fix: add utf-8 encoding to registry file open calls

Quratulain-bilal:
Add encoding='utf-8' to 4 open() calls in extensions/__init__.py and presets/__init__.py that read/write registry JSON files. On Windows with non-UTF-8 locales, these would produce UnicodeDecodeError or corrupted data.


mnriem:
Thank you!

PR #3817, opened 2026-07-28
fix: skip corrupted run state files in list_runs

Quratulain-bilal:
Wrap json.load() in workflows/engine.py list_runs with 	ry/except (json.JSONDecodeError, OSError) so a single corrupted state.json doesn't crash the entire listing.


mnriem:
Thank you!

PR #3818, opened 2026-07-28
fix: bound response read in integration catalog fetch

Quratulain-bilal:
Replace unbounded 
esp.read() with 
ead_response_limited() in integrations/catalog.py to prevent DoS via oversized catalog responses.


mnriem:
Thank you!

PR #3819, opened 2026-07-28
fix: eliminate TOCTOU race in file unlink calls

Quratulain-bilal:
Replace if path.exists(): path.unlink() with path.unlink(missing_ok=True) in extensions/__init__.py clear_cache and integrations/_helpers.py to eliminate TOCTOU race conditions.


mnriem:
Thank you!

PR #3773, opened 2026-07-28
fix(presets): escape catalog metadata in discovery output

marcelsafin:
## Description

Preset catalog values such as names, IDs, versions, descriptions, authors,
repository URLs, and licenses were interpolated directly into Rich output by
`preset search` and the catalog branch of `preset info`. Bracketed content was
therefore interpreted as markup instead of rendered literally.

This escapes those remaining catalog-derived fields with the module's existing
`_escape_markup` helper. Tag shape validation and tag rendering are intentionally
left to #3769, which already covers that separate crash path and overlapping tag
escape.

Regression tests assign distinct Rich tags to every field in scope and verify
literal rendering in both commands.

## Testing

- [x] Tested locally with `uv run specify --help`
- [x] Ran existing tests with `uv sync && uv run pytest`
- [ ] Tested with a sample project (if applicable)

Full suite: 5423 passed, 172 skipped.

## AI Disclosure

- [ ] I **did not** use AI assistance for this contribution
- [x] I **did** use AI assistance (describe below)

GitHub Copilot (GPT-5.6 Sol) autonomously identified, implemented, tested, and
self-reviewed this change on behalf of @marcelsafin.

PR #3826, opened 2026-07-28
fix(presets): escape installed preset metadata in Rich output

Noor-ul-ain001:
## Problem

`preset.yml` is user-editable, but the **installed-preset** display paths interpolate its fields straight into `console.print`, where Rich parses `[...]` as a style tag.

#3773 escaped the *catalog* branch of these commands; the local branch was left behind — so the same field renders correctly from a catalog and incorrectly once installed.

**Silent data loss.** Description `Does [stuff] nicely`:

```
  Description: Does  nicely
```

**Hard crash.** An unbalanced tag like `Broken [/red] tag` raises `rich.errors.MarkupError`, aborting with a traceback and exit 1 — the preset can't be inspected at all:

```
$ specify preset info markup-pack
  ID:          markup-pack
  Version:     1.0.0
MarkupError: closing tag '[/red]' at position 22 doesn't match any open tag
```

`preset resolve` was unescaped throughout, and it echoes **its own argument**, so this crashes on user input alone with no preset installed:

```
$ specify preset resolve 'no[/red]such'
MarkupError: closing tag '[/red]' at position 12 doesn't match any open tag
```

Separately, the composition chain's literal `[{strategy_label}]` was itself consumed as a style tag, so every chain line printed a **blank** label:

```
  Composition chain:
    1.  base-pack v1.0.0 → ...      # should read "1. [base] ..."
    2.  app-pack v1.0.0 → ...       # should read "2. [append] ..."
```

## Fix

Route the untrusted fields through the already-imported `_escape_markup`:

- `preset list` (installed branch) — name, id, version, description
- `preset info` (installed branch) — name, id, version, description, author, tags, repository, license, and the per-template description
- `preset info` (catalog branch) — the tags join the earlier sweep missed
- `preset resolve` — the `template_name` argument, resolved paths, layer sources, composition-error message
- composition chain — escaped the literal bracket as `\[`, matching the step-graph line in `workflow info`

No behaviour change beyond rendering: fields now display verbatim instead of being swallowed or crashing.

## Tests

`TestInstalledPresetRichMarkup` in `tests/test_presets.py` — 5 tests covering markup fields in list/info, the template description line, the unbalanced-tag crash, the `preset resolve` argument crash, and the composition strategy labels.

Test-the-test: with the source fix stashed, all 5 fail (4 on markup assertions, 1 on the swallowed template description); with it applied, all 5 pass.

`pytest tests/test_presets.py` → **475 passed, 7 failed, 1 error**, matching the pre-existing baseline on clean `main` for this Windows box (symlink tests requiring elevation + one unrelated GHES-asset error). None are in the touched paths.

### DIFF
--- a/src/specify_cli/presets/_commands.py
+++ b/src/specify_cli/presets/_commands.py
@@ -59,8 +59,11 @@ def preset_list():
     for pack in installed:
         status = "[green]enabled[/green]" if pack.get("enabled", True) else "[red]disabled[/red]"
         pri = pack.get('priority', 10)
-        console.print(f"  [bold]{pack['name']}[/bold] ({pack['id']}) v{pack['version']} — {status} — priority {pri}")
-        console.print(f"    {pack['description']}")
+        name = _escape_markup(str(pack['name']))
+        pack_id = _escape_markup(str(pack['id']))
+        version = _escape_markup(str(pack['version']))
+        console.print(f"  [bold]{name}[/bold] ({pack_id}) v{version} — {status} — priority {pri}")
+        console.print(f"    {_escape_markup(str(pack['description']))}")
         tags = pack.get("tags", [])
         if isinstance(tags, list) and tags:
             tags_str = _escape_markup(", ".join(str(t) for t in tags))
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

Noor-ul-ain001:
Thanks — both Copilot comments addressed in eb667ec (tests only, no source change).

**Catalog tags assertion.** `test_info_escapes_catalog_markup` now asserts the rendered tags join, so the catalog-branch escape is covered by its own branch's test rather than riding on the installed-preset one.

**Untested `preset resolve` escapes.** Added three cases patching `PresetResolver`:
- `test_resolve_escapes_layer_path_and_source` — markup through the top-layer path/source line
- `test_resolve_escapes_fallback_path_and_source` — the no-layer `resolve_with_source` branch
- `test_resolve_escapes_composition_error` — a markup-bearing `resolve_content` exception

One note on the path assertion: a closing tag can't be embedded in a mocked `Path` (the `/` in `[/red]` is read as a separator), so the path carries an opening tag for the swallowing case and the unbalanced tag rides on the adjacent `source` field printed on the same line.

Test-the-test: with `_commands.py` reverted to the pre-fix revision, 9 of the 10 markup tests now fail (was 5); with the fix applied all 10 pass. Full `tests/test_presets.py` → 478 passed, 7 failed, 1 error — unchanged pre-existing baseline on this Windows box (symlink tests needing elevation + one unrelated GHES-asset error).

mnriem (approved):

mnriem:
Thank you!

PR #3834, opened 2026-07-29
fix: add utf-8 encoding to extension and preset registry file I/O

Quratulain-bilal:
## Summary

Add explicit `encoding='utf-8'` to extension and preset registry file I/O to prevent corruption on non-UTF-8 Windows locales.

## Changes

- `extensions/__init__.py`: Added `encoding='utf-8'` to registry read/write
- `presets/__init__.py`: Added `encoding='utf-8'` to registry read/write


mnriem:
Thank you!

PR #3837, opened 2026-07-29
fix: escape Rich markup in workflow error output

Quratulain-bilal:
## Summary

Escape exception text with `_escape_markup()` to prevent Rich from interpreting square brackets as markup tags.

## Changes

- `_commands.py`: Added `_escape_markup(str(exc))` in ValueError and Exception handlers


mnriem (changes requested):

mnriem (approved):

mnriem:
Thank you!

PR #3838, opened 2026-07-29
fix: handle TOCTOU race in list_runs state file read

Quratulain-bilal:
## Summary

Handle TOCTOU race where `state.json` is deleted between `exists()` check and `open()`.

## Changes

- `engine.py`: Removed `exists()` guard, wrapped open/load in try/except to skip missing/corrupt state files


KSchlobohm:
Please address Copilot feedback and resolve the merge conflicts with `main`. Also add a negative-path test for the new skip-on-error branch (corrupt/missing `state.json` → skipped, valid runs still returned) — our standard is both a positive and negative test per new behavior branch.

mnriem:
The reported failure is already handled on current `main`: missing, malformed, and invalid-UTF-8 state files are skipped, including deletion between the existence check and opening the file. Valid sibling runs remain available.

Closing this as covered by #3814 and #3817. The additional logging and streaming-timeout changes are separate concerns and should not be included under this fix.

PR #3839, opened 2026-07-29
fix: remove TOCTOU race in RunState.load

Quratulain-bilal:
## Summary

Remove `exists()` check before `open()` and catch `FileNotFoundError` directly to prevent TOCTOU race.

## Changes

- `engine.py`: Replaced exists()+open() with open() wrapped in try/except FileNotFoundError


mnriem:
Thank you!

PR #3840, opened 2026-07-29
fix: show error details in preset catalog config read failure

Quratulain-bilal:
## Summary

Capture and display exception message when reading preset-catalogs.yml fails.

## Changes

- `_commands.py`: Added `as e` to except clause and display error details


mnriem:
Thank you!

PR #3841, opened 2026-07-29
fix: use chunked read for extension manifest hash

Quratulain-bilal:
## Summary

Replace unbounded `f.read()` with chunked iteration to prevent excessive memory allocation.

## Changes

- `extensions/__init__.py`: Use chunked read in `get_hash()` matching `integrations/manifest.py` pattern


mnriem:
Thank you!

PR #3843, opened 2026-07-29
fix: use chunked read for integration and preset manifest hash

Quratulain-bilal:
## Summary

Replace unbounded `f.read()` with chunked iteration to prevent excessive memory allocation.

## Changes

- `integrations/catalog.py`: Use chunked read in `get_hash()`
- `presets/__init__.py`: Use chunked read in `get_hash()`


mnriem:
Thank you!

PR #3844, opened 2026-07-29
fix: narrow bare except Exception in VS Code settings merge

Quratulain-bilal:
## Summary

Narrow bare `except Exception` to specific exception types to let programming errors propagate.

## Changes

- `_utils.py`: Replace `except Exception` with `(OSError, ValueError, KeyError)`


mnriem:
Thank you!

PR #3845, opened 2026-07-29
fix: use missing_ok=True in extension cache clear

Quratulain-bilal:
## Summary

Replace check-then-act pattern with `unlink(missing_ok=True)` to eliminate TOCTOU race.

## Changes

- `extensions/__init__.py`: Use `missing_ok=True` in `clear_cache()` matching the pattern used for per-URL cache files


mnriem:
Thank you!

PR #3846, opened 2026-07-29
fix: use missing_ok=True in integration JSON removal

Quratulain-bilal:
## Summary

Replace check-then-act pattern with `unlink(missing_ok=True)` to eliminate TOCTOU race.

## Changes

- `integrations/_helpers.py`: Use `missing_ok=True` in `_remove_integration_json()`


mnriem:
Thank you!

PR #3847, opened 2026-07-29
fix(workflows): validate prompt step 'timeout' like the shell step

Noor-ul-ain001:
## Problem

#3768 added a `timeout` to the prompt step and passed it straight into `subprocess.run(timeout=...)`. Neither `validate()` nor `execute()` checks it, so a bad value from a user-authored `workflow.yml` escapes as a raw exception.

```yaml
steps:
  - id: first
    type: shell
    run: echo side-effect
  - id: ask
    type: prompt
    prompt: do it
    timeout: abc
```

```
$ specify workflow run wf.yml
  ▸ [first] shell …
Workflow failed: unsupported operand type(s) for +: 'float' and 'str'
```

The engine re-raises anything a step throws, so this takes down the **whole run** — after `first` has already run its side effect — with a message naming neither the step nor the field.

| `timeout:` | Result today |
| --- | --- |
| `abc`, `[30]` | `TypeError: unsupported operand type(s) for +: 'float' and 'str'` — run aborts |
| `.nan` | `ValueError: cannot convert float NaN to integer` — run aborts |
| `0`, `-5` | immediate `TimeoutExpired` for a command that never got the time to run |
| `true` | silently becomes a **1-second** limit (bool is an int subclass) |

The sibling **shell** step already rejects exactly these values via a `_timeout_error()` helper shared by `execute()` and `validate()` — so the same workflow fails validation cleanly as a shell step and crashes as a prompt one.

## Fix

Mirror that helper onto `PromptStep`: `validate()` reports the contract error, and `execute()` re-checks it so an unvalidated run (the engine does not auto-validate step config) fails just that step instead of aborting.

```
Workflow validation failed:
  • Prompt step 'ask': 'timeout' must be a positive number of seconds, got 'abc'.
```

Caught before the first step runs. Positive int/float timeouts and an absent `timeout` are unaffected.

## Tests

12 tests in `TestPromptStep`, mirroring the shell step's:

- `test_validate_rejects_invalid_timeout` — `"30"`, `True`, `inf`, `nan`, `0`, `-5`, `["30"]`, `None`
- `test_validate_accepts_valid_timeout` — `300`, `5`, `0.5`, and an absent field
- `test_execute_fails_cleanly_on_invalid_timeout` — with `subprocess.run` patched to assert it is never reached

Test-the-test: with the source fix reverted, all 9 rejection tests fail; the 3 accepts-valid tests correctly still pass.

`pytest tests/test_workflows.py` → **725 passed, 17 failed, 67 errors**, identical to the pre-existing baseline on this Windows box (symlink tests requiring elevation + `pytest-of-E` tmpdir permission errors). None are in the touched paths.

### DIFF
--- a/src/specify_cli/workflows/steps/prompt/__init__.py
+++ b/src/specify_cli/workflows/steps/prompt/__init__.py
@@ -157,12 +157,16 @@ class PromptStep(StepBase):
         if "timeout" not in config:
             return None
         timeout = config["timeout"]
-        if (
-            isinstance(timeout, bool)
-            or not isinstance(timeout, (int, float))
-            or not math.isfinite(timeout)
-            or timeout <= 0
-        ):
+        try:
+            valid_timeout = (
+                not isinstance(timeout, bool)
+                and isinstance(timeout, (int, float))
+                and timeout > 0
+                and math.isfinite(timeout)
+            )
+        except OverflowError:
+            valid_timeout = False
+        if not valid_timeout:
             return (
                 f"Prompt step {config.get('id', '?')!r}: 'timeout' must be a "
                 f"positive number of seconds, got {timeout!r}."
--- a/src/specify_cli/workflows/steps/prompt/__init__.py
+++ b/src/specify_cli/workflows/steps/prompt/__init__.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import math
 import shutil
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

mnriem:
Please fix test & lint errors

mnriem (approved):

PR #3855, opened 2026-07-29
fix: eliminate TOCTOU race in zip packaging

Quratulain-bilal:
Open file once and derive both stat info and content from the same file descriptor to prevent race conditions where the file is modified between stat() and read_bytes() calls.


mnriem (changes requested):

mnriem (approved):

mnriem:
Thank you!

PR #3856, opened 2026-07-29
fix: narrow bare except Exception in invoke separator resolution

Quratulain-bilal:
Replace overly broad except Exception with (ImportError, ValueError, KeyError) to let programming errors like TypeError or AttributeError propagate.


mnriem:

mnriem:
Thank you!

PR #3857, opened 2026-07-29
fix: cap stdin read at 1 MiB to prevent DoS

Quratulain-bilal:
Unbounded sys.stdin.read() allowed a malicious caller to exhaust memory by sending a multi-gigabyte payload. Cap at 1 MiB and raise typer.Exit if truncated.


mnriem:
Thank you!

PR #3861, opened 2026-07-29
fix: preserve unreadable event config files

marcelsafin:
## Summary

- handle read failures while loading user-owned event configuration
- skip the merge instead of crashing or replacing existing content
- cover unreadable config paths with a regression test

## Testing

- `.venv/bin/python -m pytest -q tests/integrations/test_events.py` (104 passed)
- `.venv/bin/python -m pytest -q` (6009 passed, 173 skipped)

## AI disclosure

GitHub Copilot helped identify edge cases and review the implementation. I reproduced the failure and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3862, opened 2026-07-29
fix: normalize non-UTF-8 integration manifests

marcelsafin:
## Summary

- convert manifest decoding failures into the same public `ValueError` boundary used for malformed JSON
- report a clear invalid UTF-8 message while preserving exception chaining
- add regression coverage for non-UTF-8 manifest bytes

## Testing

- `.venv/bin/python -m pytest -q tests/integrations/test_manifest.py` (55 passed)
- `.venv/bin/python -m pytest -q` (6009 passed, 173 skipped)

## AI disclosure

GitHub Copilot helped identify the inconsistent error boundary and review the implementation. I reproduced the failure and validated the final change with targeted and full test suites.

### DIFF
--- a/src/specify_cli/integrations/manifest.py
+++ b/src/specify_cli/integrations/manifest.py
@@ -471,6 +471,10 @@ class IntegrationManifest:
         path = inst.manifest_path
         try:
             data = json.loads(path.read_text(encoding="utf-8"))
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3863, opened 2026-07-29
fix: escape workflow step metadata

marcelsafin:
## Summary

- render installed and catalog step metadata literally in list, search, and info output
- escape echoed step IDs, including the not-found path
- preserve intentional Rich formatting around untrusted values

## Testing

- `.venv/bin/python -m pytest -q tests/test_workflows.py` (822 passed)
- `.venv/bin/python -m pytest -q` (6012 passed, 173 skipped)

## AI disclosure

GitHub Copilot helped identify affected output paths and review the implementation. I reproduced the failures and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3868, opened 2026-07-29
Match task IDs beyond T999 in taskstoissues dedup regex

### LINKED ISSUE #3866: [Bug]: taskstoissues regex `\bT\d{3}\b` misses convergence task IDs at T1000 and beyond
## Bug Description

`templates/commands/taskstoissues.md` and `templates/commands/converge.md` disagree on the task-ID format, so `/speckit.taskstoissues` silently skips convergence tasks numbered `T1000` and above.

**Producer** — `converge.md:206` assigns IDs with:

```
zero-padded IDs `T{M+1:03d}, T{M+2:03d}, …`
```

`03d` is a *minimum* field width, not a cap, so once the task count passes 999 this correctly produces 4-digit IDs (`T999` → `T1000`).

**Consumer** — `taskstoissues.md:67` defines task IDs as "a `T` followed by three digits, e.g. `T001`" and matches issue titles with:

```
task ID pattern `\bT\d{3}\b`
```

`\d{3}` with a trailing `\b` matches *exactly* three digits. Given `T1000`, the `\b` after `\d{3}` fails (position between `0` and `0` is not a word boundary), so there is no match at all.

The consequence is a silent no-op rather than an error: affected tasks are neither deduplicated against existing issues nor converted into new ones.

## Steps to Reproduce

1. In a project whose `tasks.md` already contains more than 999 tasks, run `/speckit.converge`.
2. Convergence appends a `## Phase N: Convergence` section with IDs at `T1000`+ (per `converge.md` step 3).
3. Run `/speckit.taskstoissues`.
4. No issues are created for the `T1000`+ tasks, and re-running does not report them as already having issues.

A large `tasks.md` is awkward to stage by hand; the mismatch is also visible directly by testing the documented pattern against the documented ID format:

```python
import re
re.search(r"\bT\d{3}\b", "T1000: example task")  # -> None
re.search(r"\bT\d{3,}\b", "T1000: example task") # -> matches "T1000"
```

## Expected Behavior

Any task ID that `converge.md` can emit should be recognized by `taskstoissues.md` — including IDs longer than three digits — so dedup and issue creation cover every task.

## Actual Behavior

Tasks with IDs `>= T1000` are invisible to `taskstoissues`: not extracted, not deduplicated, not converted. No warning or error is emitted, so the run appears to have succeeded completely.

## Proposed Fix

Widen the quantifier to three-or-more digits in `taskstoissues.md`:

```diff
-task ID pattern `\bT\d{3}\b`
+task ID pattern `\bT\d{3,}\b`
```

and update the accompanying prose from "a `T` followed by three digits" to "at least three digits".

One note on the existing rationale, which reads:

> word boundaries so tokens like `ST001` or `T0010` are not matched by mistake

Both guards still hold under `\d{3,}`, but for slightly different reasons, so the comment is worth rewording:

- `ST001` — still unmatched. There is no `\b` between `S` and `T` (both word characters), independent of the quantifier.
- `T0010` — now captured as the whole token `T0010` rather than being rejected. The trailing `\b` forces the full digit run to be consumed, so `T001` can never match as a prefix of `T0010`. This is safe because matched IDs are intersected with the ID set built from `tasks.md`, so a token that is not a real task ID matches nothing.

It may also be worth stating the ID contract explicitly in `converge.md` (that `03d` is a floor and consumers must accept `\bT\d{3,}\b`), so the producer and consumer cannot drift apart again.

Happy to open a PR for this if useful.

## Specify CLI Version

0.14.3 (templates installed recorded `0.14.5.dev0`). Also confirmed present on `main` today and in `v0.14.4`: `templates/commands/taskstoissues.md:67` and `templates/commands/converge.md:206`.

## AI Agent

Claude Code

## Operating System

Windows 11 (10.0.22631)

## Python Version

Python 3.14.6

## Additional Context

Found during review of a fresh `specify init` rather than at runtime, so the T1000+ threshold was never actually hit in our project — the impact is latent for anyone whose `tasks.md` grows past 999 tasks. Severity is low for that reason, but the failure mode is silent, which makes it easy to miss when it does occur.

patberry1961:
Fixes #3866.

## Description

`templates/commands/converge.md` and `templates/commands/taskstoissues.md` disagree on the task-ID format, so `/speckit.taskstoissues` silently skips convergence tasks numbered `T1000` and above.

**Producer** — `converge.md` assigns IDs with `T{M+1:03d}`. The `03d` field width is a *minimum*, not a cap, so once the task count passes 999 it correctly emits 4-digit IDs (`T999` → `T1000`).

**Consumer** — `taskstoissues.md` matched issue titles with `` `\bT\d{3}\b` ``, which matches *exactly* three digits: given `T1000`, the trailing `\b` cannot match between two digits, so there is no match at all.

Affected tasks were therefore neither deduplicated against existing issues nor converted into new ones. The failure mode is a silent no-op rather than an error — the pattern simply finds nothing, so the run looks like it succeeded.

This widens the pattern to `` `\bT\d{3,}\b` `` and describes task IDs as "at least three digits". It also records the ID contract in `converge.md` (that `03d` is a floor, and consumers must accept `\bT\d{3,}\b`), so the producer and consumer can't quietly drift apart again.

### On the existing word-boundary rationale

The original comment justified the boundaries as stopping `ST001` and `T0010` from matching. Both guards still hold under `\d{3,}`, but for different reasons, so I reworded it rather than leaving a rationale that no longer describes the behaviour:

- `ST001` — still unmatched. There's no `\b` between `S` and `T` (both word characters), independent of the quantifier.
- `T0010` — now captured as the whole token `T0010` instead of being rejected. The trailing `\b` forces the entire digit run to be consumed, so `T001` can never match as a prefix of `T0010`. That's safe because matched IDs are intersected with the ID set built from `tasks.md`, so a token that isn't a real task ID matches nothing.

## Testing

Behaviour of the old vs. new pattern against `"<id>: example task"` titles:

| input | `\bT\d{3}\b` | `\bT\d{3,}\b` |
|---|---|---|
| `T001` | `T001` | `T001` |
| `T042` | `T042` | `T042` |
| `T999` | `T999` | `T999` |
| `T1000` | — | `T1000` |
| `T1001` | — | `T1001` |
| `T12345` | — | `T12345` |
| `ST001` | — | — |
| `XT1000` | — | — |
| `T01` | — | — |
| `T1` | — | — |

The first three rows confirm no regression for existing 3-digit IDs; rows 4–6 are the fix; the last four confirm the boundary guards still reject non-IDs. Prefix safety: against the title `T0010: task`, the old pattern yields `[]` and the new yields `['T0010']` — never a spurious `T001`.

Note on scope: this changes prompt text in two command templates, not executable code, so there's no unit test to add — the regex lives in instructions the agent follows. The table above is the evidence. I did not attempt an end-to-end run, since reproducing it genuinely needs a `tasks.md` with more than 999 tasks; the defect and the fix are both verifiable by inspection of the documented ID format against the documented pattern.

## AI Disclosure

This change was authored by Claude (Claude Code) working under my direction, including the diff, the commit message, and this PR description. The same applies to #3866 and to my responses on this PR unless I say otherwise.

Extent and human involvement: the bug was found by Claude while reviewing a fresh `specify init` in one of our repos, not at runtime — our `tasks.md` never approached 999 tasks, so the impact here was latent. Before filing, the analysis was verified against `main` rather than taken on trust: the line references were read from the current tree, and the old/new pattern behaviour was executed to produce the table above rather than reasoned about on paper. I reviewed the change and the reasoning before submitting.

Flagging the low real-world severity explicitly so you can weigh it appropriately — the argument for fixing it is the silent failure mode, not observed breakage.


mnriem (changes requested):

patberry1961:
Closing this — withdrawing the contribution from my side, as I am not able to commit to review follow-up or ongoing maintenance on it. Nothing to do with how it was handled; apologies for the churn.

The underlying finding is independently verifiable and I have no claim on it, so leaving the summary here for anyone who wants to pick it up:

`templates/commands/taskstoissues.md` matches issue titles with `\bT\d{3}\b` (exactly three digits, since the trailing `\b` cannot match between two digits), while `templates/commands/converge.md` emits IDs via `T{M+1:03d}` — a minimum field width that grows past three digits once the task count exceeds 999. Tasks at or beyond `T1000` are therefore silently skipped by tasks-to-issues: not extracted, not deduplicated, not converted, and no error is raised. The fix is one character — `\d{3}` becomes `\d{3,}` — and both existing word-boundary guards still hold (`ST001` has no boundary between `S` and `T`; the trailing `\b` makes `T0010` match as the whole token rather than as a `T001` prefix).

Reuse any of it freely, no attribution needed.

AI disclosure, per CONTRIBUTING: this comment — like the PR and issue — was written by Claude (Claude Code) under my direction.

PR #3870, opened 2026-07-30
fix: use missing_ok=True in extension ZIP cleanup

Quratulain-bilal:
Replace check-then-act pattern with unlink(missing_ok=True) to eliminate TOCTOU race condition in finally blocks.


mnriem (changes requested):
Please resolve conflicts and pull in upstream/main

mnriem (approved):

mnriem:
Thank you!

PR #3873, opened 2026-07-30
feat(presets): add opt-in constitution-sync preset

mnriem:
## Summary

Follow-up to #3790, which removed the consistency-propagation pass from the core `/constitution` command in favor of runtime resolution. This adds a bundled, **opt-in** `constitution-sync` preset that restores the pre-0.14.4 propagation behavior for teams that treat materialized `plan`/`spec`/`tasks` templates as reviewed, committed artifacts.

## Background

Through 0.14.3, `/constitution` ran a "consistency propagation checklist" that materialized amended constitutional guidance into the scaffolded templates and installed command files. #3790 (shipped in 0.14.4) removed that in favor of **runtime resolution** — `plan`, `tasks`, and `analyze` read `.specify/memory/constitution.md` live on every run, so templates only need to carry a pointer.

For the default workflow this is strictly better and **not a breaking change**: `/constitution` still works, and materialized templates are *preserved* on a non-forced upgrade (divergent hashes classify them as customizations). What teams lose is the automatic propagation of amended guidance into those materialized templates. This preset is the supported escape hatch for teams that depend on that.

## What's included

- **`presets/constitution-sync/`** — `preset.yml` (`requires.speckit_version: ">=0.14.4"`), a `wrap`-strategy override of `speckit.constitution` composing on `{CORE_TEMPLATE}` (forward-compatible with core changes), and a README.
- **`presets/catalog.json`** — bundled catalog entry.
- **`docs/upgrade.md`** — documents the 0.14.4 behavior change, why it isn't a functional break, the non-forced-upgrade behavior, the optional pointer-reset cleanup, and the opt-in.
- **`tests/test_presets.py`** — `TestConstitutionSyncPreset`: structural coverage (manifest, wrapper content, catalog integrity) plus a composition test asserting the wrap layer resolves atop the bundled core.

## Design notes

The wrapper only writes into the project's **own** `.specify/templates/` scaffolds and installed command files — never into stack-owned template layers. The README documents the underlying tension explicitly: auto-propagation *materializes/freezes* guidance into files, while the resolution stack *recomputes* templates from package layers on demand, so propagating into a stack-owned template would be clobbered on the next reconciliation. The preset is for project-local governed scaffolds; teams whose templates come from other presets/extensions should stay on runtime resolution.

## Testing

`tests/test_presets.py` — 507 passed (including 4 new `TestConstitutionSyncPreset` cases).

---
🤖 This PR was authored by GitHub Copilot (model: Claude Opus 4.8) on behalf of @mnriem, under human supervision. Commits carry an `Assisted-by:` trailer.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -208,6 +208,66 @@ Restart your IDE to refresh the command list.
 
 ---
 
+## Behavior change: `/constitution` no longer propagates into templates (0.14.4)
+
+Starting in **0.14.4** ([#3790](https://github.com/github/spec-kit/pull/3790)), the
+`/constitution` command is scoped to its own artifact. It updates
+`.specify/memory/constitution.md` and writes a Sync Impact Report, and **no longer edits**
+`plan-template.md`, `spec-template.md`, `tasks-template.md`, installed command files, or
+guidance docs.
+
+### Why
+
+Spec Kit uses **runtime resolution**: `plan`, `tasks`, and `analyze` read
+`.specify/memory/constitution.md` live on every run, and `analyze` is the dedicated drift
+checker. The governed templates carry a pointer, not a copy — `plan-template.md` ships
+`[Gates determined based on constitution file]`, and `/plan` fills that section from the live
+constitution each run. Propagation duplicated the single source of truth and fought the
+preset/override composition system (a `replace` preset shadows an edited core template).
+
+### Is this a breaking change for existing projects?
+
+**No — your workflow keeps working.** The templates are scaffolds, not authorities. When you
+run `/plan`, it copies the template into a per-feature `plan.md` and re-derives the Constitution
+Check from the live constitution; `/analyze` validates against it. Even if a previous
+`/constitution` run materialized concrete gate text into `.specify/templates/plan-template.md`,
+the live constitution remains the source of truth at runtime.
+
+On a **non-forced upgrade**, a materialized template is *preserved* (its hash diverges from the
+recorded managed copy, so the refresh treats it as a customization and does not overwrite it).
+Nothing regresses.
+
+### Optional cleanup — return to the runtime pointer
+
+A frozen, pre-filled Constitution Check is a slightly misleading scaffold and can bias the first
+`/plan` pass. To move fully back to runtime resolution, reset the section body in
+`.specify/templates/plan-template.md` to the pointer:
+
+```text
+## Constitution Check
+
+*GATE: Must pass before Phase 0 research. Re-check after Phase 1 design.*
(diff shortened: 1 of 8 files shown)


mnriem:
Addressed the three review comments in 19be9bb09b46f4c956840b12714102ad294c9617:

1. **Wheel packaging** — added `presets/constitution-sync` to the `[tool.hatch.build.targets.wheel]` force-include so `_locate_bundled_preset` resolves it in a released wheel; without this the `bundled: true` advertisement was unshippable. Also added `tests/contract/test_wheel_bundled_presets.py`, which asserts every `bundled` preset in `presets/catalog.json` is force-included (covers `lean` too, so this class of gap can't recur).

2. **Scope Guard conflict** — the wrapper now opens its propagation section by explicitly stating it **supersedes** the core Scope Guard for template/command propagation, while keeping the guard's other constraints (no feature implementation, no unrelated files) in force. Resolves the direct contradiction in the composed command.

3. **Composition test** — added `test_resolved_content_embeds_core_and_sync_pass`, which calls `resolve_content` and asserts `{CORE_TEMPLATE}` is substituted and the effective command contains both the core body (`## Scope Guard`) and the sync pass (`## Constitution Template Sync`, `plan-template.md`).

Full run: 623 passed (`tests/test_presets.py` + `tests/contract/`).

---
Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8). This review round was addressed autonomously under human supervision.

mnriem:
Second review round addressed in ec066d51:

1. **Frontmatter assertions** — `test_wrapper_uses_core_template_and_propagates` now parses the Markdown frontmatter as YAML and asserts `strategy == "wrap"` structurally, and checks `{CORE_TEMPLATE}` in the body section only (with a guard that `strategy: wrap` does not appear in the body). No more substring false-positive risk.

2. **`tomllib` on Python 3.10** — not applicable: `requires-python = ">=3.11"`, and the repo already imports `tomllib` bare in the sibling contract tests (`test_wheel_core_pack_scripts.py`, `test_catalog_schema.py`) and in `src/specify_cli/_assets.py`. Adding a `tomli` fallback would diverge from the established pattern, so I left it consistent.

3. **`author` casing** — kept as `"github"` to match the existing bundled `lean` preset (`presets/lean/preset.yml`) and the catalog, which both use lowercase `github`. Changing only this preset would break the consistency the comment asks for.

Ran the **full** suite: 6185 passed, 6 skipped.

---
Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8). This review round was addressed autonomously under human supervision.

PR #3874, opened 2026-07-30
feat: support tar archives for installs

mnriem:
## Summary

- add secure `.tar.gz` / `.tgz` extraction with ZIP-equivalent path, link, conflict, entry-count, and size protections
- support ZIP, `.tar.gz`, and `.tgz` across extension and preset direct URL, catalog, install, update, cleanup, and SHA-256 flows
- preserve complete workflow packages from local directories and local/remote/catalog archives with transactional directory replacement
- document workflow archive/package installation and add archive parity coverage

## Validation

- `.venv/bin/python -m pytest tests/test_download_security.py -q`
- `.venv/bin/python -m pytest tests/test_extensions.py tests/test_presets.py tests/test_workflows.py -q`

Authored autonomously on behalf of @mnriem by GitHub Copilot (model: GPT-5.6 Sol).

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -103,10 +103,17 @@ specify workflow add <source>
 
 | Option          | Description                                            |
 | --------------- | ------------------------------------------------------ |
-| `--dev`         | Install from a local workflow YAML file or directory   |
+| `--dev`         | Install from a local YAML file, package directory, or archive |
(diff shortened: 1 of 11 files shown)


mnriem:
Addressed the review feedback in `15255753`: restored the preset `force` compatibility argument, added content-based detection for suffixless workflow archives while retaining YAML size limits, and made package rollback restore the prior workflow before best-effort cleanup. Added regression coverage for each path; the full local suite passes.

PR #3879, opened 2026-07-31
fix: escape Rich markup in `workflow resolve` output

Noor-ul-ain001:
## Problem

`specify workflow resolve <id>` prints its two summary sections through `console.print`, which has Rich markup enabled. Neither line escaped its interpolated fields.

**1. The tier label never rendered — on every invocation.**

```python
f"  • [{layer.tier}] {layer.source} (priority={priority})"
```

Rich parses `[base]` and `[project-overlay]` as style tags, so the bracketed tier is swallowed. This needs no untrusted input; the column has simply never worked:

```
  •  project:ov1 (priority=10)     <- tier label missing
  •  base (priority=n/a)           <- tier label missing
```

**2. Step IDs reach Rich as markup.**

Step IDs come from base-workflow / overlay YAML and are validated only against `:` (`_parse_edit` in `overlays/schema.py`) — brackets pass validation. Two failure modes:

- Balanced `[stuff]` is silently swallowed, so `new[stuff]` prints as `new` and the step appears to vanish.
- Unbalanced `[/red]` raises `rich.errors.MarkupError` → uncaught traceback, **exit 1**. The workflow can't be inspected at all:

```
$ specify workflow resolve wf
...
Step attribution:
  • a: base
MarkupError: closing tag '[/red]' at position 7 doesn't match any open tag
```

## Fix

Route the interpolated fields through `rich.markup.escape` and escape the literal tier bracket as `\[`. This matches the pattern already used by `workflow info`'s step graph and `workflow_list`'s `\[disabled]`.

Display-only: the returned payload was already unescaped and is unchanged, so machine-readable callers see no difference.

## Tests

Three regression tests in `TestOverlayCli`, all verified to fail without the fix:

- `test_workflow_resolve_prints_tier_labels` — asserts `[base]` and `[project-overlay]` appear in output
- `test_workflow_resolve_escapes_rich_markup_in_step_id` — parametrized over `new[stuff]` (swallowed) and `new[/red]` (crashes)

`tests/workflows/test_overlay_commands.py` passes 23/23. The wider `tests/workflows/` suite shows 10 pre-existing failures, all symlink-permission tests that fail identically on a clean tree on Windows without elevation.

### DIFF
--- a/src/specify_cli/workflows/overlays/_commands.py
+++ b/src/specify_cli/workflows/overlays/_commands.py
@@ -7,6 +7,7 @@ from typing import Any
 
 import typer
 import yaml
+from rich.markup import escape as _escape_markup
 
 from ..._console import console, err_console
 from ...extensions import normalize_priority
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3887, opened 2026-07-31
fix(workflows): fail a fan-in step whose `output` is not a mapping

jawwad-ali:
## Problem

`FanInStep.execute` began with:

```python
output_config = config.get("output") or {}
if not isinstance(output_config, dict):
    output_config = {}
```

Any non-mapping `output` was silently replaced with `{}` and the step still returned **COMPLETED**. Every aggregation key the author declared vanished, and downstream `{{ steps.<id>.output.<key> }}` resolved to `None` and interpolated as an empty string.

Note also that `x or {}` masked the **falsy** shapes (`[]`, `false`, `0`, `''`) before the `isinstance` check even ran.

```
output=[]       -> status=completed  error=None
output=False    -> status=completed  error=None
output=0        -> status=completed  error=None
output=''       -> status=completed  error=None
output=['a']    -> status=completed  error=None
output='oops'   -> status=completed  error=None
output=5        -> status=completed  error=None
```

## This is already a known flaw — in this file

`validate` rejects a non-mapping `output`, and its comment names the execute-side bug outright:

```python
output = config.get("output")
if output is not None and not isinstance(output, dict):
    # execute() silently coerces a non-mapping output to {}, so the
    # author's declared aggregation keys would vanish with no error.
    # Reject at validation, mirroring the command-step (#3262) fix.
```

The engine does not auto-validate before `execute` (see `WorkflowEngine.load_workflow`), so on an unvalidated run that comment describes exactly what happens. This PR closes the execute-side half.

## Fix

Fail the step with **validate()'s own message**, mirroring the `wait_for` guard a few lines below in the same method:

```python
output_config = config.get("output")
if output_config is None:
    output_config = {}
elif not isinstance(output_config, dict):
    return StepResult(status=StepStatus.FAILED, error=..., output={"results": []})
```

This is the same "silent empty result + COMPLETED" class the `wait_for` guards in this method already reject, and the same shape as the fan-out non-list `items` guard.

**No breaking change.** An explicit `output:` (YAML null) stays valid, matching `validate` — pinned by a second new test. A mapping behaves exactly as before. The only inputs whose behaviour changes are ones that silently produced an empty aggregation.

## Verification

- **Fail-before / pass-after:** the seven parametrized cases fail on unpatched `src` and pass with the fix.
- `tests/test_workflows.py` scoped regression: failure set identical to the clean-`main` baseline captured on `81bf741` (20 pre-existing in this file, all Windows symlink-privilege).

Tests go into the existing `TestFanInStep`, beside `test_execute_non_list_wait_for_fails_loudly` — the sibling guard whose docstring describes the same wiring bug.

---

### DIFF
--- a/src/specify_cli/workflows/steps/fan_in/__init__.py
+++ b/src/specify_cli/workflows/steps/fan_in/__init__.py
@@ -20,9 +20,31 @@ class FanInStep(StepBase):
 
     def execute(self, config: dict[str, Any], context: StepContext) -> StepResult:
         wait_for = config.get("wait_for", [])
-        output_config = config.get("output") or {}
-        if not isinstance(output_config, dict):
+        output_config = config.get("output")
+        if output_config is None:
             output_config = {}
+        elif not isinstance(output_config, dict):
+            # ``validate`` rejects a non-mapping ``output`` and its comment says
+            # why: "execute() silently coerces a non-mapping output to {}, so the
+            # author's declared aggregation keys would vanish with no error."
+            # The engine does not auto-validate before ``execute``, so on an
+            # unvalidated run that is exactly what happened -- and ``x or {}``
+            # masked the falsy shapes ([], false, 0, '') before the isinstance
+            # check even ran. Every declared key vanished while the step still
+            # reported COMPLETED, so downstream ``steps.<id>.output.<key>``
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3892, opened 2026-07-31
fix(scripts): stop setup-tasks text mode crashing on a legacy stdout code page

jawwad-ali:
## Problem

`_check_file` / `_check_dir` in `scripts/python/setup_tasks.py` hard-code the U+2713 / U+2717 glyphs and `print()` them to `sys.stdout`:

```python
marker = "✓" if path.is_file() else "✗"
print(f"  {marker} {description}")
```

On Windows `sys.stdout` falls back to the ANSI code page whenever stdout is **not a console** — a pipe or a file redirect, which is exactly how an agent or a workflow step invokes these scripts. U+2713 is unencodable in cp1252, so the document listing aborts mid-report with `UnicodeEncodeError`.

## Relationship to #3890

This is the **byte-identical twin** of the block in `scripts/python/check_prerequisites.py`. I flagged it explicitly in that PR's body rather than widening that PR's scope:

> `scripts/python/setup_tasks.py:58-65` carries a byte-identical block and crashes the same way. I kept this PR to one script — happy to extend it to the twin here or as a follow-up, whichever you prefer.

This is that follow-up. The two PRs touch different source files and different test files, so they are independent and can merge in either order.

## Fix

Downgrade to ASCII only when stdout cannot encode the glyph, so a UTF-8 console is unaffected:

```python
glyph = "✓" if ok else "✗"
try:
    glyph.encode(getattr(sys.stdout, "encoding", None) or "utf-8")
except (LookupError, UnicodeEncodeError):
    return "[OK]" if ok else "[FAIL]"
return glyph
```

`[OK]` / `[FAIL]` is the ASCII rendering these markers **already have in this repo**:

| Site | Code |
|---|---|
| `scripts/powershell/common.ps1:243` | `Write-Output "  [OK] $Description"` |
| `scripts/powershell/common.ps1:246` | `Write-Output "  [FAIL] $Description"` |
| `tests/parity_helpers.py:130-131` | `text.replace("  ✓ ", "  [OK] ").replace("  ✗ ", "  [FAIL] ")` |

The PowerShell twin emits the ASCII form natively and the shared parity helper maps the glyphs onto it, so the suite already treats the two as equivalent output.

**Breaking risk:** a UTF-8-capable stdout still gets the glyphs, byte-identical to today. The only case that changes is one that previously raised and truncated the report.

## Verification

- **Fail-before / pass-after:** the new test (running the real script with `PYTHONIOENCODING=cp1252`) fails on unpatched source and passes with the fix. File: **1 failed → 3 passed, 14 skipped** (the skips are the bash-requiring parity cases on Windows).
- Scoped regression: failure set identical to the clean-`main` baseline captured on `81bf741` (0 pre-existing in scope).

---

### DIFF
--- a/scripts/python/setup_tasks.py
+++ b/scripts/python/setup_tasks.py
@@ -55,14 +55,31 @@ def _available_docs(paths: FeaturePaths) -> list[str]:
     return docs
 
 
+def _status_marker(ok: bool) -> str:
+    """Return the status glyph, downgraded to ASCII when stdout cannot encode it.
+
+    On Windows sys.stdout falls back to the ANSI code page whenever it is not a
+    console - a pipe or a file redirect, which is how agents and workflow steps
+    invoke these scripts - and U+2713 is unencodable in cp1252, so printing it
+    raised UnicodeEncodeError and aborted the report mid-listing.
+    "[OK]"/"[FAIL]" is the ASCII rendering these markers already have in-tree:
+    see Test-FileExists in scripts/powershell/common.ps1 and
+    normalize_status_text in tests/parity_helpers.py.
+    """
+    glyph = "✓" if ok else "✗"
+    try:
+        glyph.encode(getattr(sys.stdout, "encoding", None) or "utf-8")
+    except (LookupError, UnicodeEncodeError):
+        return "[OK]" if ok else "[FAIL]"
+    return glyph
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3895, opened 2026-07-31
fix(kimi): preserve non-UTF-8 user skills

marcelsafin:
## Summary

- treat a non-UTF-8 `SKILL.md` as user-owned during Kimi legacy cleanup
- preserve unreadable user skill content instead of aborting teardown
- add regression coverage for invalid UTF-8 skill bytes

## Testing

- `.venv/bin/pytest tests/integrations/test_integration_kimi.py -q` (49 passed)

## AI disclosure

GitHub Copilot helped identify the cleanup edge case and review the implementation. I reproduced the failure and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3896, opened 2026-07-31
fix(presets): tolerate non-UTF-8 legacy commands

marcelsafin:
## Summary

- treat legacy command frontmatter decoding as best-effort
- keep the default `replace` strategy when command content is not UTF-8
- add regression coverage for an invalidly encoded legacy command

## Testing

- `.venv/bin/pytest tests/test_presets.py -q` (509 passed)

## AI disclosure

GitHub Copilot helped identify the incomplete best-effort boundary and review the implementation. I reproduced the failure and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3897, opened 2026-07-31
fix(events): ignore non-UTF-8 event overrides

marcelsafin:
## Summary

- treat non-UTF-8 integration event overrides like other unreadable override files
- retain events already resolved from earlier layers instead of aborting resolution
- add regression coverage for an invalidly encoded override

## Testing

- `.venv/bin/pytest tests/integrations/test_events.py -q` (105 passed)

## AI disclosure

GitHub Copilot helped identify the missing decode-error boundary and review the implementation. I reproduced the failure and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3898, opened 2026-07-31
fix(presets): validate required manifest mappings

marcelsafin:
## Summary

- require the `preset`, `requires`, and `provides` manifest sections to be mappings
- raise `PresetValidationError` instead of leaking raw type errors from nested access
- cover invalid section shapes with regression tests

## Testing

- `.venv/bin/pytest tests/test_presets.py -q` (517 passed)
- `.venv/bin/pytest -q` (6114 passed, 176 skipped)

## AI disclosure

GitHub Copilot helped identify the missing shape guards and review the implementation. I reproduced the failures and validated the final change with targeted and full test suites.

### DIFF
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -296,6 +296,12 @@ class PresetManifest:
                 f"(expected {self.SCHEMA_VERSION})"
             )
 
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3899, opened 2026-07-31
fix(workflows): reject mismatched run state IDs

marcelsafin:
## Summary

- reject persisted run state whose `run_id` differs from the requested run directory
- prevent a corrupt state file from redirecting later saves into another run
- add regression coverage for mismatched stored identities

## Testing

- `.venv/bin/pytest tests/test_workflows.py -q` (873 passed)

## AI disclosure

GitHub Copilot helped identify the inconsistent identity boundary and review the implementation. I reproduced the failure and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3900, opened 2026-07-31
fix(events): skip non-UTF-8 extension manifests

marcelsafin:
## Summary

- skip non-UTF-8 fallback `extension.yml` files during event collection
- preserve the documented fail-soft behavior for malformed on-disk extensions
- add regression coverage for invalid manifest bytes

## Testing

- `.venv/bin/pytest tests/integrations/test_events.py -q` (105 passed)

## AI disclosure

GitHub Copilot helped identify the missing fallback boundary and review the implementation. I reproduced the failure and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3901, opened 2026-07-31
fix(workflows): refetch non-UTF-8 catalog caches

marcelsafin:
## Summary

- treat non-UTF-8 workflow and step cache payloads as invalid cache entries
- refetch both catalog types instead of leaking `UnicodeDecodeError`
- verify successful refetch replaces the invalid cache with valid JSON

## Testing

- `.venv/bin/pytest tests/test_workflows.py -q` (874 passed)
- `.venv/bin/pytest -q` (6107 passed, 176 skipped)

## AI disclosure

GitHub Copilot helped identify the missing cache decode boundary and review the implementation. I reproduced the failures and validated the final change with targeted and full test suites.

### DIFF
--- a/src/specify_cli/workflows/catalog.py
+++ b/src/specify_cli/workflows/catalog.py
@@ -514,7 +514,7 @@ class WorkflowCatalog:
                     cached = json.load(f)
                 if isinstance(cached, dict):
                     return cached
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3902, opened 2026-07-31
fix(bundler): wrap local catalog decode failures

marcelsafin:
## Summary

- route plain-path and `file://` bundle catalogs through the shared JSON loader
- convert non-UTF-8 local catalog failures into the `BundlerError` boundary
- cover both local source forms with regression tests

## Testing

- `.venv/bin/pytest tests/unit/test_bundler_adapters.py -q` (18 passed)
- `.venv/bin/pytest -q` (6107 passed, 176 skipped)

## AI disclosure

GitHub Copilot helped identify the bypassed shared loader and review the implementation. I reproduced the failures and validated the final change with targeted and full test suites.


mnriem:
Thank you!

PR #3903, opened 2026-07-31
fix: bound sys.stdin.read() to prevent OOM on oversized payloads

Quratulain-bilal:
## Problem

`sys.stdin.read()` loads all of stdin into memory with no size limit. A malfunctioning agent hook or malicious process can pipe gigabytes of data into stdin, exhausting process memory before the payload is ever consumed.

This affects both the CLI event runner (`commands/event.py`) and the standalone events entry point (`events.py`).

## Fix

Cap stdin reads at 10 MiB using a bounded chunked read loop. The function reads in 64 KiB chunks and stops once the limit is reached, preventing unbounded memory allocation.

## Testing

- Verified that normal payloads (well under 10 MiB) pass through unchanged
- Verified that the bounded reader stops at the limit instead of reading all input

### DIFF
--- a/src/specify_cli/commands/event.py
+++ b/src/specify_cli/commands/event.py
@@ -6,6 +6,29 @@ from pathlib import Path
 import sys
 import typer
 
+_MAX_STDIN_BYTES = 10 * 1024 * 1024  # 10 MiB
+
+
+def _read_stdin_bounded(max_bytes: int = _MAX_STDIN_BYTES) -> str:
+    """Read at most *max_bytes* from stdin to prevent unbounded memory use.
+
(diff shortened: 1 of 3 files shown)


mnriem:
Thanks — and this one has real precedent: capping stdin was accepted in #3857, and #4337 caps it in the generated dispatcher, so the threat model isn't purely theoretical here. Two things before it can move: (1) **disclose any AI assistance** per CONTRIBUTING; (2) it **conflicts with `main`** — please rebase. Worth checking it doesn't overlap #4337's dispatcher cap (align the limits — #3857 used 1 MiB, this uses 10 MiB; they should be consistent). With disclosure + rebase it's reviewable. Marking `author-awaiting`.

PR #3904, opened 2026-07-31
fix: skip corrupted state.json in list_runs() instead of aborting

Quratulain-bilal:
## Problem

If any single `state.json` file is corrupted, truncated, or unreadable, the unhandled exception aborts the entire iteration and all subsequent valid runs are never listed.

## Fix

Catch `OSError`, `JSONDecodeError`, and `UnicodeDecodeError` to skip bad entries gracefully.

## Testing

- Verified that valid runs are still listed correctly
- Verified that a corrupted state.json is skipped without crashing


mnriem (approved):

mnriem:
Thank you!

PR #3905, opened 2026-07-31
fix: use incremental hashing in Preset.get_hash()

Quratulain-bilal:
## Problem

`Preset.get_hash()` called `f.read()` which loads the entire file into memory before hashing. A maliciously sized preset file could cause unbounded memory allocation.

## Fix

Use incremental `hashlib.sha256().update()` with 64 KiB chunks.

## Testing

- Verified hash output matches the previous implementation for the same file


mnriem (changes requested):
Please resolve conflicts

PR #3907, opened 2026-07-31
fix: use incremental hashing in CatalogDescriptor.get_hash()

Quratulain-bilal:
## Problem

`CatalogDescriptor.get_hash()` called `fh.read()` which loads the entire file into memory before hashing. A large or maliciously sized catalog descriptor file could cause unbounded memory allocation.

## Fix

Use incremental `hashlib.sha256().update()` with 64 KiB chunks.

## Testing

- Verified hash output matches the previous implementation for the same file

### DIFF
--- a/src/specify_cli/integrations/catalog.py
+++ b/src/specify_cli/integrations/catalog.py
@@ -841,5 +841,8 @@ class IntegrationDescriptor:
 
     def get_hash(self) -> str:
         """SHA-256 hash of the descriptor file."""
+        h = hashlib.sha256()
         with open(self.path, "rb") as fh:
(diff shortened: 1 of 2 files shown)


mnriem:
Thanks, and the test is thorough. But before this can move, I need the underlying problem to be real: a `CatalogDescriptor` is a small local `integration.yml` that the user installed themselves — not untrusted or network-controlled input — so "a maliciously sized descriptor exhausts memory" isn't a threat model that applies here (you'd have to put a multi-GB file on your own disk). Could you show a realistic scenario where this file is attacker-controlled or genuinely unbounded? If it's just defensive style, that's fine to say, but then it's very low priority. Also please add the AI-disclosure per CONTRIBUTING. Marking `author-awaiting`.

PR #3908, opened 2026-07-31
fix: eliminate TOCTOU race in RunState.load()

Quratulain-bilal:
## Problem

`RunState.load()` checks `exists()` then calls `open()`. Between the two calls the file can be deleted, causing a raw `FileNotFoundError` instead of the intended custom message.

## Fix

Remove the `exists()` pre-check and wrap `open()` in `try/except FileNotFoundError`.

## Testing

- Verified custom error message is raised when state.json is missing

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -733,11 +733,10 @@ class RunState:
         the lookup path. Without this guard, a caller passing a value like
         ``../escape`` (e.g. via ``specify workflow resume`` CLI argument)
         would interpolate path-traversal segments into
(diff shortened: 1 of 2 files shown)


mnriem:
Thanks — and unlike much of the batch, this one ships a regression test, which is appreciated. Two things before it can move: (1) please **disclose any AI assistance** per CONTRIBUTING; (2) it currently **conflicts with `main`** — please rebase. On severity: the TOCTOU here only changes which error message surfaces (raw `FileNotFoundError` vs the custom one) in an extreme delete-mid-open race, so it's low-priority and sits behind proven work — but with disclosure + rebase it can be reviewed. Marking `author-awaiting`.

PR #3909, opened 2026-07-31
fix: eliminate TOCTOU race in yamlio.load_yaml()

Quratulain-bilal:
## Problem

`yamlio.load_yaml()` checks `exists()` then calls `read_text()`. The file can be deleted between the two calls, causing a raw `FileNotFoundError` instead of the clear `BundlerError`.

## Fix

Remove the `exists()` pre-check and catch `FileNotFoundError` from `read_text()`.

## Testing

- Verified `BundlerError` is raised when file is missing

### DIFF
--- a/src/specify_cli/bundler/lib/yamlio.py
+++ b/src/specify_cli/bundler/lib/yamlio.py
@@ -54,10 +54,10 @@ def load_yaml(path: Path) -> Any:
     caller to reject.
     """
     path = Path(path)
-    if not path.exists():
-        raise BundlerError(f"File not found: {path}")
     try:
         text = path.read_text(encoding="utf-8")
+    except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


mnriem:
Thanks — has a test, appreciated. Please add the **AI-disclosure** per CONTRIBUTING. Bigger-picture: this is the same TOCTOU fix as #3908 (there it's `RunState.load()`, here `yamlio.load_yaml()`) — the identical `exists()`-then-read pattern at two call sites. Rather than one PR per call site, could you **consolidate the TOCTOU/exists-precheck fixes into a single PR** covering all the affected spots? It's far faster to review as one change, and with the disclosure added it'd move quicker. Same low-severity note as #3908 (this only changes which error surfaces in an extreme race), so it sits behind proven work — but consolidated + disclosed it's reviewable. Marking `author-awaiting`.

PR #3910, opened 2026-07-31
fix: eliminate TOCTOU races in catalog_fetch()

Quratulain-bilal:
## Problem

`catalog_fetch()` checks `exists()` then calls `read_text()` for both `file://` and bare path URLs. The file can be deleted between the two calls.

## Fix

Remove the `exists()` pre-checks and catch `FileNotFoundError` from `read_text()`.

## Testing

- Verified `BundlerError` is raised when catalog file is missing

### DIFF
--- a/src/specify_cli/bundler/services/adapters.py
+++ b/src/specify_cli/bundler/services/adapters.py
@@ -143,15 +143,17 @@ def make_catalog_fetcher(*, allow_network: bool = True):
 
         if scheme == "file":
             path = _file_url_to_path(parsed)
-            if not path.exists():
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

mnriem:

Quratulain-bilal:
Hi @mnriem — I've addressed Copilot's feedback: removed the unused exists() mock setup and added mock_path.exists.assert_not_called() as a regression guard to verify the pre-check was genuinely eliminated. All 20 tests pass. Ready for re-review when you get a chance.

mnriem (approved):

mnriem:
Thank you!

PR #3911, opened 2026-07-31
fix: enforce size limit on individual files in build_bundle()

Quratulain-bilal:
## Problem

`build_bundle()` called `read_bytes()` without any size guard. A single large asset file could exhaust memory.

## Fix

Enforce `MAX_ZIP_MEMBER_BYTES` (10 MiB) limit before reading each file.

## Testing

- Verified normal files are packaged correctly
- Verified oversized files are rejected

### DIFF
--- a/src/specify_cli/bundler/services/packager.py
+++ b/src/specify_cli/bundler/services/packager.py
@@ -98,11 +98,20 @@ def build_bundle(
                 st = os.fstat(fh.fileno())
                 mode = 0o755 if st.st_mode & 0o111 else 0o644
                 info.external_attr = mode << 16
+                # Fast metadata rejection: skip files whose size exceeds the
+                # limit before touching the read path.  Then also bound the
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
Please resolve conflicts

mnriem:

Quratulain-bilal:
Hi @mnriem — I've addressed Copilot's feedback: builds now write to a temporary sibling file first, and the final artifact path is only replaced atomically after every member passes validation. This prevents a partial/corrupt archive from being left at the output path if a file exceeds the size limit. The temp file is cleaned up on any failure. Added 	est_oversized_file_does_not_corrupt_output and 	est_temp_file_cleaned_up_on_failure to verify the atomic behavior. All 19 tests pass. Ready for re-review when you get a chance.

mnriem:
Please resolve conflicts

mnriem:

mnriem:

PR #3912, opened 2026-07-31
fix(workflows): reject a retry gate whose verdict enum forbids the reset value

Noor-ul-ain001:
## Problem

A gate with `on_reject: retry` consumes a bound reject verdict before pausing by resetting the named input to `""` — documented behaviour, so a later resume prompts again. Every `resume()` re-resolves the persisted inputs through `_coerce_input`.

Those two rules collide when the bound input declares an `enum` that does not list `""`. The reset writes a value the input's own enum forbids, and the run wedges:

```yaml
inputs:
  spec_verdict:
    type: string
    enum: [approve, reject]
steps:
  - id: review
    type: gate
    options: [approve, reject]
    on_reject: retry
    verdict_input: spec_verdict
```

```console
$ specify workflow run wf --input spec_verdict=reject
Status: paused
$ specify workflow resume <run_id> --input note=b
Error: Input 'spec_verdict' value '' not in allowed values: ['approve', 'reject'].
```

The workflow validates clean and the first run looks fine, so the failure only appears at the second resume. It is also unrecoverable in practice: `_resolve_inputs` re-coerces the whole persisted map, so *any* resume that supplies an input dies on the stored `""`. Only a resume with no inputs at all still works — and that is precisely the call that cannot deliver a new verdict, which is the one thing the retry cycle exists to allow.

## Fix

Extend the existing `verdict_input` cross-check (which already confirms the name is declared) to also require that a retry-bound input's `enum` admits the reset sentinel, and report it with a fix hint. To do that, thread the input *definitions* through `_validate_steps` instead of just their names.

Rejected the alternative of popping the key instead of writing `""`: that lets the input's `default` flow back in on the next resume, so a gate the user just rejected would silently auto-approve.

Docs: note the `enum` requirement next to the reset behaviour it follows from.

## Tests

4 validation tests for the new guard (rejects a retry gate whose enum omits `""`; passes when the enum includes it, when `on_reject` is `abort`/`skip`, and when there is no enum at all), plus a characterization test that drives the engine directly to pin the wedge the guard prevents.

Verified the guard test fails without the engine change. Full `tests/test_workflows.py`: 870 passed, 20 failed — the same 20 symlink tests fail identically on an unmodified `upstream/main` checkout (they need symlink privileges this Windows box lacks), so no regressions.

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -623,6 +623,19 @@ pauses: the named stored input is reset to `""`. A later resume therefore
 prompts or pauses again until another verdict is supplied. Approve, abort, and
 skip outcomes leave the input unchanged.
 
+Because of that reset, a verdict input used with `on_reject: retry` must accept
+`""`. If it declares an `enum`, include the empty string — otherwise the reset
+value violates the input's own `enum` and the run can no longer be resumed with
+any input. `specify workflow add` reports this as a validation error.
+
+```yaml
+inputs:
+  spec_verdict:
+    type: string
+    enum: ["", approve, reject]
+    default: ""
+```
+
 ## FAQ
 
 ### What happens when a workflow hits a gate step?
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #3914, opened 2026-07-31
Add `--extension` flag to `specify init` for opting into extensions at init time

mnriem:
No way to install extensions during `specify init` — the git extension is hardcoded to auto-install, and everything else requires a separate `specify extension add` post-init. This adds a repeatable `--extension` flag so users can opt into extensions at init time (prerequisite for git no longer being enabled by default).

### `_install_extension_during_init` helper
New function in `src/specify_cli/commands/init.py` that auto-detects source type and delegates accordingly:
- **Bundled name** (`git`, `selftest`) — checks bundled package first, falls back to catalog; skips if already installed
- **Local path** — triggers on `./`, `../`, `/`, `~/`, `.\`, `..\` prefixes or any `Path.is_absolute()` match (Windows-safe)
- **HTTPS URL** — downloads ZIP, enforces HTTPS (localhost HTTP allowed), cleans up after install
- Raises `ValueError` on failure; caller converts to a tracker error without aborting init

### `specify init` updates
- Added `--extension` as a repeatable `list[str] | None` typer option
- Tracker steps pre-registered as `extension-{i}` before the Live context; installed after preset, before finalize
- Extension failures are non-fatal (recorded in tracker, init continues)

### Usage

```bash
# Bundled extension by name
specify init my-project --integration copilot --extension git

# Multiple extensions
specify init my-project --extension git --extension selftest

# Local path
specify init my-project --extension ./my-extensions/custom-ext

# URL
specify init my-project --extension https://example.com/extensions/my-ext.zip
```

## Tests
Five new tests in `TestExtensionFlag`: bundled name, multiple extensions, local absolute path, unknown extension (graceful error, no abort), and combined with `--preset`. Full `test_cli.py` suite (102 tests) passes.

## Notes
This supersedes stale PR #2396, rebased onto current `main` and adapted to the refactored `init` command (moved from `__init__.py` to `src/specify_cli/commands/init.py`).

---
This PR was prepared by GitHub Copilot (model: Claude Opus 4.8) on behalf of @mnriem.

### DIFF
--- a/src/specify_cli/commands/bundle/__init__.py
+++ b/src/specify_cli/commands/bundle/__init__.py
@@ -120,6 +120,7 @@ def _run_init(integration: str, *, script_type: str, offline: bool = False) -> N
             integration=integration,
             integration_options=None,
             extensions=None,
+            trust_extension_urls=False,
         )
     except typer.Exit as exc:
         if exc.exit_code:
--- a/src/specify_cli/commands/bundle/__init__.py
+++ b/src/specify_cli/commands/bundle/__init__.py
@@ -119,6 +119,7 @@ def _run_init(integration: str, *, script_type: str, offline: bool = False) -> N
             preset=None,
             integration=integration,
             integration_options=None,
+            extensions=None,
         )
     except typer.Exit as exc:
         if exc.exit_code:
--- a/src/specify_cli/commands/init.py
+++ b/src/specify_cli/commands/init.py
@@ -30,6 +30,58 @@ def _stdin_is_interactive() -> bool:
     return sys.stdin.isatty()
 
 
+def _ext_spec_is_url(ext_spec: str) -> bool:
+    """Return True when *ext_spec* is an http(s) URL rather than a name/path."""
+    from urllib.parse import urlparse
+
+    try:
+        return urlparse(ext_spec).scheme in ("http", "https")
+    except ValueError:
+        return False
+
+
+def _confirm_extension_url_trust(
+    url_specs: list[str], *, trust_override: bool
+) -> dict[str, bool]:
+    """Resolve trust for each URL-based extension before the Live display.
+
+    URL installs pull an arbitrary external extension, so they get the same
+    default-deny confirmation as ``extension add --from``. Returns a mapping of
+    ``url_spec -> approved``. With *trust_override* every URL is pre-approved.
+    In a non-interactive session without the override, every URL is denied
+    (the prompt cannot be answered), mirroring the default-deny posture.
+    """
+    from rich.markup import escape as _escape_markup
+    from rich.panel import Panel
+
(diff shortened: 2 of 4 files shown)


mnriem:
Addressed the review feedback and fixed the failing CI in 60fd487c63213be5a9eff1779083683f21f3c43e.

- **Second downloader (init URL branch):** extracted a shared `install_extension_from_url` helper in `extensions/_commands.py`. Both `extension add --from` and `specify init --extension <url>` now route through the same authenticated, redirect-guarded, bounded (50 MiB) download with the TOCTOU-safe transient archive. The raw `urlopen()` path in init is gone.
- **Inactive events:** init now calls `_refresh_events_and_warn` once after the extension batch installs, so an extension declaring `events:` gets its hooks activated (matching the `add` path).
- **Rich markup injection:** the user-controlled ext spec and error text are now escaped before being interpolated into StepTracker labels.
- **URL test coverage:** added two tests to `TestExtensionFlag` — non-HTTPS rejection, and a successful HTTPS ZIP install that also asserts the download cache is cleaned up.
- **CI pytest failure:** the bundler's `_run_init` was calling the init callback without `extensions`, so it received the typer `OptionInfo` sentinel (`'OptionInfo' object is not iterable`). Now passes `extensions=None`.

Full suite passes locally (6280 passed, 8 skipped); `ruff@0.15.0 check src tests` is clean.

Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8).

mnriem:
Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8).

Addressed review #4829503833 and rebased to clear conflicts. Commit `7a93e27e`.

**Trust confirmation for URL installs (review feedback):** URL-based `--extension` installs now go through the same default-deny trust gate as `extension add --from`. Interactive sessions get an "⚠ Untrusted Source" panel and a `default=False` confirm prompt; non-interactive sessions deny by default unless `--trust-extension-urls` is passed. Trust is resolved *before* the Rich `Live` display since the prompt cannot be answered under the spinner. An unconfirmed URL is skipped (not aborted), consistent with other non-fatal extension failures during init. The bundler init callback now passes `trust_extension_urls=False` explicitly.

**Conflict resolution:** Rebased onto `upstream/main`. The overlapping change was upstream rewriting the `add --from` URL block to add tar.gz/archive-format detection; I folded that into the shared `install_extension_from_url` helper (superset with zip + tar.gz via `detect_archive_format`/`archive_format_from_name`) so both `add --from` and init share one hardened path.

**Tests:** Added coverage for deny-by-default (non-interactive), interactive-confirm approval, and trusted install; updated the existing non-HTTPS and ZIP tests for the new gate. Full suite green (6351 passed, 8 skipped); `ruff@0.15.0 check src tests` clean.

PR #3916, opened 2026-07-31
fix: narrow exception in script permission check from Exception to OSError/PermissionError

Quratulain-bilal:
## Problem

Bare `except Exception` silently swallows all errors during script permission checks.

## Fix

Narrow to `(OSError, PermissionError)` which are the realistic failure modes for file open/read operations.

## Testing

- Verified scripts with shebang get executable permissions
- Verified scripts without shebang are skipped


mnriem (changes requested):

PR #3917, opened 2026-07-31
fix: narrow exceptions in get_speckit_version() from Exception to specific types

Quratulain-bilal:
## Problem

Outer catch narrowed to `PackageNotFoundError`. Inner catch narrowed to `(OSError, ValueError, KeyError)` for file/parse errors.

## Fix

Narrowed both exception handlers to specific types.

## Testing

- Verified version is returned correctly
- Verified fallback to pyproject.toml works


mnriem (changes requested):

PR #3918, opened 2026-07-31
fix: narrow exception in Content-Length header lookup from Exception to AttributeError

Quratulain-bilal:
## Problem

Bare `except Exception` would swallow `MemoryError`, `RecursionError`, etc.

## Fix

Narrow to `AttributeError` which is the only realistic failure from `getheader()`.

## Testing

- Verified Content-Length header is read correctly

- Verified missing header is handled gracefully


mnriem (changes requested):

PR #3922, opened 2026-07-31
fix: make StepRegistry atomic and raise on corruption

Quratulain-bilal:
## Problem

`StepRegistry._load()` silently returned empty defaults on corruption, causing installed steps to appear to vanish. `save()` wrote directly to the target file risking data loss.

## Fix

`_load()` now raises `OSError` on symlinked paths and corrupted files. `save()` now uses atomic writes via mkstemp + os.replace.


mnriem (changes requested):
Please address Copilot feedback and fix test and lint errors

PR #3924, opened 2026-07-31
fix: make save_init_options() write atomically

Quratulain-bilal:
## Problem

`write_text()` directly to the target file risks leaving a partially written (corrupted) file if the process is killed mid-write.

## Fix

Use `mkstemp` + `os.replace` for atomic write.


Quratulain-bilal:

PR #3925, opened 2026-07-31
fix: catch OSError in streaming dispatch_command()

Quratulain-bilal:
## Problem

The streaming branch only caught `KeyboardInterrupt` but not `OSError`. If the binary is removed between `shutil.which()` and `subprocess.run()` (TOCTOU race), an `OSError` propagates unhandled.

## Fix

Added `except OSError` handler in both base and copilot streaming dispatch.


mnriem (changes requested):

PR #3926, opened 2026-07-31
fix: log rollback errors instead of silently swallowing them

Quratulain-bilal:
## Problem

`_rollback()` caught and discarded all exceptions during component removal. If a rollback step fails, the user has no indication the rollback was incomplete.

## Fix

Add debug logging so failures are visible.

### DIFF
--- a/src/specify_cli/bundler/services/installer.py
+++ b/src/specify_cli/bundler/services/installer.py
@@ -256,5 +256,5 @@ def _rollback(
         try:
             installer.remove(project_root, component)
         except Exception:  # noqa: BLE001 - best-effort rollback
-            logger.debug("rollback failed for %s: %s", component, exc_info=True)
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

mnriem:
The logging-format issue from the earlier review is corrected.


This adds debug-only diagnostics rather than normal user-facing reporting of an incomplete rollback, so we’re keeping it as `triage-can-wait`. Please keep that distinction clear in the description.


Before further review, resolve the conflicts with current `main`. Please also state whether AI assistance was used and, if so, include the agent/tool, model, mode/settings, and extent of assistance.


mnriem:
Please resolve conflicts. Probably needs a rebase

mnriem:

PR #3929, opened 2026-07-31
fix(presets): restore core skills instead of deleting them on preset remove

### LINKED ISSUE #3928: [Bug]: Core skills get deleted when preset is removed
### Bug Description

When you remove a preset that override's core commands/skills, the corresponding .github/skills files are deleted - instead, core's skills should still be retained.

### Steps to Reproduce

Repro:

1.) Add pirate speak preset that overrides all the core's commands (for example use the /speckit-preset plugin and prompt to add the preset)
2.) Open the .github\skills folder for your project, and validate the pirate speak skills that are generated
3.) Remove the pirate speak preset (for example use the /speckit-preset plugin and prompt to remove it)
4.) Look at the contents of the .github\skills folder again:
* Expected - All of the core skills should still exist
* Actual - All of the core skills are deleted with removing the pirate speak preset
*
It seems the only way to fix this is to use project init again, to rescaffold the core skills which isn't desirable.

### Expected Behavior

All core skills should still be retained.

### Actual Behavior

All core skills get deleted

### Specify CLI Version

0.15.0

### AI Agent

GitHub Copilot

### Operating System

Windows 11

### Python Version

3.12.10

### Error Logs

```shell

```

### Additional Context

_No response_

chelsealong:
Fixes #3928

## Bug

When a preset that overrides core commands/skills is removed, the corresponding
core skills are deleted instead of being restored — the only way to get them
back is to re-scaffold with `specify init`.

## Root cause

`PresetManager._unregister_skills_in_dir` looks for the original core command
template at `self.project_root / ".specify" / "templates" / "commands"`. That
directory is never populated by `specify init` in a real project — the actual
core command templates live in the bundled `core_pack/commands/` directory
(wheel install) or the repo-root `templates/commands/` tree (source checkout),
exactly like `PresetResolver.resolve_core()` already falls back to via
`_locate_core_pack()` / `_repo_root()`.

Because the lookup always missed, every preset-removal restoration fell
through to the "no core template found" branch and deleted the skill outright.

## Fix

Restoration now falls back to the bundled core_pack/repo-root templates when
the project-local override directory doesn't have the file. This fallback is
gated behind a new `restore_from_bundled_core` flag so the pre-existing "retire
a stale skill because its command now renders as a command file elsewhere"
path (triggered by `ai_skills` mode toggles) keeps its original delete
behavior — it must not resurrect a duplicate skill that a command file has
already superseded.

## Repro (before the fix)

```
specify init --here --integration claude
specify preset add lean --dev presets/lean
ls .claude/skills          # all 10 core skills present
specify preset remove lean
ls .claude/skills          # only 5 remain — specify/plan/tasks/implement/constitution deleted
```

After the fix, all 10 skills are present again, restored with core content
after `specify preset remove lean`.

## Testing

Added `test_skill_restored_on_preset_remove_without_project_core_templates`,
which (unlike the existing `test_skill_restored_on_preset_remove`) does not
pre-seed `.specify/templates/commands` with a fake core file, matching what a
real `specify init` project looks like. Confirmed it fails without the fix
(`AssertionError: Core skill must be restored, not deleted`) via
`git checkout HEAD~1 -- src/specify_cli/presets/__init__.py`, then passes with
the fix restored.

```
$ .venv/bin/python -m pytest tests/test_presets.py -q
526 passed in 7.40s

$ .venv/bin/python -m pytest tests -q
6380 passed, 9 skipped in 396.25s

$ .venv/bin/python -m pytest tests/test_agent_config_consistency.py -q
28 passed in 0.24s

All checks passed!
```

## AI disclosure

This PR was authored primarily by an AI coding agent (Claude Code): it
diagnosed the root cause, implemented the fix, and wrote/verified the
regression test. I reviewed the diff and test output before submitting.

### DIFF
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -3281,6 +3281,8 @@ class PresetManager:
                 if current_source not in owned_sources:
                     continue
 
+            extension_restore = extension_restore_index.get(skill_name)
+
             # Try to find the core command template. Project-local overrides
             # in core_templates_dir take precedence, but that directory is
             # rarely populated — the real core commands ship in the bundled
@@ -3288,9 +3290,17 @@ class PresetManager:
             # (source checkout). Callers that want a genuine restore (a
             # preset was removed outright, not superseded by another
             # renderer) opt into that fallback via restore_from_bundled_core
-            # so the skill is restored instead of deleted (#3928).
+            # so the skill is restored instead of deleted (#3928). An
+            # installed extension providing a core-named command resolves
+            # ahead of bundled core elsewhere, so skip the bundled fallback
+            # when an extension restore exists — otherwise it would win
+            # over the higher-priority extension layer below.
             core_file = core_templates_dir / f"{short_name}.md"
-            if not core_file.exists() and restore_from_bundled_core:
+            if (
+                not core_file.exists()
+                and restore_from_bundled_core
+                and extension_restore is None
+            ):
                 from .. import _locate_core_pack, _repo_root
 
                 _core_pack = _locate_core_pack()
@@ -3343,7 +3353,6 @@ class PresetManager:
                 mutated_names.append(skill_name)
                 continue
 
-            extension_restore = extension_restore_index.get(skill_name)
             if extension_restore:
                 content = extension_restore["source_file"].read_text(encoding="utf-8")
                 frontmatter, body = registrar.parse_frontmatter(content)
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -2157,7 +2157,10 @@ class PresetManager:
             ]
             if dir_core_ext_names:
                 self._unregister_skills_in_dir(
-                    dir_core_ext_names, skills_dir, dir_agent
+                    dir_core_ext_names,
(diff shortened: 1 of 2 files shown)


chelsealong:
Fixed the Windows CI failure: the new test read the restored SKILL.md with `read_text()` (no explicit encoding), which defaults to the platform locale encoding — cp1252 ("charmap") on Windows. The bundled `specify.md` template contains a UTF-8 emoji whose bytes aren't valid cp1252, causing `UnicodeDecodeError` on windows-latest. Now reads with `encoding="utf-8"`, matching how the file is written. Full suite (6380 passed, 9 skipped) and ruff pass locally.

chelsealong:
Addressed the inline review comment: the bundled-core fallback now skips itself when `extension_restore_index` already has an entry for the skill, so an extension-owned skill restores from the extension rather than being replaced by lower-priority bundled core content. Added `test_extension_wins_over_bundled_core_on_preset_remove` (stubs the restore index since the extension command-namespace rules make a genuine end-to-end name collision hard to construct through real manifests); confirmed it fails without the fix. Full suite: 6381 passed, 9 skipped; `ruff check src tests`: all checks passed.

mnriem:
Thank you!

PR #3934, opened 2026-08-01
feat(events): context injection for opencode and JSON-envelope agent hooks

kanfil:
### 🎯 Problem & Motivation

The agent-native runtime events layer (#3704) dispatched event scripts cleanly, but for several agents the script's `stdout` was either discarded or turned into user-facing noise instead of reaching model context:

1. **opencode**: the generated TS plugin called `runEvent` with `stdio: ['pipe', 'inherit', 'inherit']` — `stdout` was inherited by the TUI console, not captured. `session_start` mapped to `session.created` (a lifecycle event with no output channel), and `user_prompt_submit` was not mapped at all.
2. **JSON-protocol agents** (`gemini`, `tabnine`, `qwen`, `devin`, `copilot`, `cursor`): these agents mandate JSON on stdout (`hookSpecificOutput.additionalContext`, `additionalContext`, or `additional_context`). The dispatcher's plain-text `stdout` either failed JSON parsing, was ignored, or (for Gemini/Tabnine) was rendered as user-facing `systemMessage` noise.

This PR adds **first-class context injection** across all 9 event-capable agent integrations.

---

#### 1. opencode Context Injection
- **`session_start`**: remapped from `session.created` to `experimental.chat.system.transform`. Handlers' output is pushed into `output.system` (system-prompt injection, re-applied per LLM request so context survives compaction).
- **Per-sessionID caching**: `experimental.chat.system.transform` fires per LLM turn, but session-start handlers should run once per session. The generated plugin caches handler output by `sessionID` and reuses it on subsequent turns; the cache is evicted on `session.deleted`.
- **Non-session guard**: OpenCode fires `experimental.chat.system.transform` for non-session operations (e.g. agent generation) with no `sessionID`. The plugin guards with `if (!input.sessionID) return;` so session-start handlers don't inject into internal prompts.
- **`user_prompt_submit`**: mapped to `chat.message`. Handlers' output is pushed into `output.parts` as a synthetic `TextPart`.
- **OpenCode Part ID Brand Compliance**: `Part.id` must start with OpenCode's `prt` brand; an invalid ID fails user-part schema validation and crashes the session. The generated plugin derives `base` from `output.parts[last].id` (inheriting `prt` even if OpenCode changes prefixes), falling back to `"prt_"` + ts36 + rand.
- **`runEvent` Output Capture**: `stdio` updated to `['pipe', 'pipe', 'inherit']` with `encoding: 'utf-8'` so `stdout` is captured and returned while `stderr` remains inherited for error visibility.

#### 2. JSON-Envelope Dispatcher Wrapping
Adds `events_context_envelope` mapping to `IntegrationBase` and per-integration classes. The native hook command passes the target envelope as a 5th argument to `.specify/events.py`:

| Integration | `session_start` | `user_prompt_submit` | Other events | Envelope Shape |
|---|---|---|---|---|
| `claude`, `codex` | plain | plain | plain | Passthrough (already inject plain `stdout`) |
| `gemini`, `tabnine` | `hookSpecificOutput` | `hookSpecificOutput` | `suppress` | `{"hookSpecificOutput": {"hookEventName": ..., "additionalContext": ...}}` (suppresses non-injectable events to avoid `systemMessage` noise) |
| `qwen`, `devin` | `hookSpecificOutput` | `hookSpecificOutput` | `suppress` | same |
| `copilot` | `additionalContext` | `additionalContext` | plain | `{"additionalContext": ...}` (top-level) |
| `cursor` | `additional_context` | `suppress` | `suppress` | `{"additional_context": ...}` (top-level, `beforeSubmitPrompt` has no context field) |
| `opencode` | TS plugin | TS plugin | TS plugin | Plugin handles injection directly |

The dispatcher template (`_EVENTS_DISPATCHER_TEMPLATE`) and `resolve_and_run_event_command` parse the 5th argument and wrap non-empty `stdout` in the requested JSON envelope before writing.

#### 3. `hookEventName` in `hookSpecificOutput`
Qwen's hooks spec marks `hookEventName` as mandatory inside `hookSpecificOutput`. The native event name (e.g. `"SessionStart"`, `"UserPromptSubmit"`) is threaded from the integration's `CANONICAL_TO_NATIVE` through `_dispatcher_command` (6th arg) to the dispatcher and included in the JSON output. Applies to all `hookSpecificOutput` agents (Gemini, Tabnine, Qwen, Devin).

#### 4. Dispatcher Positional Arg Alignment
The dispatcher always receives the timeout as the 4th positional argument, even when the caller omits `timeout_seconds` (defaults to 60s). This keeps the argv order (`command event timeout envelope native_event`) aligned so the envelope doesn't land in the timeout slot and silently fall back to plain stdout.

---

### 🧪 Verification

- `uv run python -m pytest tests/integrations/test_events.py tests/integrations/test_integration_opencode.py -v` (107 passed)
- `uv run python -m pytest tests/test_agent_config_consistency.py -q` (28 passed)
- Smoke-tested generated dispatcher script template end-to-end with all envelope options (`plain`, `hookSpecificOutput`, `additionalContext`, `additional_context`, `suppress`).
- Smoke-tested generated opencode TS plugin code output.

### DIFF
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -251,7 +251,7 @@ def _resolve_argv(template_path, project_root, ext_id):
     return [str(script_abs), *rest]
 
 
-def _run_inline(command_name, payload, project_root, timeout):
+def _run_inline(command_name, payload, project_root, timeout, envelope="plain"):
     """Resolve and run the event command with stdlib only (no specify_cli)."""
     template_path, ext_id = _find_command_template(command_name, project_root)
     if not template_path:
@@ -269,7 +269,7 @@ def _run_inline(command_name, payload, project_root, timeout):
             cwd=str(project_root),
         )
         if result.stdout:
-            sys.stdout.write(result.stdout)
+            _emit(result.stdout, envelope)
         if result.returncode != 0:
             if result.stderr:
                 sys.stderr.write(result.stderr)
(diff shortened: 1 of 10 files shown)


mnriem (changes requested):

mnriem (approved):

PR #3943, opened 2026-08-03
fix(manifests): reject non-string metadata instead of crashing on it

Noor-ul-ain001:
## Problem

`ExtensionManifest._validate()` and `PresetManifest._validate()` check only key **presence** for `id`/`name`/`version`/`description`, then feed the values straight to `re.match()` and `packaging.Version()`. Both raise a bare `TypeError` on a non-string — and `TypeError` is neither `ValidationError` nor `PresetValidationError`, so it escapes every caller that already handles a malformed manifest.

YAML makes this an easy authoring slip rather than a contrived one:

```yaml
extension:
  id: myext
  version: 1.0      # unquoted -> parses as a float, not a string
```

## Symptom

This is exactly the failure mode the in-tree comment above the section guards was written to prevent (the preset side landed as #3898). `list_installed()` degrades a bad manifest to `⚠️ Corrupted extension`, but its `except` catches only the domain error:

```python
except ValidationError:
    # Corrupted extension
```

So a single bad manifest makes `specify extension list` exit 1 with a raw `TypeError` traceback and **no output at all** — hiding every *healthy* extension too, not just the broken one. Same for `specify preset list`.

Before this change:

```
$ specify extension list
TypeError: 'float' object is not iterable      # exit 1, nothing listed
```

After:

```
$ specify extension list

Installed Extensions:

    good-ext (v1.0.0)
     ...
  ✗ bad-ext (v1.0)
     ⚠️ Corrupted extension
```

## Also unguarded on the same path

- extension `provides.commands[].name` → `TypeError` from the command-name pattern match. The sibling `file` field was already safe, since `relative_extension_path_violation()` rejects a non-string — line 404 even special-cases one for its error label, so non-string values were known to arrive here.
- preset `provides.templates[].name` / `.file` → `TypeError` from `re.match` and `os.path.normpath` respectively.

## Why this shape of fix

The third manifest twin, `IntegrationDescriptor` (`integrations/catalog.py`), is **already hardened** — it type-checks the same four fields and catches `TypeError` alongside `InvalidVersion`. This PR brings the other two in line with that existing reference implementation, so the three manifest validators stay consistent.

## Testing

68 tests added across `tests/test_extensions.py` and `tests/test_presets.py`:

- each of the four metadata fields × `float`, `int`, `None`, `list`, `dict`, `bool`
- extension `commands[].name` and preset `templates[].name`/`.file`, same matrix
- one end-to-end guard per manifest type asserting a healthy entry still lists normally while only the bad one degrades to "Corrupted"

All 68 fail with the source change reverted and pass with it. Full `tests/test_extensions.py` + `tests/test_presets.py` run shows no new failures: the 7 failures / 66 errors present are identical before and after (pre-existing Windows symlink-elevation and pytest tmpdir `WinError 5` env issues, unrelated to this change).

Assisted-by: Claude Opus 5 (1M context)

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -285,9 +285,25 @@ class ExtensionManifest:
             raise ValidationError(
                 f"Invalid extension: expected a mapping, got {type(ext).__name__}"
             )
+        # Check presence AND type: the format/version checks below feed these
+        # values straight to ``re.match`` and ``packaging.Version``, both of
+        # which raise a bare TypeError on a non-string. YAML makes that an easy
+        # authoring slip -- unquoted ``version: 1.0`` parses as a float and
+        # ``id: 2`` as an int -- and TypeError is not a ValidationError, so it
+        # escapes every caller that already handles a malformed manifest (see
+        # list_installed()'s "Corrupted extension" fallback, which catches
+        # ValidationError only, making one bad extension exit ``specify
+        # extension list`` with a raw traceback and hide the healthy ones).
+        # Mirrors the sibling IntegrationDescriptor, which already type-checks
+        # the same four fields.
         for field in ["id", "name", "version", "description"]:
             if field not in ext:
                 raise ValidationError(f"Missing extension.{field}")
+            if not isinstance(ext[field], str):
+                raise ValidationError(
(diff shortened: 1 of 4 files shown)


mnriem:
Thank you!

PR #3957, opened 2026-08-03
fix(events): return None for an unparseable script command

marcelsafin:
## Description

The script-command resolver splits the configured command with a bare `shlex.split()`, so a command string with unbalanced quotes (e.g. `command: "run 'oops"`) crashes event dispatch with a raw `ValueError: No closing quotation`. The dispatcher-template twin a few lines up already wraps the very same call in `try/except ValueError` and returns `None` so dispatch falls back cleanly.

**Fix:** wrap the split the same way and return `None`, restoring parity between the two paths. One malformed hook entry no longer takes down dispatch for every other hook.

## Testing

- [x] New regression test `test_unparseable_script_command_returns_none` (fails on main, passes with fix)

## AI Disclosure

### DIFF
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -524,7 +524,13 @@ def _resolve_event_command_argv(
     else:
         base = project_root / ".specify"
 
-    tokens = shlex.split(script_cmd, posix=(os.name != "nt"))
+    try:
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3960, opened 2026-08-03
fix(extensions): reject reinstall when a kept config cannot be read

marcelsafin:
## Description

The keep-config rescue branch of `install_from_directory()` reads each preserved config with bare `read_bytes()`/`stat()` calls, so a kept config that cannot be read (permission or I/O error) crashes the reinstall with a raw `OSError`. The sibling symlink guard four lines above already rejects with a `ValidationError` and resolution guidance for exactly the same reason: bytes that cannot be safely rescued must never reach the `rmtree` below.

**Fix:** wrap the read and raise `ValidationError` with guidance (*fix its permissions or remove it, then reinstall*) while `dest_dir` is still untouched, so the preserved bytes are never rescued half-read or lost.

## Testing

- [x] New regression test `test_reinstall_with_unreadable_kept_config_aborts_with_guidance` (fails on main with raw `PermissionError`, passes with fix)

## AI Disclosure

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -2186,10 +2186,24 @@ class ExtensionManager:
                         "a regular file or remove it — then reinstall."
                     )
                 if cfg_file.is_file():
-                    stranded_configs[cfg_file.name] = (
-                        cfg_file.read_bytes(),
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3962, opened 2026-08-03
fix(extensions): treat an unreadable staged backup as a conflict

marcelsafin:
## Description

The rescue-retry loop in `install_from_directory()` reads each staged backup with bare `stat()`/`read_bytes()` calls, so a staged config that cannot be read crashes the reinstall with a raw `OSError`. Every sibling read in this path — the live twin four lines below, the packaged baseline check, the mode sidecar — already catches `OSError`.

**Fix:** treat an unreadable staged file like an uncomparable live config: add it to the conflict set so both copies are preserved and the retry aborts with the existing resolution guidance while `dest_dir` is still untouched.

## Testing

- [x] New regression test `test_retry_with_unreadable_staged_config_aborts_and_preserves_both` (fails on main with raw `PermissionError`, passes with fix)

## AI Disclosure

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -2093,8 +2093,19 @@ class ExtensionManager:
                         _staged_modes = _loaded_modes
             for staged_name in sorted(staged_names):
                 staged_file = rescue_staging_dir / staged_name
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3963, opened 2026-08-03
fix(events): preserve a non-UTF-8 config.toml on hook install/teardown

marcelsafin:
## Description

`_merge_toml_fragment()` and `_remove_toml_entries()` read the user's `config.toml` (Codex hooks) with bare `read_text()` calls, so a non-UTF-8 (or otherwise unreadable) file crashes `install_integration_events()` and `remove_integration_events()` with a raw `UnicodeDecodeError` — and the merge path **regenerates the file from what it read**, so it would have discarded the user's bytes had it not crashed first. Every JSON merge/remove path already goes through `_load_user_json()`, which skips on an unreadable file precisely to preserve user content (#22).

**Fix:**
- merge: abort and return `False` so the caller skips tracking the untouched file (same S5 contract as `_merge_copilot_json`), leaving the user's bytes intact;
- teardown: skip the cleanup with a warning instead of crashing — the file contains only user content as far as we can tell, and the caller drops the manifest claim either way (S9).

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` (6,311 passed, 176 skipped)
- [x] Two new regression tests (install with pre-existing non-UTF-8 `config.toml`, teardown after the config was rewritten as non-UTF-8; both fail on main, pass with fix)

## AI Disclosure

### DIFF
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -1193,11 +1193,12 @@ def install_integration_events(
                 lines.append(f'timeout = {_native_timeout(integration, cfg.get("timeout", 60) + EVENT_TIMEOUT_BUFFER)}')
                 lines.append('speckit_marker = true')
                 lines.append('')
-        _merge_toml_fragment(config_path, "\n".join(lines))
-        rel = str(config_path.relative_to(project_root))
-        if rel not in manifest.files:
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #3964, opened 2026-08-03
fix: add timeout to run_command() to prevent indefinite hangs

Quratulain-bilal:
## Problem

"run_command() utility called subprocess.run() without timeout. If a spawned process hangs, the CLI blocks forever.

## Fix

- Add 	imeout parameter (default 120s)
- Handle subprocess.TimeoutExpired exception
- Remove redundant hasattr(e, 'stderr') check


mnriem (changes requested):

PR #3965, opened 2026-08-03
fix: use logger instead of print() in events.py

Quratulain-bilal:
## Problem

Three error/warning messages in events.py used print(..., file=sys.stderr) instead of the module's existing logger, bypassing logging configuration.

## Fix

Replace print() calls with logger.error()/logger.warning() for consistent logging behavior.


mnriem (changes requested):

PR #3966, opened 2026-08-03
fix: use logger instead of print() in _agent_config.py

Quratulain-bilal:
## Problem

_agent_config.py used print(..., file=sys.stderr) for warnings instead of using the logging module.

## Fix

Add logging.getLogger(__name__) and replace print() with logger.warning(). Remove unused sys import.

### DIFF
--- a/src/specify_cli/_agent_config.py
+++ b/src/specify_cli/_agent_config.py
@@ -1,9 +1,11 @@
 """Agent configuration constants derived from the integration registry."""
 from __future__ import annotations
 
+import logging
 import os
-import sys
 from typing import Any
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


mnriem (commented):
Address Copilot feedback and fix test & lint errors

mnriem:
This is a logging-consistency change rather than a demonstrated functional fix, so we’re keeping it as `triage-can-wait`.


The specific test mismatch from the earlier review remains: `test_resolve_default_init_integration_invalid_warns_and_falls_back` still checks `captured.err`. Adding a separate `caplog` test does not update that existing assertion. Please address the named test as part of the change.


Please also state whether AI assistance was used and, if so, include the agent/tool, model, mode/settings, and extent of assistance.


mnriem:

PR #3967, opened 2026-08-03
fix: add timeout to streaming subprocess.run() in base.py

Quratulain-bilal:
## Problem

The streaming branch of dispatch_command() called subprocess.run() without a timeout, which could hang indefinitely if the child process stalls.

## Fix

Apply the same timeout parameter (default 600s) as the non-streaming branch, with proper TimeoutExpired handling.


mnriem (changes requested):

PR #3969, opened 2026-08-03
fix: replace print() with logger.warning() in extensions catalog warnings

Quratulain-bilal:
## Problem

print(..., file=sys.stderr) is inappropriate for library code - it bypasses the logging framework and can't be silenced or redirected by consumers.

## Fix

Replaced with logger.warning() for proper log management. Removed unused local sys imports.

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -12,6 +12,7 @@ import copy
 import errno
 import hashlib
 import json
+import logging
 import os
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


mnriem:
Thanks — this is the right fix (stderr `print` in library code can't be silenced/redirected by consumers), and the `caplog` regression test is exactly what I want to see. Just one thing before review: please **disclose any AI assistance** per CONTRIBUTING. With that added, this is in good shape — and for reference, PRs like this one (focused + a real test) are much easier to prioritize than the untested atomic-write batch.

Quratulain-bilal:
## AI Assistance Disclosure

I used AI assistance (GitHub Copilot) to identify `print()` calls in library code that bypass the logging framework. The refactoring was reviewed and validated by me — I verified the logger output matches the original stderr behavior and added a regression test using `caplog` to confirm warnings are logged at the correct level.

@mnriem I've added the AI disclosure above. This PR was developed with Copilot assistance to identify the print-to-logger migration points, and I personally reviewed the changes and wrote the caplog regression test. Ready for re-review when you have a chance.

mnriem (changes requested):

PR #3970, opened 2026-08-03
fix: replace print() with logger.warning() in presets catalog warning

Quratulain-bilal:
## Problem

print(..., file=sys.stderr) bypasses the logging framework and cannot be silenced or redirected by consumers.

## Fix

Replaced with logger.warning() for proper log management. Removed unused sys import.


mnriem (changes requested):

PR #3971, opened 2026-08-03
fix: make _safe_write_json actually atomic with mkstemp + os.replace

Quratulain-bilal:
## Problem

Despite its name, _safe_write_json used write_text() which truncates the file before writing. A crash or power loss mid-write leaves a partial JSON file. This function is called by _merge_copilot_json, _merge_opencode_plugin_ref, etc., meaning all Copilot and OpenCode event config writes were non-atomic.

## Fix

Now uses 	empfile.mkstemp + os.replace for atomic writes, matching the pattern used in _utils.py, shared_infra.py, and other safe-write utilities in the codebase.

### DIFF
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -2017,8 +2017,9 @@ def _safe_write_json(dst: Path, data: dict) -> None:
     fd, tmp = tempfile.mkstemp(
         dir=str(dst.parent), prefix=f".{dst.name}.", suffix=".tmp"
     )
-    os.chmod(tmp, 0o644)
     try:
+        if dst.exists() and hasattr(os, "fchmod"):
+            os.fchmod(fd, dst.stat(follow_symlinks=False).st_mode & 0o7777)
         with os.fdopen(fd, "w", encoding="utf-8") as f:
             json.dump(data, f, indent=2)
             f.write("\n")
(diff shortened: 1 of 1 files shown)


mnriem (changes requested):

mnriem:
Thanks. Two things before this can be reviewed, per CONTRIBUTING: (1) please **disclose any AI assistance** and its extent; (2) this needs a **test** — assert the file round-trips after a successful write, and that a failure mid-write leaves the original intact and cleans up the temp file. One correctness note from the review: `os.replace` makes publication atomic but isn't durable without an `fsync` of the file and the parent directory — so as written it doesn't fully close the crash/power-loss window it targets. Worth adding those syncs (or reusing an existing atomic-write helper like `_utils.atomic_write_json`). Marking `author-awaiting`.

Quratulain-bilal:
## AI Assistance Disclosure

I used AI assistance to identify that `_safe_write_json` used non-atomic `write_text()` despite its name. The atomic rewrite (`mkstemp` + `os.replace` + `os.fchmod`) was reviewed by me. I noted the maintainer feedback about `fsync` for durability and will address that along with the required regression test.

Quratulain-bilal:

PR #3972, opened 2026-08-03
fix: make _merge_toml_fragment and _remove_toml_entries use atomic writes

Quratulain-bilal:
## Problem

Both _merge_toml_fragment and _remove_toml_entries used write_text() which truncates the file before writing. A crash or power loss mid-write leaves a partial TOML file.

## Fix

Now uses 	empfile.mkstemp + os.replace for atomic writes, matching the pattern used elsewhere in the codebase.

### DIFF
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -1731,8 +1731,9 @@ def _merge_toml_fragment(dst: Path, fragment: str) -> None:
     fd, tmp = tempfile.mkstemp(
         dir=str(dst.parent), prefix=f".{dst.name}.", suffix=".tmp"
     )
-    os.chmod(tmp, 0o644)
     try:
+        if dst.exists() and hasattr(os, "fchmod"):
+            os.fchmod(fd, dst.stat(follow_symlinks=False).st_mode & 0o7777)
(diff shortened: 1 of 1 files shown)


mnriem:
Thanks — the atomic mkstemp+os.replace pattern is correct and matches the house style. Two things before review, per CONTRIBUTING: (1) please disclose any AI assistance and its extent; (2) this needs a test — assert the file still contains the expected TOML after a successful write, and that a mid-write failure leaves the original intact and cleans up the temp file. Optional: reuse the existing atomic-write helpers (_utils.atomic_write_json, shared_infra._write_shared_bytes) rather than inlining the pattern a third time. Marking author-awaiting once addressed.

Quratulain-bilal:
## AI Assistance Disclosure

I used AI assistance to identify non-atomic write patterns in `_merge_toml_fragment` and `_remove_toml_entries`. The solution was reviewed by me I verified each path uses `mkstemp` + `os.replace` with permission preservation, and I'm working on a crash-injection test to prove the original file survives mid-write failures.

Quratulain-bilal:

PR #3973, opened 2026-08-03
fix: add ValueError guard for timeout in _build_opencode_plugin

Quratulain-bilal:
## Problem

int(cfg.get('timeout', 60)) crashes with an unhandled ValueError when a user provides a non-numeric timeout value in event configuration. Two other call sites in the same file already handle this gracefully with 	ry/except (TypeError, ValueError).

## Fix

Added the same 	ry/except guard, making the third call site consistent.


mnriem (changes requested):

Quratulain-bilal:
Hi @mnriem — I've addressed Copilot's feedback: the regression test now asserts the complete 
unEvent() call with the 60-second fallback, not just the handler name. Ready for re-review when you get a chance.

mnriem (approved):

PR #3974, opened 2026-08-03
fix: use atomic writes for composed preset command files

Quratulain-bilal:
## Problem

Three write_text() calls for composed command files were not atomic. A crash mid-write leaves a partial .md file which can cause errors when the command is later read.

## Fix

Now uses a shared _atomic_write_text helper with 	empfile.mkstemp + os.replace.

### DIFF
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -67,8 +67,9 @@ def _atomic_write_text(path: Path, content: str) -> None:
     fd, tmp = tempfile.mkstemp(
         dir=str(path.parent), prefix=f".{path.name}.", suffix=".tmp"
     )
-    os.chmod(tmp, 0o644)
     try:
+        if path.exists() and hasattr(os, "fchmod"):
+            os.fchmod(fd, path.stat(follow_symlinks=False).st_mode & 0o7777)
         with os.fdopen(fd, "w", encoding="utf-8") as f:
             f.write(content)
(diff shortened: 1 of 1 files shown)


mnriem:
Thanks for all the hardening work — the shared `_atomic_write_text` helper here is the right shape. A process point, though: you currently have **37 open PRs**, most of them the same class of small defensive change (atomic writes, exception narrowing, timeouts). Per CONTRIBUTING, accounts past 3 open PRs may have further submissions deprioritized in the review queue, and this volume of near-identical PRs is genuinely hard to review one by one.

Could you **consolidate the related changes into a single PR** (e.g. one "atomic writes across preset/manifest/events" PR), and for that PR: (1) **disclose any AI assistance** per CONTRIBUTING, and (2) include **tests** — a crash/failure-injection test showing the original file survives a mid-write failure and the temp file is cleaned up? A focused, tested, disclosed PR is far more likely to land than many small ones. I'm marking the batch `author-over-cap` and will hold the rest pending that consolidation.

Quratulain-bilal:
## AI Assistance Disclosure

I used AI assistance to identify non-atomic `write_text()` calls for composed preset command files. The solution (`mkstemp` + `os.replace` with `os.fchmod` for permission preservation) was reviewed and validated by me. I verified the fix matches the existing atomic-write pattern used in `_utils.py` and `shared_infra.py`.

Quratulain-bilal:

PR #3975, opened 2026-08-03
fix: log progress tracker refresh errors instead of silently swallowing

Quratulain-bilal:
## Problem

The bare except Exception: pass in StepTracker._maybe_refresh() completely hid rendering bugs in the Rich progress display.

## Fix

Now logs at DEBUG level with full traceback for diagnostics.


mnriem:
Thank you!

PR #3980, opened 2026-08-04
fix(manifests): reject non-string requires.speckit_version

Noor-ul-ain001:
## Problem

`requires.speckit_version` is checked for **presence** but never for **type** in both the extension and preset manifest validators. An unquoted YAML `speckit_version: 1.0` — an easy authoring slip, since YAML parses it as a float — passes validation and reaches `SpecifierSet(required)` in `check_compatibility()`.

That call is guarded by `except InvalidSpecifier` alone, which a non-string escapes **two different ways**:

| value | what happens |
|---|---|
| `1.0`, `5`, `True`, `None` | `TypeError: 'float' object is not iterable` — raised by the `SpecifierSet` constructor, not `InvalidSpecifier` |
| `[">=0.1.0"]`, `{"min": "0.1"}` | **iterable**, so `SpecifierSet` accepts it; fails much later as `AttributeError: 'str' object has no attribute 'filter'` from inside `.contains()` |

Neither is a `CompatibilityError`/`PresetCompatibilityError`, so both bypass the `Compatibility Error` handlers in `extensions/_commands.py:1101` and `presets/_commands.py:275`, and exit 1 with a raw traceback that **names no field** — leaving the author no hint which manifest key is wrong. The list/dict case is the nastier one: the error surfaces deep inside `packaging` with a message that mentions neither the manifest nor the key.

Reproduced on `main` with an otherwise fully valid manifest:

```
validate() PASSED. requires_speckit_version = 1.0
...
  File "src/specify_cli/extensions/__init__.py", line 1855, in check_compatibility
    SpecifierSet(required)  # Just to validate
TypeError: 'float' object is not iterable
```

## Fix

- **Both validators** now require a non-empty string, so the failure is a clean `ValidationError`/`PresetValidationError` naming the field.
- **Both `check_compatibility()` methods** additionally reject a non-string up front. These are public and reachable with a hand-built or mutated manifest (`test_check_compatibility_invalid` already mutates `manifest.data` this way), so the guard belongs at both layers.

This mirrors the sibling `IntegrationDescriptor`, which [already requires a non-empty string](https://github.com/github/spec-kit/blob/main/src/specify_cli/integrations/catalog.py#L735) for the very same key, and completes the type-checking pass started in #3943 for the neighbouring `extension`/`preset` fields — `id`/`name`/`version`/`description` are now type-checked while `requires.speckit_version` was still presence-only.

Blank strings are rejected too, so `speckit_version: "   "` can't silently mean "any version".

## Testing

33 regression tests added across both modules, covering every escape path (scalars, iterables, and blank strings) at both the validator and manager layer. **26 fail without the source change** (the rest are existing-behaviour assertions that pass either way).

```
33 passed
```

Full `tests/test_extensions.py` + `tests/test_presets.py`: `7 failed, 952 passed, 78 errors` both **with and without** this change — identical counts. Those pre-existing failures are Windows-local `PermissionError (WinError 5)` from symlink/tmpdir restrictions in my environment, unrelated to this change.

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -341,6 +341,25 @@ class ExtensionManifest:
             )
         if "speckit_version" not in requires:
             raise ValidationError("Missing requires.speckit_version")
+        # Presence alone is not enough: check_compatibility() feeds this value to
+        # ``SpecifierSet(required)``, guarded only by ``except InvalidSpecifier``,
+        # which a non-string escapes two different ways. A float/int/bool/None
+        # raises TypeError from the constructor, while a list or dict is an
+        # *iterable*, so SpecifierSet accepts it and the failure surfaces much
+        # later as ``AttributeError: 'str' object has no attribute 'filter'`` from
+        # inside .contains(). Neither is a CompatibilityError, so both bypass the
+        # CLI's "Compatibility Error" handler and exit 1 with a raw traceback
+        # naming no field. An unquoted ``speckit_version: 1.0`` is an easy YAML
+        # slip. Mirrors the sibling IntegrationDescriptor, which already requires
+        # a non-empty string here.
+        if (
+            not isinstance(requires["speckit_version"], str)
+            or not requires["speckit_version"].strip()
+        ):
+            raise ValidationError(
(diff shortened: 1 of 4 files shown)


mnriem:
Thank you!

Issue #4010, opened 2026-08-07
[Feature]: Allow extensions to declare templates and scripts in their manifest


mnriem:
## Problem Statement

Extensions today can only formally declare **commands** (plus `config`, `hooks`, `events`) under `provides` — see `ExtensionManifest` (`src/specify_cli/extensions/__init__.py:606-621`) and the validation requiring "at least one command, hook, or event" (`:386-387`). There is no `provides.templates` or `provides.scripts`.

An extension *can* still ship `templates/<name>.md` or `scripts/<name>.sh` files, which the resolver picks up by **convention** (`PresetResolver.collect_all_layers` → `_find_in_subdirs`, `src/specify_cli/presets/__init__.py:5417-5455`). But those contributions are:

- **Undeclared** — no `id`, `name`, `description`; matched purely by filename.
- **Forced to `replace`** (`:5452`) — no author control, and no way to make the intent explicit.
- **Invisible to tooling** — not counted or shown by any `list`/`info` command.

This blocks a consumer (e.g. a setup wizard) from enumerating extension-contributed templates/scripts with real metadata, and it means presets and extensions have asymmetric manifest schemas for the same kinds of artifacts.

## Proposed Solution

Add **opt-in, declarative** `provides.templates` and `provides.scripts` sections to the extension manifest, mirroring the preset shape (`provides.templates` with `type/name/file/description`, `src/specify_cli/presets/__init__.py:400-453`) — **minus any authorable `strategy`**, since extension contributions are always `replace` (see below).

```yaml
provides:
  commands:
    - name: speckit.myext.report
      file: commands/speckit.myext.report.md
      description: "..."
  templates:
    - name: myext-template
      file: templates/myext-template.md
      description: "Report scaffold contributed by myext"
  scripts:
    - name: myext-collect
      file: scripts/bash/myext-collect.sh
      description: "Data-collection helper"
      runtimes: [bash, powershell, python]   # declared, not inferred
```

Behavior:

- **Declared entries are authoritative; convention lookup remains the fallback** — same pattern presets already use (`:5380-5383`), so nothing breaks for existing extensions.
- New `ExtensionManifest.templates` / `ExtensionManifest.scripts` properties + validation (shape, path-safety via `relative_extension_path_violation`, name format), consistent with the existing command/preset validators.
- Declared templates/scripts become countable and describable, so downstream `list --json` / `info --json` (tracked separately) can report per-kind `provides` counts and per-artifact detail.

[...]

PR #4012, opened 2026-08-07
feat(extensions): accept provides.templates and provides.scripts in manifest

chelsealong:
## Summary

Extensions could only formally declare **commands** under `provides` (plus `config`/`hooks`/`events`) — see `ExtensionManifest` (`src/specify_cli/extensions/__init__.py`). Templates and scripts an extension shipped were picked up purely by filename **convention** during preset/template resolution, with no `id`, `name`, or `description`, and forced to `replace` with no way to make that intent explicit.

This PR implements #4010 end to end — manifest schema, validation, and resolver wiring:

- `provides.templates` and `provides.scripts` are now optional, validated sections of `extension.yml`, mirroring the shape `PresetManifest` already uses for its own templates (`name`/`file`/`description`), minus an authorable `strategy` — extension-provided artifacts always resolve as `replace`, and a manifest that includes a `strategy` key on one of these entries now raises a clear `ValidationError` instead of silently accepting (and ignoring) it.
- `provides.scripts[].runtimes` accepts an optional list restricted to `bash`/`powershell`/`python` — declared metadata, not inferred from the file extension.
- `ExtensionManifest` gains `templates` and `scripts` properties (parity with the existing `commands`/`config` properties) so tooling (e.g. a setup wizard, `extension info`) can enumerate an extension's declared artifacts directly from the manifest.
- The "must provide at least one command, hook, or event" rule is relaxed so an extension may satisfy it with only a declared template or script — otherwise the new sections would be useless for an extension whose only contribution is a template/script.
- `PresetResolver.collect_all_layers`, `resolve()`, and `resolve_with_source()` (`src/specify_cli/presets/__init__.py`) now consult `ExtensionManifest.templates`/`.scripts` for `template_type in {"template", "script"}` the same way they already did for `.commands`, via the shared `_extension_manifest_declared_template()` helper. **The manifest is authoritative and is checked before convention-based lookup**: a declared entry whose `file` doesn't sit at the conventional path (`templates/<name>.md` / `scripts/<name>.sh` directly under the extension root) still resolves, a declared entry wins over a stale conventional file at the same name, and a declared-but-missing file does not fall back to convention. Undeclared on-disk files keep resolving via convention exactly as before (no regression).
- `provides.templates`/`provides.scripts` entries must have unique names within their section — a duplicate is rejected with a `ValidationError`, since the resolver returns the first entry matching a name and a later duplicate would otherwise be silently unreachable.
- All new fields are additive under `schema_version: "1.0"`; no version bump needed.
- Docs: `extensions/EXTENSION-API-REFERENCE.md` manifest schema section and Python API property list updated. The doc's "Always resolve as 'replace'" claim now matches actual resolver behavior for templates/scripts, not just commands. `extensions/EXTENSION-DEVELOPMENT-GUIDE.md`'s `provides` section now distinguishes its own sub-fields (`commands`/`templates`/`scripts`) from `hooks`/`events`, which are top-level manifest fields, not nested under `provides`.

## Testing

New test class `TestExtensionManifestTemplatesAndScripts` in `tests/test_extensions.py` covers: valid declarations, templates/scripts-only extensions (no commands/hooks/events), section-type validation, per-entry shape/path-safety/name-format validation, rejected `strategy` key, rejected duplicate names within a section, `runtimes` validation (type + allowed values), and `description` type-checking.

Three new tests in `tests/test_presets.py::TestWrapStrategy` cover resolver precedence per the issue's acceptance criteria: `test_extension_template_resolves_via_manifest_when_filename_differs` and `test_extension_script_resolves_via_manifest_when_filename_differs` assert a manifest-declared template/script resolves when its file lives away from the naming convention; `test_extension_template_convention_lookup_unaffected_when_undeclared` asserts an undeclared on-disk template still resolves via the pre-existing convention (no regression).

Confirmed all new tests fail against the pre-fix source (`git checkout HEAD~1 -- src/specify_cli/extensions/__init__.py src/specify_cli/presets/__init__.py`) — 19/20 manifest tests with `AttributeError`/`DID NOT RAISE`, and the two new resolver-precedence tests with an empty layer list — then pass again after restoring the fix. The duplicate-name rejection added after review (`git checkout HEAD -- src/specify_cli/extensions/__init__.py` against the pre-fix version of that file) was confirmed to fail with `DID NOT RAISE ValidationError` on both parametrized cases before the fix, and pass after.

```
$ .venv/bin/python -m pytest tests/test_extensions.py -k "TestExtensionManifestTemplatesAndScripts" -q
====================== 22 passed, 499 deselected in 0.35s ======================

$ .venv/bin/python -m pytest tests/test_extensions.py tests/test_presets.py -q
====================== 1109 passed, 2 warnings in 9.49s ======================

$ .venv/bin/python -m pytest tests -q
=========== 6623 passed, 9 skipped, 45 warnings in 370.28s (0:06:10) ===========

All checks passed!
```

## AI Disclosure

Implemented autonomously by Claude Code (model: Claude Sonnet 5) under human direction: read the issue and related open PRs first to confirm no overlap, implemented the manifest schema/validation/properties and resolver wiring, wrote failing-first regression tests, and verified the full test suite and `ruff` locally before pushing.

### DIFF
--- a/extensions/EXTENSION-API-REFERENCE.md
+++ b/extensions/EXTENSION-API-REFERENCE.md
@@ -40,11 +40,24 @@ requires:
       required: boolean    # Optional, default: false
 
 provides:
-  commands:              # Required, at least one command
+  commands:              # At least one of commands/templates/scripts/hooks/events required
     - name: string       # Required, pattern: ^speckit\.[a-z0-9-]+\.[a-z0-9-]+$
       file: string       # Required, relative path to command file
       description: string # Required
       aliases: [string]  # Optional, same pattern as name; namespace must match extension.id and must not shadow core or installed extension commands
+
+  templates:             # Optional, array of declared templates. Always resolve
+                          # as "replace" -- 'strategy' is not an authorable field here.
+    - name: string       # Required, pattern: ^[a-z0-9-]+$
+      file: string       # Required, relative path to template file
+      description: string # Optional
+
+  scripts:               # Optional, array of declared scripts. Always resolve
+                          # as "replace" -- 'strategy' is not an authorable field here.
+    - name: string       # Required, pattern: ^[a-z0-9-]+$
+      file: string       # Required, relative path to script file
+      description: string # Optional
+      runtimes: [string] # Optional, subset of: bash, powershell, python
 
   config:                # Optional, array of config files
     - name: string       # Config file name
@@ -111,6 +124,29 @@ defaults:                # Optional, default configuration values
 - **Examples**: `speckit.jira.specstoissues`, `speckit.linear.sync`
 - **Invalid**: `jira.specstoissues`, `speckit.command`, `speckit.jira.CreateIssues`
 
+#### `provides.templates[].name` / `provides.scripts[].name`
+
+- **Type**: string
+- **Pattern**: `^[a-z0-9-]+$`
+- **Description**: Unlike commands, templates and scripts are not invoked by
+  name, so they use the same plain slug pattern as `extension.id` rather than
+  the namespaced command pattern.
+- **Examples**: `myext-template`, `myext-collect`
+
+#### `provides.templates[].strategy` / `provides.scripts[].strategy`
+
+- Not an authorable field. Extension-contributed templates and scripts are
+  always resolved as `replace`; a manifest that includes a `strategy` key on
+  one of these entries is rejected with a `ValidationError`. Composable
+  strategies (`wrap`/`prepend`/`append`) are preset-only.
+
+#### `provides.scripts[].runtimes`
+
+- **Type**: array of strings
+- **Values**: `bash`, `powershell`, `python`
+- **Description**: Declares which runtimes the script supports. Purely
+  informational metadata — it is not used to select or invoke the script.
+
 #### `hooks`
 
 - **Type**: object
@@ -143,6 +179,8 @@ manifest.version                   # str: Version
 manifest.description               # str: Description
 manifest.requires_speckit_version  # str: Required spec-kit version
 manifest.commands                  # List[Dict]: Command definitions
+manifest.templates                 # List[Dict]: Declared template definitions
+manifest.scripts                   # List[Dict]: Declared script definitions
 manifest.hooks                     # Dict: Hook definitions
 ```
 
--- a/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
+++ b/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
@@ -177,9 +177,11 @@ Compatibility requirements.
 
 What the extension provides.
 
-**Optional sub-fields**:
+**Optional sub-fields** (at least one of `commands`, `templates`, `scripts`, `hooks`, or `events` is required):
 
-- `commands`: Array of command objects (at least one command or hook is required)
+- `commands`: Array of command objects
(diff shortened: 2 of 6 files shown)


mnriem (changes requested):

chelsealong:
Addressed the Copilot feedback in 0070d25:

- Manifest lookup is now checked *before* convention-based lookup for extension templates/scripts (both in `resolve()` and `collect_all_layers()`, via a new shared `_extension_manifest_declared_template()` helper), so a declared entry at a non-conventional path wins over a stale conventional file, and a declared-but-missing file is no longer silently masked by convention.
- `resolve()`/`resolve_with_source()` are now manifest-aware for templates/scripts, matching `collect_all_layers()`/`resolve_content()`.
- Added regression tests covering: manifest wins when both a stale conventional file and a declared file exist, a declared-but-missing file doesn't fall back, and `resolve()`/`resolve_with_source()` parity for scripts.
- Updated `extensions/EXTENSION-DEVELOPMENT-GUIDE.md`'s `provides` section, which still said only commands were supported and that a command or hook was required.

Full suite (6621 passed, 9 skipped) and `ruff check` pass.

mnriem:

chelsealong:
Fixed in 467c0f7: \`_extension_manifest_declared_template()\` was calling \`.resolve()\` on the returned candidate path, which follows symlinks in \`ext_dir\`'s ancestors (e.g. the symlinked tmp dir on macOS CI runners) and diverges from the unresolved paths convention-based lookup returns for the same directory — that's what the 4 macOS/ubuntu/windows pytest failures were. Now resolution is only used for the path-traversal containment check; the returned path stays unresolved, matching convention lookup. Reproduced the exact failures locally by pointing `TMPDIR` at a symlink, confirmed they're gone after the fix, then ran the full suite (6621 passed, 9 skipped) and `ruff check src tests` (all checks passed) with the fix in place.

mnriem (approved):

mnriem:
Closes #4010

chelsealong:
**Edit:** this PR merged (684b3d8) before 1ac3c3c was pushed, so that commit did not make it into `main`. Opened #4016 to carry it forward instead.

---

Addressed the remaining Copilot feedback in 1ac3c3c:

- `provides.templates`/`provides.scripts` entries with a duplicate `name` within the same section are now rejected with a `ValidationError` (the resolver returns the first entry matching a name, so a later duplicate was silently unreachable). Added regression coverage for both sections.
- `EXTENSION-DEVELOPMENT-GUIDE.md`'s `provides` section now distinguishes its own sub-fields (`commands`/`templates`/`scripts`) from `hooks`/`events`, which are top-level manifest fields, not nested under `provides`.
- Updated the PR description: the resolver bullet now correctly says the manifest is checked *before* convention-based lookup (it previously described the pre-0070d25 behavior).

Full suite (6623 passed, 9 skipped) and `ruff check` pass.

Re: the AGENTS.md acceptance-criterion item — I intentionally left that file alone. It documents only the AI-agent integration subsystem (adding Claude/Gemini/Copilot/etc. support); none of the other subsystems (presets, the extension system itself) are covered there either. The extension manifest schema is documented in `extensions/EXTENSION-API-REFERENCE.md` and `EXTENSION-DEVELOPMENT-GUIDE.md`, both already updated. Happy to add a pointer there if a maintainer prefers otherwise.

PR #4013, opened 2026-08-07
fix(bundle): wrap malformed YAML in a local .zip bundle manifest

Noor-ul-ain001:
## Problem

`_local_manifest_source` accepts three local bundle sources. Two of them reach YAML through `BundleManifest.from_file` → `load_yaml`, which wraps a parse failure:

```python
try:
    has_node = yaml.compose(text) is not None
    data = yaml.safe_load(text)
except yaml.YAMLError as exc:
    raise BundlerError(f"Invalid YAML in {path}: {exc}") from exc
```

The `.zip` branch parses inline instead, with no guard:

```python
data = _yaml.safe_load(io.BytesIO(raw))
return BundleManifest.from_dict(data)
```

`yaml.YAMLError` derives **directly from `Exception`** — it is neither a `ValueError` nor an `OSError`:

```
yaml.error.YAMLError -> Exception -> BaseException -> object
```

So it slips past `bundle_install`'s `except BundlerError` and surfaces as a raw traceback.

## Reproduction

```bash
python -c "import zipfile; zipfile.ZipFile('bad.zip','w').writestr('bundle.yml','bundle: [unclosed\n  id: x\n')"
specify bundle install ./bad.zip
```

For the *identical* malformed `bundle.yml`, the three local sources disagree:

| Source | Result |
|---|---|
| `./bundle-dir` | `Error: Invalid YAML in bundle.yml: ...` — exit 1 |
| `./bundle.yml` | `Error: Invalid YAML in standalone.yml: ...` — exit 1 |
| `./bundle.zip` | **`yaml.parser.ParserError` traceback** |

## Why this is the local path's bug, not a missing global handler

The remote counterpart of this exact call — `_download_manifest`, which parses a manifest fetched from a catalog — already handles it, and even names the exception explicitly:

```python
except _yaml.YAMLError as exc:
    raise BundlerError(
        f"Downloaded content for bundle '{entry_id}' from {_source_desc} "
        f"is not valid YAML: {exc}"
    ) from exc
```

So the same corrupt manifest is reported cleanly when it arrives over the network but crashes when installed from disk. This fix brings the local branch in line with both its in-function siblings and its remote twin.

## Tests

Two regression tests in `tests/integration/test_bundler_local_install.py`:

- `test_local_zip_wraps_malformed_manifest_yaml` — pins the `BundlerError` contract on the zip branch.
- `test_malformed_manifest_yaml_fails_alike_for_every_local_source` — drives all three local sources through the CLI and asserts they now fail identically. This is the one that encodes the actual invariant; on `main` it fails with `artifact.zip leaked ParserError` while the other two sources pass.

Both fail before the change and pass after. `tests/integration/test_bundler_local_install.py` + `tests/contract/test_bundle_cli.py`: 49 passed. `ruff check` clean.

### DIFF
--- a/src/specify_cli/commands/bundle/__init__.py
+++ b/src/specify_cli/commands/bundle/__init__.py
@@ -771,7 +771,17 @@ def _local_manifest_source(arg: str):
                 error_type=BundlerError,
                 label="bundle manifest",
             )
-        data = _yaml.safe_load(io.BytesIO(raw))
+        try:
+            data = _yaml.safe_load(io.BytesIO(raw))
+        except _yaml.YAMLError as exc:
+            # The sibling directory/bundle.yml branches reach YAML through
+            # load_yaml(), which turns a parse failure into a BundlerError. This
+            # branch parses inline, so without this it raises a raw YAMLError --
+            # neither a ValueError nor an OSError -- which escapes
+            # bundle_install()'s `except BundlerError` as a traceback.
+            raise BundlerError(
+                f"Invalid YAML in bundle.yml inside '{candidate}': {exc}"
+            ) from exc
         return BundleManifest.from_dict(data)
 
     if candidate.name == "bundle.yml" or candidate.suffix in (".yml", ".yaml"):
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4016, opened 2026-08-07
fix(extensions): reject duplicate provides.templates/scripts names

chelsealong:
## Summary

Follow-up to #4012 (#4010), which merged before the last round of Copilot review feedback could be addressed. This carries forward the two outstanding items that are still valid against `main`:

- `provides.templates`/`provides.scripts` entries with a duplicate `name` within the same section are now rejected with a `ValidationError`. The resolver (`PresetResolver._extension_manifest_declared_template`) returns the first entry matching a name, so a later duplicate was silently unreachable while still being exposed by `ExtensionManifest.templates`/`.scripts` — this closes that gap at manifest-validation time instead. Added parametrized regression tests for both `templates` and `scripts`.
- `extensions/EXTENSION-DEVELOPMENT-GUIDE.md`'s `provides` section now distinguishes its own sub-fields (`commands`/`templates`/`scripts`) from `hooks`/`events`, which are top-level manifest fields read from the manifest root, not nested under `provides` — the prior wording could lead an author to indent them under `provides`, where they wouldn't satisfy validation.

The other two items from that last review round were doc-only / description-only and don't apply here: #4012's PR description has already been corrected in place, and the AGENTS.md acceptance-criterion item was intentionally left alone since that file documents only the AI-agent integration subsystem, not the extension system (which has its own docs, already updated).

## Testing

New parametrized test `test_provides_entry_duplicate_name_rejected` in `tests/test_extensions.py::TestExtensionManifestTemplatesAndScripts` covers both `templates` and `scripts`. Confirmed it fails with `DID NOT RAISE ValidationError` against the pre-fix source (`git checkout HEAD~1 -- src/specify_cli/extensions/__init__.py`), then passes after restoring the fix.

```
$ .venv/bin/python -m pytest tests/test_extensions.py tests/test_presets.py -q
====================== 1110 passed, 2 warnings in 11.30s ======================

All checks passed!
```

## AI Disclosure

Implemented autonomously by Claude Code (model: Claude Sonnet 5) under human direction, in response to outstanding automated review feedback on #4012 that had not been incorporated before that PR merged.

### DIFF
--- a/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
+++ b/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
@@ -177,11 +177,16 @@ Compatibility requirements.
 
 What the extension provides.
 
-**Optional sub-fields** (at least one of `commands`, `templates`, `scripts`, `hooks`, or `events` is required):
+**Optional sub-fields:**
 
 - `commands`: Array of command objects
 - `templates`: Array of template objects
 - `scripts`: Array of script objects
+
+`hooks` and `events` are separate top-level manifest fields (siblings of
+`provides`, not nested under it — see [`hooks`](#hooks) below). At least one
+of `provides.commands`, `provides.templates`, `provides.scripts`, `hooks`, or
+`events` is required.
 
 **Command object**:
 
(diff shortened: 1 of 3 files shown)


mnriem:
Thank you!

PR #4030, opened 2026-08-10
Fix bug-test Python dependency provisioning

### LINKED ISSUE #3997: [Bug]: bug-test agentic workflow cannot run tests — no Python/uv provisioning and PyPI blocked by egress firewall
## Bug Description

The `bug-test` agentic workflow (stage 3 of the assess → fix → test bug pipeline) cannot execute this repository's test suite. Runs come back **inconclusive** with an environment/setup failure rather than a real pass/fail, because the workflow environment can neither find `pytest`/`uv` preinstalled nor install them (PyPI is blocked by the egress firewall).

This is a CI/workflow-infrastructure bug, not a `specify` CLI bug. It was surfaced on issue #3990: the `bug-test` run ([actions run 31021548784](https://github.com/github/spec-kit/actions/runs/31021548784)) for PR #3995 reported:

> **Setup failure (blocking):** `pytest` is not installed in the CI runner. All pip install attempts failed with `403 Forbidden` (PyPI access blocked by network policy). The venv was created successfully but dependency installation was not possible.

Net effect: the fix could only be verified by code inspection, and the pipeline's third stage provides no actual test signal for any Python change.

## Steps to Reproduce

1. Label a Python-fix issue `bug-test` so `.github/workflows/bug-test.lock.yml` runs (e.g. issue #3990 / PR #3995).
2. Observe the agent follow `bug-test.md` Step 3, which instructs it to run `uv sync --extra test` then `uv run pytest` for uv-based Python projects.
3. The install fails: `pytest`/`uv` are not provisioned and outbound requests to PyPI return `403 Forbidden`.
4. The workflow posts a `⚠️ inconclusive` report and applies `tests-inconclusive`.

## Expected Behavior

The `bug-test` workflow provisions the Python test toolchain and runs `pytest`, producing a real passing/failing result for Python fixes.

## Actual Behavior

Tests never run. The result is always `inconclusive` for Python changes because of two compounding gaps in `.github/workflows/bug-test.lock.yml` (and its source `.github/workflows/bug-test.md`):

1. **No Python/test toolchain is provisioned.** The job runs `actions/setup-node` and installs the Copilot CLI, but there is **no `actions/setup-python` / `astral-sh/setup-uv` step and no `uv sync` / `pip install`** before the agent starts. `pytest` and `uv` are listed as allowed tools but are not actually present. spec-kit requires them — `pytest` is under the `test` optional-dependencies extra in `pyproject.toml`.

2. **The egress firewall blocks PyPI.** The gh-aw squid firewall `allowDomains` list (the `GH_AW_ALLOWED_DOMAINS` / `awf-config.json` `network.allowDomains` in the lock file) permits `registry.npmjs.org`, the Ubuntu/apt mirrors, `github.com`, Google APIs, etc., but **does not include `pypi.org` or `files.pythonhosted.org`.** So the agent's `uv sync --extra test` / `pip install` attempts return 403. Notably JS's npm registry is allowed while Python's PyPI is not — the environment appears tuned for JS ecosystems and was never fully wired for this Python repo.

The `bug-test.md` prompt itself (Step 3) tells the agent to run `uv sync --extra test`, a command guaranteed to fail behind this allowlist.

## Proposed Fix

Either (or both) of:

- **Preinstall before the agent stage (preferred/most robust):** add `actions/setup-python` + `astral-sh/setup-uv` and run `uv sync --extra test` (or `pip install -e .[test]`) as a normal workflow step while the network is open, so the venv is populated and the agent can run `uv run pytest` offline.
- **Allowlist PyPI:** add `pypi.org` and `files.pythonhosted.org` to the firewall `allowDomains` so the agent's own `uv sync` / `pip install` succeeds at agent time.

Since these workflows are compiled, the change belongs in the source `.github/workflows/bug-test.md` (and the shared gh-aw config that produces the `allowDomains` list), followed by recompiling the `.lock.yml`. The same gap likely affects `bug-fix` and any other Python-executing agentic workflows and should be fixed consistently.

## Additional Context

- Affected files: `.github/workflows/bug-test.md`, `.github/workflows/bug-test.lock.yml`.
- Evidence run: https://github.com/github/spec-kit/actions/runs/31021548784 (comment on #3990).
- The `bug-test` agent behaved correctly — it located the fix, fell back to code inspection, and honestly reported `inconclusive` rather than fabricating a result. The defect is purely in the workflow's environment setup.

WOLIKIMCHENG:
Closes #3997.

## Summary

- Provision uv and Python before the `bug-test` agent runs.
- Install repository test dependencies with `uv sync --extra test`.
- Allow PyPI package downloads through the gh-aw firewall so the agent can refresh dependencies after checking out a fix ref.
- Add static coverage for the source workflow and compiled lock.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 .venv/bin/python -m pytest -p no:cacheprovider tests/test_github_workflows.py -q`
- `PYTHONDONTWRITEBYTECODE=1 .venv/bin/python -m pytest -p no:cacheprovider tests/test_security_workflow.py::TestDependencyAuditWorkflow::test_setup_python_pin_matches_repo_standard tests/test_security_workflow.py::TestDependencyAuditWorkflow::test_setup_uv_pin_matches_repo_standard -q`
- `uvx ruff@0.15.0 check tests/test_github_workflows.py`
- `git diff --check`
- `/tmp/gh-aw-0.79.8.sGWKqX/linux-amd64 compile bug-test --no-emit --approve`


mnriem:
Pushed `62899020` — syncs gh-aw action-pin **metadata** to the latest pins across all workflows so this PR stays in lockstep.

**Why:** Dependabot already keeps the executing `uses:` pins current and uniform across all 6 workflows (checkout v7.0.1, setup-node v7.0.0, setup-python v7.0.0, setup-uv v9.0.0), but it does not touch gh-aw's own metadata — the per-file `gh-aw-manifest` blob and the shared `.github/aw/actions-lock.json` cache. Those lagged at checkout v6.0.3 / setup-node v6.4.0 in the files not recently recompiled. Since no CI recompiles the locks, a plain `gh aw compile` reads the stale cache and can silently downgrade the `uses:` lines back, undoing Dependabot's bumps and breaking lockstep.

**Change (metadata only, no workflow-body changes, `uses:` pins untouched):**
- `.github/aw/actions-lock.json`: checkout v6.0.3→v7.0.1, setup-node v6.4.0→v7.0.0, and added the setup-python v7.0.0 + setup-uv v9.0.0 entries this PR introduces via bug-test.
- `gh-aw-manifest` blobs in the 5 non-bug-test lock files: checkout + setup-node bumped to match their own `uses:` lines (bug-test was already current).

Verified: no stale SHAs remain; all 6 manifests uniform at latest; `actions-lock.json` valid JSON; each lock file changed exactly one metadata line; workflow tests pass (incl. the full-SHA pin-enforcement test).

---
Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8). This commit was agent-generated under maintainer supervision.

mnriem:
Thank you!

PR #4045, opened 2026-08-11
fix(claude): make argument-hint injection fold-aware for long descriptions

### LINKED ISSUE #4044: [Bug]: #3991 still reproduces on 0.16.2 — Claude ARGUMENT_HINTS fallback injects into a folded description scalar (wrap presets without explicit argument-hint)
## Bug Description

#3991 (preset wrap composition drops core `argument-hint` and leaks its value into `description`) was closed by #3996, but on 0.16.2 the corruption still reproduces for the configuration that motivated the original report: **Claude integration + a `strategy: wrap` preset whose `description` is long enough for the YAML dumper to fold it across multiple lines**.

Two facts combine:

1. **The #3996 inheritance fix is a no-op for bundled core commands.** #3996 adds `argument-hint` to the wrap-composition inheritance allowlist (copied from `core_frontmatter` when the preset frontmatter doesn't declare it). But the bundled core command templates (e.g. `core_pack/commands/specify.md`) declare no `argument-hint` frontmatter at all — for Claude the hints live in the hardcoded `ARGUMENT_HINTS` dict in `src/specify_cli/integrations/claude/__init__.py`. The docstring of `agents.py::apply_argument_hint` states this explicitly: *"Built-in templates carry no `argument-hint`, so this is a no-op for the core path."* So there is nothing to inherit, and the composed frontmatter reaches serialization without the key.
2. **The Claude `ARGUMENT_HINTS` fallback then injects by raw string insertion into already-serialized YAML.** `ClaudeIntegration.post_process_skill_content` → `inject_argument_hint` inserts `argument-hint: "…"` directly after the first line starting with `description:`. With a short one-line description this happens to work (which is presumably why #3996's regression test passes). With a description long enough that `dump_frontmatter` folds it into a multi-line scalar, the injected line lands **inside the folded scalar**.

Depending on whether the dumper quoted the scalar, the failure takes one of two forms (both reproduced on 0.16.2):

- **Plain (unquoted) folded scalar → the frontmatter becomes invalid YAML** (`yaml.parser.ParserError: expected <block end>, but found '<scalar>'`).
- **Quoted folded scalar** (e.g. the description contains `#`, forcing single quotes) **→ the YAML stays valid but the `argument-hint` key does not exist and its line is absorbed into the `description` string** — the same silent corruption as #3991.

## Steps to Reproduce

1. Fresh project on 0.16.2:

   ```bash
   mkdir demo && cd demo && git init -q && git commit -q --allow-empty -m init
   specify init --here --integration claude --script sh --ignore-agent-tools --force
   ```

2. Create a wrap preset whose command frontmatter declares only a **long** `description` (~150+ chars, so it folds when dumped) and `strategy: wrap` — no `argument-hint`:

   `presets-src/demo-wrap/commands/speckit.specify.md`

   ```markdown
   ---
   description: "Create or update the feature specification from a natural language feature description. Also accepts an issue URL resolved via gh CLI (demo customization)."
   strategy: wrap
   ---

   <!-- demo preamble -->

   {CORE_TEMPLATE}
   ```

   plus a matching `preset.yml` with `provides.templates: [{type: command, name: speckit.specify, file: commands/speckit.specify.md, strategy: wrap}]`.

3. Apply and inspect:

   ```bash
   specify preset add --dev ./presets-src/demo-wrap
   head -8 .claude/skills/speckit-specify/SKILL.md
   python3 -c "import yaml; print(yaml.safe_load(open('.claude/skills/speckit-specify/SKILL.md').read().split('---')[1]))"
   ```

## Expected Behavior

The composed frontmatter contains the preset's `description` exactly as declared, plus `argument-hint: Describe the feature you want to specify` as its own key (from the Claude `ARGUMENT_HINTS` mapping).

## Actual Behavior

With the description above (plain scalar), the composed frontmatter is **invalid YAML**:

```yaml
name: speckit-specify
description: Create or update the feature specification from a natural language feature
argument-hint: "Describe the feature you want to specify"
  description. Also accepts an issue URL resolved via gh CLI (demo customization).
```

```
yaml.parser.ParserError: while parsing a block mapping
  ...
expected <block end>, but found '<scalar>'
```

If the description contains a character that forces quoting (e.g. change it to `... Also accepts a GitHub issue/PR URL or #N reference resolved via gh CLI (demo).`), the output parses but is silently corrupted:

```yaml
description: 'Create or update the feature specification from a natural language feature
argument-hint: "Describe the feature you want to specify"
  description. Also accepts a GitHub issue/PR URL or #N reference resolved via gh
  CLI (demo).'
```

Parsed result: no `argument-hint` key; the hint line is absorbed into the `description` string.

## Workaround

Declare the hint explicitly in the preset command frontmatter (`argument-hint: "Describe the feature you want to specify"`). It is then carried structurally through `apply_argument_hint` before serialization, and `inject_argument_hint` skips injection (key already present).

## Specify CLI Version

0.16.2

## AI Agent

Claude Code

## Operating System

macOS 15 (Apple Silicon, Darwin 25.5.0)

## Python Version

3.14

## Additional Context

- Related history: #2903 / PR #2916, PR #2978, #3991 / PR #3996, #3391.
- The automated bug-test run on PR #3996 reported *inconclusive*, consistent with the regression test passing via the `ARGUMENT_HINTS` fallback rather than the inheritance path it was meant to guard.
- Suggested fix: make the Claude `ARGUMENT_HINTS` fallback add the key to the frontmatter dict **before** YAML serialization (the same structured path as `apply_argument_hint`), instead of post-serialization line insertion — or make `inject_argument_hint` fold-aware. A regression test should compose against the actual bundled core template with a fold-length description.

chelsealong:
## Summary

Fixes #4044 — `ClaudeIntegration.inject_argument_hint` spliced `argument-hint: "..."` in as a raw text line immediately after the *first* line starting with `description:`. That's safe for a short, single-line description, but when a description is long enough for the YAML dumper to fold it across multiple indented continuation lines, the splice landed **inside** that scalar instead of after it:

- Plain (unquoted) folded scalar → invalid YAML (`yaml.parser.ParserError`).
- Quoted folded scalar → parses, but `argument-hint` doesn't exist as a key — its line is silently absorbed into the `description` string.

This is the case #3996 didn't cover: bundled core commands carry no `argument-hint` in their source frontmatter, so `CommandRegistrar.apply_argument_hint`'s structural inheritance path is a no-op for them, and this raw-text fallback in `post_process_skill_content` is what actually runs for every core Claude skill.

## Fix

`inject_argument_hint` now skips past every continuation line of the `description` scalar (anything more indented than the `description:` key itself — this covers both plain and quoted folded styles) before inserting `argument-hint:`, so the new key always lands after the whole scalar ends rather than in the middle of it. No change to the injected line's format (`argument-hint: "<hint>"`), so short, non-folding descriptions behave exactly as before.

## Test plan

Added two regression tests in `tests/integrations/test_integration_claude.py::TestClaudeArgumentHints` that build frontmatter with a ~150+ char description via `yaml.safe_dump` (reproducing the fold from the issue) for both the plain and forced-quoted cases, run it through `inject_argument_hint`, and assert the result is valid YAML with `argument-hint` present as its own key and `description` unchanged.

Confirmed both new tests fail against the pre-fix code with exactly the two failure modes from the issue:

```
FAILED test_inject_argument_hint_survives_folded_description
  yaml.parser.ParserError: while parsing a block mapping
  ... expected <block end>, but found '<scalar>'

FAILED test_inject_argument_hint_survives_quoted_folded_description
  KeyError: 'argument-hint'
```

With the fix:

```
tests/integrations/test_integration_claude.py .......................... [ 61%]
................                                                         [100%]
============================== 42 passed in 1.35s ==============================
```

Full suite (`pytest tests -q`): 6786 passed, 9 skipped, 10 failed — all 10 failures reproduce identically on unmodified `main` (composed-template python/shell parity and a rich-markup width assertion, unrelated to this change and to the Claude integration).

`ruff check` on both changed files reports the same 17 pre-existing findings as on unmodified `main`; none introduced by this diff.

## AI assistance disclosure

This PR was written by an autonomous AI coding agent (Claude, via an agent harness), including the diagnosis, fix, and regression tests. It was validated by running the reproduction from the issue, the existing and new test suites, and by confirming pre-existing failures reproduce unmodified on `main`.

### DIFF
--- a/src/specify_cli/integrations/claude/__init__.py
+++ b/src/specify_cli/integrations/claude/__init__.py
@@ -70,11 +70,13 @@ class ClaudeIntegration(SkillsIntegration):
         """Insert ``argument-hint`` after the ``description:`` scalar in YAML frontmatter.
 
         A long ``description`` gets folded by the YAML dumper across
-        indented continuation lines (plain or quoted). Inserting the new
-        line right after the *first* line of that scalar — instead of after
-        the whole scalar — either produces invalid YAML or gets silently
-        absorbed into the description string (#4044), so every continuation
-        line (anything more indented than the key itself) is skipped first.
+        indented continuation lines (plain or quoted), and an embedded
+        paragraph break can add unindented blank lines inside a quoted
+        scalar. Inserting the new line right after the *first* line of
+        that scalar — instead of after the whole scalar — either produces
+        invalid YAML or gets silently absorbed into the description
+        string (#4044), so every continuation line (indented, or blank)
+        is skipped first.
 
         Skips injection if ``argument-hint:`` already exists in the
         frontmatter to avoid duplicate keys.
@@ -113,7 +115,11 @@ class ClaudeIntegration(SkillsIntegration):
                 i += 1
                 # Skip past folded/quoted continuation lines of the scalar
                 # before inserting, so the new key lands after it ends.
-                while i < n and lines[i][:1] in (" ", "\t"):
+                # Blank lines count too: PyYAML emits unindented blank
+                # lines for embedded "\n\n" inside a quoted scalar.
+                while i < n and (
+                    lines[i][:1] in (" ", "\t") or lines[i].rstrip("\r\n") == ""
+                ):
                     out.append(lines[i])
                     i += 1
                 # Preserve the exact line-ending style (\r\n vs \n)
--- a/src/specify_cli/integrations/claude/__init__.py
+++ b/src/specify_cli/integrations/claude/__init__.py
@@ -67,7 +67,14 @@ class ClaudeIntegration(SkillsIntegration):
 
     @staticmethod
     def inject_argument_hint(content: str, hint: str) -> str:
-        """Insert ``argument-hint`` after the first ``description:`` in YAML frontmatter.
+        """Insert ``argument-hint`` after the ``description:`` scalar in YAML frontmatter.
+
+        A long ``description`` gets folded by the YAML dumper across
+        indented continuation lines (plain or quoted). Inserting the new
+        line right after the *first* line of that scalar — instead of after
+        the whole scalar — either produces invalid YAML or gets silently
+        absorbed into the description string (#4044), so every continuation
+        line (anything more indented than the key itself) is skipped first.
 
         Skips injection if ``argument-hint:`` already exists in the
         frontmatter to avoid duplicate keys.
@@ -90,15 +97,25 @@ class ClaudeIntegration(SkillsIntegration):
         in_fm = False
         dash_count = 0
         injected = False
-        for line in lines:
+        i = 0
+        n = len(lines)
+        while i < n:
+            line = lines[i]
             stripped = line.rstrip("\n\r")
             if stripped == "---":
                 dash_count += 1
                 in_fm = dash_count == 1
                 out.append(line)
+                i += 1
                 continue
             if in_fm and not injected and stripped.startswith("description:"):
                 out.append(line)
+                i += 1
+                # Skip past folded/quoted continuation lines of the scalar
+                # before inserting, so the new key lands after it ends.
+                while i < n and lines[i][:1] in (" ", "\t"):
+                    out.append(lines[i])
+                    i += 1
                 # Preserve the exact line-ending style (\r\n vs \n)
                 if line.endswith("\r\n"):
                     eol = "\r\n"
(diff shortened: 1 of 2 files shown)


chelsealong:
Fixed: `inject_argument_hint` now also treats unindented blank lines as continuation lines, since PyYAML emits those for an embedded `\n\n` inside a quoted `description` scalar. Added `test_inject_argument_hint_survives_multi_paragraph_description`, which reproduces the absorbed-key failure on the prior fix and passes now.

mnriem:
Thank you!

Issue #4065, opened 2026-08-12
[Feature]: Stable, block-allocated identifiers (FR/SC/T/CHK) so edits never force a renumber


ira-at-work:
### Problem Statement

`FR-###`, `SC-###`, `T###` and `CHK###` are **references**, not just labels. Once a spec exists, its identifiers get cited from `plan.md`, `tasks.md` dependency lines, checklist items (`[Spec §FR-001]`), `analyze.md` findings tables, `converge.md` source-refs, the GitHub issue titles created by `/speckit.taskstoissues`, commit messages, and PR review comments.

But the templates and commands tell agents to allocate them **densely and sequentially**:

- `templates/commands/tasks.md:160` — "**Task ID**: Sequential number (T001, T002, T003...) in execution order"
- `templates/commands/checklist.md:258` — "globally incrementing IDs starting at CHK001"
- `templates/checklist-template.md:45` — "Items are numbered sequentially for easy reference"
- `templates/spec-template.md` — `FR-001`…`FR-007`, `SC-001`…`SC-004` run as one dense sequence

Dense + sequential means **any insertion or deletion forces a renumber**, and a renumber silently invalidates every citation elsewhere. This shows up in practice in several ways:

1. **Cross-artifact reference rot.** Insert `FR-004` into an existing spec and every later FR shifts. `checklists/requirements.md`, `plan.md`, and any open PR comment now point at the wrong requirement. Nothing errors — the references just quietly mean something else.
2. **Token cost, twice over.** Adding one task at position 10 of a 50-task file rewrites 40 task lines plus every dependency reference (`depends on T012, T013`) and every summary block. Deleting a task is worse: the renumber-and-re-reference pass is pure waste, and it contaminates the agent's context with a diff that carries no information.
3. **Agents get it wrong at scale.** [#1497](https://github.com/github/spec-kit/issues/1497) reported exactly this: after deleting three tasks from a 100-task file the task lines were renumbered "more or less correctly" but the `By Type: … (T009-T021, T036-T039, T046-T049)` summaries went out of sync and were hard to repair.
4. **It contradicts a rule the repo already has.** `templates/commands/converge.md:77` forbids the agent to "rewrite, renumber, reorder, or delete any existing task", and `:219` says "Never reuse or renumber existing IDs." That invariant is exactly right — but it currently only applies to convergence tasks, while the dense allocation scheme in `tasks.md` / `spec-template.md` / `checklist.md` actively pushes agents in the opposite direction everywhere else.

[...]

Issue #1497, opened 2026-01-20
Hierarchical Task Numbering System


Vamshi-Bantaram:
When LLMs add new tasks to `tasks.md`, they must renumber **all subsequent tasks** in the entire file. This is inefficient:

- **High token consumption**: Adding 1 task at position 10 in a 50-task file requires renumbering 40+ tasks (2000+ tokens)
- **Slow and error-prone**: Sequential numbering (T001, T002, T003...) forces cascading renumbering
- **Maintenance overhead**: Updates require changing references throughout the document

## Proposed Solution

Use **hierarchical task numbering** (e.g., T1.1, T1.2, T1.3) grouped by phase/story. Adding a task only requires renumbering within the same parent group.

### Format Example

**Current (Sequential):**
```markdown
- [ ] T001 Install uv
- [ ] T002 Initialize project
- [ ] T003 Add pytest
- [ ] T004 Create tests directory
- [ ] T005 Create schema parser  ← Adding here requires T005-T043 renumbering
```

**Proposed (Hierarchical):**
```markdown
## Phase 1: Setup
- [ ] T1.1 Install uv
- [ ] T1.2 Initialize project
- [ ] T1.3 Add pytest
- [ ] T1.4 Create tests directory

## Phase 2: Foundational
- [ ] T2.1 Create schema parser  ← Adding T2.1.1 only affects T2.1.x tasks
```

**Adding a new task:**
```markdown
## Phase 1: Setup
- [ ] T1.1 Install uv
- [ ] T1.2 Initialize project
- [ ] T1.2.1 [NEW] Verify Python version  ← Only T1.3+ need renumbering
- [ ] T1.3 Add pytest  ← Renumbered from T1.3
```

miszczu-drako:
yup, that would be great.
Today I asked LLM to remove 3 tasks, and it got really messy, because summaries at the bottom of the tasks.md were hard to update
Let's say initially there were tasks T001-T100, and I asked to remove T61, T71 and T82. 
Renumbering of the existing tasks was done more or less correctly, but summaries were not
```md
### By Type:
- Configuration: 2 tasks (T001-T002)
- Foundation: 6 tasks (T003-T008)
- Test Tasks: 18 tasks (T009-T021, T036-T039, T046-T049)
- Implementation: 16 tasks (T022-T028, T040-T042, T050-T054) - added cache metrics task
- Verification: 17 tasks (T029-T035, T043-T045, T055-T058)
[...]

PR #4066, opened 2026-08-12
feat(templates): allocate FR/SC/T/CHK identifiers sparsely and treat them as permanent

ira-at-work:
Implements #4065. See that issue for the full problem statement and the alternatives that were weighed.

> Per `CONTRIBUTING.md` this touches templates and command prompts, so it needs maintainer agreement before merge. Opening it alongside the issue so the proposal is concrete rather than hypothetical — happy to trim, split, or drop it.

## Description

Spec Kit identifiers are **references, not labels**. `plan.md`, `tasks.md` dependency lines, checklist items (`[Spec §FR-001]`), `analyze.md` tables, `converge.md` source-refs, the GitHub issues created by `/speckit.taskstoissues`, commit messages and PR comments all cite them.

The templates nevertheless allocate them densely and sequentially (`tasks.md:160` — "Sequential number (T001, T002, T003...)"; `checklist.md:258` — "globally incrementing IDs starting at CHK001"). Dense + sequential means every insert or delete forces a renumber, and a renumber silently invalidates every citation elsewhere while burning tokens rewriting lines that carry no new information.

[#1497](https://github.com/github/spec-kit/issues/1497) reported the failure mode concretely — deleting three tasks from a 100-task file renumbered the task lines "more or less correctly" but left the `(T009-T021, T036-T039, T046-T049)` summaries pointing at the wrong work. It was closed as stale, not rejected on the merits.

The repo already has the right invariant: `converge.md:77` forbids the agent to "rewrite, renumber, reorder, or delete any existing task" and `:219` says "Never reuse or renumber existing IDs." It just applies to convergence tasks only, while everything around it models the opposite behaviour. This PR generalises that invariant and gives it the room it needs to hold.

## What changes

**Two levels of spacing, each doing a different job.**

| | rule | what it buys |
|---|---|---|
| **Group** | each FR category / task phase / checklist category starts at the next multiple of **1000** | groups are independent — appending to Phase 1 never touches Phase 2 |
| **Item** | items step by **10** within a group | positional insert: a task between `T1010` and `T1020` becomes `T1015`, IDs stay ascending, nothing after it moves |
| **Delete** | remove the line and stop | no renumber, no re-reference pass, no context churn — gaps are the steady state |

```diff
 ## Phase 1: Setup
-- [ ] T001 Create project structure per implementation plan
-- [ ] T002 Initialize [language] project with [framework] dependencies
-- [ ] T003 [P] Configure linting and formatting tools
+- [ ] T1000 Create project structure per implementation plan
+- [ ] T1010 Initialize [language] project with [framework] dependencies
+- [ ] T1020 [P] Configure linting and formatting tools

 ## Phase 2: Foundational
-- [ ] T004 Setup database schema and migrations framework
+- [ ] T2000 Setup database schema and migrations framework
```

Applied to `FR`, `SC`, `T` and `CHK` alike:

- `templates/spec-template.md` — FR samples grouped by category (`FR-1000`… / `FR-2000`…), SC on the same scheme, both with the rule stated in a template comment
- `templates/tasks-template.md` — sample tasks renumbered per phase; `## Format` documents block, step, insert and delete
- `templates/checklist-template.md` — `CHK1000` / `CHK2000` per category; closing note changes from "numbered sequentially for easy reference" to "stable references — do not renumber"
- `templates/commands/{specify,clarify,tasks,checklist,converge}.md` — generation rules updated to allocate, insert into gaps, and preserve
- `docs/concepts/complex-features.md` — the one example range that would otherwise contradict the templates

**Convergence phases** now open their own block (next unused multiple of 1000) instead of continuing from `T{M+1:03d}`, so a converge run cannot land inside a phase's range.

**One dependent fix.** `taskstoissues.md:67` matched issue titles with `` `\bT\d{3}\b` `` — *exactly* three digits. Given `T1000` the trailing `\b` cannot fall between two digits, so there is no match at all and those tasks are silently neither deduplicated nor created. This is already reachable on `main` ([#3866](https://github.com/github/spec-kit/issues/3866)) because `converge.md`'s `03d` is a floor rather than a cap; four-digit IDs make it unavoidable. Widened to `` `\bT\d{3,}\b` ``, with the contract recorded in `converge.md` so producer and consumer can't drift apart again.

| input | `\bT\d{3}\b` | `\bT\d{3,}\b` |
|---|---|---|
| `T001` | `T001` | `T001` |
| `T1000` | — | `T1000` |
| `T12345` | — | `T12345` |
| `ST1000` | — | — |
| `T1` | — | — |

Note the word boundaries still do their original job: `T100` can never match *inside* `T1000`, because the trailing `\b` forces the whole digit run to be consumed.

## Compatibility

- **ID shape is unchanged** (`T` + digits, `FR-` + digits). Nothing that parses the existing format breaks — unlike the hierarchical `T1.1` scheme proposed in #1497.
- **Existing artifacts are not retroactively renumbered.** Old `T001`-style files keep working; the rule applies to newly generated and newly edited artifacts.
- `tests/hooks/tasks.md` is deliberately left at `T001` for that reason.

## Testing

```
.venv/bin/python -m pytest tests -q -p no:randomly
```

- **with this branch**: `13 failed, 6616 passed, 177 skipped`
- **on clean `main`**: `13 failed, 6616 passed, 177 skipped`

Identical — the 13 are pre-existing failures unrelated to this change (branch-slug generation in `test_timestamp_branches.py` / `test_git_extension.py`, and template-composition parity in `test_resolve_template_python_parity.py` and friends). Verified by stashing the diff and re-running.

`tests/test_specify_template_numbering.py` (top-level step ordinals in `specify.md`) passes — that test governs markdown list numbering, which this PR does not touch.


mnriem:
See comment in #4065

PR #4088, opened 2026-08-12
fix(presets): reject falsy malformed catalog roots on load

marcelsafin:
## Description

`PresetCatalog._load_catalog_config()` used `yaml.safe_load(...) or {}`, so falsy non-mapping documents such as `[]`, `0`, `false`, and an empty string were silently accepted as empty configuration. That hides corrupt catalog files and makes them behave differently from truthy malformed roots.

This preserves the existing empty/null no-op while rejecting every present non-mapping root consistently.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` (6,655 passed, 177 skipped)
- [x] Added regressions for four falsy non-mapping YAML roots

## AI Disclosure

Implemented autonomously by GitHub Copilot CLI (model: GPT-5.6 Sol) under human direction; failing regressions were added before the fix, then the full suite and lint were run locally. Commit includes `Assisted-by` and `Co-authored-by` trailers.

### DIFF
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -4276,11 +4276,13 @@ class PresetCatalog:
         if not config_path.exists():
             return None
         try:
-            data = yaml.safe_load(config_path.read_text(encoding="utf-8")) or {}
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4089, opened 2026-08-12
fix(events): skip unreadable extension manifests

marcelsafin:
## Description

The raw-manifest fallback in `collect_extension_events()` skips malformed YAML and undecodable text, but an `OSError` from reading the same optional `extension.yml` escapes and aborts event collection.

This handles read failures at the existing best-effort boundary, so one inaccessible extension manifest does not block events from the rest of the project.

## Testing

- [x] Added an unreadable-manifest regression using a targeted `Path.read_text` failure

## AI Disclosure

Implemented autonomously by GitHub Copilot CLI (model: GPT-5.6 Sol) under human direction; a failing regression was added before the fix, then the full suite and lint were run locally. Commit includes `Assisted-by` and `Co-authored-by` trailers.

### DIFF
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -1014,7 +1014,7 @@ def collect_extension_events(project_root: Path) -> ResolvedEvents:
                 continue
             try:
                 data = yaml.safe_load(ext_yml.read_text(encoding="utf-8")) or {}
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4091, opened 2026-08-12
fix(bundler): reject non-string manifest list members

marcelsafin:
## Description

Bundle manifest fields declared as string arrays (`tags`, `requires.tools`, and `requires.mcp`) currently coerce every member with `str()`. Invalid values such as integers, booleans, and objects therefore enter the model as plausible strings instead of failing schema validation.

This validates each member and preserves valid strings without coercion.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` (6,654 passed, 177 skipped)
- [x] Added contract regressions for non-string tags, tools, and MCP entries

## AI Disclosure

Implemented autonomously by GitHub Copilot CLI (model: GPT-5.6 Sol) under human direction; failing contract regressions were added before the fix, then the full suite and lint were run locally. Commit includes `Assisted-by` and `Co-authored-by` trailers.


mnriem:
Thank you!

PR #4092, opened 2026-08-12
fix(presets): fail closed on unreadable provenance

marcelsafin:
## Description

`_constitution_provenance_matches_preset()` already treats missing, invalid JSON, and undecodable provenance as a non-match, but an `OSError` while reading the same metadata escapes and can interrupt preset handling.

This fails closed for read errors as well: unreadable provenance never authorizes a preset match.

## Testing

- [x] Added a direct unreadable-provenance regression using a targeted `Path.read_text` failure

## AI Disclosure

Implemented autonomously by GitHub Copilot CLI (model: GPT-5.6 Sol) under human direction; a failing regression was added before the fix, then the full suite and lint were run locally. Commit includes `Assisted-by` and `Co-authored-by` trailers.

### DIFF
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -107,7 +107,7 @@ def _constitution_provenance_matches_preset(
         return False
     try:
         metadata = json.loads(provenance.read_text(encoding="utf-8"))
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4093, opened 2026-08-12
fix(bundler): reject non-string catalog tags

marcelsafin:
## Description

Bundle catalog tags are declared as a list of strings, but `_parse_tags()` currently applies `str()` to every member. Mixed lists such as `["valid", 1]` are silently normalized instead of rejected as invalid catalog data.

This validates every tag member and returns the original strings without coercion.

## Testing

- [x] Added a contract regression for mixed string/non-string tags

## AI Disclosure

Implemented autonomously by GitHub Copilot CLI (model: GPT-5.6 Sol) under human direction; a failing contract regression was added before the fix, then the full suite and lint were run locally. Commit includes `Assisted-by` and `Co-authored-by` trailers.

### DIFF
--- a/src/specify_cli/bundler/models/catalog.py
+++ b/src/specify_cli/bundler/models/catalog.py
@@ -117,7 +117,11 @@ def _parse_tags(value: Any, entry_id: str) -> tuple[str, ...]:
         raise BundlerError(
             f"Catalog entry '{entry_id}': 'tags' must be a list of strings."
         )
-    return tuple(str(t) for t in value)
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
Please resolve conflicts

marcelsafin:
Closing because #4318 merged the same non-string catalog-tag validation and regression coverage first. Posted on behalf of @marcelsafin by GitHub Copilot (model: GPT-5.6 Sol).

PR #4094, opened 2026-08-12
fix(presets): reject malformed catalog roots during mutation

marcelsafin:
## Description

Preset catalog add/remove commands assume the YAML document root is a mapping. A sequence or scalar root therefore raises a raw type error instead of reporting corrupt configuration, and mutation must not rewrite that file.

This validates the root shape before reading or changing catalog entries. Empty and null documents retain their existing empty-config behavior.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` (6,653 passed, 177 skipped)
- [x] Added parameterized add/remove regressions for non-mapping YAML roots

## AI Disclosure

Implemented autonomously by GitHub Copilot CLI (model: GPT-5.6 Sol) under human direction; failing regressions were added before the fix, then the full suite and lint were run locally. Commit includes `Assisted-by` and `Co-authored-by` trailers.

### DIFF
--- a/src/specify_cli/presets/_commands.py
+++ b/src/specify_cli/presets/_commands.py
@@ -757,10 +757,15 @@ def preset_catalog_add(
     # Load existing config
     if config_path.exists():
         try:
-            config = yaml.safe_load(config_path.read_text(encoding="utf-8")) or {}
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4101, opened 2026-08-13
fix(taskstoissues): widen task-ID regex to match IDs longer than 3 digits

ira-at-work:
## Summary
- `/speckit.converge` assigns new task IDs with `T{M+1:03d}` — `:03d` is a zero-pad floor, not a width cap, so once `tasks.md` has more than 999 tasks the generated IDs are four digits or longer (e.g. `T1000`).
- `/speckit.taskstoissues` matches issue titles against `\bT\d{3}\b` to dedupe. That pattern requires exactly three digits followed by a word boundary, which can never fall between two digits — so a title containing `T1000` doesn't match at all, and that task is silently neither deduplicated nor created.
- Widens the pattern to `\bT\d{3,}\b` and updates the accompanying explanation. Word boundaries still prevent `ST001`-style false positives, and the digit run is still consumed in full so `T100` can't match inside `T1000`.

Split out of the discussion on #4065, which noted this bug is already reachable on `main` today independent of that proposal (see also #3866).

## Test plan
- [ ] Manually walk through `/speckit.taskstoissues` against a `tasks.md` with >999 tasks (or a converge-appended ID past `T999`) and confirm the regex now matches and dedupes correctly.
- [ ] Confirm three-digit IDs (`T001`) still match as before.


mnriem (changes requested):

ira-at-work:
I hope this fixes everything.

mnriem (approved):

mnriem:
Thank you!

PR #4102, opened 2026-08-13
feat(presets): preset stacks — named, reusable preset lineups

ira-at-work:
## Summary
- Adds `.specify/preset-stacks.yml` support: named, ordered lists of preset entries (id/priority/optional source), equivalent to a saved sequence of `specify preset add` calls.
- A stack named `default` is applied automatically at `specify init` (no flags needed); `specify init --preset-stack <name>|none` selects a specific stack or opts out. `--preset` and `--preset-stack` are mutually exclusive.
- Adds `specify preset stack list/add/remove` (alongside the existing `install`), mirroring `preset catalog add/remove`'s exact edit pattern — `add`/`remove` only ever touch the config file, never install/uninstall anything.
- Reapplying a stack diffs against its previous state and uninstalls dropped entries, unless another currently-applied stack still lists them. A run with any failing entry defers the uninstall half until the stack applies cleanly, so a transient failure can never remove a working preset.
- A stack entry with no explicit `source` bypasses the discovery-only `install_allowed` gate when resolved through the catalog, since listing a preset in your own stack is itself the trust decision.

## Scope
Everything new lives under `presets/` and `src/specify_cli/presets/`. Two files outside that tree are touched, both required by `--preset-stack` itself and nothing else:

- `src/specify_cli/commands/init.py` (+40): the `--preset-stack` option and its call into `presets.stacks`. Stack selection and output rendering live in `presets/stacks.py` (`select_stack()`, `render_apply_result()`), shared with `specify preset stack install`.
- `src/specify_cli/commands/bundle/__init__.py` (+1): `_run_init()` invokes init's Typer callback with a fully-enumerated kwarg list, bypassing Click's option-default resolution, so every new init parameter has to be listed there too. Without the line, `preset_stack` keeps the raw `typer.Option` sentinel and bundle-driven bootstrap fails on a string comparison.

## Test plan
- [x] `pytest tests/test_preset_stacks.py tests/test_presets.py` — all green (641 passed)
- [x] Full suite (`pytest`) — 6682 passed, 4 pre-existing/unrelated failures (branch-naming short-word retention, confirmed via `git stash` baseline diff, not caused by this change)
- [x] Manually walked all 8 steps of `specs/001-preset-stacks/quickstart.md` against a real scratch project with the built CLI
- [x] Updated `presets/README.md` and `presets/ARCHITECTURE.md`

### DIFF
--- a/presets/ARCHITECTURE.md
+++ b/presets/ARCHITECTURE.md
@@ -144,6 +144,42 @@ flowchart TD
 
 Catalogs are fetched with a 1-hour cache (per-URL, SHA256-hashed cache files). Each catalog entry has a `priority` (for merge ordering) and `install_allowed` flag.
 
+## Preset Stacks
+
+`.specify/preset-stacks.yml` holds named, ordered lists of preset entries (`preset`, `priority`,
+optional `source`). `load_stacks_config()` parses and validates the file (unique stack names,
+`default`/`none` reserved, no duplicate `preset` within a stack); `apply_stack()` then drives the
+same install path as `specify preset add` for each entry, in priority order.
+
+```mermaid
+flowchart TD
+    A["apply_stack(project_root, stack)"] --> B["load stack-state.json\n(prior applied stack, if any)"]
+    B --> C{"entry has explicit source?"}
+    C -- Yes --> D["install_from_directory / install_from_archive"]
+    C -- No --> E["catalog.download_pack(bypass_install_allowed=True)"]
+    E --> D
+    D --> F["record entry result (success/error)"]
+    F --> G{"more entries?"}
+    G -- Yes --> C
+    G -- No --> H["diff this stack's preset IDs vs\nprior stack-state for this name"]
+    H --> I["uninstall presets dropped from the stack,\nunless still listed by another applied stack"]
+    I --> J["write updated stack-state.json"]
+```
+
+Per-entry failures are collected but never abort the run — `apply_stack()` returns a result with
+one entry per attempted install plus the list of removed preset IDs; the caller (CLI command or
+`specify init`) renders success/failure lines and exits non-zero only if at least one entry failed.
+
+`specify init`'s implicit-default behavior and `specify preset stack install <name>` share this
+same `apply_stack()` call — resolving which stack to apply (named, implicit `default`, or none) is
+the only logic that differs between the two entry points.
+
+- **Python**: `load_stacks_config()`, `apply_stack()`, `_resolve_entry_source()` in
+  `src/specify_cli/presets/stacks.py`
+- **CLI**: `specify preset stack list/install/add/remove` in
+  `src/specify_cli/presets/_commands.py`; `--preset-stack` on `specify init` in
+  `src/specify_cli/commands/init.py`
+
 ## Repository Layout
 
 ```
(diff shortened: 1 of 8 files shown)


mnriem:
Please address Copilot feedback and limit the changes to presets as indicated by the PR title and description

ira-at-work:
Thanks — Copilot's six comments are all addressed, and I've clarified the scope question in the PR description.

**Copilot feedback**

Two of them were real defects in the sync logic, now fixed in `presets/stacks.py` with regression tests:

- *Failed entry treated as dropped* — stack membership now comes from `stack.entries`, with `StackEntryResult` recording the attempt's outcome separately, so a transient failure can no longer make a still-listed preset look dropped and uninstall a working installation.
- *Installed manifest ID never checked* — a successful entry is now tracked under the ID its manifest actually declares (what `PresetManager` installs and removes under) and reported as `Preset 'x' installed (as 'y')` when the two differ, so dropping the entry later removes the right preset instead of orphaning it.
- Following from those two: a run with any failing entry now defers uninstalls entirely (`deferred_removals`, surfaced in the CLI output) rather than guessing at attribution, since a failed entry yields no manifest ID and removal is destructive. The next clean apply performs them.

The other four were documentation inaccuracies, all corrected:

- Only `none` is reserved; `default` is an ordinary definable stack name that `init` picks when no `--preset-stack` is given.
- `apply_stack()` installs in listed order; `priority` is the resolver precedence recorded on the install, not an install order.
- The non-zero-exit claim now applies only to `specify preset stack install`. `specify init` treats stack application as best-effort — failures warn and init still exits zero, matching how a failing `--preset` already behaves there. Documented rather than changed, so init keeps one consistent contract for optional preset installs.

**Scope**

Everything new is under `presets/` and `src/specify_cli/presets/`. Two files outside that tree are touched, both required by `--preset-stack` itself:

- `commands/init.py` — the flag and its call into the presets package. I moved stack selection and result rendering into `presets/stacks.py` (`select_stack()`, `render_apply_result()`), shared with `specify preset stack install`; that cut this diff from +83 to +40 lines.
- `commands/bundle/__init__.py` (one line) — `_run_init()` invokes init's Typer callback with a fully-enumerated kwarg list, bypassing Click's option-default resolution, so every new init parameter has to be listed there. Without it, `preset_stack` keeps the raw `typer.Option` sentinel and bundle-driven bootstrap fails on a string comparison. It was a separate commit before, which made it read as an unrelated fix; it's now folded into the `--preset-stack` commit where it belongs.

Happy to split the `init` wiring into a follow-up PR if you'd rather land the engine and CLI verbs on their own — just say the word.

Full suite: 6682 passed, with the same 4 pre-existing branch-naming failures present on `main`.

mnriem:
Apologies, @ira-at-work — I should have raised this earlier in the review rather than after you'd built the whole thing out.

Digging into this, I think `specify bundle install` may already cover the use case here, and I'd like you to evaluate it before we take on a parallel presets-only subsystem:

- **It scaffolds an empty project.** Run in a directory with no `.specify/`, `bundle install` resolves the manifest, picks the integration (from `--integration` or the bundle manifest), runs the compatibility gates, then calls the real `specify init` before installing components. So "empty dir → scaffolded project + curated preset lineup" is one command today.
- **Bundles already carry ordered presets with `priority`, `strategy`, and `source`** (`ComponentRef`), installed through the same preset primitives, with provenance records, refcounted removal, conflict detection, and `bundle update` refresh.
- **Dev/local install works** — `bundle install ./path` accepts a `bundle.yml`, a bundle directory, or a built `.zip`, bypassing the catalog (the analog of `preset add --dev`).

Could you try modeling your target workflow as a bundle and running `specify bundle install` (including on a fresh dir and from a local path), and clarify whether it solves it for you? If it does, we can retire this in favor of the existing path. If it doesn't, please tell us specifically why not so we can figure out the right fix.

*Posted by @mnriem via GitHub Copilot (model: Claude Opus 4.8).*

ira-at-work:
yup, looks like the only part left uncovered is the default bundle option. ok then...

PR #4129, opened 2026-08-14
feat(scripts): add SPECIFY_FEATURE_NO_PERSIST env var to suppress feature.json writes (#4128)

### LINKED ISSUE #4128: [Feature]: Multi-agent isolation protocol — process-level feature context without shared-state races
## Problem Statement

When multiple AI agents run Spec Kit pipelines concurrently within the same repository checkout (e.g., Antigravity subagents, Claude Code sub-agents, or parallel Cursor Composer tabs), they share a single `.specify/feature.json` file as their feature context pointer. This creates a **write-write race condition**:

```
Timeline:
  t0  Agent-A: SPECIFY_FEATURE_DIRECTORY="specs/003-auth" → setup-plan.sh
      → get_feature_paths() persists "specs/003-auth" to feature.json  ✓
  t1  Agent-B: SPECIFY_FEATURE_DIRECTORY="specs/004-perf" → setup-tasks.sh
      → get_feature_paths() persists "specs/004-perf" to feature.json  ← overwrites A's value
  t2  Agent-A: (new process, no env var) → check-prerequisites.sh
      → reads feature.json → resolves "specs/004-perf"  ← WRONG FEATURE
```

The root issue: `get_feature_paths()` in `common.sh` (L191-192) **always persists** `SPECIFY_FEATURE_DIRECTORY` back to `feature.json` unless the caller explicitly passes `--no-persist`. Since most scripts (`setup-plan.sh`, `setup-tasks.sh`) call `get_feature_paths` without `--no-persist`, every agent invocation silently overwrites the shared singleton.

### Impact

- **Silent cross-contamination**: Agent B's plan/tasks get written into Agent A's feature directory (or vice versa) without any error signal.
- **Non-reproducible failures**: The behavior depends on timing — sometimes it works, sometimes it doesn't, making debugging extremely difficult.
- **Blocks multi-agent orchestration**: Features like `/speckit-implement-waves` (#3507), which propose running phases in parallel subagents, cannot work safely without solving this shared-state problem first.

### Real-world reproduction

We encountered this in [TTZip](https://github.com/wittkung/TTZip) (a macOS archive utility, 525+ tests, 28 design patterns) while running Antigravity subagents to parallelize a sorting-bugfix TDD suite alongside a 7z compression optimization. Both agents used `SPECIFY_FEATURE_DIRECTORY` correctly in their own processes, but the persist-on-read side effect in `get_feature_paths` caused each agent to clobber the other's `feature.json` entry on every script call.

---

## Root Cause Analysis

The feature resolution chain in `common.sh` `get_feature_paths()` (L163-231) has a correct **read** priority:

```
1. SPECIFY_FEATURE_DIRECTORY env var  (explicit override)
2. .specify/feature.json              (persisted fallback)
3. Error                              (no context)
```

But it has an **unconditional write side effect** on the env-var branch (L191-192):

```bash
if [[ "$no_persist" != true ]]; then
    _persist_feature_json "$repo_root" "$SPECIFY_FEATURE_DIRECTORY"
fi
```

The `--no-persist` guard (added in #3025) is a function-level parameter, not an environment-level control. Scripts that are "just resolving paths" but don't know they should pass `--no-persist` (like `setup-plan.sh`, `setup-tasks.sh`) trigger the persist unconditionally.

### What already works

Credit to the maintainers — the infrastructure for multi-agent isolation is **already in place**:

| Mechanism | Status | Issue |
|:----------|:-------|:------|
| `SPECIFY_FEATURE_DIRECTORY` env var priority | ✅ Working | — |
| `--no-persist` read-only resolution | ✅ Working | #3025 |
| `CURRENT_BRANCH` fallback from feature dir basename | ✅ Working | #3026 |
| `SPECIFY_INIT_DIR` for monorepo project scoping | ✅ Working | — |
| Parser fallback chain (jq → python3 → grep/sed) | ✅ Working | #3304 |

What's **missing** is the guidance layer: documentation, agent skill instructions, and an environment-level `no-persist` toggle.

---

### 1. Official multi-agent documentation (`docs/multi-agent.md`)

A new document covering:
- The race condition scenario (as above)
- The **Multi-Agent Isolation Protocol**: always inject `SPECIFY_FEATURE_DIRECTORY` per-process, never rely on `feature.json` for read
- Integration-specific examples (Antigravity subagents, Claude Code sub-agents, Cursor multi-tab, CI matrix)
- FAQ: "Do I need git worktrees?" → No, env-var isolation is sufficient for same-checkout concurrency

### 2. Update agent skill templates to stop instructing direct `feature.json` writes

Currently, the `specify` command template (the upstream equivalent of `speckit-specify/SKILL.md`) instructs agents to:

> *Persist the resolved path to `.specify/feature.json`: `{"feature_directory": "<resolved feature dir>"}`*

This instruction should be replaced with:

> *Pass the resolved feature directory to downstream commands via `SPECIFY_FEATURE_DIRECTORY` environment variable prefix. Example: `SPECIFY_FEATURE_DIRECTORY="specs/003-auth" .specify/scripts/bash/setup-plan.sh --json`*

The `feature.json` persistence should remain as an **automatic side effect** of `get_feature_paths()` for single-agent backward compatibility, but agents should not be told to write it directly (which bypasses the script's own idempotency guards in `_persist_feature_json`).

### 3. (Optional) `SPECIFY_NO_PERSIST` environment variable

Add an environment-level equivalent of the `--no-persist` function parameter:

```bash
# In get_feature_paths(), after the --no-persist argument check (L167-171):
if [[ "${SPECIFY_NO_PERSIST:-}" == "1" || "${SPECIFY_NO_PERSIST:-}" == "true" ]]; then
    no_persist=true
fi
```

This allows CI pipelines and agent orchestrators to set `SPECIFY_NO_PERSIST=1` globally, ensuring that no script invocation can accidentally write `feature.json` — even scripts that don't pass `--no-persist` internally.

---

## Backward Compatibility

This proposal is **fully backward compatible**:

| Scenario | Before | After |
|:---------|:-------|:------|
| Single agent, no env var | Reads `feature.json` | Identical behavior |
| Single agent, with env var | Reads env var, persists to `feature.json` | Identical behavior |
| Multi-agent, each sets env var | Race on `feature.json` (bug) | Each agent's reads are short-circuited by env var; persistence is harmless |
| Multi-agent + `SPECIFY_NO_PERSIST=1` | N/A | No `feature.json` writes at all |
| `specify integration upgrade` | Overwrites managed files | `docs/multi-agent.md` is not in manifest; protocol rules live in user-space |

No existing scripts, templates, or workflows change behavior. The persist side effect is still there (it's a "last writer wins" overwrite that's harmless when every reader uses env vars). `SPECIFY_NO_PERSIST` is strictly additive.

---

## Reference Implementation

We've been running this protocol in production at [TTZip](https://github.com/wittkung/TTZip) with Antigravity (Google DeepMind's agentic coding tool) subagents. Our implementation consists of:

1. **Project-level rule file** ([`.agents/rules/speckit-multiagent.md`](https://github.com/wittkung/TTZip/blob/main/.agents/rules/speckit-multiagent.md)): Instructs all agents to inject `SPECIFY_FEATURE_DIRECTORY` per-process and never read/write `feature.json` directly.
2. **Global user rule**: Gates (hard state machine gating) that prevent any agent from writing production code before spec/plan/tasks artifacts exist under the declared feature directory.
3. **Concurrent verification**: Validated that two agents operating on `specs/003-sorting-fix/` and `specs/006-7z-conquest/` simultaneously produce zero cross-contamination.

The protocol adds zero overhead to single-agent workflows and requires no upstream code changes to function — it's purely a documentation and guidance contribution. The optional `SPECIFY_NO_PERSIST` env var is a small, additive improvement to `common.sh`.

---

## Related Issues

- #3507 — `/speckit-implement-waves`: Needs this isolation protocol as a prerequisite for safe parallel phase execution
- #1476 — Git worktree isolation: Our approach is complementary (env-var isolation within a single checkout vs. filesystem isolation across worktrees)
- #3025 — `--no-persist` for read-only resolution: Foundation we build on
- #3026 — `CURRENT_BRANCH` fallback: Foundation we build on
- #752 — Claude Code subagent feature execution: Would benefit from this protocol

## Component

Core scripts (`common.sh`), Documentation, Agent skill templates

chelsealong:
## Problem

Fixes #4128.

When multiple Spec Kit script invocations run concurrently against the same
checkout (e.g. multiple agents/subagents each working on a different
feature), each one typically sets its own `SPECIFY_FEATURE_DIRECTORY`. Most
core scripts (`setup-plan`, `setup-tasks`) call `get_feature_paths()` without
`--no-persist`, so every invocation persists its `SPECIFY_FEATURE_DIRECTORY`
to the shared `.specify/feature.json` — the last writer wins, silently
overwriting another process's pinned feature directory.

`--no-persist` (added for #3025) already exists as a per-call opt-out, but
it's a parameter each call site has to pass — scripts that don't know they
need it (like `setup-plan`/`setup-tasks`) never do.

## Fix

Add `SPECIFY_FEATURE_NO_PERSIST` as the environment-level equivalent of
`--no-persist` / `-NoPersist` / `no_persist=True`. When set to `1` or `true`,
it suppresses the `.specify/feature.json` write in `get_feature_paths()`
regardless of whether the calling script passes the per-call flag. An
orchestrator can set it once for a whole process tree so no script
invocation in that tree can write `feature.json`, even ones that don't know
to opt out themselves.

Implemented identically in all three script variants per the project's
parity rule (`AGENTS.md`):
- `scripts/bash/common.sh` — `get_feature_paths()`
- `scripts/powershell/common.ps1` — `Get-FeaturePathsEnv`
- `scripts/python/common.py` — `get_feature_paths()`

Also documented the new variable in `docs/reference/core.md`'s environment
variables table, next to `SPECIFY_FEATURE_DIRECTORY`.

Fully backward compatible: existing behavior when `SPECIFY_FEATURE_NO_PERSIST` is
unset is unchanged in every scenario.

## Test plan

Added `tests/test_specify_no_persist.py`, covering all three script
variants via `setup-plan` (which calls `get_feature_paths()` without
`--no-persist`, so it's a script that previously had no way to opt out):

- Baseline: `SPECIFY_FEATURE_DIRECTORY` set, no `SPECIFY_FEATURE_NO_PERSIST` →
  `feature.json` is written (existing behavior, unchanged).
- `SPECIFY_FEATURE_NO_PERSIST=1`/`true` → `feature.json` write is suppressed, for
  bash, PowerShell, and Python.
- Two-agent race: agent A persists `specs/001-a`, agent B runs with
  `SPECIFY_FEATURE_DIRECTORY=specs/002-b` and `SPECIFY_FEATURE_NO_PERSIST=1` →
  `feature.json` still points at agent A's `specs/001-a` (the pin isn't
  clobbered).

Confirmed the new tests fail without the fix (`git checkout HEAD~1 --
scripts/bash/common.sh scripts/powershell/common.ps1 scripts/python/common.py`,
then re-ran):

```
tests/test_specify_no_persist.py::test_bash_persists_by_default PASSED
tests/test_specify_no_persist.py::test_bash_specify_no_persist_suppresses_write FAILED
tests/test_specify_no_persist.py::test_bash_specify_no_persist_does_not_clobber_existing_pin FAILED
tests/test_specify_no_persist.py::test_ps_specify_no_persist_suppresses_write FAILED
tests/test_specify_no_persist.py::test_py_specify_no_persist_suppresses_write FAILED
4 failed, 1 passed in 1.27s
```

And pass with the fix restored:

```
tests/test_specify_no_persist.py::test_bash_persists_by_default PASSED
tests/test_specify_no_persist.py::test_bash_specify_no_persist_suppresses_write PASSED
tests/test_specify_no_persist.py::test_bash_specify_no_persist_does_not_clobber_existing_pin PASSED
tests/test_specify_no_persist.py::test_ps_specify_no_persist_suppresses_write PASSED
tests/test_specify_no_persist.py::test_py_specify_no_persist_suppresses_write PASSED
5 passed in 1.28s
```

Also ran:

- `shellcheck --severity=error` on all tracked `.sh` files — clean.
- `markdownlint-cli2` on `docs/reference/core.md` — same 2 pre-existing
  `MD028` findings as on unmodified `main` (unrelated blockquotes further
  down the file, confirmed by diffing lint output before/after).
- Full `pytest tests/ -q` (6944 tests): 6921 passed, 9 skipped, 4 pre-existing
  failures unrelated to this change (`test_presets.py`,
  `test_resolve_template_python_parity.py` — template-composition/locale
  edge cases), confirmed present on `main` before this change too.

No manual agent-driven testing was done for this change since it's a
core-script/docs-only change with no slash-command template modifications.

## AI disclosure

This PR was written primarily by an autonomous AI coding agent (Claude Code,
Claude Sonnet 5), including the code change, tests, and this description.


wittkung:
Issue author here. Reviewed the diff — the implementation across all three language variants (bash/ps/python) matches the proposal in #4128 exactly, and the clobber-prevention test (`test_bash_specify_no_persist_does_not_clobber_existing_pin`) covers the core race scenario we described.

We backported the bash change locally and ran 8 end-to-end scenarios against `get_feature_paths()` directly — all passing:

- Default persist ✓
- `--no-persist` suppresses ✓
- `SPECIFY_NO_PERSIST=1` suppresses ✓
- `SPECIFY_NO_PERSIST=true` suppresses ✓
- Agent-A pin not clobbered by Agent-B with NO_PERSIST ✓
- Path resolution unaffected by NO_PERSIST ✓
- `SPECIFY_NO_PERSIST=0` does not suppress (boundary) ✓

Happy to contribute a follow-up `docs/guides/multi-agent.md` guide once this lands — covering the full isolation protocol with integration examples for Antigravity, Claude Code, Cursor, and CI matrix jobs.

chelsealong:
Fixed: `create-new-feature` (bash/ps/py) wrote `.specify/feature.json` directly via `_persist_feature_json`/`Save-FeatureJson`/`persist_feature_json`, bypassing `get_feature_paths()` entirely, so `SPECIFY_NO_PERSIST` had no effect there. All three variants now check `SPECIFY_NO_PERSIST` before that write, with parity tests added in `tests/test_specify_no_persist.py` (confirmed failing before the fix, passing after). Full `pytest tests/` run: 6937 passed, 9 skipped, 1 pre-existing unrelated failure (verified present without this change too).

mnriem (changes requested):
Please address Copilot feedback and adjust the variable name to SPECIFY_FEATURE_NO_PERSIST

chelsealong:
Renamed `SPECIFY_NO_PERSIST` to `SPECIFY_FEATURE_NO_PERSIST` across bash/PowerShell/Python (`common.*`, `create-new-feature.*`, tests, docs) in 7e92dfb. The Copilot inline comment about `create-new-feature` bypassing the variable was already addressed in bc3341a. `tests/test_specify_no_persist.py` (8 tests) and shellcheck both pass with the new name.

chelsealong:
Updated the PR title and description to use `SPECIFY_FEATURE_NO_PERSIST` throughout (the code/docs were already renamed in 7e92dfb; only the PR metadata still referenced the old `SPECIFY_NO_PERSIST` name). No code changes needed.

mnriem:
Please resolve conflicts

chelsealong:
Merged upstream/main and resolved the conflict in `docs/reference/core.md` (upstream had reworded the `SPECIFY_FEATURE` row; kept upstream's wording and kept our `SPECIFY_FEATURE_NO_PERSIST` row after it). No other files conflicted. `tests/test_specify_no_persist.py` (8 tests) and the full `pytest tests/` (8208 passed, 12 skipped) pass post-merge; shellcheck clean.

mnriem:
The implementation now addresses the original feedback, including the direct `create-new-feature` write paths across Bash, PowerShell, and Python. Before re-review, please update the PR’s test evidence: `tests/test_specify_no_persist.py` now contains eight tests, while the description still reports the earlier five-test run and omits the three `create-new-feature` regression tests. Please include the current eight-test result and then resolve the addressed review threads.


Posted on behalf of @mnriem by GitHub Copilot (model: GPT-5.6 Sol, autonomous); comment fully AI-drafted.

chelsealong:
Updated the PR description's test plan to reflect the current 8 tests in `tests/test_specify_no_persist.py` (including the 3 `create-new-feature` regression tests), with a fresh passing run: 8 passed in 8.41s. Resolved the two addressed review threads (the `create-new-feature` bypass, fixed in bc3341a, and the naming inconsistency, fixed in 7e92dfb).

mnriem (approved):

PR #4135, opened 2026-08-15
fix(bundler): pass explicit option values when delegating to `workflow_add`

jawwad-ali:
## Problem

**`specify bundle install` can never install a workflow component.** It fails 100% of the time, with an error about `--dev` that has nothing to do with what the user asked for.

`bundler/services/primitives.py` delegates to the Typer command callables in-process, as its module docstring describes:

```python
lambda: workflow_add(component.id)
```

But `workflow_add` declares two `typer.Option` parameters (`workflows/_commands.py:1712-1716`):

```python
def workflow_add(
    source: str = typer.Argument(..., help="Workflow ID, URL, or local path"),
    dev: bool = typer.Option(False, "--dev", ...),
    from_url: str | None = typer.Option(None, "--from", ...),
):
```

Called from Python rather than through Typer, an omitted option parameter keeps its **`OptionInfo` sentinel** as the value:

```
dev default      -> OptionInfo   truthy=True
from_url default -> OptionInfo   is None=False
```

So `if dev:` is always true and the function takes the local-path branch.

```
>>> primitive_manager("workflows", project, allow_network=True).install(
...     ComponentRef(kind="workflows", id="code-review", version="1.0.0"))

Error: --dev source must be a workflow YAML file, supported archive, or
directory containing workflow.yml: code-review

BundlerError: Failed to install workflow 'code-review'.
```

The catalog install path is unreachable.

## Why only this call site

A signature survey of all four delegated callables:

| callable | parameters |
|---|---|
| `workflow_add` | `source=ArgumentInfo, dev=OptionInfo, from_url=OptionInfo` |
| `workflow_remove` | `workflow_id=ArgumentInfo` |
| `workflow_step_add` | `step_id=ArgumentInfo` |
| `workflow_step_remove` | `step_id=ArgumentInfo` |

`workflow_add` is the **only** one that declares options, and it is the only call site that omits them — so the other three are safe exactly as written, and this fix is correctly scoped to one line.

## Fix

```python
lambda: workflow_add(component.id, dev=False, from_url=None),
```

**No breaking change.** No input that works today behaves differently — today *every* bundler workflow install fails; afterwards it reaches the catalog path the code always intended.

## Verification

- **Fail-before / pass-after:** the new test fails on unpatched `src` and passes with the fix — **1 failed → 21 passed**.
- The new test asserts the *values*, not just the call: `dev is False` and `from_url is None`, so a future regression to sentinels is caught rather than silently swallowed by a `**kwargs` stub.
- One existing stub (`lambda wid: ...`) had to accept the kwargs; I made it capture them instead of discarding them.
- Scoped regression over `tests/unit`: no new failures vs a clean-`main` baseline captured on `bf88c9f9`.

---

### DIFF
--- a/src/specify_cli/bundler/services/primitives.py
+++ b/src/specify_cli/bundler/services/primitives.py
@@ -337,7 +337,16 @@ class _WorkflowKindManager:
         with _chdir(self._root):
             _delegate_command(
                 "install", f"workflow '{component.id}'",
-                lambda: workflow_add(component.id),
+                # Pass the options explicitly. Called in-process rather than
+                # through Typer, an omitted ``typer.Option`` parameter keeps its
+                # ``OptionInfo`` sentinel as the value -- which is TRUTHY and is
+                # not ``None`` -- so ``workflow_add``'s ``if dev:`` took the
+                # local-path branch for every catalog install and failed with
+                # "--dev source must be a workflow YAML file ...". It is the only
+                # one of the four delegated commands that declares options;
+                # workflow_remove / workflow_step_add / workflow_step_remove take
+                # a bare ``typer.Argument`` and are safe as written.
+                lambda: workflow_add(component.id, dev=False, from_url=None),
             )
 
     def refresh(self, component: ComponentRef) -> None:
--- a/src/specify_cli/bundler/services/primitives.py
+++ b/src/specify_cli/bundler/services/primitives.py
@@ -347,6 +347,15 @@ class _WorkflowKindManager:
         with _chdir(self._root):
             _delegate_command(
                 "install", f"workflow '{component.id}'",
+                # Pass the options explicitly. Called in-process rather than
+                # through Typer, an omitted ``typer.Option`` parameter keeps its
+                # ``OptionInfo`` sentinel as the value -- which is TRUTHY and is
+                # not ``None`` -- so ``workflow_add``'s ``if dev:`` took the
+                # local-path branch for every catalog install and failed with
+                # "--dev source must be a workflow YAML file ...". It is the only
+                # one of the four delegated commands that declares options;
+                # workflow_remove / workflow_step_add / workflow_step_remove take
+                # a bare ``typer.Argument`` and are safe as written.
                 lambda: workflow_add(component.id, dev=False, from_url=None),
             )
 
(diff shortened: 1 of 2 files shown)


jawwad-ali:
Rebased onto current `main` — the conflict is resolved and this is `MERGEABLE` again.

**In fairness, you should know this is now largely superseded.** While it sat, #4284 (041faeea, merged Sep 1) landed the identical source fix:

```python
lambda: workflow_add(component.id, dev=False, from_url=None),
```

and folded the same regression assertion into the existing test:

```python
assert calls == [("bundled-wf", False, None)]
```

That is exactly why this branch conflicted — `main` had already grown my change by another route.

Resolving the conflict, I took **main's side wherever it was equal or better**: its `test_offline_workflow_allows_bundled` captures `(wid, dev, from_url)` as a tuple, which is a stronger assertion than my original `**kwargs`-swallowing version, so I dropped mine in favour of it.

What remains unique here is small and entirely additive:

1. An explanatory comment at the call site recording *why* the explicit values are required — that outside Typer an omitted `typer.Option` keeps its `OptionInfo` sentinel, which is **truthy and not `None`**, so `workflow_add`'s `if dev:` took the local-path branch for every catalog install. It also notes that `workflow_add` is the only one of the four delegated commands declaring options; `workflow_remove` / `workflow_step_add` / `workflow_step_remove` take a bare `typer.Argument` and are safe as written.
2. `test_workflow_install_passes_explicit_typer_options`, a dedicated test naming the sentinel trap. Its coverage **overlaps** what #4284 already asserts, so it is a documentation-grade duplicate rather than new protection.

Entirely your call whether that is worth carrying — I'm happy to close this if you would rather not review a near-duplicate, and I won't be offended either way. Verified after the rebase: 26 passed in `tests/unit/test_bundler_primitives.py`, `uvx ruff@0.15.0 check src tests` clean.

mnriem:
Thanks @jawwad-ali for explicitly pointing out that this was superseded and for preserving the existing coverage during the rebase.


Confirmed: #4284 landed the same fix, and current `main` already checks the explicit option values. The original report was valid; it was resolved through that other PR.


Rather than carry overlapping coverage, I’m closing this as superseded. Thank you for the investigation and the clear explanation of what remained.

*Drafted for @mnriem  by GitHub Copilot (model: GPT-6 Astra).*

PR #4136, opened 2026-08-15
fix(bundler): treat an explicit-null records field as missing, not the text `"None"`

jawwad-ali:
## Problem

`InstalledBundleRecord.from_dict` and `_component_from_dict` read the required fields with:

```python
str(data.get("bundle_id", "")).strip()
```

The `""` default covers only a **missing** key. A key that is *present but null* — how a hand-edited or corrupt `.specify/bundle-records.json` spells an empty field — yields `None`, and `str(None)` is the literal `"None"`. That text is non-empty, so it sails straight past the `if not bundle_id:` / `if not version:` / `if not cid:` guards that exist precisely to reject such a file.

```
bundle_id=null -> ACCEPTED: bundle_id='None' version='1.0.0'
version=null   -> ACCEPTED: bundle_id='a'    version='None'
id=null        -> ACCEPTED: components=[('presets', 'None')]
```

The record is accepted as a bundle literally **named `None` at version `None`**, contributing a component whose id is `None`.

That phantom is not inert: it feeds the collateral-protection refcount, so `components_still_needed()` reports `('presets', 'None')` as protected, and `install_bundle()`'s `other_tracked` set will attribute real components to it.

## Fix

Reuse `_text` from the sibling parser in the same package. Its docstring describes this exact trap:

> A `.get(key, "")` default only covers a *missing* key. A key that is present but null — how YAML spells an empty field (`author:` with nothing after it) — yields `None`, and `str(None)` is the literal `"None"`. That text is non-empty, so it sailed past the `if not value` required-field checks…

This is the second half of a hardening sweep that was already applied here in part: the falsy-non-list guards landed in this file as **#3666**, and the explicit-null fix landed for `bundle.yml` as **#3798** — but was never mirrored into `records.py`. Using the shared helper keeps the two parsers from drifting again.

**No breaking change.** Only null-valued inputs behave differently; every string input is byte-identical. `installed_at` is optional, so its null correctly becomes `""` rather than an error — pinned by a test.

## Verification

- **Fail-before / pass-after:** 4 new-vs-baseline failures with the source reverted, all passing with the fix — **5 failed → 1 failed** (that one is a pre-existing Windows symlink test).
- Normal records still load unchanged: `bundle_id='demo' version='1.0.0' comps=[('presets','p1')]`.
- Scoped regression over `tests/unit`: no new failures vs a clean-`main` baseline captured on `bf88c9f9`.

Tests sit beside the existing `test_load_records_rejects_record_missing_bundle_id` / `_missing_version` / `_rejects_component_missing_id`, which cover the missing-key half of the same contract.

---

### DIFF
--- a/src/specify_cli/bundler/models/records.py
+++ b/src/specify_cli/bundler/models/records.py
@@ -13,7 +13,7 @@ from typing import Any
 
 from .. import BundlerError
 from ..lib.yamlio import dump_json, ensure_within, load_json
-from .manifest import COMPONENT_KINDS, ComponentRef
+from .manifest import COMPONENT_KINDS, ComponentRef, _text
 
 RECORDS_FILENAME = "bundle-records.json"
 RECORDS_SCHEMA_VERSION = "1.0"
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4143, opened 2026-08-15
fix(workflows): strip the resolved value before matching switch cases

jawwad-ali:
## Problem

`SwitchStep.execute` matches with:

```python
str_value = str(value) if value is not None else ""
```

No strip. But the values a switch dispatches on are overwhelmingly **captured command output**, and `ShellStep` stores `proc.stdout` verbatim (`shell/__init__.py:67`). So `run: echo approve` resolves to `"approve\n"`, which matches no `approve:` case key.

```
stdout stored : 'approve\n'
matched_case  : '__default__'      <-- silently wrong
next steps    : ['fallback']
```

The switch silently falls through to `default:` — or, with no default, completes having dispatched nothing — while still reporting **COMPLETED**.

A workflow author cannot work around it: the registered filters are `default` / `join` / `map` / `contains` / `from_json`. There is no `trim`.

## This is already treated as a bug everywhere else

spec-kit strips before matching a resolved string against declared literals in every comparable site:

| Site | Code |
|---|---|
| `evaluate_condition` | `result.strip().lower()` — fixed for this exact shell-newline reason |
| `InitStep._resolve_bool` | `resolved.strip().lower()` |
| `workflows/catalog.py` | `raw_install.strip().lower() in (...)` |

Switch **case keys are such declared literals**, and this was the only site not stripping.

## Fix

One line, plus the class docstring which said "exact match, string-coerced" and is now accurate.

**Narrow.** `expression_value` still reports the **raw** value, so nothing downstream loses information — pinned by an assertion. A genuine mismatch still falls through: a second test checks `"approve-later\n"` still routes to `default`.

The only inputs whose behaviour changes are ones that previously took the wrong branch.

## Verification

- **Fail-before / pass-after:** 2 new-vs-baseline failures with the source reverted; all pass with the fix. Parametrized over `"approve\n"`, `"approve\r\n"`, `"  approve  "` and `"approve"`.
- Tests go into the **existing** `TestSwitchStep` class — several open PRs touch this file, and a duplicate class name silently shadows.
- Scoped regression: no new failures vs a clean-`main` baseline captured on `bf88c9f9` (21 pre-existing in scope).

One note on the run: an unrelated `test_resume_rejects_malformed_run_state_origin_fields` case failed once during verification. I checked rather than assumed — it fails on a *different* parametrization each time and passes 4/4 in isolation both with and without this change, so it is a pre-existing Windows flake (`os.replace PermissionError`). A clean re-run of the full gate passed.

---

### DIFF
--- a/src/specify_cli/workflows/steps/switch/__init__.py
+++ b/src/specify_cli/workflows/steps/switch/__init__.py
@@ -12,7 +12,8 @@ class SwitchStep(StepBase):
     """Multi-branch dispatch on an expression.
 
     Evaluates ``expression:`` once, matches against ``cases:`` keys
-    (exact match, string-coerced).  Falls through to ``default:`` if
+    (exact match; the resolved value is string-coerced and stripped of
+    surrounding whitespace first).  Falls through to ``default:`` if
     no case matches.
     """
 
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4150, opened 2026-08-15
fix(cline): stop unrelated prose from suppressing the hook command note

jawwad-ali:
## Problem

`ClineIntegration._inject_hook_command_note` guards idempotency with a **whole-document substring scan**:

```python
if "replace dots" in content:
    return content
```

The shared `SkillsIntegration._inject_hook_command_note` helper (`base.py:1637-1642`) instead compares only the line immediately above each match:

```python
previous_lines = content[:m.start()].splitlines()
if previous_lines and previous_lines[-1] == indent + note:
    return m.group(0)
```

Cline never picked up that change.

```
A. unrelated prose -> note injected? False      (base: True)
B. two hook sections, one already noted -> notes: 1   (want 2)
C. indented instruction after a blank line:
   '## Hooks\n\n  <note>\n\n  - For each executable hook, ...'   <-- spurious blank line
```

**(a)** Any command or extension markdown whose prose happens to contain the phrase *"replace dots"* — e.g. `"When normalizing table names, replace dots with underscores."` — loses the note entirely. The generated workflow then tells the agent to emit `/speckit.git.commit`, a **dotted command Cline never registers** (Cline installs `speckit-git-commit.md`). The note exists precisely to prevent that.

**(b)** A document with one already-noted hook section never gets a note on a second, un-noted one — exactly the case the base helper was changed to handle.

**(c)** Secondary: cline's regex captures indentation with `^(\s*)` where the base uses `^([ \t]*)`. Because `\s` matches newlines, the captured "indent" can swallow a preceding blank line, which is then re-emitted **between** the note and the instruction. The per-instruction check cannot line up until this is fixed, so both belong in one change.

## Fix

Adopt the base helper's per-instruction check and its `[ \t]*` capture. After the fix:

```
A. unrelated prose -> note injected? True
B. two sections -> notes: 2 (want 2)
C. indented+blank -> note sits directly above the instruction, indentation preserved
D. idempotent -> True
```

## Verification

- **Fail-before / pass-after:** 3 new-vs-baseline failures with the source reverted → **35 passed** with the fix.
- Idempotency is asserted explicitly (re-running on already-noted content is a no-op) — the property the old guard was there to provide is kept.
- Full `tests/integrations` scope: **18 failed, 2627 passed** — identical to the clean-`main` baseline's 18 (all Windows symlink-privilege), plus my 3 new tests.

**No breaking change.** Content that already has the note above every instruction is returned byte-identical; only documents that were previously *missing* a note now gain one.

---

### DIFF
--- a/src/specify_cli/integrations/cline/__init__.py
+++ b/src/specify_cli/integrations/cline/__init__.py
@@ -101,14 +101,25 @@ class ClineIntegration(MarkdownIntegration):
 
         Targets the line ``- For each executable hook, output the following``
         and inserts the note on the line before it, matching its indentation.
-        Skips if the note is already present.
+        Skips individual instructions that already have the note immediately
+        above them.
         """
-        if "replace dots" in content:
-            return content
+        note = _HOOK_COMMAND_NOTE.rstrip("\n")
 
         def repl(m: re.Match[str]) -> str:
             indent = m.group(1)
             instruction = m.group(2)
+            # Check the line immediately above this instruction, mirroring the
+            # shared ``SkillsIntegration`` helper. The previous whole-document
+            # ``if "replace dots" in content`` scan meant any command or
+            # extension markdown whose prose merely contained that phrase lost
+            # the note entirely -- leaving the agent told to emit a dotted
+            # ``/speckit.git.commit``, which Cline never registers -- and a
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4155, opened 2026-08-16
fix(console): fail fast in select_with_arrows when stdin is not a TTY

### LINKED ISSUE #4152: specify init hangs on select_with_arrows prompts in non-interactive environments — no flag reaches them, --force does not bypass them
## Summary

When `specify init` is run in an environment without an interactive TTY (an AI agent harness, CI, or any piped/captured stdin), initialization can hang indefinitely at an arrow-key selection prompt (`select_with_arrows`). The prompt waits for keypresses that will never arrive, renders nothing useful to the caller, and never times out.

The existing CLI flags do not cover this:

- `--script` and other explicit option flags (`--here`, `--integration`, `--ai`) pre-answer *their own* questions, but any selection **not** covered by a flag still falls through to `select_with_arrows` — there is no flag that reaches every prompt.
- `--force` suggests "proceed without asking", but it only suppresses confirmation-style prompts; it does **not** provide a default answer to selection prompts, so the hang is unreachable by `--force` too.
- There is no global `--non-interactive` / `--defaults` mode, and no TTY detection: with stdin not a TTY, the tool still enters the interactive selector instead of failing fast.

## Reproduction

1. Run `specify init --here --integration claude --force` (or any invocation that leaves at least one selection unanswered by flags) with stdin/stdout not attached to a TTY — e.g. piped, under CI, or driven by an agent harness.
2. Initialization reaches the first selection that has no corresponding flag and blocks in `select_with_arrows` waiting for arrow-key input.
3. No timeout, no error, no output indicating what is being asked — from the caller's side the process simply hangs.

## Expected behavior (any of these would resolve it)

1. **Fail fast when non-interactive:** if stdin is not a TTY and a selection has no flag-supplied answer, exit with a clear error naming the missing option and the flag that would supply it — instead of blocking on keyboard input.
2. **A `--non-interactive` (or `--defaults`) mode** that answers every selection with its documented default, and errors loudly on any selection that has no safe default.
3. **Flag coverage for every `select_with_arrows` prompt**, so a fully scripted init is expressible — at which point `--force` + explicit flags genuinely means "no prompts".

## Why it matters

Spec Kit is a natural fit for agent-driven workflows, and those are precisely the environments with no TTY. Today a scripted `specify init` is not reliably expressible: whether it completes depends on whether the chosen flag combination happens to pre-answer every prompt, and when it doesn't, the failure mode is a silent hang rather than an actionable error.

## Environment

- `specify` 0.16.3.dev0
- Invoked non-interactively (AI agent harness; stdin not a TTY)
- Workaround used: ran the interactive selection out-of-band and supplied every flag the CLI exposes; the remaining prompt(s) had to be avoided rather than answered.

yunaremaia:
Fixes #4152

## Problem

`specify init` can hang **indefinitely** at arrow-key selection prompts when stdin is not attached to a TTY (agent harness, CI, piped input). `select_with_arrows` blocks in `readkey()` waiting for keypresses that never arrive — no timeout, no error, no output. This matters because agent-driven workflows are precisely the environments with no TTY, and Spec Kit is a natural fit for them.

## Fix (defense in depth)

`select_with_arrows` now detects non-interactive stdin (`sys.stdin.isatty()`) and:

1. **Resolves to `default_key` immediately** when one is provided — making a fully scripted init expressible without needing a flag for every prompt (issue option 2).
2. **Raises `ValueError`** naming the missing interactive session when no default exists — fail fast with an actionable message instead of hanging (issue option 1).

Both call sites in `init.py` already guard with `_stdin_is_interactive()`, so this change cannot alter existing behavior for guarded paths — it closes the hang for **any future or unguarded caller** of the selector (issue: "there is no flag that reaches every prompt").

## Tests

New file `tests/test_console_non_interactive.py` (3 tests):

- **no-TTY without default** → raises `ValueError` ("not a terminal")
- **no-TTY with default** → resolves to the default (no hang)
- **TTY path** → interactive loop untouched (get_key mocked to Enter)

Verified: `8/8` console tests green. `tests/integrations/test_cli.py` failures are **pre-existing in this environment** (84 failures with and without this change — identical count), so no regressions introduced.

### DIFF
--- a/src/specify_cli/_console.py
+++ b/src/specify_cli/_console.py
@@ -162,6 +162,12 @@ def select_with_arrows(
 
     Returns:
         Selected option key
+
+    Raises:
+        ValueError: If stdin is not a TTY and no ``default_key`` is given —
+            the interactive loop would otherwise block forever waiting for
+            keypresses that will never arrive (agent harness, CI, piped
+            input). See issue #4152.
     """
     if not options:
         raise ValueError("select_with_arrows() requires at least one option.")
@@ -171,6 +177,17 @@ def select_with_arrows(
         selected_index = option_keys.index(default_key)
     else:
         selected_index = 0
+
+    # Fail fast in non-interactive environments instead of hanging on
+    # readkey() forever. With a default, resolve to it immediately.
+    if not sys.stdin.isatty():
+        if default_key and default_key in option_keys:
+            return default_key
+        raise ValueError(
+            "select_with_arrows() requires an interactive TTY, but stdin is "
+            "not a terminal. Pass an explicit option flag to pre-answer this "
+            "selection, or run in an interactive session."
+        )
 
     selected_key = None
 
--- a/tests/test_console_non_interactive.py
+++ b/tests/test_console_non_interactive.py
@@ -1,0 +1,54 @@
+"""
+Tests for non-interactive (no-TTY) behavior of select_with_arrows.
+
+When stdin is not a TTY (agent harness, CI, piped input), the interactive
+arrow-key selector must fail fast with a clear error instead of blocking
+forever on readkey(). See issue #4152.
+"""
+
+import sys
+
+import pytest
+
+from specify_cli._console import select_with_arrows
+
+
+class FakeStdin:
+    """Stdin-like object that reports not a TTY."""
+
+    def isatty(self) -> bool:
+        return False
+
+
+def test_select_with_arrows_fails_fast_without_tty_and_no_default(monkeypatch):
+    monkeypatch.setattr("sys.stdin", FakeStdin())
+    with pytest.raises(ValueError, match="not a terminal"):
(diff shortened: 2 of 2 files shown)


yunaremaia:
Hi maintainers — the workflow runs on this PR show `action_required` (first-time contributor policy). The change is small and self-contained:

- `select_with_arrows` now checks `sys.stdin.isatty()`: with a `default_key` it resolves immediately (no hang), without one it raises a clear `ValueError` instead of blocking forever on `readkey()` (fixes #4152).
- 3 new tests in `tests/test_console_non_interactive.py` — 8/8 console tests green locally, no regressions (the pre-existing `test_cli.py` failures in my environment are identical with/without the change).

Could you approve the runs when you get a chance? Thanks!

mnriem (changes requested):

PR #4160, opened 2026-08-16
feat(analyze): detect closed vocabulary drift

### LINKED ISSUE #4106: [Feature]: /speckit.analyze should flag closed vocabularies that are enumerated differently in different artifacts
**Correction, 14 Aug 2026:** the paragraph beginning "One of the four is this class" misdescribes the delivered artefact. In the delivered contract both enumerations are of six and agree with each other; the enumeration of two existed earlier in the run and should not have been presented as the shipped contract. Errata: https://doi.org/10.5281/zenodo.21926137 — see my comment below for what this does and does not cost the request. The original text is left intact.

---

Disclosure, per the Spec Kit AI contributions policy: this issue was drafted with AI assistance. Extent: I directed the investigation and chose the framing; an AI assistant read the source material and wrote the draft; I reviewed every claim against the sources cited and I am responsible for the content. Replies from me are my own.

---

### Problem statement

When a spec declares a closed set of values — statuses, categories, admissible kinds — and more than one part of the spec refers to that set, the set gets restated in each place. There is usually nowhere to declare it once. So it multiplies, and any error in it multiplies with it.

This comes from a measured case rather than a hunch.

I ran an experiment where a 40-module system was generated entirely from a contract — no module written or edited by hand, 44 model calls — and verified against reference trajectories published beforehand in an independent formalism.

Four defects survived complete automated verification. None was in the 391 generated statements. All four were in the 30 hand-written contract signatures, and all four were the same error: mis-transcribing a definition that was available in the project. The model wrote correctly what the contract declared, including what the contract declared wrongly.

One of the four is this class. Three parts of the contract declared which invariant classes were admissible. A correction reached one of the three; the other two kept the original wording. The delivered system carried two incompatible vocabularies for the same closed set — one enumerating two values, the others six — and all three enumerations were wrong, in three different ways.

Two verifications could have caught it and neither did. Structural verification compares the contract against the code: it checks which module decides what, not which vocabulary each signature declares in prose. The trajectory oracle validates the running dynamics against published reference traces: it never reached it, because the reference instances do not exercise that part of the formalism. Trace validation validates what the trace traverses.

The contradiction was real, present in the contract, and inert over the validation corpus. It would have surfaced the first time someone ran an instance outside the two validated ones.

The gap, stated precisely: the verification covered the structure of the artifacts and the dynamics of execution. The vocabulary declared in prose inside the specification was covered by neither. I had the analogous check for numeric constants — a threshold declared twice with two values — and it caught those. The same defect over a set of names went straight through.

### Proposed solution

In `/speckit.analyze`, which already does cross-artifact consistency and coverage, add a check for closed vocabularies: flag when the same named set of values is enumerated in more than one artifact or section and the enumerations are not identical, and report the locations and the difference between them.

It needs no domain understanding — only to notice that a named set is enumerated in more than one place and that the enumerations differ. And it is invisible to every other stage: it is not a coverage gap, since every requirement is covered, and not a structural inconsistency between spec, plan and tasks, since the artifacts agree about what exists. It is two statements of the same closed set that disagree about its members, sitting inert until an execution path finally traverses it.

A warning rather than a gate seems right. Legitimate divergence exists — one section sometimes intends a subset of another — and the point is that today the divergence is invisible either way.

A second signal in the same family, if it helps: the same multiplication happened to a procedure rather than a vocabulary. Six signatures repeated the identical construction recipe for one object, because the contract language had no place to declare how a structure is built. Fields had somewhere to go; the construction procedure did not, so it migrated into the signatures and multiplied. Wherever a spec language lacks a home for something, that thing ends up copied, and the copies drift.

Reference, if the case is of interest: When the Defect Is in the Specification: Trajectory-Level Validation of Generated Code, https://doi.org/10.5281/zenodo.21879238 — §8.3 is this defect with its three transcriptions, §10.1 states the check as pending work. The report declares its own limits plainly: one formalism, one model, one operator, and the author of the oracle is the operator of the experiment.

AnkitPorwal04:
## Summary

- inventory explicitly named finite sets during semantic modeling
- compare repeated declarations and report their symmetric member difference
- avoid false positives for documented subsets and unnamed example lists
- expose a closed-vocabulary mismatch metric in the analysis report
- add a focused command-template contract test

Fixes #4106

## Testing

- .venv\\Scripts\\python.exe -m pytest tests\\test_analyze_closed_vocabulary.py tests\\test_specify_template_numbering.py -q (3 passed)
- git diff --check

## AI assistance disclosure

This pull request was implemented, tested, and drafted autonomously by OpenAI Codex (GPT-5) on behalf of @AnkitPorwal04. The human operator selected the overall goal but did not line-by-line author or review the patch before submission.

### DIFF
--- a/templates/commands/analyze.md
+++ b/templates/commands/analyze.md
@@ -111,6 +111,7 @@ Create internal representations (do not include raw artifacts in output):
 - **User story/action inventory**: Discrete user actions with acceptance criteria
 - **Task coverage mapping**: Map each task to one or more requirements or stories (inference by keyword / explicit reference patterns like IDs or key phrases)
 - **Constitution rule set**: Extract principle names and MUST/SHOULD normative statements
+- **Closed-vocabulary inventory**: Record explicitly named finite sets (for example statuses, roles, categories, or enum-like value lists), their members, and each location where they are declared
 
 ### 4. Detection Passes (Token-Efficient Analysis)
 
@@ -150,6 +151,14 @@ Focus on high-signal findings. Limit to 50 findings total; aggregate remainder i
 - Task ordering contradictions (e.g., integration tasks before foundational setup tasks without dependency note)
 - Conflicting requirements (e.g., one requires Next.js while other specifies Vue)
 
+#### G. Closed Vocabulary Consistency
+
+- Find explicitly named finite sets that are enumerated in more than one artifact or section
+- Compare repeated declarations after normalizing case, quoting, and presentation order while preserving the actual member names
+- Flag declarations of the same named set when members are missing, added, or renamed, and report both locations plus the symmetric difference
+- Treat an explicitly identified subset as intentional when the artifact states the subset relationship; otherwise report the divergence as a warning rather than assuming it is an error
+- Do not infer closed vocabularies from unrelated prose or unnamed example lists
+
 ### 5. Severity Assignment
 
 Use this heuristic to prioritize findings:
@@ -187,6 +196,7 @@ Output a Markdown report (no file writes) with the following structure:
 - Coverage % (requirements with >=1 task)
 - Ambiguity Count
 - Duplication Count
+- Closed Vocabulary Mismatch Count
 - Critical Issues Count
 
 ### 7. Provide Next Actions
--- a/tests/test_analyze_closed_vocabulary.py
+++ b/tests/test_analyze_closed_vocabulary.py
@@ -1,0 +1,16 @@
+"""Contract checks for closed-vocabulary analysis guidance."""
+
+from pathlib import Path
+
+
+ANALYZE_TEMPLATE = Path(__file__).parent.parent / "templates" / "commands" / "analyze.md"
+
+
+def test_analyze_checks_repeated_closed_vocabularies() -> None:
+    content = ANALYZE_TEMPLATE.read_text(encoding="utf-8")
+
+    assert "Closed-vocabulary inventory" in content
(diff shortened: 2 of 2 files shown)


mnriem (changes requested):

yunusdim:
Author of #4106 and of the closed-vocabulary preset. The summary here covers the divergence side well. Two things from building and mechanically testing the same check, both about the honesty of an empty result rather than the divergence metric itself.

Coverage, not just the mismatch count. Report declarations inspected against declarations recognised on every run, including zero-divergence ones. A clean result and a run that recognised nothing produce identical output otherwise. Building the preset, my first extractor reported zero over a corpus that held a real divergence in plain sight, because it parsed 4 of 112 declarations. The zero was only honest once the recognised count sat beside it.

Pin the unit of declarations inspected. Lines, sentences and candidate fragments are three defensible readings, and since this pass is prose a model executes, two runs on the same spec report different coverage and both are right. One defined unit, or the number is not comparable across runs.

mnriem:
Thanks @yunusdim — both points are useful. Recognition counts and a defined counting unit would make an empty result more interpretable, although they would not establish that every relevant declaration was detected.


@AnkitPorwal04, I also want to clarify my earlier request to address the review feedback: the preset-first rollout concern was part of that feedback. Addressing the individual implementation comments would not, by itself, establish approval to promote this into core.


The preset is now catalogued. Let’s defer this core change while we gather actual before-and-after agent evidence through that opt-in path, rather than expand this PR with further prompt changes now.

PR #4166, opened 2026-08-17
test(presets): normalize whitespace in resolve output assertion to prevent terminal line-wrap failures

aoright:
### Summary
- In `tests/test_presets.py::TestInstalledPresetRichMarkup::test_resolve_accepts_dotted_command_name`, the assertion `assert "constitution.md" in strip_ansi(result.output)` failed when running in environments with constrained terminal column widths or long paths because Rich console output wrapped `templates/commands/constituti\non.md` across lines.
- Normalize whitespace in the stripped output before asserting substring inclusion, ensuring test determinism regardless of path length or terminal width.

### DIFF
--- a/tests/test_presets.py
+++ b/tests/test_presets.py
@@ -13672,7 +13672,7 @@ class TestInstalledPresetRichMarkup:
         )
 
         assert result.exit_code == 0, (result.output, result.exception)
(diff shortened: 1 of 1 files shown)


mnriem:
Thank you!

PR #4181, opened 2026-08-18
fix(workflows): validate dispatch defaults

WOLIKIMCHENG:
## Summary
- validate workflow-level integration, model, and options defaults before dispatch
- reject malformed defaults before unvalidated execution can create run state
- cover invalid, valid, and null default shapes

## Verification
- All checks passed!
- focused workflow default validation tests (16 passed)


mnriem (changes requested):

mnriem (approved):

PR #4187, opened 2026-08-18
fix(integrations): report a falsy non-mapping integration descriptor as a shape error

Noor-ul-ain001:
## Summary

`IntegrationDescriptor._load` (in `src/specify_cli/integrations/catalog.py`) does `yaml.safe_load(fh) or {}`. `_validate` opens with an `isinstance(self.data, dict)` check, so a truthy non-mapping (`- a`, `hello`) is reported correctly as a shape error — but `or {}` replaces falsy non-mappings with an empty mapping *before* that check ever runs, so those descriptors were reported as `Missing required field: schema_version` instead of the actual problem:

| `integration.yml` content | Before | After |
|---|---|---|
| `false` | `Missing required field: schema_version` | `Descriptor root must be a YAML mapping, got bool` |
| `0` | `Missing required field: schema_version` | `Descriptor root must be a YAML mapping, got int` |
| `''` | `Missing required field: schema_version` | `Descriptor root must be a YAML mapping, got str` |
| `[]` | `Missing required field: schema_version` | `Descriptor root must be a YAML mapping, got list` |
| `null` / `~` / `NULL` | `Missing required field: schema_version` | `Descriptor root must be a YAML mapping, got NoneType` |
| *(empty file)* | `Missing required field: schema_version` | unchanged — still `Missing required field: schema_version` |

`safe_load` returns `None` for both an explicit null scalar and a genuinely empty document, so a plain `data is None` normalization can't tell them apart either. This uses `yaml.compose`, which yields no node only for a genuinely empty document, to keep the empty-file case reporting its missing fields while every other non-mapping shape (including explicit null) now surfaces the real error.

This is the same bug class a maintainer fixed the same day in the sibling overlay-manifest loader (`ProjectOverlaySource.collect`, commit 39c36c4 / #3884) — that PR's description even calls out that the pattern should be checked elsewhere. This is the unfixed twin in the integration catalog's descriptor loader.

## Test plan

- [x] Added `TestIntegrationDescriptor::test_falsy_non_mapping_descriptor_reports_shape_error` (parametrized over `[]`, `false`, `0`, `''`, `null`, `~`, `NULL`, `- a`, `hello`) and `test_empty_document_still_reports_missing_fields`
- [x] Verified the new tests fail without the fix (7 of 10 cases) and pass with it
- [x] `pytest tests/integrations/test_integration_catalog.py` — 133 passed

### DIFF
--- a/src/specify_cli/integrations/catalog.py
+++ b/src/specify_cli/integrations/catalog.py
@@ -686,14 +686,14 @@ class IntegrationDescriptor:
             # explicit null scalar (``null``, ``~``, ``Null``, ``NULL``), so it
             # cannot tell them apart on its own. ``compose`` yields no node
             # only for a genuinely empty document.
-node = yaml.compose(text)
-data = yaml.safe_load(text)
-is_empty_document = node is None or (
-    data is None
-    and isinstance(node, yaml.nodes.ScalarNode)
-    and node.value == ""
-    and node.start_mark.index == node.end_mark.index
-)
+            node = yaml.compose(text)
+            data = yaml.safe_load(text)
+            is_empty_document = node is None or (
+                data is None
+                and isinstance(node, yaml.nodes.ScalarNode)
+                and node.value == ""
+                and node.start_mark.index == node.end_mark.index
+            )
         except yaml.YAMLError as exc:
             raise IntegrationDescriptorError(f"Invalid YAML in {path}: {exc}")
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

mnriem (approved):

mnriem:
Thank you!

PR #4189, opened 2026-08-18
fix(utils): narrow bare except Exception in merge_json_files

Noor-ul-ain001:
## Summary

`merge_json_files`'s read of the existing JSON file wraps `json5.load` in a bare `except Exception as e:` that returns `None` (meaning "preserve existing settings, skip merge") on *any* failure — including a real programming bug like a `TypeError`/`AttributeError`, which is silently treated identically to an expected I/O or parse error and, without `verbose`, surfaces nothing at all.

The only expected outcomes for this block are:
- `OSError` — the file is inaccessible (already special-cased for `FileNotFoundError` just above)
- `ValueError` — malformed JSON5 (json5's decode error is a `ValueError` subclass)

Anything else should propagate instead of being swallowed as if it were a parse failure.

This is the same bug, same fix shape, as its caller `handle_vscode_settings`, whose own bare `except Exception` was narrowed to `(OSError, ValueError, KeyError)` in commit 16f4577 (#3844) with the identical rationale — "let programming errors like TypeError or AttributeError propagate while still handling expected I/O and parse errors gracefully." That PR's own regression test monkeypatched `merge_json_files` itself to prove the *caller's* narrowing works, but the callee it monkeypatched around still had the original bare-except bug.

## Test plan

- [x] Added `test_merge_json_files_propagates_programming_errors` to `tests/test_merge.py`, monkeypatching `json5.load` to raise `TypeError` and asserting it propagates out of `merge_json_files` instead of returning `None`
- [x] Verified the new test fails without the fix (`DID NOT RAISE <class 'TypeError'>`) and passes with it
- [x] `pytest tests/test_merge.py` — 13 passed

### DIFF
--- a/src/specify_cli/_utils.py
+++ b/src/specify_cli/_utils.py
@@ -249,7 +249,7 @@ def merge_json_files(existing_path: Path, new_content: Any, verbose: bool = Fals
         except FileNotFoundError:
             # Handle race condition where file is deleted after exists() check
             exists = False
-        except Exception as e:
+        except (OSError, ValueError) as e:
             if verbose:
                 console.print(f"[yellow]Warning: Could not read or parse existing JSON in {existing_path.name} ({e}).[/yellow]")
             # Skip merge to preserve existing file if unparseable or inaccessible (e.g. PermissionError)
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

Issue #4164, opened 2026-08-17
[Feature]: Stop grep-based spec alignment, structured IDs + context packs


harsha09:
### Problem Statement

I'm frustrated when a product change (especially a UI change) has to be folded into an existing feature. Agents find related work by grepping similar wording in spec.md, plan.md, and tasks.md. After a reword “charge on Pay” vs “show a confirm modal” the live AC/FR is often not retrieved, so another /speckit.specify pass or a manual edit adds a new requirement instead of updating the existing one.

The result is two live lines for the same behavior, or two live lines that contradict each other. /speckit.analyze is supposed to catch duplicates and conflicts, but it only reports what the model loaded. If the related line was never in context, the report is clean. I end up rerunning analyze over and over; it is not a reliable alignment step.

There is no script-owned inventory of live IDs (FR-, US/AC, T-xxx) that specify, analyze, and implement all share. Commands dump whole Markdown files (wasted tokens) and still miss the one line that mattered. I need a way to keep the spec set unambiguous no silent duplicates, no contradictory live requirements without another round of hope-and-grep.

### Proposed Solution

Keep spec.md / plan.md / tasks.md as the human artifacts. Add a script-owned live inventory so specify, analyze, and implement stop grepping prose to decide what exists.

1. Structured sidecar for tasks (and later spec/plan records)
Alongside tasks.md, write tasks.yaml (YAML on disk, JSON on stdout). Each task is a record: id, status, story, files, covers: [US1/AC2, FR-007]. A Python script is the only mutator: parse Markdown once, then list / get / mark-done / coverage. Agents do not regex checkboxes.

2. Complete live inventory, every command
The script emits every live FR- / AC / SC / T- for the current feature. That list is the source of “what exists.” Specify consults it before adding a requirement. Analyze builds its report from it. Implement loads one task plus the records it covers. No command should discover the spec by grepping similar text.

3. Context pack instead of whole files
specify context --task T014 (or equivalent) prints a small JSON pack: that task, the live ACs/FRs it cites, matching plan bullets, and optional research hits. /speckit.implement and /speckit.analyze drive from the pack/inventory, not from dumping three Markdown files. That removes the missed-line failure and cuts tokens.

4. Local, per-feature recall only (optional second phase)
[...]

PR #4220, opened 2026-08-20
community: add speckit-inventory extension and inventory-alignment preset for #4164

Yash-Chindam:
## Description

Lists a paired community **extension** and **preset** that deliver Phase 1 of #4164 without any change to Spec Kit core, following the design [@mnriem proposed in that issue](https://github.com/github/spec-kit/issues/4164#issuecomment-5348616852). Per CONTRIBUTING's note on large changes, that comment is the prior maintainer agreement for this direction.

### The problem this solves

A user reworks a UI requirement mid-feature. The agent looks for related work by grepping similar wording across `spec.md`, `plan.md`, and `tasks.md`. After the reword — "charge on Pay" becomes "show a confirm modal" — nothing matches, so the agent adds a *second* live requirement instead of updating the first. `/speckit.analyze` cannot catch it, because the contradicting line was never loaded into context. The user is left with duplicate and contradictory requirements to reconcile by hand.

The fix is to stop discovering requirements by wording and start discovering them by ID: derive the complete set of live IDs from the artifacts that already exist, then make the agent classify every proposed requirement against that set before it writes.

### What is being listed

Both ship from one standalone repo, [`spec-kit-inventory-alignment`](https://github.com/Yash-Chindam/spec-kit-inventory-alignment), at [v0.1.0](https://github.com/Yash-Chindam/spec-kit-inventory-alignment/releases/tag/v0.1.0).

**`speckit-inventory`** — extension, category `visibility`, effect `read-only`:

- Regenerates every live `FR-`/`NFR-`/`SC-`/`AC-`/`T-` ID from the existing `spec.md` and `tasks.md` on each run and prints JSON. Nothing is persisted, so a `tasks.yaml` to `tasks.md` sync problem cannot arise — there is no second source of truth to drift.
- `speckit.speckit-inventory.list` returns the complete inventory; `speckit.speckit-inventory.context` returns a focused context pack (one task plus only the requirements it covers) instead of a whole-file dump.
- Optional `before_specify` and `before_analyze` hooks, both `optional: true` so they prompt rather than fire automatically.

**`inventory-alignment`** — preset:

- `wrap` overrides of `speckit.specify` and `speckit.analyze` using the `{CORE_TEMPLATE}` placeholder, prepending a pre-pass that classifies each proposed requirement as already-true / edit-existing / conflict / genuinely-new, then falling through to the original command body.
- `append` override of `tasks-template` documenting the optional `covers:` field that links a task to the requirement IDs it implements.

### Design constraints held deliberately

- **No second source of truth** — the inventory is derived, never stored.
- **No mutation** — the extension only reads; there are no write calls in the script.
- **No core changes** — this PR touches only catalog and docs files.
- **No embeddings** — that was Phase 2 in the original proposal and is out of scope.

### One naming deviation, flagged deliberately

The issue comment asks for `speckit.inv.context`, but that name cannot coexist with the extension id `speckit-inventory` the same comment specifies. The extension loader requires a command's namespace segment to equal the manifest id, so that combination fails at install:

```
Validation Error: Command 'speckit.inv.context' must use extension namespace 'speckit-inventory'
```

I kept the id and named the commands `speckit.speckit-inventory.list` / `.context`. If the shorter invocation is preferred, renaming the extension id to `inv` gets there — happy to do that instead; it is a one-line manifest change plus new release assets.

### Known gap, not blocking

`preset.yml`'s `requires:` can only declare `speckit_version`, so the preset cannot formally require the extension. The wrappers degrade safely — without the extension the core workflow runs unchanged — and the pairing is documented in both READMEs. Happy to leave the `requires: extensions:` enhancement as its own issue, as @mnriem suggested.

### Verified against the published release archives

Not just a local checkout — these ran against the actual `download_url`s in the catalog entries.

| Check | Result |
|---|---|
| `specify extension add --from <release url>` | Installs; both commands registered; 2 agent skills auto-registered |
| `specify preset add --from <release url>` | Installs at priority 10 |
| `specify preset resolve tasks-template` | Chain resolves `1. [base] core` then `2. [append] inventory-alignment v0.1.0` |
| `wrap` composition | Pre-pass precedes the full core body (355 / 271 lines); zero unresolved `{CORE_TEMPLATE}`; core frontmatter `scripts:` preserved in `speckit.analyze` |
| Hook registration | Both written to `.specify/extensions.yml` as `optional: true` |
| Release asset integrity | Both URLs return 200; downloaded bytes match the `sha256` in the catalog entries |
| Extension unit tests | 9/9 pass |

### Worked example

Given a `spec.md` defining `FR-001`, `FR-002`, `NFR-001`, `SC-001`, and a `tasks.md` whose `T010` covers `SC-001` and a nonexistent `FR-009`, the inventory reports the dangling reference by source rather than silently accepting it:

```json
{ "id": "FR-009", "source": "tasks.md", "line": 6,
  "text": "- [ ] T010 [US2] Add retry telemetry in src/telemetry/ (covers: SC-001, FR-009)" }
```

`spec.md` is scanned first, so it wins as the definition site; anything whose `source` is `tasks.md` is a reference to a requirement that was never defined. The `speckit.analyze` pre-pass instructs the agent to report exactly that case.

The context pack returns only the requirements a given task covers, not the whole spec — which is the token-cost argument for the approach:

```
--task T002  ->  task T002 plus FR-001, NFR-001   (2 of 5 requirements)
--task T001  ->  task T001 plus nothing           (unlinked setup task, valid)
```

### Spec Kit's own suite

630 passed for catalog- and community-selected tests on this branch. Two failures are `WinError 1314` symlink-privilege failures from my Windows environment; they reproduce identically with these changes stashed, so they are unrelated to this PR.

### What was not tested

Command *composition and registration* are verified above, but `/speckit.specify` and `/speckit.analyze` were not run end-to-end inside a coding agent to observe the pre-pass and hook prompt in a live session. This PR changes no core command behavior, so I read CONTRIBUTING's manual-testing requirement as not applying to it — but the matrix is written and I will run it and post results if a maintainer wants that evidence before merge.

## My own analysis of the developer experience

Honest assessment, including where I think this is weak.

**What works well.** The read-only, derive-every-time design is the part I would defend hardest. The original issue proposed YAML sidecars, and working through it convinced me that would have been a mistake — a sidecar makes the sync problem the user's problem, and the sync problem is worse than the bug being fixed. Deriving on every run means the inventory is never stale and there is nothing to migrate for existing projects, which also answers the migration question I raised earlier in the issue.

**What is genuinely uncertain.** The value depends on the agent actually honoring the pre-pass. The extension guarantees the *data* is complete and cheap to load; it cannot guarantee the model classifies correctly against it. I expect this to reduce duplicate requirements substantially rather than eliminate them, and I would rather say that plainly than oversell it. The `covers:` field is likewise only as good as the agent's discipline in emitting it — the `append` override documents it, but nothing enforces it, and a task without `covers:` degrades to an empty context pack.

**The rough edge users will hit first.** The two-part install is awkward. Because `preset.yml` cannot declare an extension dependency, a user who installs only the preset gets prompts that silently do nothing — the workflow still runs correctly, but the feature appears broken. Both READMEs lead with this, and it is the strongest argument for the `requires: extensions:` enhancement being worth doing.

**Scope I deliberately did not take.** Embeddings would help with genuine paraphrase matching, which is the harder half of the original problem. Leaving them out keeps this dependency-free and reviewable, and I would rather see whether the ID-based half earns adoption first.

## AI Disclosure

Filed by @Yash-Chindam. Substantial AI assistance — code generation, not just documentation or comments — in two distinct passes.

**Pass 1 — ChatGPT (Codex, model: GPT-5).** Drafted the initial extension, preset, and tests in a separate repository, and wrote my earlier scoping comment on #4164.

**Pass 2 — Claude Code (model: Claude Opus 5), autonomous.** Reviewed that draft against this issue thread and the repo's own guidance, and found real defects in it:

- The `append` override of `tasks-template` and the `before_specify` / `before_analyze` hooks from @mnriem's comment were missing entirely. The `covers:` field was load-bearing for the extension, but nothing taught the template to emit it.
- Requirement IDs were harvested from `spec.md` only, rather than from `spec.md` and `tasks.md`, so dangling references were invisible.
- The CLI rejected unpadded task IDs — `--task T14` failed on a task literally written as `T14`, because the parser normalized to `T014` but the lookup did not.

It fixed those, added the `list` command, expanded the tests from 2 to 9, produced these catalog entries and documentation rows, and wrote this description.

**Attribution of the testing above, stated precisely.** Every check in the Testing section was executed by Claude Code on my machine at my direction, and I have reviewed the results. I am not claiming I re-ran each command by hand. Per AGENTS.md, the commit on this branch carries an `Assisted-by: Claude Code (model: Claude Opus 5, autonomous)` trailer, and I will disclose agent involvement again in each review-round comment rather than relying on this section to cover them.

This account had no other open pull requests in this repository when this one was filed.

### One process note

`add-community-extension.md` and `add-community-preset.md` automate these catalogs from `[Extension]:` / `[Preset]:` issue submissions, and `EXTENSION-PUBLISHING-GUIDE.md` says not to open a pull request that edits `extensions/catalog.community.json` directly. I raised this as a PR because that is what the issue asked for — very happy to close it and refile through the two issue templates if you prefer the automated path.


mnriem:
@Yash-Chindam Can you please file 2 separate issues, one for the extension submission and one for the preset submission. We have issue template for these that run agentic workflows that vet the submissions. Thanks! Great work :)

Yash-Chindam:
Filed both submissions:

- #4226 — `[Extension]: Add Spec Inventory (speckit-inventory)`
- #4227 — `[Preset]: Add Inventory Alignment (inventory-alignment)`

Closing this PR, since the two workflows open their own PRs against the same four files and theirs carry the `Closes` reference to the submission issues.

Both issues are unlabeled. The template's `needs-triage` label doesn't exist in the repo so creation with labels failed, and I deliberately didn't apply `extension-submission` / `preset-submission` myself since that label is the vetting gate — they'll need a maintainer to trigger the workflows.

One change to the submitted artifacts while preparing these, in `Yash-Chindam/spec-kit-inventory-alignment@312f1a3`: both READMEs now document the `specify extension add --from <download-url>` / `specify preset add --from <download-url>` install form rather than only `--dev`, which is what the preset submission checklist asks for. The Documentation URLs in both issues point at `main` to reflect that. Release assets and their hashes are unchanged from v0.1.0.

Also flagged in #4226, since it's a deviation from your comment rather than something I'd want discovered during vetting: `speckit.inv.context` can't coexist with the extension id `speckit-inventory`, because the loader requires a command's namespace segment to equal the manifest id — installing that pairing fails with `Command 'speckit.inv.context' must use extension namespace 'speckit-inventory'`. I kept the id and used `speckit.speckit-inventory.list` / `.context`. If you'd prefer the shorter invocation, renaming the extension id to `inv` gets there in one manifest line plus new assets.

Disclosure: this comment and the two submissions were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. Repository commits carry `Assisted-by:` trailers.

PR #4230, opened 2026-08-20
fix(workflows): stop offering a condition correction that inverts it

ntdatt812:
Follow-up to #4182. That PR merged while this last fix was in flight — by about a minute — so the defect below is still live on `main`.

## Problem

`format_condition_correction` wraps whatever it is handed. That is right for a formatter, but the three step validators advertise its output as a paste-ready correction, and for two inputs it cannot repair the suggestion is actively harmful.

I measured what pasting each suggestion does rather than reasoning about it:

| condition | validator reports | correction it suggests | pasting that gives |
|---|---|---|---|
| `"   "` | always true | `"{{ }}"` | **False** |
| `{{ inputs.name == 'abc` | always true | `"{{ inputs.name == 'abc }}"` | **False** |

Both **invert** the condition rather than repair it:

- the blank core interpolates to the empty string;
- the open quote survives wrapping, so the raw-close fallback evaluates a truncated comparison whose result is the string `"False"`, which `evaluate_condition` reads as the `false` keyword before `bool()` is ever reached.

So an author who follows the advice trades an always-true condition for an always-false one, having been told it was the fix.

## Change

`format_condition_remediation` builds the advice, and the three validators call it instead of assembling the sentence themselves. It offers a correction only when wrapping would genuinely repair the input, and otherwise names the fault — the same call already made for `condition_has_malformed_expression_block`, which deliberately offers no suggestion.

```
'   '                    -> ... There is no expression here to wrap: use the literal
                             true or false, since an empty '{{ }}' block evaluates to
                             the empty string and would silently invert the condition.

{{ inputs.name == 'abc   -> ... Close the unbalanced quote first: wrapping it as
                             written leaves the quote open, so the raw-close fallback
                             evaluates a truncated comparison rather than the one
                             written, and its result can silently invert the condition.

inputs.count > 100       -> ... Wrap the expression: "{{ inputs.count > 100 }}".
```

`_has_unbalanced_quote` uses the same left-to-right scan as `_find_block_close` and `_strip_stray_delimiters`, so "inside a string" means one thing throughout the module.

## Note on one of my own claims

My first draft of the unbalanced-quote message said the wrapped form *"stays always true"*. The new test failed and showed it returns `False` — `evaluate_condition` recognises the residual `"False"` as the keyword before coercion. The message and docstring say *inverted*, which is what the measurement shows. Flagging it because the distinction is the whole point of the change.

## Verification — Python 3.11, on this branch off `main`

```
tests/unit + tests/test_workflows.py   1233 passed  (main: 1216)
```

The 22 failures in that run are identical with and without this commit — all `TestWorkflowCliAlignment` symlink tests, which need elevation on Windows. I ran the suite against unmodified `main` to confirm that rather than assume it.

**Mutation-checked:** removing either gate fails exactly the 9 new parametrised cases and nothing else, so the fixtures pin this defect rather than passing incidentally.

Credit to the Copilot reviewer on #4182 — it raised this as a suppressed comment there and was right.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -879,3 +879,59 @@ def format_condition_correction(condition: Any) -> str:
     # double-spaced "{{  }}" that string concatenation would otherwise produce.
     body = "{{ " + core + " }}" if core else "{{ }}"
     return json.dumps(body, ensure_ascii=False)
+
+
+def _has_unbalanced_quote(text: str) -> bool:
+    """True when a quote opened in *text* is never closed.
+
+    Same left-to-right, first-quote-wins scan the rest of this module uses, so the
+    answer agrees with what ``_find_block_close`` and ``_strip_stray_delimiters``
+    consider "inside a string".
+    """
+    quote: str | None = None
+    for ch in text:
+        if quote is not None:
+            if ch == quote:
+                quote = None
+        elif ch in ("'", '"'):
+            quote = ch
+    return quote is not None
+
+
+def format_condition_remediation(condition: Any) -> str:
+    """The advice sentence for a condition that is never evaluated.
+
+    ``format_condition_correction`` wraps whatever it is handed, which is the right
+    behaviour for a formatter but the wrong thing to *advertise* for two inputs it
+    cannot actually repair. Both were being offered as paste-ready corrections:
+
+    * a blank core -- ``condition: "   "`` corrected to ``"{{ }}"``, which evaluates
+      to the empty string. The author is told to paste something that flips an
+      always-true condition to always-false, which is a different bug rather than a
+      fix.
+    * an unbalanced quote -- ``{{ inputs.name == 'abc`` corrected to
+      ``"{{ inputs.name == 'abc }}"``. The quote is still open, so the raw-close
+      fallback evaluates a truncated comparison and yields the string ``"False"``,
+      which ``evaluate_condition`` then reads as the ``false`` keyword. The author
+      pastes the correction and the condition flips from always-true to
+      always-false -- inverted, not repaired.
+
+    Naming what is wrong beats handing back something that looks authoritative and
+    is not -- the same call already made for
+    ``condition_has_malformed_expression_block``, which offers no correction at all.
+    """
+    core = _strip_stray_delimiters(str(condition)).strip()
+    if not core:
+        return (
+            "There is no expression here to wrap: use the literal true or false, "
+            "since an empty '{{ }}' block evaluates to the empty string and would "
+            "silently invert the condition rather than repair it."
+        )
+    if _has_unbalanced_quote(core):
+        return (
+            "Close the unbalanced quote first: wrapping it as written leaves the "
+            "quote open, so the raw-close fallback evaluates a truncated comparison "
+            "rather than the one written, and its result can silently invert the "
+            "condition instead of repairing it."
+        )
+    return "Wrap the expression: " + format_condition_correction(condition) + "."
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -904,9 +904,24 @@ def _has_unbalanced_quote(text: str) -> bool:
     return quote is not None
 
 
+_BRACKET_PAIRS = {")": "(", "]": "[", "}": "{"}
+
+# The operators the evaluator delimits with spaces; derived so the check cannot
+# drift from _COMPARISON_OPERATORS.
+_WORD_OPERATORS = tuple(
+    op for op in (" or ", " and ") + _COMPARISON_OPERATORS if op.startswith(" ")
+)
+
+
 def _has_unbalanced_bracket(text: str) -> bool:
-    """True when a bracket opened outside a quoted operand is never closed."""
-    depth = 0
+    """True when brackets outside a quoted operand do not nest and match.
+
+    A depth counter is not enough: it calls ``inputs.f(]`` balanced, because the
+    ``]`` cancels the ``(``. The evaluator then resolves that body to ``None`` and
+    the comparison is false, which is the inversion this module is trying to keep
+    out of the suggested correction. Track the opener types instead.
+    """
+    stack: list[str] = []
     quote: str | None = None
     for ch in text:
         if quote is not None:
@@ -915,33 +930,50 @@ def _has_unbalanced_bracket(text: str) -> bool:
         elif ch in ("'", '"'):
             quote = ch
         elif ch in "([{":
-            depth += 1
-        elif ch in ")]}":
-            depth -= 1
-            if depth < 0:
+            stack.append(ch)
+        elif ch in _BRACKET_PAIRS:
+            if not stack or stack.pop() != _BRACKET_PAIRS[ch]:
                 return True
-    return depth != 0
+    return bool(stack)
 
 
 def _has_incomplete_operand(text: str) -> bool:
-    """True when a top-level operator in *text* is missing an operand.
+    """True when an operator in *text* is missing an operand on either side.
 
-    Reads ``_COMPARISON_OPERATORS`` and the boolean keywords from the evaluator
-    rather than restating them, so the check cannot drift from what
-    ``_evaluate_simple_expression`` will actually split on.
+    Splits on **every** top-level occurrence rather than the first. Checking only
+    the first is the same defect this module exists to reject one level up: it let
+    ``inputs.a == inputs.b ==`` through, because the leading ``==`` has operands on
+    both sides and the scan stopped there.
+
+    Reads ``_COMPARISON_OPERATORS`` from the evaluator rather than restating it, so
+    the check cannot drift from what ``_evaluate_simple_expression`` splits on.
     """
+    stripped = text.strip()
+    if not stripped:
+        return True
+
+    # `not x` is a valid prefix form; `and x` and `or x` are not, and none of the
+    # three is valid alone or trailing. The keyword scans below use bare words
+    # because a leading operator has no space in front of it to match on.
+    if stripped in ("and", "or", "not") or stripped.endswith(" not"):
+        return True
+    # Word operators lose their delimiting space at the ends of a stripped core, so
+    # a trailing "not in" or a leading "and" needs matching without it. Derived from
+    # the evaluator's own table rather than restated.
+    for op in _WORD_OPERATORS:
+        if stripped.endswith(op.rstrip()) or stripped.startswith(op.lstrip()):
+            return True
+
     for op in (" or ", " and ") + _COMPARISON_OPERATORS:
-        idx = _find_top_level(text, op)
-        if idx == -1:
+        if _find_top_level(stripped, op) == -1:
             continue
-        if not text[:idx].strip() or not text[idx + len(op):].strip():
+        if any(not segment.strip() for segment in _split_top_level(stripped, op)):
             return True
-    if _find_top_level(text, "|") != -1:
-        if any(not segment.strip() for segment in _split_top_level(text, "|")):
+
+    if _find_top_level(stripped, "|") != -1:
+        if any(not segment.strip() for segment in _split_top_level(stripped, "|")):
             return True
-    return text in ("not", "or", "and") or text.endswith((" not", " or", " and"))
-
-
+    return False
 def _wrapping_would_not_repair(core: str) -> str | None:
     """Why wrapping *core* in ``{{ }}`` would not yield the expression intended.
 
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -1077,7 +1077,24 @@ def _unresolvable_term(text: str) -> str | None:
 
     if _find_top_level(stripped, "|") != -1:
         segments = _split_top_level(stripped, "|")
-        return _unresolvable_term(segments[0])
+        reason = _unresolvable_term(segments[0])
+        if reason is not None:
+            return reason
+        # A filter argument is an ordinary operand to `_apply_filter`, which
+        # evaluates it with `_evaluate_simple_expression` like any other. Skipping
+        # it let `inputs.tags | join(bogus)` be offered as paste-ready: `bogus` is
+        # no namespace root, resolves to None, and the wrapped form then raises
+        # `join: expected a string separator, got NoneType`. Parse with the same
+        # pattern `_apply_filter` uses, so a form this does not recognize is left
+        # to the evaluator probe rather than guessed at here.
+        for segment in segments[1:]:
+            match = re.fullmatch(r"(\w+)\((.+)\)", segment.strip())
+            if match is None:
+                continue
+            reason = _unresolvable_term(match.group(2))
+            if reason is not None:
+                return reason
+        return None
 
     for op in (" or ", " and "):
         idx = _find_top_level(stripped, op)
@@ -1097,6 +1114,24 @@ def _unresolvable_term(text: str) -> str | None:
             )
 
     if _is_literal(stripped):
+        return None
+
+    # A list literal is a term the evaluator understands, and it recurses into the
+    # elements rather than resolving the brackets as a name. Not mirroring that
+    # denied the correction to `inputs.tag in ['x', 'y']` -- a condition wrapping
+    # repairs completely -- while reporting the list as an unresolvable name. The
+    # empty-segment skip matches `_evaluate_simple_expression`, which drops them so
+    # `[1, 2,]` is `[1, 2]` rather than `[1, 2, None]`.
+    if stripped.startswith("[") and stripped.endswith("]"):
+        inner = stripped[1:-1].strip()
+        if not inner:
+            return None
+        for element in _split_top_level_commas(inner):
+            if not element.strip():
+                continue
+            reason = _unresolvable_term(element)
+            if reason is not None:
+                return reason
         return None
 
     segments = _split_top_level(stripped, ".")
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -1029,15 +1029,31 @@ def _evaluator_rejects(text: str) -> str | None:
 
 
 def _looks_numeric(text: str) -> bool:
+    """Mirror the evaluator's numeric literal test exactly.
+
+    `_evaluate_simple_expression` only calls `float()` when a `.` is present and
+    `int()` otherwise, so `1e3` is not a number to it -- it falls through to a path
+    lookup and resolves to None. A bare `float()` here accepted `1e3` and the
+    correction turned a truthy condition false.
+    """
     try:
-        float(text)
-    except ValueError:
+        if "." in text:
+            float(text)
+        else:
+            int(text)
+    except (ValueError, TypeError):
         return False
     return True
 
 
 def _is_literal(text: str) -> bool:
-    if len(text) >= 2 and text[0] in ("'", '"') and text[-1] == text[0]:
+    """Mirror the evaluator's literal tests exactly.
+
+    The string case is the opening quote's *matching close being the final
+    character*, not first/last-character equality: `'a' 'b'` passes the latter but
+    is two literals to the evaluator, which falls through to a path lookup.
+    """
+    if text[:1] in ("'", '"') and text.find(text[0], 1) == len(text) - 1:
         return True
     return text.lower() in ("true", "false", "none", "null") or _looks_numeric(text)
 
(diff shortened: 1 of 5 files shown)


mnriem (changes requested):
Please address Copilot feedback. Thanks for all the great work!

ntdatt812:
Both suppressed findings were correct. Fixed in `6944920`.

**1. The quote-balance gate was not sufficient.** `inputs.name ==` has a non-empty, quote-balanced core, so a correction was still advertised:

```
inputs.name ==   ->   "{{ inputs.name == }}"      True -> False
```

Same harm as the two cases this PR already covered — the author is told the condition is always true and hands back an always-false one.

**2. The message named the wrong mechanism.** I said the wrapped form goes through the raw-close fallback. Measured:

```
_is_single_expression("{{ inputs.name == 'abc }}")  ->  True
```

So it takes the typed fast path instead. The diagnosis was wrong even though the conclusion — that the result can invert the condition — was right.

## What changed

Enumerating broken shapes was the mistake; it is why this needed three rounds. `_wrapping_would_not_repair` now reports the first reason wrapping cannot yield the intended expression — empty core, unclosed quote, unbalanced bracket, or an operator missing an operand — and the advice names that instead of offering a suggestion.

`_has_incomplete_operand` reads `_COMPARISON_OPERATORS`, extracted from `_evaluate_simple_expression`, so the check cannot drift from what the evaluator actually splits on.

The messages now describe the text itself rather than the interpolator path it will take. Asserting an internal route is exactly what made my previous two versions wrong, twice, so they no longer do it.

## The test that matters

`test_every_offered_correction_is_a_complete_expression` states the property rather than the shapes: whatever is advertised as paste-ready must survive both validators. A new malformed shape is caught by the invariant now, not by adding another fixture row.

## Verification — Python 3.11

```
tests/unit/test_condition_expression_block.py      182 passed  (was 133)
tests/unit + tests/test_workflows.py              1282 passed  (was 1233)
```

22 failures identical before and after — the `TestWorkflowCliAlignment` symlink tests that need elevation on Windows, confirmed by running against the stashed tree.

## One mutation initially escaped, and that is worth reporting

Removing `"<"` from `_COMPARISON_OPERATORS` left all tests green. The operator test parametrised over the constant it was checking, so shrinking the constant shrank the test — the same can't-fail-when-it-matters shape this whole PR exists to reject, in my own test. It is hard-coded now and that mutation fails.

Final mutation results, each gate against its own cases: operand gate → 12 failures, bracket gate → 3, dropped operator → 1.

mnriem:

ntdatt812:
@mnriem — done, in `24c71a6`. Both live Copilot findings were correct.

**1. `_has_incomplete_operand` inspected only the first occurrence of each operator**, and its end-of-string check covered only trailing boolean keywords:

```
inputs.a == inputs.b ==   ->  correction offered   True -> False
and inputs.ready          ->  correction offered   True -> False
```

Worth naming plainly: that is the *same* defect this PR's parent commit fixed one level up — stopping at the first match — reintroduced inside the gate written to prevent it. It now splits on every top-level occurrence and requires every operand to be non-empty.

A stripped core also loses the space that delimits a word operator, so `inputs.a not in` matched nothing at all. `_WORD_OPERATORS` is derived from `_COMPARISON_OPERATORS` and matched against both ends without that space.

**2. `_has_unbalanced_bracket` counted depth**, so mismatched types cancelled out:

```
inputs.f(]   ->  correction offered   True -> False
```

It now tracks opener types on a stack and rejects a non-matching closer.

## On the third thread

The docstring one is marked outdated and refers to the text before `6944920`. The line it points at now reads:

> Each branch names something observable about the text itself, deliberately not the interpolator path it will take: two earlier versions of this message asserted an internal route — the raw-close fallback — and were wrong, because `_is_single_expression` accepts the wrapped form and sends it down the typed fast path instead.

So it does not attribute the inversion to the raw-close fallback; it records that earlier versions did and states the actual path, which matches what the reviewer described. I left it as the explanation for why the messages avoid naming internal routes. Happy to reword if you would rather it not mention the fallback at all.

## Verification — Python 3.11

```
tests/unit/test_condition_expression_block.py      207 passed  (was 182)
tests/unit + tests/test_workflows.py              1307 passed  (was 1282)
```

22 failures identical before and after — the `TestWorkflowCliAlignment` symlink tests needing elevation on Windows, confirmed against the stashed tree.

**Mutation-checked:** depth-only brackets → 9 failures, first-occurrence-only → 3, dropping the end-of-core word scan → 16.

mnriem (dismissed):

ntdatt812:
That was the strongest point raised on this PR, and it was right. Addressed in `6c8ca2c`.

`reason is None` only excluded four structural shapes, and structural shapes cannot establish that wrapping yields a working expression. Both examples reproduce:

```
inputs.items | length     ->  offered;  wrapped form raises
                              ValueError("unknown filter 'length'")
he said "hi"\nthen left   ->  offered;  wrapped form resolves to None,  True -> False
```

The first replaces an always-true condition with a crash; the second inverts it.

## Two checks, both reading the evaluator rather than guessing

- `_unregistered_filter` walks the top-level `|` segments and reports the first name missing from `_REGISTERED_FILTERS` — the same tuple `_apply_filter` raises on, so the validator cannot bless a filter the evaluator will reject.
- `_reads_as_prose` reports a core that is several bare terms with nothing joining them. Quoted spans and bracketed groups are skipped, so `inputs.f('a b')` and `inputs.name == 'two words'` are unaffected, and a `not ` prefix is allowed.

## The fixture was mine, and it was wrong in two ways

`he said "hi"\nthen left` was only in the offered-correction set because I built that fixture as `TRICKY_CONDITIONS + [...]`. `TRICKY_CONDITIONS` exists to exercise the formatter's quoting and deliberately contains prose — reusing it asserted something it was never meant to claim. The list is explicit now, with the tricky-quoting entries that really are expressions carried over by hand, so adding prose to that fixture can no longer widen this invariant.

`inputs.tags | length > 0` was also mine, and `length` is not a registered filter. It is `join(',')` now. My own fixture was asserting a correction the evaluator would have crashed on.

## Verification — Python 3.11

```
tests/unit/test_condition_expression_block.py      212 passed  (was 207)
tests/unit + tests/test_workflows.py              1312 passed  (was 1307)
```

22 failures identical before and after — the `TestWorkflowCliAlignment` symlink tests needing elevation on Windows, confirmed against the stashed tree.

**Mutation-checked:** dropping either new gate fails 3 cases and nothing else.

## One thing worth your call, @mnriem

This is the fourth round in which the *suggested correction* — not the validation itself — was wrong in a new way. The rejection of never-evaluated conditions has been solid since #4182; it is the paste-ready suggestion that keeps needing another gate.

I think the gates are now in the right place, because the last two read the evaluator's own tables rather than restating structure. But if you would rather not carry that surface at all, dropping the suggestion and keeping only the diagnosis would delete four helpers and remove this whole class of defect permanently. Happy to cut it down if that is the call — it is your maintenance burden, not mine.

ntdatt812:
Thanks for the approval @mnriem — one more push after it, because the Copilot comment that arrived two minutes later was correct and I would rather not leave it. `a61ff6c`.

Two more shapes got through:

```
inputs.tags | join   ->  offered;  `join` is registered, but with no argument
                         _apply_filter raises ValueError
inputs.count+1       ->  offered;  the evaluator has no arithmetic, reads it as a
                         key named "count+1", wrapped form resolves to None,
                         True -> False
```

That is the fifth shape in four rounds, which is the argument against enumerating shapes. So this commit stops:

- **`_evaluator_rejects`** runs the core through `_evaluate_simple_expression` against a probe namespace and returns the evaluator's own error. Any filter under an unknown name or in an unsupported form is now reported by the code that will actually run. `_unregistered_filter`, which restated the filter table, is deleted.
- **`_is_not_a_bare_path`** covers the part a probe cannot: a single-term core is resolved as a path lookup, so every dotted segment must be an identifier. `count+1` is not, and neither is prose — so `_reads_as_prose` is deleted too.

The probe namespace resolves roots but not leaves on purpose. I tried the permissive version first, and a namespace that answers every lookup also answers `inputs.count+1` — it hides the exact shape the probe exists to expose.

Net: two helpers fewer, and no restatement of the evaluator's tables anywhere in the gate.

## Verification — Python 3.11

```
tests/unit/test_condition_expression_block.py      230 passed  (was 212)
tests/unit + tests/test_workflows.py              1330 passed  (was 1312)
```

22 failures identical before and after — the `TestWorkflowCliAlignment` symlink tests needing elevation on Windows, confirmed against the stashed tree.

**Mutation-checked:** dropping either check fails 6 cases and nothing else.

Since this landed after your approval, please re-approve or tell me to revert to `6c8ca2c` — I did not want to sit on a reproducible defect, but I also do not want to slip a change past a review you had already given.

mnriem:

ntdatt812:
@mnriem — two of the three were right and are fixed in `a2b1c8b`. The third I have to push back on, with the command output.

## 1. The probe rejected valid expressions — correct, and the worst finding so far

```
steps.emit.output.stdout | from_json      ->  refused
inputs.tags | join(inputs.separator)      ->  refused
```

Both are valid; the first is exercised in `tests/test_workflows.py`. The probe hands `from_json` a dict and it raises, so treating every probe error as a rejection blamed the author for the placeholder's type. That is a false positive — withholding help from a condition that was already correct — which is worse than the false negatives the earlier rounds fixed.

`_evaluator_rejects` now reports only the two failures `_apply_filter` raises about the expression itself: an unknown filter name, and a registered filter used in an unsupported form. Everything else a probe run raises is about probe values, not the author's text.

## 2. The suffix pattern was too permissive — correct

`_resolve_dot_path` matches `^([\w-]+)\[(\d+)\]$` — one numeric index. My pattern accepted any bracket contents and repeats, so `inputs.tags[foo]` and `inputs.matrix[0][1]` passed as paths, resolved to `None`, and were offered a correction that turns a truthy condition false.

`_PATH_SEGMENT` is that grammar now. It also replaces `str.isidentifier`, which was wrong in the other direction — the resolver allows a hyphen and a leading digit in a key name, so `inputs.2bad` is a legitimate lookup and my test asserting otherwise was wrong.

## 3. E305 — this one does not reproduce

```
$ ruff rule E305
warning: Selection `E305` has no effect because preview is not enabled.

$ ruff check --select E305 src/specify_cli/workflows/expressions.py
All checks passed!
```

`[tool.ruff.lint]` in `pyproject.toml` extends the default selection with `S602`/`S604`/`S605` only, and preview is off, so E305 never fires in `ruff check src tests`.

The blank lines were genuinely wrong, though, and are fixed. And the comment pointed at something real that it did not name: this branch had **added 3 ruff errors** — two `SIM102` and a `PIE810`/`UP037` on the new code. `ruff check` on this file is back to the 5 pre-existing errors on `main`, all in code this PR does not touch:

```
8:1   I001    (import block)        <- pre-existing
48, 67, 105, 123  TRY004           <- pre-existing
Found 5 errors.
```

## Verification — Python 3.11

```
tests/unit/test_condition_expression_block.py      236 passed  (was 230)
tests/unit + tests/test_workflows.py              1336 passed  (was 1330)
```

22 failures identical before and after — the `TestWorkflowCliAlignment` symlink tests needing elevation on Windows.

**Mutation-checked:** treating every probe error as a rejection fails 2 cases; loosening the path grammar fails 2. (The first time I ran the second mutation my `perl` edit silently did not apply and I nearly recorded it as uncaught — verified the constant actually changed before trusting the result.)

## Standing offer, sharpened

This is round six, and every round has been about the *suggested correction*, never about the validation itself — that has been correct since #4182 merged. If you would rather not carry this surface, say the word and I will cut the suggestion entirely: it deletes five helpers and ends this class permanently, leaving the diagnosis that was always the valuable part.

ntdatt812:
All three addressed in `b56486f`, and the first one explains why this has taken so many rounds.

```
inputs.a === inputs.b          ->  offered;  splits cleanly on `==`, evaluator reads
                                   `= inputs.b` as a path and resolves it to None
bogus == 'x'                   ->  offered;  unknown root, same result
inputs.payload | from_json()   ->  offered;  raises at run time
```

Every gate so far inspected only the shape it was written for, so an operand one level down walked past all of them.

## Recursing to the leaves

`_unresolvable_term` replaces `_is_not_a_bare_path` and walks operands the way `_evaluate_simple_expression` does — filters, then `or`/`and`/`not`, then comparisons — to the leaves. A leaf must be a literal or a dotted path rooted in `_NAMESPACE_ROOTS`, the roots `_build_namespace` actually supplies (`inputs`, `steps`, `item`, `fan_in`, `context`).

Both shapes above fall out of that without either being named, which is the property the previous rounds kept failing to reach.

## The filter split, done from the source this time

`_evaluator_rejects` now keeps only what `_apply_filter` raises about the filter *expression*. Reading the raise sites, those all quote the segment back as `got '| ...'`, while the value errors name the type they received:

```
from_json: expected '| from_json' with no arguments or trailing tokens, got '| from_json()'
from_json: expected a JSON string, got NoneType
join: expected a string separator, got NoneType
```

Under a probe, that type is the placeholder — not anything the author wrote. My previous prefix list missed the first line, and when I widened it by filter name it wrongly rejected `steps.emit.output.stdout | from_json` again, undoing the fix from the round before. The marker is derived from the messages rather than guessed.

## One case no review raised

`_find_top_level` matches `" and "` with literal spaces, so a newline before the keyword is not an operator:

```
{{ inputs.x == 1\nand inputs.name == 'abc' }}   ->  False
{{ inputs.x == 1 and inputs.name == 'abc' }}    ->  True
```

That condition was in my offered-correction fixture. It is a refusal case now.

## Verification — Python 3.11

```
tests/unit/test_condition_expression_block.py      253 passed  (was 236)
tests/unit + tests/test_workflows.py              1353 passed  (was 1336)
ruff check src/specify_cli/workflows/expressions.py   5 errors — the ones already on main
```

22 failures identical before and after — the `TestWorkflowCliAlignment` symlink tests needing elevation on Windows.

**Mutation-checked:** dropping the recursion fails 15 cases, dropping the namespace-root check fails 5, treating every probe error as a rejection fails 2. (The namespace-root mutation initially reported as uncaught because my `perl` edit silently did not apply — I check the constant actually changed before trusting a mutation result now, having been bitten by that once already on this PR.)

## The offer still stands

Round seven, still entirely about the suggested correction. I think recursing to the leaves is finally the right shape rather than another patch, because it is defined by the evaluator's grammar instead of by the examples found so far. But if you would rather not carry this surface, one word and I will cut the suggestion — five helpers go with it, and the diagnosis that was always the valuable part stays.

mnriem:
Lets get this one through the review and then if you can describe what specifically would make it easier / cleaner / faster then lets address that by describing it in an issue so we can asses it? Hope that is OK? Thanks!

mnriem:

ntdatt812:
@mnriem — two of the three are fixed in `967fa66`. The third I could not fix without doing the thing this PR has spent seven rounds backing away from, so I want to put a direct question to you rather than keep going.

## Fixed: my checks approximated the evaluator instead of matching it

```
1e3        ->  offered;  no "." so the evaluator calls int(), which fails, and it
                         falls through to a path lookup. float() alone accepted it.
'a' 'b'    ->  offered;  the evaluator requires the opening quote's match to be the
                         final character; first/last-character equality is not that.
inputs[0]  ->  offered;  _build_namespace hands back mappings, so an indexed root
                         resolves to None however the index is written.
```

All three are truthy before wrapping and `False` after. `_looks_numeric` and `_is_literal` now use the evaluator's own tests verbatim, and the root segment is matched without stripping an index off it first.

## Not fixed: `inputs.tags | join(5)`

`join` always raises for a non-string separator, so this is a real hole. But it is a **type** rule, and `_evaluator_rejects` deliberately ignores value errors because under a probe they usually describe the placeholder rather than the author's text — that is the fix from two rounds ago, which stopped `steps.emit.output.stdout | from_json` being wrongly rejected.

The two cannot be separated from the message:

```
join: expected a string separator, got int        <- the author's literal
join: expected a string separator, got NoneType   <- my probe's placeholder
```

They differ only in a type name the probe may have supplied. I tried keying on whether the filter segments reference a namespace root, but that misclassifies `steps.emit.output.stdout | from_json`, where the probe-caused error is about the filter *input* rather than its argument. Catching `join(5)` means encoding each filter's argument types in the validator.

## The question

Eight rounds now, every one of them about the *suggested correction*. The validation itself has been correct since #4182 merged; what keeps breaking is the paste-ready string offered alongside it, because getting it right means reimplementing the evaluator's grammar and now its type rules inside a validator.

So, plainly — which do you want?

1. **Keep the suggestion.** I encode the filter argument types next, and we accept that this surface tracks the evaluator by hand and will drift again.
2. **Drop the suggestion.** The error keeps naming the fault and stops handing back a string to paste. That deletes five helpers and the whole class of defect, permanently. `format_condition_correction` stays exported and tested for anyone who wants it.

I have offered (2) three times without wanting to remove a shipped behaviour unilaterally, so I would rather you decide. Either way I will do it in one pass.

## Verification — Python 3.11

```
tests/unit/test_condition_expression_block.py      267 passed  (was 253)
tests/unit + tests/test_workflows.py              1367 passed  (was 1353)
ruff check src/specify_cli/workflows/expressions.py   5 errors — the ones already on main
```

22 failures identical before and after — the `TestWorkflowCliAlignment` symlink tests needing elevation on Windows.

**Mutation-checked:** restoring the bare `float()` fails 2 cases; restoring the first/last-character quote test fails 3.

ntdatt812:
@mnriem Understood on both counts — through review first, and the broader design point as a separate issue. I will write that one up once this lands.

Both suppressed Copilot findings are addressed in `a5970cf`. I checked each against the evaluator before changing anything, because they point in opposite directions and one of them was described at the wrong entry point.

**List literals — the correction was withheld from a valid condition.**

`_unresolvable_term` resolved `['x', 'y']` as a path, so:

```
inputs.tag in ['x', 'y']
  -> No correction is offered because "['x', 'y']" is not a name the evaluator can resolve
  but {{ inputs.tag in ['x', 'y'] }} evaluates to True
```

`_evaluate_simple_expression` treats `[...]` as a term and recurses into the elements, so the leaf check now does the same, including the empty-segment skip that makes `[1, 2,]` be `[1, 2]` rather than `[1, 2, None]`.

The finding named `format_condition_correction`, which always wraps whatever it is handed; the gate is `_wrapping_would_not_repair`, reached through `format_condition_remediation`. Same defect, different entry point.

**Filter arguments — a correction was offered for something that raises.**

```
inputs.tags | join(bogus)
  -> Wrap the expression: "{{ inputs.tags | join(bogus) }}"
  but that raises ValueError: join: expected a string separator, got NoneType
```

`_apply_filter` evaluates the argument with `_evaluate_simple_expression` like any other operand. The check now parses it with the same `(\w+)\((.+)\)` pattern `_apply_filter` uses, and a form that pattern does not match is left to the evaluator probe rather than guessed at here.

One correction to the finding's reasoning, since it affects the fix: not every unresolvable argument raises. `default` tolerates the None, so `inputs.count | default(bogus)` evaluates fine and is withheld on the same policy that already withholds `bogus == 'x'` — which also evaluates fine. I split the tests accordingly rather than asserting a crash that does not happen; my first version of that test asserted it and failed.

## Verification

Windows, Python 3.13. Every case is asserted against what the evaluator does with the wrapped form, not against a restatement of the check, so a check that drifts from the evaluator fails the test rather than agreeing with itself.

```
tests/unit/test_condition_expression_block.py          280 passed
tests/test_extensions.py + test_workflows.py
  + test_condition_expression_block.py                 1718 passed, 22 failed
```

Those 22 are pre-existing and platform-bound. Measured on both sides — reverting this PR's two files to their parent revision gives **the identical 22 failures**, `diff` clean, with 1705 passed against 1718 here. The +13 is exactly the cases added. They are the same symlink and bash-parity classes that account for all 168 failures in the full local run (7005 passed): unelevated Windows cannot create symlinks, and the `*_python_parity` files shell out to the bash scripts.

**Mutation-checked**, after confirming each edit actually applied:

```
skip every filter argument      -> 3 failed  (join(bogus), map(bogus), default(bogus))
list literals fall through      -> 5 failed  (all five list-literal cases)
```

Nothing else moves in either case, so the new cases pin these two behaviours rather than passing incidentally.

mnriem:

ntdatt812:
Addressed in `b0c7704`. This one is a correction to something **I** got wrong earlier in this branch, not just a gap.

Two rounds ago I replaced an index-stripping root check with a plain membership test, on the reasoning that `_build_namespace` hands back mappings so an indexed root always resolves to None. That is true of four roots and false of the fifth. `StepContext.item` is `Any`, and a fan-out assigns the item value itself, so when the item is a list `_resolve_dot_path` takes its index branch and resolves:

```
item[0] == 'x'   item = ["x", "y"]   ->  evaluates True    correction offered: False   <- wrong
inputs[0]        always a mapping    ->  evaluates False   correction offered: False   <- right
steps[1]         always a mapping    ->  evaluates False   correction offered: False   <- right
fan_in[0]        always a mapping    ->  evaluates False   correction offered: False   <- right
```

So the index is stripped for `item` alone rather than for roots in general, and there is a paired test for the other four so the narrowing cannot later widen into "any indexed root". Both directions are asserted against what the evaluator returns, not against the check.

I should have caught this when I wrote that line — the comment I put above it asserted a property of all five roots that I had only verified for `inputs`.

## Verification

```
tests/unit/test_condition_expression_block.py           286 passed
tests/test_extensions.py + test_workflows.py
  + test_condition_expression_block.py                  1724 passed, 22 failed
```

Same 22 as the unpatched tree, `diff` clean against the baseline I measured for the previous push — the pre-existing symlink and bash-parity classes on unelevated Windows. 1705 -> 1724 passed is exactly the 19 cases added across both pushes.

**Mutation-checked**, after confirming the edit applied:

```
never strip the index   ->  2 failed  (item[0], item[1])
```

The always-mapping cases stay green under that mutation, which is the point — they do not depend on the strip, so they still guard the other side.

mnriem:
Thank you!

Issue #4213, opened 2026-08-19
[Feature]: Add --json output to preset info and extension info with fully-expanded per-contribution detail (per-pack detail view; complements list --json summary counts)


nicolehaugen:
### Problem Statement

Today `specify preset info <id>` and `specify extension info <id>` emit only text. `speckit-wizard-canvas` reads the raw `preset.yml` / `extension.yml` files with `js-yaml` and re-normalizes the shape itself in `composition/collect.mjs::parseProvidesEntries` and `::parseHookDeclarations` — including strategy inference from shorthand keys (`replaces` vs. `wraps` vs. `prepends` vs. `appends`), hook-phase normalization (`phase` vs. `trigger`, `command` vs. `targetCommand`), and script-runtime inference by globbing `scripts/{bash,powershell,python}/*.{sh,ps1,py}`. All of this is server-side data being reconstructed on the client.

Downstream consumers need the structured shape returned by the CLI itself so they can stop reconstructing it.

**How this differs from the companion `preset list --json` / `extension list --json` issue.** The `list` variant operates on **the collection of installed packs** and returns a **JSON array**, one row per pack, where `provides` is **just integer counts** (`{ commands: 4, templates: 2, scripts: 1, hooks: 3 }`) — a summary/catalog view suitable for "here's every preset the project has". `info` operates on **one pack, addressed by id**, and returns a **single JSON object** where `provides` is **fully expanded** — every command, template, script, and hook enumerated with its full per-contribution schema (`id`, `name`, `description`, `artifact`, `optional`, `handoffs`, `strategy`, `sourcePath`, `runtimes`, etc.). A wizard rendering "here's what `speckit.git` contributes to my project" needs the `info` detail; `list`'s counts are not enough. Both surfaces are required; neither is a subset of the other.

### Proposed Solution

Add `--json` to both `info` commands. Output shape:

```json
{
  "id": "…", "name": "…", "description": "…", "version": "…",
  "author": "…", "priority": 100, "enabled": true, "source": { "…": "…" },
  "commands": [
    { "id": "…", "name": "speckit.plan", "description": "…",
      "artifact": "specs/{feature}/plan.md", "optional": false,
      "handoffs": [ { "to": "speckit.tasks", "when": "…", "message": "…" } ],
      "strategy": "wrap",
      "source": { "layer": "preset", "presetId": "…" },
      "sourcePath": "commands/speckit.plan.md" }
  ],
  "templates": [
[...]

Issue #4212, opened 2026-08-19
[Feature]: Add specify artifact command exposing composition stacks as JSON


nicolehaugen:
### Problem Statement

`PresetResolver.collect_all_layers()` (`presets/__init__.py:5460`) and `resolve_content()` (`presets/__init__.py:5688`) already implement the composition math needed to expose an artifact's layer stack — ordered layer stack, per-layer strategy, winner selection, and shadowing — but nothing serializes it. `speckit-wizard-canvas` resorts to shelling `specify preset resolve <id>` and parsing text output one line at a time (`composition/collect.mjs::batchResolveTemplates`), and re-implements strategy resolution and shadowing itself in `composition/assembler.mjs` (~23 KB). The wizard's own header comment on `composition/collect.mjs` says:

> "Once the CLI grows equivalent commands (e.g. `specify composition list --json`), this whole file goes away and the scanner calls those commands directly."

This is a direct in-code TODO for this issue.

### Proposed Solution

Add a new `specify artifact` command group:

- `specify artifact list --json` — every artifact Spec Kit knows about (commands, templates, scripts), one row each:

  ```json
  { "id": "command:speckit.constitution", "name": "speckit.constitution", "kind": "command",
    "description": "…" }
  ```

- `specify artifact info <name> --json` — the full stack for one artifact:

  ```json
  {
    "id": "command:speckit.constitution", "name": "speckit.constitution", "kind": "command",
    "description": "…",
    "stack": [
      { "id": "command:speckit.constitution",
        "layer": "preset", "sourceId": "speckit.corp",
        "presetId": "speckit.corp", "presetName": "Corp Preset",
        "strategy": "wrap", "active": true, "hidden": false,
        "manifestPath": ".specify/presets/speckit.corp/preset.yml",
        "lookupId": "preset:speckit.corp:command:speckit.constitution" },
      { "id": "command:speckit.constitution",
        "layer": "preset", "sourceId": "speckit.core",
        "presetId": "speckit.core", "presetName": "…",
        "strategy": "replace", "active": false, "hidden": true,
        "manifestPath": ".specify/presets/speckit.core/preset.yml",
        "lookupId": "preset:speckit.core:command:speckit.constitution" },
      { "id": "command:speckit.constitution",
        "layer": null, "sourceId": null,
        "presetId": null, "presetName": null,
        "strategy": "replace", "active": false, "hidden": true,
        "manifestPath": null, "lookupId": null }
    ]
  }
  ```

Stack rules:

[...]

Issue #4210, opened 2026-08-19
[Feature]: Add deterministic contribution IDs and stack lookup IDs for resolved artifacts


nicolehaugen:
### Problem Statement

Spec Kit identifies contributed commands, templates, scripts, and hooks primarily by `name`. Names can collide across artifact kinds and source layers, and a name alone cannot reliably link a resolved artifact-stack layer to the exact manifest contribution that supplied it. Consumers such as `preset info --json`, `extension info --json`, and `specify artifact` need stable contribution identifiers and a join key that works across reinstalls and machines.

### Proposed Solution

Introduce computed, opaque `id` fields for every command, template, script, and hook returned by public preset/extension manifest and info APIs, plus a `lookupId` field on every provenance-backed non-built-in resolved artifact-stack layer.

For named preset, extension, and project-override contributions, derive IDs using:

```text
{layer}:{sourceId}:{kind}:{name}
```

Use `project`, `preset`, or `extension` for `layer`; `_` only for the project-override source ID; and the preset or extension manifest ID for manifest-declared sources. Use `command`, `template`, or `script` for `kind`. Examples include `preset:speckit.core:command:speckit.plan`, `extension:speckit.git:template:pr-body`, and `project:_:template:spec-template`.

For hooks, use `{eventName}:{command}` as the name component. Hooks are valid only for preset and extension layers. If duplicate event/command hook entries are valid, add an appropriate stable discriminator or reject duplicates; do not use array position.

Built-in artifacts do not have an originating manifest contribution and therefore do not receive a contribution `lookupId`. Every artifact, including built-ins, instead has a source-agnostic public ID of the form `{kind}:{name}`. Built-in stack rows are recognized by absent provenance fields (`layer`, `sourceId`, and `lookupId` are null), and round-trip through the public artifact ID.

Compute IDs at read, serialization, or resolution time rather than persisting them in authored or installed manifests. For manifest-declared preset and extension layers, `lookupId` must exactly match the originating contribution's `id`. Preserve existing name-based behavior and document IDs as opaque strings.

### Alternatives Considered

Install-time UUIDs are unsuitable because they differ across machines and reinstalls. Existing names are insufficient because they collide across sources and kinds. Content hashes are unsuitable because IDs would change whenever artifact content is edited. Array indexes are unsuitable for hooks because reordering entries would change their IDs.

[...]

PR #4240, opened 2026-08-20
feat(core): add `specify core info --json` command

### LINKED ISSUE #4215: [Feature]: Add specify core inventory command exposing the baked-in commands, templates, and scripts as JSON
### Problem Statement

Consumers of SpecKit sometimes need "the baseline set of commands, templates, and scripts baked into SpecKit itself — what a project has with zero presets and zero extensions installed." Command names are already enumerable via `CORE_COMMAND_NAMES` at `extensions/__init__.py:119` — and, importantly, this is not a static list. It's populated by `_load_core_command_names()` at `extensions/__init__.py:78-108`, which **already dynamically scans** `core_pack/commands` (wheel-time) or `templates/commands` (source checkout) at import time, using the canonical `_locate_core_pack()` / `_repo_root()` asset resolvers. Current commands in the baseline (from that scan): `analyze`, `checklist`, `clarify`, `constitution`, `converge`, `implement`, `plan`, `specify`, `tasks`, `taskstoissues`.

There is **no equivalent scan for the shipped templates or scripts**, so consumers who want the baseline have to either:

1. Enumerate the filesystem directly (fragile — internal directory layout isn't a stable contract), or
2. Maintain their own hardcoded list. `speckit-wizard-canvas` does exactly this — `composition/collect.mjs::loadCoreInventory` imports `pipeline/canonical.mjs`, which is a hand-maintained copy of what SpecKit ships. Every SpecKit release risks drift.

A companion issue (`specify artifact info --json`, filed separately) already surfaces core as `layer: "core"` rows inside each artifact's stack, but that's a *per-artifact* view. For "what does a bare-bones SpecKit look like?" a consumer needs an *inventory* view too.

### Proposed Solution

Add a `specify core inventory --json` command that returns the full baseline inventory:

```json
{
  "commands": [
    { "id": "core:_:command:speckit.plan",
      "name": "speckit.plan",
      "description": "…",
      "sourcePath": "<internal-path-relative-to-package>",
      "artifact": "specs/{feature}/plan.md",
      "optional": false,
      "handoffs": [ "…" ] }
  ],
  "templates": [
    { "id": "core:_:template:constitution",
      "name": "constitution",
      "description": "…",
      "sourcePath": "…" }
  ],
  "scripts": [
    { "id": "core:_:script:setup",
      "name": "setup",
      "description": "…",
      "sourcePath": "…",
      "runtimes": ["bash", "powershell", "python"] }
  ]
}
```

The entry shape mirrors the per-contribution schema returned by the companion `preset info --json` / `extension info --json` issue, minus `strategy` (core is always the base layer). Under the hood the command:

- Reuses `CORE_COMMAND_NAMES` for the command list.
- **Extends the existing file-scan pattern** used by `_load_core_command_names()` — same `_locate_core_pack()` + `_repo_root()` canonical asset resolvers — to scan `core_pack/templates` and `core_pack/scripts` (with the equivalent source-checkout fallbacks under `templates/` and `scripts/`). No new asset-discovery paths.
- Runs a small frontmatter parser over the enumerated command / template markdown files (and reads script headers), extracting `name` / `description` / `artifact` / `optional` / `handoffs` / (script) `runtimes`, and emits the JSON.
- Uses the same stable-`id` scheme (`core:_:kind:name`) as the companion `specify artifact` issue so consumers can join this output to that command's `layer: "core"` stack rows via `lookupId`.

### Alternatives Considered

- Fold entirely into the companion `specify artifact` command's `layer: "core"` rows. Rejected — that command gives per-artifact stacks; this one gives the flat inventory needed for "here's what you get out of the box" listings. Both views point at the same data with the same `id`.
- Return only names (no `description` / `sourcePath` / `handoffs`). Rejected — the wizard's `pipeline/canonical.mjs` needs the metadata, which is why it exists in the first place.
- Ship the inventory as a static JSON file inside the package. Rejected — still requires the wizard to know the package path; a CLI command is more stable.

### Component

### AI Agent (if applicable)

Not applicable

### Use Cases

1. `speckit-wizard-canvas` deletes `pipeline/canonical.mjs` and replaces `composition/collect.mjs::loadCoreInventory` with `JSON.parse(execFileSync("specify", ["core", "inventory", "--json"]))`.
2. A "SpecKit tour" UI shows first-time users exactly which commands / templates / scripts they get with zero installs.
3. A CI check asserts "no preset is shadowing a core artifact I depend on" by cross-referencing this inventory with the companion `specify artifact info --json` stacks and the companion `preset list --json` installed set.

### Acceptance Criteria

- [ ] `specify core inventory --json` emits `{commands, templates, scripts}` arrays; each entry has `id`, `name`, `description`, `sourcePath`, and the kind-specific fields (`artifact`/`optional`/`handoffs` on commands, `runtimes` on scripts).
- [ ] `id` values follow the `core:_:kind:name` scheme (matching the stable-id scheme used by the companion `specify preset info --json` / `specify extension info --json` / `specify artifact info --json` commands) so consumers can join this inventory to those commands' output via `lookupId`.
- [ ] Command list is populated from `CORE_COMMAND_NAMES`; template and script lists come from a scan that **reuses the existing `_locate_core_pack()` / `_repo_root()` canonical asset resolvers** — not `Path(__file__)` arithmetic. (Path-arithmetic discovery has silently broken in this codebase once already; see #3014 / #3274.) No filesystem paths outside the package leak into the output.
- [ ] Deterministic: same SpecKit version always produces identical output (stable order + stable ids).
- [ ] Non-zero exit + stderr JSON error on packaging problems (missing shipped file, malformed frontmatter).
- [ ] Tests: enumeration parity with `CORE_COMMAND_NAMES` for commands; count parity with the shipped file set for templates/scripts; `lookupId` cross-reference with the companion `specify artifact info --json` output on a bare-bones project.
- [ ] Docs: new `specify core inventory` section in the CLI reference plus a note that it complements the companion `specify artifact` command.

### Additional Context

Direct replacement for `plugins/spec-kit-copilot-wizard/extensions/speckit-wizard-canvas/pipeline/canonical.mjs` and `composition/collect.mjs::loadCoreInventory` in `github/spec-kit-copilot`. Depends on the companion "stable id / lookupId" issue; benefits from the companion "artifact/optional/handoffs" issue.

nicolehaugen:
Closes #4215

Introduces `specify core info --json` — a read-only command that emits Spec Kit's baked-in inventory (commands, templates, scripts) as a deterministic JSON document. Downstream tools (starting with `speckit-wizard-canvas`) can now consume the baseline programmatically instead of maintaining a hand-copied duplicate.

## What ships

- **New package** `src/specify_cli/core/` with `build_core_inventory()` + `CoreInventoryError`, wired to the existing `_locate_core_pack()` / `_repo_root()` helpers so wheel-installed and source-checkout layouts both work.
- **New Typer group** `specify core` with one leaf command, `info --json`. `--json` is required by design (the surface is programmatic); invoking without it exits 2 with a pointer at the flag.
- **Deterministic output**: alphabetically sorted by `name` within each section; fixed top-level key order (`commands`, `templates`, `scripts`); forward-slash source paths on every platform; byte-identical across runs.
- **Stable ids** follow the [#4210](https://github.com/github/spec-kit/issues/4210) grammar `core:_:<kind>:<name>`, so the output is join-ready against the future `specify artifact info` surface tracked in [#4212](https://github.com/github/spec-kit/issues/4212).
- **Fail-fast on packaging errors**: JSON envelope on stderr and exit 1 on missing files, unparseable frontmatter, wheel/source mismatches, or type mismatches — no partial output ever escapes.
- **Docs** at `docs/reference/core-info.md`, wired into `docs/toc.yml`; `CHANGELOG.md` entry added.

## Tests

- `tests/test_core_inventory.py` — unit tests for path resolution, frontmatter parsing, typed empty defaults, handoffs normalization, partial script runtimes, template-description fallback, determinism, sorting, stable-id grammar, and every fail-fast branch.
- `tests/test_core_info_command.py` — CLI-level tests via `CliRunner`: exit codes, help surface, schema-conformance, determinism, path shape, project-isolation (runs outside a Spec Kit project and ignores installed presets), and fail-fast envelopes.
- `tests/contract/test_core_inventory_wheel.py` — count/name parity between the source-checkout resolution and a simulated wheel-`core_pack/` layout.

> ⚠️ **Local test run was blocked** by a TLS handshake failure preventing `uv sync --extra test` (or `pip install pytest`) from installing pytest into the venv. All logic was validated with inline `python -c` scripts against the live repo, but pytest itself did not run in this environment — I'm relying on CI to execute the test suites.

## Two documented deviations

Both are called out in `docs/reference/core-info.md`:

1. Four of six shipped helper scripts (`create-new-feature`, `setup-plan`, `setup-tasks`, `resolve-template`) currently have no header comment. The command emits a filename-derived fallback description (`"Baseline helper script: <name>."`) rather than fail hard. Adding real headers to those scripts is left as a follow-up.
2. Real-world frontmatter for `handoffs:` in `templates/commands/*.md` is a list of mappings with `agent`, `label`, and `prompt` keys — not a bare list of strings. The inventory surfaces only the `agent` values (`list[str]`) to match the schema; consumers that need labels/prompts should read the frontmatter directly.

## Companion issues

- Uses the id grammar from [#4210](https://github.com/github/spec-kit/issues/4210).
- Complements the `preset info` / `extension info` JSON surfaces from [#4213](https://github.com/github/spec-kit/issues/4213).
- The forward-looking join against project-scoped artifacts is deferred to [#4212](https://github.com/github/spec-kit/issues/4212); a `TODO(#4212)` marker is left at the top of `tests/test_core_info_command.py`.

---

*Assisted-by: GitHub Copilot (model: claude-opus-4.7, supervised) — every commit on this branch carries the same trailer.*

### DIFF
--- a/docs/reference/core-info.md
+++ b/docs/reference/core-info.md
@@ -31,7 +31,7 @@ The response is a JSON object with three top-level arrays, always in this order 
 Every entry carries a stable identifier suitable for cross-referencing:
 
 - `id`: `core:_:<kind>:<name>` (grammar defined in [#4210](https://github.com/github/spec-kit/issues/4210); `<kind>` is `command`, `template`, or `script`)
-- `name`: the base name (no extension, hyphenated form for scripts — e.g. `setup-plan` not `setup_plan`)
+- `name`: the logical artifact name. Commands are namespaced as `speckit.<stem>`; templates use their filename stem; scripts use a hyphenated stem (for example, `setup-plan`, not `setup_plan`).
 - `description`: short human-readable summary
 - `sourcePath`: package-relative, forward-slash path (e.g. `templates/commands/plan.md`)
 
--- a/docs/reference/core-info.md
+++ b/docs/reference/core-info.md
@@ -68,7 +68,7 @@ specify core info --json | jq '.commands[] | select(.name == "speckit.plan")'
 
 - Alphabetical ordering by `name` within each section.
 - Fixed top-level key order: `commands`, `templates`, `scripts`.
-- Fields inside each entry are emitted in a fixed order (see the schema at [`specs/001-core-info/contracts/core-info-schema.json`](https://github.com/github/spec-kit/blob/main/specs/001-core-info/contracts/core-info-schema.json)).
+- Fields inside each entry are emitted in a fixed order (see the entry construction in [`src/specify_cli/core/__init__.py`](https://github.com/github/spec-kit/blob/main/src/specify_cli/core/__init__.py)).
 - Path separators are always forward-slashes, even on Windows.
 
 ## Failure Modes
--- a/docs/reference/core-info.md
+++ b/docs/reference/core-info.md
@@ -30,7 +30,7 @@ The response is a JSON object with three top-level arrays, always in this order 
 
 Every entry carries a stable identifier suitable for cross-referencing:
 
-- `id`: `core:_:<kind>:<name>` (grammar defined in [#4210](https://github.com/github/spec-kit/issues/4210); `<kind>` is `command`, `template`, or `script`)
+- `id`: `core:_:<kind>:<name>` (`<kind>` is `command`, `template`, or `script`)
 - `name`: the logical artifact name. Commands are namespaced as `speckit.<stem>`; templates use their filename stem; scripts use a hyphenated stem (for example, `setup-plan`, not `setup_plan`).
 - `description`: short human-readable summary
 - `sourcePath`: package-relative, forward-slash path (e.g. `templates/commands/plan.md`)
@@ -62,43 +62,37 @@ specify core info --json | jq '.commands[] | select(.name == "speckit.plan")'
 
 - Building tools that need to know what the shipped baseline contains without unpacking the wheel by hand.
 - Auditing which commands, templates, or scripts a given Spec Kit release ships.
-- Cross-referencing baseline artifacts against extension- or preset-provided ones. A companion command exposing the equivalent view for project-scoped artifacts is tracked in [#4212](https://github.com/github/spec-kit/issues/4212).
+- Cross-referencing baseline artifacts against extension- or preset-provided ones.
 
 ## Determinism Guarantees
 
 - Alphabetical ordering by `name` within each section.
 - Fixed top-level key order: `commands`, `templates`, `scripts`.
-- Fields inside each entry are emitted in a fixed order (see the entry construction in [`src/specify_cli/core/__init__.py`](https://github.com/github/spec-kit/blob/main/src/specify_cli/core/__init__.py)).
+- Fields inside each entry are emitted in a fixed order.
 - Path separators are always forward-slashes, even on Windows.
 
 ## Failure Modes
 
-The command is designed to fail fast rather than silently emit a partial or malformed inventory. On any packaging error — a missing shipped file, unparseable frontmatter, or a mistyped field — it exits with code `1` and prints a JSON error envelope on stderr:
+The command fails fast on packaging errors it can detect at read time — unparseable frontmatter, a mistyped field, or a script missing its canonical bash variant — rather than silently emitting a malformed inventory. On any such error it exits with code `1` and prints a JSON error envelope on stderr:
 
 ```json
 {
-  "error": "core_inventory.missing_file",
-  "message": "Baseline command 'plan' is registered but its source file is missing.",
-  "artifact": { "kind": "command", "name": "plan" }
+  "error": "core_inventory.frontmatter_parse",
+  "message": "Could not parse YAML frontmatter of command 'plan': ...",
+  "artifact": { "kind": "command", "name": "plan", "sourcePath": "templates/commands/plan.md" }
 }
 ```
 
 Common `error` codes:
 
-| Code                                 | Meaning                                                        |
-| ------------------------------------ | -------------------------------------------------------------- |
-| `core_inventory.assets_missing`      | Neither the wheel-shipped `core_pack/` nor the source checkout was found. |
-| `core_inventory.missing_file`        | A baseline command file expected to be shipped is absent.      |
-| `core_inventory.frontmatter_parse`   | YAML frontmatter on a baseline file could not be parsed.       |
-| `core_inventory.frontmatter_shape`   | Frontmatter parsed but a field has the wrong type.             |
-| `core_inventory.missing_canonical`   | A script ships a non-bash variant but no canonical bash one.   |
-| `core_inventory.invalid_source_path` | An emitted `sourcePath` failed the shape check (absolute or contained a backslash). |
+| Code                                | Meaning                                                        |
+| ----------------------------------- | -------------------------------------------------------------- |
+| `core_inventory.assets_missing`     | Neither the wheel-shipped `core_pack/` nor the source checkout was found. |
+| `core_inventory.frontmatter_parse`  | YAML frontmatter on a baseline file could not be parsed, or a field has the wrong type. |
+| `core_inventory.missing_description`| A baseline command, template, or script has no usable description. |
+| `core_inventory.missing_canonical`  | A script ships a non-bash variant but no canonical bash one.   |
+| `core_inventory.invalid_source_path`| An emitted `sourcePath` failed the shape check (absolute or contained a backslash). |
 
 These conditions represent packaging bugs and should be reported.
 
-## Known Deviations
-
-Two small deviations from the specification are noted here for transparency:
-
-- **Script descriptions**: Only 2 of the 6 shipped scripts (`check-prerequisites`, `common`) currently have header comments. For the other four (`create-new-feature`, `setup-plan`, `setup-tasks`, `resolve-template`), the command emits a filename-derived fallback (`"Baseline helper script: <name>."`). Adding real header comments to those scripts is tracked separately.
-- **Command `handoffs`**: The source frontmatter for `handoffs:` is a list of mappings with `agent`, `label`, and `prompt` keys. The core inventory surfaces only the `agent` values (a list of strings) — the same shape the schema exposes. Consumers that need the labels and prompts should read the frontmatter directly.
+Note: templates and scripts are enumerated from whatever files are present on disk, so an asset that is missing entirely from the installation is simply omitted from the output rather than raised as an error.
--- a/docs/reference/core-info.md
+++ b/docs/reference/core-info.md
@@ -73,7 +73,7 @@ specify core info --json | jq '.commands[] | select(.name == "speckit.plan")'
 
 ## Failure Modes
 
-The command fails fast on packaging errors it can detect at read time — unparseable frontmatter, a mistyped field, or a script missing its canonical bash variant — rather than silently emitting a malformed inventory. On any such error it exits with code `1` and prints a JSON error envelope on stderr:
+The command fails fast on packaging errors it can detect at read time — unparseable frontmatter, a mistyped field, or a script missing its canonical bash variant — rather than silently emitting a malformed inventory. On any such error it exits with code `1` and prints a JSON error envelope on stderr. Note that templates and scripts are enumerated from whatever files are present on disk, so an asset that is missing entirely from the installation is simply omitted from the output rather than raised as an error.
 
 ```json
 {
@@ -94,5 +94,3 @@ Common `error` codes:
 | `core_inventory.invalid_source_path`| An emitted `sourcePath` failed the shape check (absolute or contained a backslash). |
 
 These conditions represent packaging bugs and should be reported.
-
-Note: templates and scripts are enumerated from whatever files are present on disk, so an asset that is missing entirely from the installation is simply omitted from the output rather than raised as an error.
--- a/docs/reference/core-info.md
+++ b/docs/reference/core-info.md
@@ -1,10 +1,10 @@
-# Core Inventory
+# Core Info
 
 The `specify core` command group exposes read-only information about Spec Kit's baked-in ("core") baseline — the commands, templates, and helper scripts that ship inside the CLI package itself. Presets, extensions, integrations, and project-scoped assets are intentionally **not** included; use their dedicated commands (`specify preset`, `specify extension`, `specify integration`) for those.
 
-Core-inventory output is deterministic and byte-identical across runs, so downstream tools can safely diff, hash, or cache it.
+Core info output is deterministic and byte-identical across runs, so downstream tools can safely diff, hash, or cache it.
 
-## List the Core Inventory
+## Inspect Core Info
 
 ```bash
 specify core info --json
@@ -12,7 +12,7 @@ specify core info --json
 
 | Option        | Description                                                          |
 | ------------- | -------------------------------------------------------------------- |
-| `--json`      | **Required.** Emit the inventory as JSON on stdout.                  |
+| `--json`      | **Required.** Emit core info as JSON on stdout.                      |
 
 The `--json` flag is mandatory. A human-readable table format is intentionally not offered — the command is designed for programmatic consumption by external tools and other Spec Kit surfaces. Invoking `specify core info` without `--json` exits non-zero with a hint pointing at the flag.
 
@@ -68,7 +68,7 @@ specify core info --json | jq '.commands[] | select(.name == "plan")'
 
 - Alphabetical ordering by `name` within each section.
 - Fixed top-level key order: `commands`, `templates`, `scripts`.
-- Fields inside each entry are emitted in a fixed order (see the schema at [`specs/001-core-inventory/contracts/core-info-schema.json`](https://github.com/github/spec-kit/blob/main/specs/001-core-inventory/contracts/core-info-schema.json)).
+- Fields inside each entry are emitted in a fixed order (see the schema at [`specs/001-core-info/contracts/core-info-schema.json`](https://github.com/github/spec-kit/blob/main/specs/001-core-info/contracts/core-info-schema.json)).
 - Path separators are always forward-slashes, even on Windows.
 
 ## Failure Modes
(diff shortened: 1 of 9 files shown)


docs/reference/core-info.md
```diff
@@ -0,0 +1,104 @@
...
+## Determinism Guarantees
+
+- Alphabetical ordering by `name` within each section.
+- Fixed top-level key order: `commands`, `templates`, `scripts`.
+- Fields inside each entry are emitted in a fixed order (see the entry construction in [`src/specify_cli/core/__init__.py`](https://github.com/github/spec-kit/blob/main/src/specify_cli/core/__init__.py)).
+- Path separators are always forward-slashes, even on Windows.
+
+## Failure Modes
+
+The command is designed to fail fast rather than silently emit a partial or malformed inventory. On any packaging error — a missing shipped file, unparseable frontmatter, or a mistyped field — it exits with code `1` and prints a JSON error envelope on stderr:
```
nicolehaugen: This comment is no longer accurate - it doesn't catch every packaging error, such as if the installed commands are "corrupt" by deleting them from disk after the commands are inititally detected.

docs/reference/core-info.md
```diff
@@ -0,0 +1,104 @@
...
+  "optional": false,
+  "handoffs": ["speckit.tasks", "speckit.checklist"]
+}
+```
+
+## When to Use It
+
+- Building tools that need to know what the shipped baseline contains without unpacking the wheel by hand.
+- Auditing which commands, templates, or scripts a given Spec Kit release ships.
+- Cross-referencing baseline artifacts against extension- or preset-provided ones. A companion command exposing the equivalent view for project-scoped artifacts is tracked in [#4212](https://github.com/github/spec-kit/issues/4212).
```
nicolehaugen: Remove reference to other issue.

docs/reference/core-info.md
```diff
@@ -0,0 +1,104 @@
...
+{
+  "commands":  [ /* baseline slash-commands (e.g. speckit.plan) */ ],
+  "templates": [ /* markdown templates (spec, plan, tasks, ...) */ ],
+  "scripts":   [ /* helper scripts (bash/powershell/python)      */ ]
+}
+```
+
+Every entry carries a stable identifier suitable for cross-referencing:
+
+- `id`: `core:_:<kind>:<name>` (grammar defined in [#4210](https://github.com/github/spec-kit/issues/4210); `<kind>` is `command`, `template`, or `script`)
```
nicolehaugen: Remove reference to issue - any information that is valuable to include here, needs to be explicitly added concisely.

docs/reference/core-info.md
```diff
@@ -0,0 +1,104 @@
...
+
+- Building tools that need to know what the shipped baseline contains without unpacking the wheel by hand.
+- Auditing which commands, templates, or scripts a given Spec Kit release ships.
+- Cross-referencing baseline artifacts against extension- or preset-provided ones. A companion command exposing the equivalent view for project-scoped artifacts is tracked in [#4212](https://github.com/github/spec-kit/issues/4212).
+
+## Determinism Guarantees
+
+- Alphabetical ordering by `name` within each section.
+- Fixed top-level key order: `commands`, `templates`, `scripts`.
+- Fields inside each entry are emitted in a fixed order (see the entry construction in [`src/specify_cli/core/__init__.py`](https://github.com/github/spec-kit/blob/main/src/specify_cli/core/__init__.py)).
```
nicolehaugen: remove file refernce.

docs/reference/core-info.md
```diff
@@ -0,0 +1,104 @@
...
+- Path separators are always forward-slashes, even on Windows.
+
+## Failure Modes
+
+The command is designed to fail fast rather than silently emit a partial or malformed inventory. On any packaging error — a missing shipped file, unparseable frontmatter, or a mistyped field — it exits with code `1` and prints a JSON error envelope on stderr:
+
+```json
+{
+  "error": "core_inventory.missing_file",
```
nicolehaugen: I don't believe this is applicable any more.  If a file is missing, it simply won't be included.

docs/reference/core-info.md
```diff
@@ -0,0 +1,104 @@
...
+  "message": "Baseline command 'plan' is registered but its source file is missing.",
+  "artifact": { "kind": "command", "name": "plan" }
+}
+```
+
+Common `error` codes:
+
+| Code                                 | Meaning                                                        |
+| ------------------------------------ | -------------------------------------------------------------- |
+| `core_inventory.assets_missing`      | Neither the wheel-shipped `core_pack/` nor the source checkout was found. |
```
nicolehaugen: Update this to remove those error codes that are no longer applicable based on the actual implementation - such as missing file no longer being applicable.

docs/reference/core-info.md
```diff
@@ -0,0 +1,104 @@
...
+| `core_inventory.missing_file`        | A baseline command file expected to be shipped is absent.      |
+| `core_inventory.frontmatter_parse`   | YAML frontmatter on a baseline file could not be parsed.       |
+| `core_inventory.frontmatter_shape`   | Frontmatter parsed but a field has the wrong type.             |
+| `core_inventory.missing_canonical`   | A script ships a non-bash variant but no canonical bash one.   |
+| `core_inventory.invalid_source_path` | An emitted `sourcePath` failed the shape check (absolute or contained a backslash). |
+
+These conditions represent packaging bugs and should be reported.
+
+## Known Deviations
```
nicolehaugen: Remove the known deviations section.

src/specify_cli/core/__init__.py
```diff
@@ -0,0 +1,541 @@
+"""Core (baseline) asset inventory for the specify-cli package.
+
+Public entry point: :func:`build_core_inventory` returns a fixed-shape dict
+enumerating the baseline commands, templates, and scripts that ship inside
+the wheel (`specify_cli/core_pack/`) or the source checkout (repo root).
+
+The output is the exact shape validated by
+``specs/001-core-info/contracts/core-info-schema.json``. See
```
nicolehaugen: Remove this reference to the .json file that was used for specs.

src/specify_cli/core/__init__.py
```diff
@@ -0,0 +1,541 @@
...
+``specs/001-core-info/data-model.md`` for authoritative field types.
```
nicolehaugen: Remove reference to this file.

src/specify_cli/core/__init__.py
```diff
@@ -0,0 +1,541 @@
...
+_RUNTIMES: tuple[tuple[str, str, str], ...] = (
+    # (runtime name, source-layout subdir, filename suffix)
+    ("bash", "bash", ".sh"),
+    ("powershell", "powershell", ".ps1"),
+    ("python", "python", ".py"),
+)
+
+
+class CoreInventoryError(Exception):
+    """Raised when the shipped baseline is inconsistent (FR-012)."""
```
nicolehaugen: Remove requirement id references

src/specify_cli/core/__init__.py
```diff
@@ -0,0 +1,541 @@
...
+
+@dataclass(frozen=True)
+class _CoreLayout:
+    """Filesystem layout for the resolved core-asset base.
+
+    Two shapes are supported. Wheel install: ``core_pack/commands/``,
+    ``core_pack/templates/``, ``core_pack/scripts/``. Source checkout:
+    ``templates/commands/``, ``templates/``, ``scripts/``. The emitted
+    ``sourcePath`` is always the source-layout form so wheel and source
+    outputs are byte-identical (FR-010, SC-002).
```
nicolehaugen: remove requirement references

src/specify_cli/core/__init__.py
```diff
@@ -0,0 +1,541 @@
...
+
+def build_core_inventory() -> dict[str, Any]:
+    """Return the baseline inventory as a fixed-shape dict.
+
+    Shape (validated by ``contracts/core-info-schema.json``):
+    ``{"commands": [...], "templates": [...], "scripts": [...]}``.
+
+    All three top-level lists are sorted alphabetically by ``name``. Two
+    consecutive calls MUST return byte-identical JSON when serialized
+    with ``json.dumps(..., sort_keys=False)`` (SC-002).
```
nicolehaugen: remove requirement id references

src/specify_cli/core/__init__.py
```diff
@@ -0,0 +1,541 @@
...
+
+    Raises :class:`CoreInventoryError` on any packaging inconsistency
+    (FR-012). No partial inventory is ever returned.
```
nicolehaugen: remove requirement id references.

src/specify_cli/core/__init__.py
```diff
@@ -0,0 +1,541 @@
...
+    """
+    layout = _resolve_core_root()
+
+    from ..extensions import CORE_COMMAND_NAMES  # lazy: avoid import cycle
+
+    # A deleted bundled command cannot be detected because this set is derived from those files.
+    # That would already corrupt the installation, so an independent manifest is not warranted.
```
nicolehaugen: remove second line in comment - just keep the first point about deleted bundled command.

tests/test_core_info_command.py
```diff
@@ -0,0 +1,193 @@
+"""CLI-level tests for ``specify core info --json`` via Typer's CliRunner.
+
+Contract references:
+- ``specs/001-core-info/contracts/core-info-cli.md``
```
nicolehaugen: remove contract references section.

tests/test_core_info_command.py
```diff
@@ -0,0 +1,193 @@
...
+- ``specs/001-core-info/contracts/core-info-schema.json``
+
+Deferred (companion #4212): cross-reference integration tests that join this
```
nicolehaugen: removed comments and TODOs for #4212 test - this will be addressed separately

tests/test_core_info_command.py
```diff
@@ -0,0 +1,193 @@
...
+
+def test_core_info_source_paths_are_package_relative() -> None:
+    parsed = json.loads(_invoke(["core", "info", "--json"]).output)
+    for section in ("commands", "templates", "scripts"):
+        for entry in parsed[section]:
+            assert _PACKAGE_REL.match(entry["sourcePath"]), entry["sourcePath"]
+
+
+# ---------------------------------------------------------------------------
+# T023 — project-isolation guarantees (FR-013)
```
nicolehaugen: remove all references to req id

tests/test_core_inventory.py
```diff
@@ -0,0 +1,396 @@
+"""Unit and behavioral tests for ``specify_cli.core.build_core_inventory``.
+
+Test targets:
+
+- Path resolution across wheel and source layouts
+- Frontmatter parsing with typed empty defaults (Clarifications Q2)
+- Alphabetical ordering & determinism (research §R6, SC-002)
```
nicolehaugen: remove all references to req ids

tests/test_core_inventory.py
```diff
@@ -0,0 +1,396 @@
...
+- Stable id grammar per companion #4210 (US2 T028)
```
nicolehaugen: remove reference to companion issue

tests/test_core_inventory.py
```diff
@@ -0,0 +1,396 @@
...
+    _write(tmp_path / "scripts" / "bash" / "solo.sh", "#!/usr/bin/env bash\n# solo\n")
+    # No powershell variant. Only python.
+    _write(tmp_path / "scripts" / "python" / "solo.py", "# py\n")
+    override_layout(tmp_path)
+    inv = build_core_inventory()
+    solo = next(s for s in inv["scripts"] if s["name"] == "solo")
+    assert solo["runtimes"] == ["bash", "python"]
+
+
+def test_script_without_bash_fails_fast(tmp_path: Path, override_layout) -> None:
```
nicolehaugen, line 233: Fail fast behavior no longer applicable - re-review tests to ensure they are testing applicable scenarios based on actual implementation.

nicolehaugen (commented):
Do analysis to remove references to req ids and references to other companion issues.

tests/test_core_inventory.py
nicolehaugen: @copilot ensure all my feedback is incorporated: Absolute paths still leak in error envelopes.
_read_leading_frontmatter() still uses source_path=str(md_path) at [#L157-L173](https://github.com/github/spec-kit/blob/0368835f59bce2731d7150bc9cc02bb5c81815de/src/specify_cli/core/__init__.py#L157-L173). This is the issue from [your linked review comment](https://github.com/github/spec-kit/pull/4240#discussion_r3825007353).

The documentation still describes command names generically.
It says name is the “base name,” even though commands are namespaced as speckit.plan, at [docs/reference/core-info.md#L31-L36](https://github.com/github/spec-kit/blob/0368835f59bce2731d7150bc9cc02bb5c81815de/docs/reference/core-info.md#L31-L36).

The documentation example is still incorrect.
It uses .name == "plan" and shows core:_:command:plan, but the implementation emits speckit.plan, at [docs/reference/core-info.md#L43-L59](https://github.com/github/spec-kit/blob/0368835f59bce2731d7150bc9cc02bb5c81815de/docs/reference/core-info.md#L43-L59).

The schema link is still broken.
The documentation still links to specs/001-core-info/contracts/core-info-schema.json, but that file is not present.

The independent package-manifest concern remains unresolved.
The current implementation still enumerates the files it finds, so it cannot detect every possible omitted packaged asset. As you noted, that is likely a packaging-integrity concern rather than something this command needs to own, but the PR’s documentation still claims broader fail-fast behavior than the implementation guarantees.

nicolehaugen:
Closing both PR and associated issue - in discussion with @mnriem, an upcoming to change to SpecKit is to move the "core" commands/templates/scripts into their own preset, so that core functionality can be treated like any other preset.  This means that to retrieve info about core, you would use the existing preset/extension info commands - as a result, there isnt' the need to have this separate command specifically to get core info.

PR #4250, opened 2026-08-21
feat(presets): let a preset declare a required extension

### LINKED ISSUE #4231: [Feature]: Let a preset declare a required extension in `requires.extensions`
### Problem Statement

A preset cannot declare that it depends on an extension, so a preset that is useless without one installs silently and appears broken.

@mnriem identified this in https://github.com/github/spec-kit/issues/4164#issuecomment-5348616852 while reviewing the `speckit-inventory` / `inventory-alignment` pair, and suggested tracking it separately rather than blocking that submission:

> One genuine gap worth noting: `preset.yml`'s `requires:` only declares `speckit_version`, so there's no formal "requires extension X" dependency to guarantee the extension is installed alongside the preset. That's a small, separable core enhancement I'd track on its own issue rather than block this on.

Filing it as suggested. There are three concrete parts to the gap:

**1. The manifest cannot express it.** `PresetManifest` validation requires `requires.speckit_version` and validates nothing else in that section, so there is no supported key for an extension dependency. `ExtensionManifest` has the same limitation.

**2. The catalog already expresses it anyway.** Three community presets in `presets/catalog.community.json` currently carry a `requires.extensions` array:

| Preset | `requires.extensions` |
|---|---|
| `aide-in-place` | `["aide"]` |
| `inventory-alignment` | `["speckit-inventory"]` |
| `mde` | `["mde"]` |

So the field is already a de-facto convention in the catalog, and the preset submission template even has a "Required Extensions (optional)" input that feeds it. But it exists only as catalog metadata — it cannot be declared in the preset package itself, which means it is absent when a preset is installed with `--dev` or `--from <url>`, i.e. exactly the paths that bypass the catalog.

**3. Nothing enforces it.** I could not find any code path that reads `requires.extensions` to gate or warn on installation. It appears to be purely descriptive.

The user-visible result: install one of those three presets without its extension and the install succeeds, the templates resolve, and the command overrides register — but the parts that call into the extension quietly do nothing. The workflow still runs correctly, so nothing errors; the feature just appears not to work, with no signal pointing at the missing extension.

### Proposed Solution

Make the dependency declarable in the manifest and checked at install time.

**Schema.** Accept an optional `requires.extensions` in `preset.yml`, mirroring the shape already used in the catalog:

```yaml
requires:
  speckit_version: ">=0.9.0"
  extensions:
    - id: speckit-inventory
      version: ">=0.1.0"   # optional
      required: true        # optional, default true
```

A plain list of id strings (`extensions: ["speckit-inventory"]`) should also be accepted, since that is what the three existing catalog entries use.

**Validation.** Validate the section like `speckit_version` is validated today: reject non-list values, non-string/non-mapping members, and ids that do not match the existing `^[a-z0-9-]+$` id pattern. `#3980` ("reject non-string `requires.speckit_version`") is the precedent for the strictness level here.

**Install-time check.** On `specify preset add`, compare declared dependencies against installed extensions and:

- if a `required: true` dependency is missing, warn clearly and name the extension plus the exact `specify extension add` command that would satisfy it;
- if a version constraint is unsatisfied, warn with both the required and installed versions.

I would suggest **warn, not hard-fail**, at least initially. These preset overrides are written to degrade safely — without the extension, the core workflow runs unchanged — so blocking the install would be harsher than the situation warrants, and would break the three catalog entries above for anyone who deliberately wants the preset alone. A `--force`-style escape hatch would be needed if it ever became a hard failure.

**Extensions too, if desired.** `extension.yml` has the identical limitation. Extending both keeps the two manifest schemas symmetric, though the preset side is where the demonstrated need is.

### Alternatives Considered

**Document the pairing in the README and stop there.** This is the current workaround, and what the `inventory-alignment` and `aide-in-place` READMEs do. It fails because the person who needs the warning is the one who did not read the README — and nothing in the install output points at the problem.

**Enforce from the catalog entry only.** The data is already there for three presets, so this is the smallest change. But it does nothing for `--dev` and `--from <url>` installs, which is how presets are tested and how anything not yet listed is installed. The dependency belongs with the package that has it.

**Hard-fail the install on a missing dependency.** Cleaner semantics, but it would immediately break installs of the three presets already declaring the field, and it overrides an author who deliberately designed for safe degradation. Better as a later opt-in (`required: true` already leaves that door open) than as the initial behavior.

**Leave it to a package manager.** #2566 proposed a community package manager with conflict detection and was closed. Dependency declaration is a manifest concern and much smaller in scope than that proposal.

### Component

### AI Agent (if applicable)

All agents

### Use Cases

**1. A read-only primitive plus the preset that wires it in.** This is the shape @mnriem recommended in #4164 and that `speckit-inventory` + `inventory-alignment` now implement: an extension provides the capability, and a preset overrides core commands to call it. The preset is inert alone. Same shape as `aide-in-place` + `aide` and `mde` + `mde`.

**2. Testing a pair locally before submitting.** With `--dev` installs the catalog is not consulted at all, so an author gets no feedback that they have installed half of their own pair — the case most likely to produce a confused bug report.

**3. Maintainer vetting.** The submission template collects "Required Extensions" today, but nothing downstream can verify the claim against the package. A manifest field makes it checkable.

### Acceptance Criteria

- [ ] `preset.yml` accepts an optional `requires.extensions`, in both the id-string-list and the mapping form.
- [ ] Invalid shapes are rejected with a clear message, consistent with the existing `requires.speckit_version` validation.
- [ ] Omitting `requires.extensions` remains valid; every existing preset continues to install unchanged.
- [ ] `specify preset add` warns when a declared `required: true` dependency is not installed, naming the extension and the command that installs it.
- [ ] `specify preset add` warns when a declared version constraint is unsatisfied, showing required vs installed.
- [ ] The warning appears for `--dev`, `--from <url>`, and catalog installs alike.
- [ ] Unit tests cover: valid declaration in both forms, each invalid shape, dependency satisfied, dependency missing, and version mismatch.
- [ ] `presets/PUBLISHING.md` and the preset schema docs document the new field.
- [ ] Decide whether the three catalog entries already carrying `requires.extensions` should be reconciled with the manifests in their source repos, or whether catalog and manifest are allowed to state it independently.

### Additional Context

I ran into this as the author of the `inventory-alignment` preset (#4227, merged in #4229). The robot that added my catalog entry had to patch the dependency back in by hand — its follow-up commit reads *"Restore the required `speckit-inventory` dependency and pin the submitted release archive SHA-256"* — which is itself a small sign that the information wants to live somewhere more reliable than a catalog entry assembled from an issue form.

I searched open and closed issues for this and did not find an existing report; the closest is #2566, which was closed and much broader in scope. Happy to be pointed at a duplicate if I missed one.

I am willing to implement this if the approach and the warn-versus-fail decision are acceptable — please confirm the direction before I open a PR, and let me know if you would prefer the `extension.yml` side included or left out of the first change.

**AI disclosure.** Filed on behalf of @Yash-Chindam by Claude Code (model: Claude Opus 5), acting autonomously. Per CONTRIBUTING, this used AI assistance for research and drafting: Claude Code inspected the preset and extension manifest validation, searched for prior issues, surveyed the community catalogs for existing `requires.extensions` usage, and wrote this issue. I reviewed it before filing.

Yash-Chindam:
## Description

Closes #4231.

A preset whose command overrides call into an extension is inert without it — but the overrides fall through to the core workflow, so nothing errors. The feature just silently does less than the user expects, with nothing in the install output pointing at the cause. Until now the only place that dependency could be stated was the README, which fails exactly the user who did not read it.

Three community presets already declare `requires.extensions` in `catalog.community.json` (`aide-in-place`, `inventory-alignment`, `mde`), and the preset submission template already collects the field — but `preset.yml` had no supported key for it and nothing read it. This adds the manifest field and the install-time check.

### What changed

**Schema.** `preset.yml` accepts an optional `requires.extensions`, taking either a bare id or a mapping:

```yaml
requires:
  speckit_version: ">=0.9.0"
  extensions:
    - "companion-extension"        # required, any version
    - id: "other-extension"
      version: ">=1.2.0"           # optional PEP 440 specifier
      required: false              # optional, defaults to true
```

**Validation.** Follows the `requires.speckit_version` strictness established by #3980, for the same reason: an unvalidated value reaches `re.match` or `SpecifierSet` later and surfaces as a bare `TypeError` that no caller handles as a malformed manifest. A non-list, a member that is neither string nor mapping, a missing or non-string or badly-shaped `id`, a non-string/blank/unparseable `version`, and a non-boolean `required` each raise `PresetValidationError` naming the offending index.

**Install-time check.** `PresetManager.find_unmet_extension_dependencies()` reports rather than raises. `specify preset add` warns once per unsatisfied dependency:

```
!  This preset depends on extensions that are not satisfied:
    speckit-inventory 0.1.0 does not satisfy >=9.0.0
      Install with: specify extension add speckit-inventory

The preset is installed and safe to use; the parts that rely on these
extensions will do nothing until they are present.
```

The check sits at the single point where the `--dev`, `--from`, and catalog paths converge, so all three behave identically rather than drifting.

### Design decisions

These follow the positions I set out in [the issue](https://github.com/github/spec-kit/issues/4231#issuecomment-5364742294) against the assessment's carried-forward questions. Each is easy to change if you'd rather go the other way.

- **Warn, not fail.** These presets are written to degrade safely, and three catalog entries already declare the dependency — hard-failing would break installs that work today. The `required` flag leaves the door open for an opt-in hard-fail later.
- **Version constraints included in v1.** The mapping form has to be parsed and validated anyway to accept `version`, so the incremental cost is the comparison itself. Deferring it would ship a field that validates but is silently ignored — the same shape as the problem this issue is about.
- **Preset side only; `extension.yml` left alone.** It has the identical limitation, but no extension in the catalog declares a dependency on another extension, so there's no demonstrated need. Happy to mirror it here or in a follow-up.
- **Manifest authoritative, catalog a mirror.** The catalog isn't consulted for `--dev` or `--from <url>` installs, so the manifest is the only copy present on every path. Documented in `PUBLISHING.md` rather than mechanically reconciled; validating the catalog field against the packaged manifest seems better as its own change against the `add-community-preset` workflow.

The field is optional, so every existing preset stays valid and silent.

### Worth flagging

The warning will fire for nobody on day one. The three presets carrying `requires.extensions` do so only in their catalog entries, not in their packaged `preset.yml`. I maintain `inventory-alignment` and will add it there; the other two need their authors. Not a blocker, but the feature starts with no live coverage.

### Automated

`tests/test_presets.py`: **624 passed**, 8 failed. All 8 failures are `WinError 1314` symlink-privilege failures from my Windows environment and reproduce identically on `main` with this branch stashed. `tests/test_extensions.py` + `tests/test_extension_registration.py`: **540 passed**, 2 failed, same symlink cause.

22 new tests:

- `requires.extensions` absent stays valid and reports no dependencies
- both declaration forms normalize to the same shape
- 12 parametrized malformed-input cases (not-a-list, bad member type, missing/non-string/bad-pattern `id`, non-string/blank/unparseable `version`, non-boolean `required`)
- dependency missing / installed / version satisfied / version unsatisfied
- `required: false` never reported
- registry entry with an unusable version is not invented into a mismatch
- multiple dependencies evaluated independently

### Manual

Per the mapping rules, `src/specify_cli/*.py` → test the affected CLI command. This changes `specify preset add`; it is not an init/scaffolding change, so no slash command is affected.

**Agent**: n/a (CLI change) | **OS/Shell**: Windows 11 / Git Bash, Python 3.11.14

| Command tested | Notes |
|----------------|-------|
| `specify preset add --dev` | Dependency missing → warning naming the extension and install command. Preset still installs. |
| `specify preset add --dev` | Extension installed, no constraint → no warning. |
| `specify preset add --dev` | `>=0.1.0` against installed `0.1.0` → no warning. |
| `specify preset add --dev` | `>=9.0.0` against installed `0.1.0` → warning showing both versions. |
| `specify preset add --dev` | Preset declaring nothing → unchanged output. |
| `specify preset add <id>` | Bundled preset (`lean`) via the `preset_id` branch → installs clean, no warning, no regression. |
| `specify preset add --from <url>` | Covered by the existing `test_preset_add_from_url_reads_in_bounded_chunks`, which executes the new call site on the `--from` path. |

All three install branches are therefore exercised, which is what the shared call site is meant to guarantee. Re-verified after rebasing onto `main` at 1.0.1.dev0.

## AI Disclosure

Filed by @Yash-Chindam. This change was written by **Claude Code (model: Claude Opus 5)**, acting autonomously on my behalf — code generation, tests, and this description, not just comments. The commit carries an `Assisted-by: Claude Code (model: Claude Opus 5, autonomous)` trailer.

Worth recording one thing it caught in its own work: the first version of `find_unmet_extension_dependencies()` broke `test_preset_add_from_url_reads_in_bounded_chunks`, which passes a duck-typed `SimpleNamespace` manifest with no `requires_extensions` attribute. That is a real robustness gap rather than a test artifact — the method is public and reachable with a hand-built manifest, the same case `check_compatibility()` already guards against — so the fix went into the code, not the test.

Every check in the Testing section was executed by Claude Code on my machine at my direction and the results reviewed by me; I am not claiming I re-ran each command by hand. I will disclose agent involvement again in each review-round comment rather than relying on this section to cover them.

### DIFF
--- a/presets/PUBLISHING.md
+++ b/presets/PUBLISHING.md
@@ -101,9 +101,9 @@ If your preset overrides commands that call into an extension, declare it in
 `requires.extensions`. Without the extension the preset still installs and the
 overrides fall through to the core workflow, so nothing errors — the feature
 just silently does less than the user expects. Declaring the dependency makes
-`specify preset add` say so, and name the command that fixes it.
+`specify preset add` say so, and spell out how to resolve it.
 
-Use a bare id, or a mapping when you need a version floor or an optional
+Use a bare id, or a mapping when you need a version constraint or an optional
 dependency:
 
 ```yaml
@@ -112,14 +112,19 @@ requires:
   extensions:
     - "companion-extension"              # required, any version
     - id: "other-extension"
-      version: ">=1.2.0"                 # optional PEP 440 specifier
+      version: ">=1.2.0,<2"              # optional PEP 440 specifier
       required: false                    # optional, defaults to true
 ```
+
+`version` accepts any PEP 440 specifier, not just a lower bound — upper bounds
+(`<2`), exact pins (`==1.2.0`), and exclusions (`!=1.3.0`) all work.
 
 Notes:
 
 - The field is optional. A preset that declares nothing behaves exactly as before.
-- A missing or version-unsatisfied dependency produces a **warning, not a failure** — the install still succeeds.
+- A dependency that is missing, disabled, or version-unsatisfied produces a **warning, not a failure** — the install still succeeds.
+- A disabled extension counts as unmet. Resolution skips disabled extensions, so the preset is just as inert as if it were absent.
+- The warning names an exact command for the missing and disabled cases. For a version mismatch it states the constraint to satisfy rather than naming a command, because `specify extension update` only moves forward to the catalog release and cannot satisfy an upper bound, a pin, or a downgrade.
 - `required: false` documents an enhancing-but-optional extension and is never warned about.
 - Declare it in `preset.yml`, not only in your catalog entry. The catalog is not consulted for `--dev` and `--from <url>` installs, so the manifest is the only copy present on every install path.
 
--- a/presets/PUBLISHING.md
+++ b/presets/PUBLISHING.md
@@ -122,9 +122,10 @@ requires:
 Notes:
 
 - The field is optional. A preset that declares nothing behaves exactly as before.
-- A dependency that is missing, disabled, or version-unsatisfied produces a **warning, not a failure** — the install still succeeds.
-- A disabled extension counts as unmet. Resolution skips disabled extensions, so the preset is just as inert as if it were absent.
-- The warning names an exact command for the missing and disabled cases. For a version mismatch it states the constraint to satisfy rather than naming a command, because `specify extension update` only moves forward to the catalog release and cannot satisfy an upper bound, a pin, or a downgrade.
+- A dependency that is missing, stale, disabled, or version-unsatisfied produces a **warning, not a failure** — the install still succeeds.
+- A disabled extension counts as unmet, and so does one whose registry entry survives after its files were removed. Resolution skips both, so the preset is just as inert as if the extension were absent.
+- The warning names an exact command for the missing, stale, and disabled cases. For a version mismatch it states the constraint to satisfy rather than naming a command, because `specify extension update` only moves forward to the catalog release and cannot satisfy an upper bound, a pin, or a downgrade.
+- A recorded version that cannot be parsed is treated as uncomparable rather than as a mismatch, so an extension whose registry version reads `unknown` is not reported as failing a constraint it was never evaluated against.
 - `required: false` documents an enhancing-but-optional extension and is never warned about.
 - Declare it in `preset.yml`, not only in your catalog entry. The catalog is not consulted for `--dev` and `--from <url>` installs, so the manifest is the only copy present on every install path.
 
--- a/presets/PUBLISHING.md
+++ b/presets/PUBLISHING.md
@@ -126,6 +126,7 @@ Notes:
 - A disabled extension counts as unmet, and so does one whose registry entry survives after its files were removed. Resolution skips both, so the preset is just as inert as if the extension were absent.
 - The warning names an exact command for the missing, stale, and disabled cases. For a version mismatch it states the constraint to satisfy rather than naming a command, because `specify extension update` only moves forward to the catalog release and cannot satisfy an upper bound, a pin, or a downgrade.
 - A recorded version that cannot be parsed is treated as uncomparable rather than as a mismatch, so an extension whose registry version reads `unknown` is not reported as failing a constraint it was never evaluated against.
+- An extension present on disk but absent from the registry counts as satisfied. Resolution admits unregistered directories, so the preset works and warning about it would be a false alarm — though with no recorded version, a `version` constraint cannot be checked against it.
 - `required: false` documents an enhancing-but-optional extension and is never warned about.
 - Declare it in `preset.yml`, not only in your catalog entry. The catalog is not consulted for `--dev` and `--from <url>` installs, so the manifest is the only copy present on every install path.
 
--- a/presets/PUBLISHING.md
+++ b/presets/PUBLISHING.md
@@ -68,6 +68,8 @@ preset:
 
 requires:
   speckit_version: ">=0.1.0"      # Required spec-kit version
+  extensions:                      # Optional: extensions this preset needs
+    - "companion-extension"
 
 provides:
   templates:
@@ -92,6 +94,34 @@ tags:                              # 2-5 relevant tags
 - ✅ Template names are lowercase with hyphens only
 - ✅ Command names use dot notation (e.g. `speckit.specify`)
 - ✅ Tags are lowercase and descriptive
+
+#### Declaring extension dependencies
+
+If your preset overrides commands that call into an extension, declare it in
+`requires.extensions`. Without the extension the preset still installs and the
+overrides fall through to the core workflow, so nothing errors — the feature
+just silently does less than the user expects. Declaring the dependency makes
+`specify preset add` say so, and name the command that fixes it.
+
+Use a bare id, or a mapping when you need a version floor or an optional
+dependency:
+
+```yaml
+requires:
+  speckit_version: ">=0.9.0"
+  extensions:
+    - "companion-extension"              # required, any version
+    - id: "other-extension"
+      version: ">=1.2.0"                 # optional PEP 440 specifier
+      required: false                    # optional, defaults to true
+```
+
+Notes:
+
+- The field is optional. A preset that declares nothing behaves exactly as before.
+- A missing or version-unsatisfied dependency produces a **warning, not a failure** — the install still succeeds.
+- `required: false` documents an enhancing-but-optional extension and is never warned about.
+- Declare it in `preset.yml`, not only in your catalog entry. The catalog is not consulted for `--dev` and `--from <url>` installs, so the manifest is the only copy present on every install path.
 
 ### 3. Test Locally
 
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -381,6 +381,14 @@ class PresetManifest:
                 f"got {type(requires['speckit_version']).__name__}"
             )
 
+        # Validate the optional extension dependency list. A preset that
+        # overrides commands calling into an extension is inert without it, and
+        # until now the only place that could be said was the README -- see
+        # issue #4231. Absent means "no dependencies", so every existing preset
+        # stays valid.
+        if "extensions" in requires:
+            self._validate_requires_extensions(requires["extensions"])
+
         # Validate provides section
         provides = self.data["provides"]
         if "templates" not in provides:
@@ -524,10 +532,115 @@ class PresetManifest:
         """Get preset author."""
         return self.data["preset"].get("author", "")
 
+    @staticmethod
+    def _validate_requires_extensions(declared: Any) -> None:
+        """Validate the optional ``requires.extensions`` list.
+
+        Accepts either a bare extension id or a mapping carrying an optional
+        version specifier and an optional ``required`` flag:
+
+        .. code-block:: yaml
+
+            requires:
+              extensions:
+                - speckit-inventory
+                - id: other-ext
+                  version: ">=1.2.0"
+                  required: false
+
+        Raises:
+            PresetValidationError: If the list or any entry is malformed.
+        """
+        if not isinstance(declared, list):
+            raise PresetValidationError(
+                "Invalid requires.extensions: expected a list, "
+                f"got {type(declared).__name__}"
+            )
+
+        for index, entry in enumerate(declared):
+            label = f"requires.extensions[{index}]"
+
+            if isinstance(entry, str):
+                entry = {"id": entry}
+            elif not isinstance(entry, dict):
+                raise PresetValidationError(
+                    f"Invalid {label}: expected a string or a mapping, "
+                    f"got {type(entry).__name__}"
+                )
+
+            if "id" not in entry:
+                raise PresetValidationError(f"Missing {label}.id")
+            extension_id = entry["id"]
+            if not isinstance(extension_id, str):
+                raise PresetValidationError(
+                    f"Invalid {label}.id: expected a string, "
+                    f"got {type(extension_id).__name__}"
+                )
+            # Same id shape the extension loader enforces, so a dependency can
+            # never name something that could not be installed in the first place.
+            if not re.match(r'^[a-z0-9-]+$', extension_id):
+                raise PresetValidationError(
+                    f"Invalid {label}.id '{extension_id}': "
+                    "must be lowercase alphanumeric with hyphens only"
+                )
+
+            if "version" in entry:
+                constraint = entry["version"]
+                # Mirrors the requires.speckit_version reasoning: a non-string
+                # escapes InvalidSpecifier two ways -- scalars raise TypeError
+                # from the constructor, and a list/dict is iterable so it
+                # constructs and only fails later inside .contains().
+                if not isinstance(constraint, str) or not constraint.strip():
+                    raise PresetValidationError(
+                        f"Invalid {label}.version: expected a non-empty string, "
+                        f"got {type(constraint).__name__}"
+                    )
+                try:
+                    SpecifierSet(constraint)
+                except InvalidSpecifier:
+                    raise PresetValidationError(
+                        f"Invalid {label}.version '{constraint}': "
+                        "not a valid version specifier"
+                    )
+
+            if "required" in entry and not isinstance(entry["required"], bool):
+                raise PresetValidationError(
+                    f"Invalid {label}.required: expected a boolean, "
+                    f"got {type(entry['required']).__name__}"
+                )
+
     @property
     def requires_speckit_version(self) -> str:
         """Get required spec-kit version range."""
         return self.data["requires"]["speckit_version"]
+
+    @property
+    def requires_extensions(self) -> List[Dict[str, Any]]:
+        """Get declared extension dependencies, normalized to mappings.
+
+        Returns:
+            One entry per dependency with ``id``, ``version`` (``None`` when
+            unconstrained), and ``required`` (defaulting to ``True``). Empty
+            when the manifest declares no dependencies.
+        """
+        declared = self.data["requires"].get("extensions")
+        if not isinstance(declared, list):
+            return []
+
+        normalized: List[Dict[str, Any]] = []
+        for entry in declared:
+            if isinstance(entry, str):
+                entry = {"id": entry}
+            if not isinstance(entry, dict) or not isinstance(entry.get("id"), str):
+                continue
+            normalized.append(
+                {
+                    "id": entry["id"],
+                    "version": entry.get("version"),
+                    "required": entry.get("required", True),
+                }
+            )
+        return normalized
 
     @property
     def templates(self) -> List[Dict[str, Any]]:
@@ -840,6 +953,65 @@ class PresetManager:
             )
 
         return True
+
+    def find_unmet_extension_dependencies(
+        self,
+        manifest: PresetManifest
+    ) -> List[Dict[str, Any]]:
+        """Find declared extension dependencies that are not satisfied.
+
+        Reports rather than raises. A preset whose overrides call into an
+        extension is written to degrade safely -- without the extension the
+        core workflow still runs -- so a missing dependency is a warning, not
+        an install failure. See issue #4231.
+
+        Args:
+            manifest: Preset manifest to inspect
+
+        Returns:
+            One entry per unsatisfied dependency, each with ``id``, the
+            requested ``version`` specifier (``None`` when unconstrained), the
+            ``installed`` version (``None`` when absent), and a ``reason`` of
+            either ``"missing"`` or ``"version"``. Optional dependencies
+            (``required: false``) are never reported.
+        """
+        # Defense in depth, mirroring check_compatibility(): this method is
+        # public and also reachable with a hand-built manifest object that
+        # predates this field. A manifest without it declares nothing.
+        candidates = getattr(manifest, "requires_extensions", None)
+        if not isinstance(candidates, list):
+            return []
+
+        declared = [
+            dep for dep in candidates
+            if isinstance(dep, dict) and dep.get("required", True)
+        ]
+        if not declared:
+            return []
+
+        registry = ExtensionRegistry(self.project_root / ".specify" / "extensions")
+        unmet: List[Dict[str, Any]] = []
+
+        for dep in declared:
+            metadata = registry.get(dep["id"])
+            if metadata is None:
+                unmet.append({**dep, "installed": None, "reason": "missing"})
+                continue
+
+            constraint = dep["version"]
+            if not constraint:
+                continue
+
+            installed = metadata.get("version")
+            if not isinstance(installed, str):
+                # A registry entry without a usable version cannot be compared.
+                # Treat it as satisfied rather than inventing a failure, since
+                # the extension is demonstrably installed.
+                continue
+            if not version_satisfies(installed, constraint):
+                unmet.append({**dep, "installed": installed, "reason": "version"})
+
+        return unmet
 
     def _register_commands(
         self,
--- a/src/specify_cli/presets/__init__.py
+++ b/src/specify_cli/presets/__init__.py
@@ -591,10 +591,15 @@ class PresetManifest:
                     f"got {type(extension_id).__name__}"
                 )
             # Same id shape the extension loader enforces, so a dependency can
-            # never name something that could not be installed in the first place.
-            if not re.match(r'^[a-z0-9-]+$', extension_id):
+            # never name something that could not be installed in the first
+            # place. fullmatch rather than match with an anchored pattern: `$`
+            # also matches before a trailing newline, so "demo-ext\n" would
+            # otherwise validate here while PresetResolver._is_safe_registry_id
+            # (which uses fullmatch) rejects it, and the newline would land in
+            # a suggested command.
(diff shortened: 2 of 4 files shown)


mnriem (changes requested):

Yash-Chindam:
Both findings addressed in 853e67d. I verified each against the code before changing anything, and both were correct.

**Remediation command didn't match the reason.** Confirmed: `install_from_directory` raises `Extension '<id>' is already installed` without `--force`, so `specify extension add <id>` could only fail for a version mismatch. Reproduced it directly:

```
Error: Extension 'speckit-inventory' is already installed. Use 'specify
extension remove speckit-inventory' first, or retry with --force to overwrite.
```

The remedy now follows the reason — `add` for missing, `update` for a version mismatch, `enable` for a disabled one — and the label changed from "Install with:" to "Fix with:", since two of the three are no longer installs.

**Disabled extensions were treated as satisfied.** Also confirmed, and the consequence is worse than a missed warning: `_collect_extension_layers` skips disabled extensions (`presets/__init__.py:5337`, matching `extensions/__init__.py:1000`), so the preset is exactly as inert as if the extension were absent, while the registry entry made the check report success. Now reported as a distinct `disabled` reason, ordered ahead of the version check — enabling is the prerequisite, and the version may well be fine once it is.

End-to-end, with the extension installed but disabled:

```
!  This preset depends on extensions that are not satisfied:
    speckit-inventory is installed but disabled
      Fix with: specify extension enable speckit-inventory
```

Running that suggestion enables the extension, and re-installing the preset is then silent.

One thing I changed beyond the two comments: the closing line read "will do nothing until they are present", which is wrong for a disabled extension that *is* present. It now reads "until this is resolved".

Two new tests cover the disabled case and its ordering against a version mismatch, and the multi-dependency test now mixes present, absent, disabled, and optional. `tests/test_presets.py`: 626 passed, 8 failed — all 8 are `WinError 1314` symlink-privilege failures from my Windows environment that reproduce on `main` with this branch stashed.

The four CI workflows are still showing `action_required`, so they have not run yet — could you approve them when you get a chance?

_Disclosure: this comment and 853e67d were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

mnriem:

Yash-Chindam:
The remaining Copilot finding is addressed in 972cb57. I checked the version-constraint behavior before changing the remediation text.

**Version remediation could over-promise.** Confirmed: the manifest accepts general PEP 440 constraints such as `<2`, `==1.2`, exclusions, and possible downgrades, while `extension update` only moves forward to the catalog release. The warning now tells users to install a release satisfying the declared constraint, without claiming that `extension update` will resolve every valid constraint.

**Regression coverage.** Added a focused test that verifies an upper-bound mismatch does not suggest `specify extension update` and instead reports the required constraint explicitly.

The focused preset dependency tests pass (3 passed), and Ruff reports no lint errors. The full preset-file run was also attempted; its unrelated failures are the existing Windows symlink-privilege cases and Typer compatibility failures documented in the PR, with the new test passing.

_Disclosure: this comment and 972cb57 were written by ChatGPT (model: GPT-5), acting under the direction of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

Yash-Chindam:
Addressed in db9a90d, and the branch is rebased onto current `main`.

**"Version floor" understated the field.** Correct — `version` is passed to `SpecifierSet`, so it accepts any PEP 440 specifier. I verified `<2`, `==1.2.0`, `!=1.3.0`, `>=1.2.0` and the bounded range `>=1.2.0,<2` all validate and round-trip through `requires_extensions`. The guide now says "version constraint", the example shows a bounded range, and the accepted forms are stated explicitly.

Two adjacent claims in the same section had also drifted from the behaviour, so I corrected them in the same pass rather than leave them for a later round:

- It promised the warning would "name the command that fixes it". That stopped being true for a version mismatch once that branch began stating the constraint instead of naming a command, per the previous round.
- The notes listed only missing and version-unsatisfied dependencies as producing a warning, and never mentioned disabled ones, which became a third reason earlier in this review.

**Rebase.** `main` had moved 39 commits, four of which touch `presets/` (#4088, #4341, #4094, #3843). Rebased cleanly with no conflicts and re-verified afterwards rather than assuming: `tests/test_presets.py` is 634 passed / 8 failed, the 8 being the `WinError 1314` symlink-privilege cases from my Windows environment that reproduce on `main` unmodified. End-to-end, a bounded range still reports correctly against an installed extension:

```
speckit-inventory 0.1.0 does not satisfy >=0.0.1,<0.0.9
  Fix with: install a release of speckit-inventory satisfying >=0.0.1,<0.0.9
```

Note the PR still has no CI results — the four workflows have never run, so nothing here has been checked on Linux or macOS. Whenever you get a chance to approve the workflow runs, that would be useful before merge.

_Disclosure: this comment and db9a90d were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

mnriem (changes requested):

Yash-Chindam:
Both findings addressed in 47a5bb6. I confirmed each against the code first; both were real.

**Stale registry entries read as satisfied.** Confirmed — `PresetResolver` guards on `ext_dir.is_dir()` in both template lookup and layer collection, so once `.specify/extensions/<id>` is gone the extension contributes nothing, while the surviving registry entry made the check report success. That is the same silent-inertness this feature exists to surface, one layer down.

Now reported as a distinct `stale` reason, checked ahead of `disabled` and the version comparison, since restoring the files is the prerequisite for either of those mattering:

```
speckit-inventory is registered but its files are missing
  Fix with: specify extension add speckit-inventory --force
```

`--force` rather than a plain `add`, because the registry entry is still present and a plain `add` would be refused as already installed.

**Unparseable versions were reported as mismatches.** Also confirmed. The guard only covered non-string values, so a string like `"unknown"` passed it and reached `version_satisfies()`, which catches `InvalidVersion` and returns `False` — indistinguishable from a genuine mismatch, and contradicting the documented behaviour that unusable versions are not invented into failures. Parseability is now checked before comparing, via a small `_is_comparable_version()` helper. Verified end-to-end that `"unknown"` against `>=9.0.0` produces no warning, while a real `0.1.0` against `>=9.0.0` still does.

**Test-helper correction worth flagging.** The existing helper only wrote the registry entry, never the extension directory — so every prior dependency test was, unknowingly, exercising the stale state. It now creates the directory as the real installer does, with an opt-out used by the stale tests. Without that, the new check would have silently passed for the wrong reason.

Added tests: `stale` detection, its precedence over `disabled` and `version`, its remediation text, and a parametrized uncomparable-version case covering `None`, a non-string, `"unknown"`, `""`, and `"latest"`.

`tests/test_presets.py`: 641 passed, 8 failed — the same `WinError 1314` symlink-privilege cases from my Windows environment, which reproduce on unmodified `main`.

The PR still has no CI results; the workflows have never run, so none of this has been checked on Linux or macOS. Approving the workflow runs when convenient would be worthwhile before merge.

_Disclosure: this comment and 47a5bb6 were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

mnriem (changes requested):

Yash-Chindam:
Addressed in b822bf8. Confirmed before changing anything, and it was correct.

**An absent registry entry was reported as missing.** `_get_all_extensions_by_priority()` admits a safe on-disk directory as an unregistered extension at implicit priority 10, so its artifacts resolve and the preset works. The warning was telling users to install something already in use.

A matching directory now counts as present, guarded the same way resolution guards itself — the id must pass `_is_safe_registry_id`, and the registry must not be corrupt, since `is_corrupt()` makes that path fail closed and contribute nothing. I reused the resolver's own predicate rather than restating the pattern, so the two cannot drift apart.

An unregistered extension has no recorded version, so a declared constraint against it is uncomparable rather than unsatisfied — consistent with how an unparseable recorded version is now handled.

This is the mirror of the stale case from the previous round, and worth stating as a single rule: **the check now agrees with resolution in both directions.** A registry entry without files is unmet; files without a registry entry are satisfied. Both previously disagreed with what resolution actually does, in opposite directions.

Three regression tests added, as requested: an unregistered directory with no constraint, one with a constraint that cannot be evaluated, and one where a corrupt registry means the directory is not usable after all and `missing` remains correct. Verified end-to-end too — an unregistered directory produces no warning, while removing it still produces the missing warning with its install command.

`tests/test_presets.py`: 644 passed, 8 failed — the same `WinError 1314` symlink-privilege cases from my Windows environment, which reproduce on unmodified `main`.

Still no CI on this PR. Six review rounds in, the workflows have never run, so none of this has been exercised on Linux or macOS — only on Windows, where those eight symlink tests cannot pass regardless. Approving the workflow runs would be worth doing before merge.

_Disclosure: this comment and b822bf8 were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

mnriem:

Yash-Chindam:
Both findings addressed in 15f97d2, including the suppressed one. Confirmed each against the code first; both were correct.

**Corrupted registry entries reached the unregistered-directory fallback.** `get()` returns `None` for a corrupted non-dict entry exactly as it does for an absent one, while `keys()` deliberately retains the id so resolution does not re-admit its directory. `is_corrupt()` does not help here — it validates only the registry container, not individual entries. So a corrupted entry whose directory survived read as satisfied while contributing nothing. The fallback now requires the id to be absent from `keys()`, which is the same signal resolution uses, so it can no longer revive what resolution excludes.

**`extension add <id>` is rejected for discovery-only catalogs.** This is the one worth dwelling on: the default community catalog is `install_allowed=False`, and all three extensions this feature exists to serve — `aide`, `mde`, and `speckit-inventory` — are listed only there. The first live warnings would have pointed users at a command that exits 1. I maintain `speckit-inventory`, so that would have been my own preset sending users to a dead end.

The warning now notes the `--from <archive-url>` form once, after the list, for the two reasons that suggest an install by id:

```
!  This preset depends on extensions that are not satisfied:
    speckit-inventory is not installed
      Fix with: specify extension add speckit-inventory

The preset is installed and safe to use; the parts that rely on these
extensions will do nothing until this is resolved.
An extension listed only in a discovery-only catalog cannot be installed by id;
pass --from <archive-url> with its release archive.
```

I took the "explicit source guidance" option rather than emitting a source-aware `--from <download_url>`, because resolving each dependency's catalog status means a catalog fetch, and `preset add --dev` currently touches no network at all. Adding one to a warning path risks hanging an offline install to improve the wording of a message. Happy to switch to the source-aware form if you would rather have the exact URL and consider the fetch acceptable.

Four tests added: a corrupted entry with its directory still present, the discovery-only note appearing for missing and stale, and its absence for a version mismatch, which suggests no install by id.

`tests/test_presets.py`: 647 passed, 8 failed — the same `WinError 1314` symlink cases from my Windows environment, which reproduce on unmodified `main`.

On that last point, and offered as a practical observation rather than a complaint: seven review rounds in, this PR has never run CI. Every result above is Windows-only, on a machine where eight of these tests cannot pass whatever the code does. Several findings in this review have been cases where my check disagreed with what resolution actually does — the kind of thing a green cross-platform run plus a reviewer's eye would triangulate faster than review alone. Approving the workflow runs whenever convenient would help.

_Disclosure: this comment and 15f97d2 were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

Yash-Chindam:
Addressed in f69a16d. Confirmed first, and it was correct.

`^[a-z0-9-]+$` admits a leading hyphen, so a dependency id such as `--force` produced `specify extension add --force`, which Typer parses as an option rather than the positional extension argument. The advertised fix could not run at all, and the stale, disabled, and version remedies shared the flaw.

Rather than add a second implementation of the rule, I reused `_command_safe_id` from `extensions/_commands.py` as you suggested — it already rejects a leading hyphen and falls back to a `<extension-id>` placeholder, and reusing it means the two cannot drift apart. The displayed id keeps plain Rich escaping; only the copyable command needs to survive the parser.

Parametrized across all four reasons. Worth noting the first version of that test asserted `"--force" not in remedy`, which failed on the stale case because its remedy legitimately ends in `--force` — the id is now a string that cannot collide with a real flag, so a match means genuine interpolation rather than a coincidence.

`tests/test_presets.py`: 652 passed, 8 failed — the Windows symlink cases from my environment, which CI's `windows-latest` jobs pass.

Two notes on state:

The branch is now rebased onto current `main` (it had fallen 16 behind, including #4092 which touches `presets/`). The previous run on `15f97d2` was green on 13 of 14 checks — ruff, markdownlint, shellcheck, CodeQL, dependency audit, and pytest on ubuntu 3.13/3.14, windows 3.13/3.14, and macos 3.14 all passed. The single non-pass was `pytest (macos-latest, 3.13)`, which was **cancelled rather than failed**: its annotation reads "The job was not acquired by Runner of type hosted even after multiple attempts", so no test executed. macOS 3.14 passing on the same commit rules out a platform issue.

Workflow runs need approving again — the fork-contributor gate re-triggers per push, so the runs for this commit are sitting at `action_required`. I cannot re-run or approve them myself (`gh run rerun` reports "Must have admin rights to Repository"). Whenever you approve the run, that also gives macOS 3.13 a fresh chance at a runner.

_Disclosure: this comment and f69a16d were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

mnriem (changes requested):

Yash-Chindam:
All four findings from the last review are addressed in b61050e, along with two more I found while re-reading the same code. Each was confirmed against the source before changing anything.

**The footer was inaccurate for a version result.** Correct, and the wording overstated things in two directions at once: it claimed dependent features "will do nothing", which is false when the extension is installed and enabled, and it asserted the preset was "safe to use", which is not something to promise about an untested version combination. The note is now split by consequence:

```
The preset is installed.
Where only a version constraint is unmet the extension is still used, so it may
not behave as the preset expects.
```

An unavailable dependency still gets the "does nothing" wording, since that one is accurate.

**OSError from the registry could crash a completed install.** This was the most serious of the four. `_load()` recovers from malformed content but deliberately lets `OSError` through, and `is_corrupt()` re-reads the file. Because the check runs after the install has finished and `preset_add` handles only preset-domain errors, an unreadable registry turned a successful install into a traceback — a warning crashing the operation it exists to annotate. An unreadable registry now yields no results.

**Corrupted entries were conflated with absent ones.** Confirmed: `get()` returns `None` for both, while `is_installed()` still counts the key, so `extension add <id>` is refused as already installed. Now a distinct `corrupt` reason with a forced reinstall:

```
speckit-inventory has an unreadable registry entry
  Fix with: specify extension add speckit-inventory --force
```

**The `re.match` vs `fullmatch` issue** I had independently found and fixed before this review arrived — `$` also matches before a trailing newline, so `"demo-ext\n"` validated here while `PresetResolver._is_safe_registry_id` rejects it, and the newline would have landed in a printed command. Now `fullmatch`, and the error quotes the id so an invisible character is visible.

Also included, found in the same pass: declaring one dependency twice warned twice. Exact repeats now collapse, while two entries for one id with different constraints are still checked separately, since both have to hold.

Two of my own tests needed correcting rather than the code — one asserted `missing` where `corrupt` is now right, and another matched prose that Rich wraps across lines. Worth stating plainly rather than quietly amending.

Verified end-to-end for the corrupt and version cases, not only in unit tests. `tests/test_presets.py`: 662 passed, 8 failed — the Windows symlink cases from my environment, which CI's `windows-latest` jobs pass. `ruff` clean.

CI has not run since the approval on `15f97d2`; the fork gate re-triggers per push, so the runs for this commit will need approving again. That earlier run was green on 13 of 14, the exception being a macOS 3.13 job that was cancelled because no runner was ever allocated, not a test failure.

_Disclosure: this comment and b61050e were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

mnriem (changes requested):

Yash-Chindam:
Addressed in 74bbef1. This one was a fair hit on a compromise I made two rounds ago and did not follow through properly.

The per-dependency line labelled the bare-id command "Fix with", while the closing note said an archive URL is required. Those contradicted each other, and for the three extensions that motivated this feature — `aide`, `mde`, `speckit-inventory`, all listed only in the discovery-only community catalog — the advertised command is refused outright. So the warning asserted a fix and then undercut it in the next breath.

Two changes:

**The label now names the action rather than claiming a fix.** "Install with", "Reinstall with", "Enable with", and "Needs" for a version constraint, which was never a command to begin with:

```
speckit-inventory is not installed
  Install with: specify extension add speckit-inventory
```

**The note now describes what actually happens** instead of contradicting the line above it:

```
If an extension is listed only in a discovery-only catalog, that command is
refused and prints the --from <archive-url> form to use instead.
```

That is verifiable: `extensions/_commands.py` prints the exact `specify extension add <id> --from <archive-url>` invocation when it rejects a discovery-only entry. The bare command is therefore a signpost rather than a dead end, and describing it that way is both accurate and more useful than a warning that quietly omits the extra step.

I kept the note unconditional rather than resolving each dependency's catalog, for the reason given earlier: that means a catalog fetch on an install path that currently touches no network, and hanging an offline `preset add` to improve a message is a bad trade. It is now emitted only when a suggested command actually resolves through the catalogs, so a disabled-only or version-only warning no longer carries it. If you would rather have the exact URL and consider the fetch acceptable, say so and I will switch it.

Three of my own tests needed updating for the new labels, and one of those was asserting against the whole output rather than the remedy — the description line legitimately shows the raw id, so only the copyable command needs to be checked. Corrected rather than loosened.

Verified end-to-end. `tests/test_presets.py`: 662 passed, 8 failed — the Windows symlink cases from my environment, which CI's `windows-latest` jobs pass. `ruff` clean, `git diff --check` clean.

CI has not run since `15f97d2`; the fork gate re-triggers per push, so these commits are waiting on approval.

_Disclosure: this comment and 74bbef1 were written by Claude Code (model: Claude Opus 5), acting autonomously on behalf of @Yash-Chindam. The commit carries an `Assisted-by:` trailer._

mnriem (approved):

mnriem:
Thank you!

PR #4284, opened 2026-08-23
fix(bundler): pass explicit workflow_add options from bundle install

### LINKED ISSUE #4282: [Bug]: bundle install cannot install workflows from a catalog (Typer option default leaks into direct call)
### What happened

`specify bundle install` cannot install any workflow from a catalog. Every workflow component fails with an error about `--dev`, even though `--dev` was never passed and the workflow resolves correctly from the catalog.

```
$ specify bundle install lean-full-lifecycle
Error: --dev source must be a workflow YAML file, supported archive, or
directory containing workflow.yml: lifecycle-greenfield-bootstrap
Error: Failed to install workflow 'lifecycle-greenfield-bootstrap'.
```

The same workflow installs fine through the CLI:

```
$ specify workflow add lifecycle-greenfield-bootstrap
✓ Workflow 'Greenfield Bootstrap' (lifecycle-greenfield-bootstrap) installed
```

Presets and extensions in the same bundle install without trouble; only workflows fail.

### Cause

`WorkflowPrimitive.install` in `src/specify_cli/bundler/services/primitives.py` calls the Typer command as a plain Python function:

```python
from ... import workflow_add

with _chdir(self._root):
    _delegate_command(
        "install", f"workflow '{component.id}'",
        lambda: workflow_add(component.id),
    )
```

`workflow_add` is declared in `src/specify_cli/workflows/_commands.py` as:

```python
def workflow_add(
    source: str = typer.Argument(..., help="Workflow ID, URL, or local path"),
    dev: bool = typer.Option(False, "--dev", help="Install from a local workflow YAML file or directory"),
    from_url: str | None = typer.Option(None, "--from", help="Install from a custom URL"),
):
```

Called directly, Typer never binds the parameters, so `dev` keeps its `typer.OptionInfo` default. That object is truthy, so the function takes the local-path branch and interprets the workflow **id** as a filesystem path.

Reproducible in a Python shell:

```python
>>> import inspect
>>> from specify_cli.workflows._commands import workflow_add
>>> d = inspect.signature(workflow_add).parameters["dev"].default
>>> type(d).__name__, bool(d)
('OptionInfo', True)
```

`workflow_remove` is unaffected: it declares only a `typer.Argument`, which the bundler supplies positionally.

### Version

Reproduced on `v1.0.1` (macOS, Python 3.13). The same code is present on `main` at the time of writing.

### Suggested fix

Pass the options explicitly at both call sites in `primitives.py`:

```python
lambda: workflow_add(component.id, dev=False, from_url=None),
```

Extracting the body into a plain function that the Typer command wraps would prevent the whole class of bug, since any future option added to the command would otherwise reintroduce it silently.

### Impact

Any bundle that provides workflows cannot be installed as a bundle. The workaround is to install workflows through `specify workflow add <id>` before `specify bundle install`, which lets Typer bind `dev=False`. They are still catalog-sourced, but `bundle install` then reports them as "already present" and does not attribute them to the bundle, so `specify bundle remove` leaves them installed.

chelsealong:
## Description

Fixes #4282

`specify bundle install` failed to install any workflow from a catalog. The
root cause is in `_WorkflowKindManager.install` (`src/specify_cli/bundler/services/primitives.py`):
it calls the Typer command `workflow_add` directly as a plain Python function,
`workflow_add(component.id)`. Typer only binds `dev`/`from_url` to their
declared defaults (`False` / `None`) when it invokes the command itself —
calling the function directly leaves `dev` bound to the raw (truthy)
`typer.OptionInfo` object. That makes the local-path branch trigger, so the
workflow's catalog ID is treated as a filesystem path and installation fails
with:

```
Error: --dev source must be a workflow YAML file, supported archive, or
directory containing workflow.yml: lifecycle-greenfield-bootstrap
```

This passes `dev=False, from_url=None` explicitly at the call site, matching
the fix suggested in the issue.

## Testing

- [x] Ran existing tests with `uv sync --extra test && uv run pytest -q` — 7308 passed, 10 skipped
- [x] `uvx ruff@0.15.0 check src tests` — all checks passed
- [x] Added a regression assertion to `tests/unit/test_bundler_primitives.py::test_offline_workflow_allows_bundled` that captures the `dev`/`from_url` values `workflow_add` is actually called with

Verified the added assertion fails without the fix:

```
$ git checkout HEAD~1 -- src/specify_cli/bundler/services/primitives.py
$ uv run pytest tests/unit/test_bundler_primitives.py -k test_offline_workflow_allows_bundled -v
...
FAILED tests/unit/test_bundler_primitives.py::test_offline_workflow_allows_bundled
AssertionError: assert [('bundled-wf', <object object at 0x...>, <object object at 0x...>)] == [('bundled-wf', False, None)]
$ git checkout HEAD -- src/specify_cli/bundler/services/primitives.py
$ uv run pytest tests/unit/test_bundler_primitives.py -v
...
20 passed in 0.21s
```

## AI Disclosure

Implemented autonomously by Claude Code (Sonnet 5) under automated direction. The regression test was written and confirmed to fail before the fix, then the full test suite and lint were run locally as shown above.

### DIFF
--- a/src/specify_cli/bundler/services/primitives.py
+++ b/src/specify_cli/bundler/services/primitives.py
@@ -337,7 +337,7 @@ class _WorkflowKindManager:
         with _chdir(self._root):
             _delegate_command(
                 "install", f"workflow '{component.id}'",
-                lambda: workflow_add(component.id),
+                lambda: workflow_add(component.id, dev=False, from_url=None),
             )
 
     def refresh(self, component: ComponentRef) -> None:
--- a/tests/unit/test_bundler_primitives.py
+++ b/tests/unit/test_bundler_primitives.py
@@ -77,13 +77,17 @@ def test_offline_workflow_allows_bundled(tmp_path: Path, monkeypatch):
     monkeypatch.setattr(
         assets, "_locate_bundled_workflow", lambda wid: tmp_path / "wf"
     )
-    calls: list[str] = []
-    monkeypatch.setattr(specify_cli, "workflow_add", lambda wid: calls.append(wid))
+    calls: list[tuple] = []
+    monkeypatch.setattr(
+        specify_cli,
+        "workflow_add",
+        lambda wid, dev=object(), from_url=object(): calls.append((wid, dev, from_url)),
+    )
 
     manager = primitive_manager("workflows", tmp_path, allow_network=False)
     manager.install(_component("workflows", "bundled-wf"))
 
-    assert calls == ["bundled-wf"]
+    assert calls == [("bundled-wf", False, None)]
 
 
 def test_assert_pinned_version_matches_passes():


mnriem:
Thank you!

PR #4285, opened 2026-08-23
fix: scaffold extension config when installing via bundler

### LINKED ISSUE #4283: [Bug]: bundle install does not scaffold extension config templates
### What happened

`specify bundle install` installs an extension without scaffolding the config templates the extension declares in `provides.config`. Installing the same extension with `specify extension add` does scaffold them.

Extension manifest:

```yaml
provides:
  config:
    - name: github-lifecycle-config.yml
      template: config-template.yml
```

After `specify extension add --dev <path>`:

```
Config scaffolded:
  • .specify/extensions/github-lifecycle/github-lifecycle-config.yml
```

After `specify bundle install <bundle-id>`, with the same extension resolved from a catalog:

```
$ ls .specify/extensions/github-lifecycle/
CHANGELOG.md  README.md  commands  config-template.yml  extension.yml  scripts
```

The template is present, but the config file it is meant to produce is not.

### Cause

`ExtensionManager.scaffold_config` is called from the `extension add` command flow in `src/specify_cli/extensions/_commands.py`:

```python
# Scaffold config templates automatically
deployed, skipped, failed = manager.scaffold_config(manifest.id)
```

`ExtensionPrimitive` in `src/specify_cli/bundler/services/primitives.py` calls `install_from_directory` / `install_from_zip` directly and never calls `scaffold_config`, so the bundler path skips it.

### Version

Reproduced on `v1.0.1` (macOS, Python 3.13).

### Impact

An extension installed as part of a bundle silently has no configuration. Any command or script that reads the config finds nothing, and the difference is invisible unless the user compares the two install paths. It also means an extension behaves differently depending on how it was installed, which makes bundle installation not equivalent to installing its components individually.

### Suggested fix

Call `scaffold_config` from `ExtensionPrimitive` after install, and surface the same deployed/skipped/failed reporting the command flow produces.

### Related

The naming rule enforced by `_target_follows_preserved_convention` (a config target must be a top-level `*-config.yml` or `*-config.local.yml`) is not documented in `EXTENSION-API-REFERENCE.md`. A manifest that declares any other target name is accepted at publish time and silently never scaffolds, reporting only `Warning: Config templates not scaffolded: <name>` with no explanation of the constraint. Documenting the rule, or naming it in that warning, would make the failure self-explanatory.

chelsealong:
## Summary

Fixes #4283.

`specify bundle install` installs an extension by calling
`ExtensionManager.install_from_directory` / `install_from_zip` directly
from `_ExtensionKindManager._do_install` in
`src/specify_cli/bundler/services/primitives.py`, but never called
`ExtensionManager.scaffold_config` afterwards. The `specify extension add`
command flow (`src/specify_cli/extensions/_commands.py`) does call it, so
an extension installed as part of a bundle ended up with its
`provides.config` templates present on disk but never deployed — the
extension silently had no configuration, while installing the same
extension directly worked correctly.

## Change

- `src/specify_cli/bundler/services/primitives.py`: call
  `self._manager.scaffold_config(manifest.id)` right after a successful
  `install_from_directory`/`install_from_zip` call in
  `_ExtensionKindManager._do_install`, using the id from the manifest
  returned by the install call (matching what the `extension add` command
  flow does).
- `tests/unit/test_bundler_primitives.py`: added
  `test_bundled_extension_install_scaffolds_config`, which installs a real
  (unmocked) extension source with a `provides.config` entry through
  `primitive_manager("extensions", ...)` and asserts the config file lands
  at `.specify/extensions/<id>/<name>`. Also updated the existing
  `install_from_directory` mocks in this file to return an object with an
  `.id` attribute (as the real method does), since the new call site reads
  `manifest.id` off the return value.

Out of scope: the issue's related documentation nit about
`_target_follows_preserved_convention` naming conventions not being
documented in `EXTENSION-API-REFERENCE.md`, and surfacing per-file
deployed/skipped/failed console reporting through the bundler's CLI
summary (the bundler currently reports install results only as install
counts, not per-component detail — plumbing that through is a larger,
separate change).

## Test plan

Confirmed the new test fails without the fix (temporarily reverted just
the production file):

```
$ git stash push -- src/specify_cli/bundler/services/primitives.py
$ python3 -m uv run pytest tests/unit/test_bundler_primitives.py -q
...
FAILED tests/unit/test_bundler_primitives.py::test_bundled_extension_install_scaffolds_config - AssertionError: assert False
 +  where False = exists()
 +    where exists = PosixPath('.../project/.specify/extensions/my-ext/my-ext-config.yml').exists
1 failed, 20 passed in 0.28s
$ git stash pop
```

With the fix:

```
$ python3 -m uv run pytest tests/unit/test_bundler_primitives.py -q
21 passed in 0.27s
```

Full suite (unmodified elsewhere):

```
$ python3 -m uv run pytest -q
7309 passed, 10 skipped, 48 warnings in 467.66s
```

`ruff check` on the touched files reports 4 pre-existing issues (unused
`noqa` on lines untouched by this change, import-order in two unrelated
test functions) that are also present on unmodified `upstream/main` —
confirmed by stashing this change and re-running the same `ruff check`
invocation, which reported the identical 4 findings at the same
locations.

## AI disclosure

This change was implemented with AI assistance (Claude Code), including
the fix, the regression test, and this PR description. All commands
listed above were run and their output verified directly.

### DIFF
--- a/src/specify_cli/bundler/services/primitives.py
+++ b/src/specify_cli/bundler/services/primitives.py
@@ -263,9 +263,10 @@ class _ExtensionKindManager:
                 component.version,
                 _bundled_manifest_version(bundled / "extension.yml", "extension"),
             )
-            self._manager.install_from_directory(
+            manifest = self._manager.install_from_directory(
                 bundled, speckit_version, priority=priority, force=force
             )
+            self._manager.scaffold_config(manifest.id)
             return
 
         if not self._allow_network:
@@ -293,9 +294,10 @@ class _ExtensionKindManager:
         )
         zip_path = catalog.download_extension(component.id)
         try:
-            self._manager.install_from_zip(
+            manifest = self._manager.install_from_zip(
                 zip_path, speckit_version, priority=priority, force=force
             )
+            self._manager.scaffold_config(manifest.id)
         finally:
             with contextlib.suppress(Exception):
                 if zip_path.exists():
--- a/tests/unit/test_bundler_primitives.py
+++ b/tests/unit/test_bundler_primitives.py
@@ -7,6 +7,7 @@ offline-first).
 from __future__ import annotations
 
 from pathlib import Path
+from types import SimpleNamespace
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


KSchlobohm:
Please add a regression test for the catalog/ZIP install path. The current test covers only an extension already bundled with Spec Kit, while the reported reproduction uses an extension downloaded from the catalog. The new test should confirm that installing the downloaded ZIP creates the extension's configuration file. Once updated, we'll take another look.

chelsealong:
Added `test_catalog_extension_install_scaffolds_config` in `tests/unit/test_bundler_primitives.py`, which forces the catalog/ZIP branch (`_locate_bundled_extension` returns `None`, `ExtensionCatalog.download_extension` supplies a real ZIP built from an extension with a `provides.config` entry) and asserts the scaffolded config file lands on disk. Confirmed it fails without the fix and passes with it; full suite (7310 passed) and ruff on touched files are unchanged from before (same 4 pre-existing findings noted in the PR description).

lilisalinas936-design (approved):

mnriem:
Thank you!

PR #4310, opened 2026-08-25
fix: exclusive mkdir so concurrent create-new-feature cannot overwrite spec.md

### LINKED ISSUE #4270: [Bug]: create-new-feature.sh reserves feature numbers non-atomically — concurrent invocations can share/overwrite a spec directory
## Version

spec-kit v1.0.1 (`.specify/scripts/bash/create-new-feature.sh`; the flow is also described in `.claude/skills/speckit-specify/SKILL.md`, "Create the directory and spec file")

## Description

Sequential numbering scans existing `specs/` directories, picks max+1, checks for existence, then uses `mkdir -p` and writes `spec.md`. Nothing in that sequence is atomic: two concurrent invocations (parallel agents on worktrees sharing a specs dir, or two terminals) can both scan, both select the same number, both pass the existence check (`mkdir -p` succeeds either way), and both write `spec.md` into the same directory — the second silently overwrites the first's starting specification.

This matters more now that multi-agent setups routinely run more than one spec-kit session against the same repository.

## Expected behavior

Reserve the directory with plain `mkdir` (no `-p`) so creation is exclusive; on `EEXIST`, discard the selected number, rescan, and retry before writing `spec.md`. A lock file would also work.

## Related

Prior sequential-numbering issues (#935, #975, #1332) covered scan-logic bugs in single-invocation scenarios; this one is specifically about the missing atomicity under concurrency.

BetterAndBetterII:
## Problem

`create-new-feature` picks the next sequential spec number, checks that `FEATURE_DIR` is absent, then creates it with `mkdir -p` (bash), `New-Item -Force` (PowerShell), or `Path.mkdir(exist_ok=True)` (Python). Those creates are not exclusive.

Two concurrent invocations can therefore scan the same max number, both pass the exists check, both create the same directory, and the second write clobbers `spec.md`.

## Fix

Create `FEATURE_DIR` exclusively. On `EEXIST`, rescan `specs/` for the next sequential number and retry before writing `spec.md`. Timestamp mode still errors on collision, matching the previous exists-check behavior.

Fixes #4270

### DIFF
--- a/scripts/bash/create-new-feature.sh
+++ b/scripts/bash/create-new-feature.sh
@@ -358,39 +358,17 @@ FEATURE_DIR="$SPECS_DIR/$BRANCH_NAME"
 SPEC_FILE="$FEATURE_DIR/spec.md"
 
 if [ "$DRY_RUN" != true ]; then
-    if [ -d "$FEATURE_DIR" ] && [ "$ALLOW_EXISTING" != true ]; then
-        if [ "$USE_TIMESTAMP" = true ]; then
-            >&2 echo "Error: Feature directory '$FEATURE_DIR' already exists. Rerun to get a new timestamp or use a different --short-name."
-        else
-            >&2 echo "Error: Feature directory '$FEATURE_DIR' already exists. Please use a different feature name or specify a different number with --number."
-        fi
-        exit 1
-    fi
-
-    NEEDS_SPEC=false
-    SPEC_TEMPLATE_FOUND=false
-    SPEC_TEMPLATE_CONTENT=""
-    if [ ! -f "$SPEC_FILE" ]; then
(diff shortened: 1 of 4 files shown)


mnriem (changes requested):

mnriem:
Thanks for the update. The reservation still occurs after existence and `spec.md` state checks, leaving TOCTOU windows that can bypass retries, overwrite an existing spec under `--allow-existing`, or leave a retried directory without a spec.

Please reserve/retry the directory first, then derive all `spec.md` state from the final path. Also add deterministic PowerShell concurrency coverage and handle Python collisions where the target is a regular file.

mnriem:

PR #4313, opened 2026-08-25
fix(implement): count checkbox markers outside code fences only (#4272)

### LINKED ISSUE #4272: [Bug]: /speckit-implement counts checkbox markers inside fenced code blocks — example checkboxes can falsely block implementation
## Version

spec-kit v1.0.1 (`.claude/skills/speckit-implement/SKILL.md`, "Check checklists status")

## Description

The checklist scan counts every `- [ ]` / `- [x]` / `- [X]` line in every checklist file, including lines inside fenced code blocks. A checklist that *documents* the checkbox format with an example fence therefore reports phantom unchecked items, and implementation is blocked by checkboxes that aren't real work items.

The clarify command already handles this correctly — `.claude/skills/speckit-clarify/SKILL.md` specifies parsing Markdown outside code fences — so this is an internal inconsistency between the two commands' parsers as much as a bug.

## Expected behavior

Count checkbox markers only outside fenced code blocks (same contract as the clarify parser), keeping the existing per-checklist total/checked/unchecked reporting.

ntdatt812:
Closes #4272.

`templates/commands/implement.md` told the agent to count checkbox markers like this:

```
- Total items: All lines matching `- [ ]` or `- [X]` or `- [x]`
```

Every line, fenced blocks included. So a checklist that *documents* the checkbox format — an example fence showing what `- [ ]` looks like — reported unchecked items that no reviewer can ever tick, and the gate stops on a non-zero unchecked count. Writing down the format blocked implementation.

`templates/commands/clarify.md` already had the right rule and the right words for it:

> lines matching `- [ ]`, `- [x]`, or `- [X]` … **outside of code fences**

So this was also two commands disagreeing about what a checklist item *is*, which is the part worth fixing rather than just the count.

Only `templates/commands/implement.md` needed the change: `grep` for the scan instruction across `templates/commands/` and `presets/*/commands/` returns exactly these two files, and the lean preset's `speckit.implement.md` does not carry a counting block.

## Test

`tests/unit/test_checklist_scan_contract.py` turns the rule into a contract rather than a one-time edit. It walks every command template, finds each line that defines what counts as a checkbox marker, and asserts that line also excludes code fences. It is parametrized per instruction, so the failure names the file, the line number and the offending text:

```
FAILED …::test_marker_scans_exclude_code_fences[implement.md-63-     - Total items: All lines matching `- [ ]` …]
```

Two details that make it a real guard rather than a passing test:

- a companion `test_the_contract_is_actually_stated_somewhere` fails if the regex ever matches nothing, so deleting or rewording the scan instruction cannot make the suite green by making the parametrization empty;
- it scans `presets/*/commands` too, so a preset that grows its own counting block is held to the same rule.

**Mutation-checked** — restoring the old wording in `implement.md` fails exactly one case, the implement one, and the other three stay green:

```
1 failed, 3 passed
```

```
python -m pytest tests/unit/test_checklist_scan_contract.py -q
4 passed
```

No behaviour outside the checklist scan changes: the per-checklist total/checked/unchecked reporting and the status table are untouched.


mnriem (changes requested):

ntdatt812:
Closing in favour of #4520, which consolidates this with #4314 and #4330 as asked on #4460. This fix is carried over unchanged, in its own commits.

Copilot's point is addressed there: the scan-instruction pattern only recognised a definition whose first marker is unchecked, so the `- [X]` checked-items definition was never guarded, and removing its code-fence exclusion left the suite green (measured: 4/4 passed). It now matches either marker, four definitions are guarded instead of three, and that removal fails.

Issue #2968, opened 2026-06-13
[Bug]: /speckit-taskstoissues creates duplicate GitHub issues when issues already exist


yucer:
## Describe the bug

`/speckit-taskstoissues` does not check for existing GitHub issues before creating new ones. If the command is run a second time (e.g. after re-generating `tasks.md`, or accidentally re-invoking the skill), it creates duplicate issues for every task without warning.

## Steps to reproduce

1. Run `/speckit-taskstoissues` on a project — issues are created for all tasks (e.g. T001–T027).
2. Run `/speckit-taskstoissues` again on the same project without any changes.
3. Observe that a second set of identical issues is created.

## Expected behavior

Before creating an issue, the skill should search for an existing open (or closed) issue whose title contains the task ID (e.g. `T001`, `[T001]`). If a match is found, skip creation and optionally log "Issue for T001 already exists (#NNN) — skipping."

## Actual behavior

A duplicate issue is created for every task, with no warning or deduplication check. The user must manually close/delete the duplicates.

## Suggested fix

Add a deduplication step before the issue-creation loop:

```
gh issue list --repo <owner>/<repo> --limit 500 --state all
```

Parse the result to extract task IDs already present in issue titles (matching patterns like `T001`, `[T001]`, `T001:`). Skip `gh issue create` for any task whose ID is already found.

## Environment

- spec-kit version: `0.8.14.dev0`
- Skill: `speckit-taskstoissues`
- Host: Claude Code (claude-sonnet-4-6)

PR #4314, opened 2026-08-25
fix(taskstoissues): scope issue dedup to the feature the tasks belong to (#4271)

### LINKED ISSUE #4271: [Bug]: /speckit-taskstoissues dedup matches bare task IDs across features — valid tasks are silently skipped
## Version

spec-kit v1.0.1 (`.claude/skills/speckit-taskstoissues/SKILL.md`, dedup step around the `list_issues` flow)

## Description

Task IDs like `T001` are local to one feature's `tasks.md` and restart from T001 in every feature. The dedup step, however, matches repository issues by task ID without scoping to the current feature: if feature `001-auth` already produced an issue titled with `T001`, then running `/speckit-taskstoissues` for feature `002-billing` sees "T001 exists" and **skips creating the issue for billing's T001** — a silent gap in the tracker, in exactly the multi-feature repos this command targets.

This is the inverse of #2968 (which was about creating duplicates): the fix direction there makes the matching aggressive enough to now suppress valid tasks.

## Expected behavior

- Include a stable feature identifier (the `NNN-name` spec dir) in each created issue's title or a label, and require **both** the feature identity and the task ID to match before skipping.
- Re-check for existing feature+task issues immediately before creation so two concurrent invocations don't race the fetch-then-create sequence.

ntdatt812:
Closes #4271.

Task IDs are local to a feature — every `tasks.md` restarts at `T001` — but the dedup step matched existing issues on the bare ID:

> For each issue title, match it against the task ID pattern `\bT\d{3,}\b` … and, when it matches one of your task IDs, mark that ID as already having an issue.

So once `001-auth` had an issue titled `T001: …`, running the command for `002-billing` saw "T001 exists" and skipped it. The issue was never created and the report said `T001 already has an issue, skipping`, which reads like success. Every feature after the first loses its low-numbered tasks, and it degrades with repo age rather than showing up on the first run.

This is the inverse of #2968: tightening the matching to stop duplicates made it aggressive enough to suppress valid tasks.

## What changed

**The canonical title now carries the feature.** `FEATURE_DIR` is already parsed in step 1, so its basename — the `NNN-name` spec directory — goes in front:

```
[002-billing] T001: Create project structure
```

The ID keeps its own word boundaries inside that prefix, so the `\bT\d{3,}\b` matching #2968 established still works unchanged, including the four-digit case.

**A task is skipped only when both match.** Feature identity *and* task ID. A `T001` belonging to another feature is a different task and no longer suppresses this one.

**Existing issues still count.** Anything filed before the prefix existed carries a bare `T001: …`. Those are treated as matching only when no `[<feature>]`-prefixed issue for that ID is present in the fetched set, so upgrading does not re-create work that is already tracked — which would have been the obvious way to turn this fix into a fresh duplicate problem.

## On the choice of title over label

The issue offers "title or a label". Title, for two reasons: the dedup already reads titles from `list_issues`, so nothing new is fetched and no extra call is added to a step that is explicitly written to bound its pagination; and a label needs create-label permission and a repo-side convention, which is a heavier requirement for a command whose whole contract is "create issues on the remote". A label would be the better choice if the identifier had to be filterable in the GitHub UI — happy to switch if you would rather have that.

I did not implement the second bullet of the issue (re-checking immediately before creation so two concurrent invocations cannot race). It is a real gap, but it is a different change — the fetch-then-create window — and it is worth its own PR rather than being folded in here where it would be untested.

## Tests

`tests/unit/test_taskstoissues_feature_scope.py`, three cases over the template text:

- the canonical title rule names `[<feature>] T001:` **and** says where `<feature>` comes from
- the skip rule requires *both* the feature and the ID
- the pre-existing-issue clause is still present, so an upgrade cannot start duplicating

Each assertion fails with the offending instruction line quoted, and `_line_containing` raises if the instruction disappears entirely — so deleting the step cannot make the suite pass vacuously.

**Mutation-checked** — dropping `[<feature>]` from the canonical title fails exactly one case:

```
1 failed, 2 passed
```

```
python -m pytest tests/unit/ -q
469 passed, 2 skipped, 2 failed
```

The two failures are `test_add_source_refuses_symlinked_specify_escape` and `test_save_records_refuses_symlinked_specify_escape`. Both fail identically on `main` on this machine — creating a symlink on Windows needs elevation — measured on a clean checkout before this branch, not assumed.


mnriem (changes requested):
Please address Copilot feedback. Note that the change should backwards compatible as it will impact anyone using this feature already

ntdatt812:
The red was mine, and it was not what it looked like. Fixed in `df9488a`.

`pytest (ubuntu-latest, 3.13)` failed on `test_yaml_has_prompt`:

```
AssertionError: speckit.taskstoissues.yaml missing prompt block scalar
```

Nothing to do with the dedup change. The sentence I added — *"so the `\bT\d{3,}\b` matching above is unchanged by the prefix"* — reached the file with two literal **U+0008 BACKSPACE bytes** where `\b` was meant; my editing pipeline interpreted the escape instead of passing it through. Invisible in the diff, in the rendered markdown, and in the PR view.

A control character cannot live in a YAML block scalar, so the generator fell back to a double-quoted flow scalar for the entire prompt, and the goose parity test — which asserts every generated command uses `prompt: |` — failed on this one file.

Pinned rather than assumed, both directions:

- with the two bytes present: `1 failed`, with that exact assertion message
- with them written as `\b`: `tests/integrations/test_integration_goose.py` **38 passed**, `tests/unit/test_taskstoissues_feature_scope.py` **3 passed**

The rendered sentence is unchanged — it always meant to read `\b`. The other five `pytest` jobs were cancelled by fail-fast, not failing on their own.

mnriem:

ntdatt812:
Closing in favour of #4520, which consolidates this with #4313 and #4330 as asked on #4460. The feature-scoped titles are carried over unchanged, in their own commits.

Both review points are addressed there. Copilot was right that the legacy rule brought #4271 back: treating a bare `T001: ...` title as this feature's whenever no scoped title exists is exactly the state of every feature on its first run after upgrading. A bare title names no feature, so no inference from the tracker can decide it, and deciding either way breaks someone: skipping drops a task silently, creating duplicates one an existing user already tracks. For backward compatibility, the command now lists every ID that matched only a bare title and asks before creating anything; confirmed issues are skipped, and retitled to the scoped form only if the user agrees, so the question does not recur. The U+0008 bytes were already fixed in `df9488a`.

PR #4319, opened 2026-08-25
fix(workflows): reject falsy non-mapping catalog config in add/remove

Noor-ul-ain001:
## Summary
- `WorkflowCatalog.remove_catalog`, `StepCatalog.add_catalog`, and `StepCatalog.remove_catalog` (all in `src/specify_cli/workflows/catalog.py`) read their config file with `yaml.safe_load(config_path.read_text(...)) or {}`.
- That `or {}` coerces a **falsy non-mapping** top-level document (`[]`, `false`, `0`, `''`) to `{}` *before* the `isinstance(data, dict)` guard runs, so a corrupted config silently becomes "no catalogs" instead of raising `"...corrupted (expected a mapping)"`. A *truthy* non-mapping (`5`, a bare list with items) already raised correctly — this was an inconsistency, and for `remove_catalog` it surfaces as a misleading `"Catalog index N out of range"` error instead.
- `WorkflowCatalog._load_catalog_config` (the shared loader behind `get_active_catalogs`) and `WorkflowCatalog.add_catalog` already guard against exactly this, with an explanatory comment (`# Do NOT coerce with or {} here: ...`). The other three call sites in the same file reimplement the read inline and were missed.
- Same bug shape as #4094 (preset catalog add/remove) and #4187 (integration descriptor loading) — falsy-or-coerce masking a shape error.

## Test plan
- [x] Added `test_remove_catalog_rejects_falsy_non_mapping_config` to `TestWorkflowCatalog` and `test_add_catalog_rejects_falsy_non_mapping_config` / `test_remove_catalog_rejects_falsy_non_mapping_config` to `TestStepCatalog`, each parametrized over `[]`, `false`, `0`, `''`.
- [x] Verified all 12 new parametrized cases fail without the fix (stashed only the source change) and pass with it.
- [x] Ran `TestWorkflowCatalog` + `TestStepCatalog` in full — 100 passed, no regressions.


mnriem:
Thanks — good fix, but it's already in `main`. The `or {}` → explicit-`isinstance` guard for falsy non-mapping catalog config landed via another change (see `catalog.py:347`, which now carries exactly this fix and comment). After your recent merge of `main`, this PR's net diff is empty, so I'm closing it as already-applied. Appreciate the contribution — the outcome you wanted is in place.

PR #4320, opened 2026-08-25
fix(presets): reject falsy non-mapping catalog config shapes

Noor-ul-ain001:
## Summary
`PresetCatalog._load_catalog_config` (`src/specify_cli/presets/__init__.py`) has two instances of the same "emptiness check runs before the shape check" bug, both of which mask a corrupted `preset-catalogs.yml` as an empty/no-op config instead of raising a clear error:

1. **Top level**: `data = yaml.safe_load(config_path.read_text(...)) or {}` coerces a FALSY non-mapping document (`[]`, `false`, `0`, `''`) to `{}` *before* the `isinstance(data, dict)` guard runs — silently swallowed as "no config", while a TRUTHY non-mapping (a bare string) correctly raises `"expected a mapping at root"`.
2. **One level down**: `catalogs_data = data.get("catalogs", [])` followed by `if not catalogs_data: return None` checks emptiness *before* checking `isinstance(catalogs_data, list)`. A FALSY non-list `catalogs:` value (`{}`, `''`, `0`, `false`) is silently treated as "no catalogs", while a TRUTHY non-list (`catalogs: "not-a-list"`) correctly raises `"must be a list"`.

`WorkflowCatalog._load_catalog_config` and `StepCatalog._load_catalog_config` (`src/specify_cli/workflows/catalog.py`) already guard against exactly both cases, with explanatory comments (`# Do NOT coerce with or {} here...` / `# Same asymmetry ... one nesting level down`). This preset sibling wasn't updated to match. Fix reuses the same pattern and comments.

## Test plan
- [x] Added `test_load_catalog_config_rejects_falsy_non_mapping_root` (parametrized over `[]`, `false`, `0`, `''`) and `test_load_catalog_config_rejects_falsy_non_list_catalogs` (parametrized over `catalogs: {}`, `''`, `0`, `false`) to `TestPresetCatalogMultiCatalog`.
- [x] Verified all 8 new parametrized cases fail without the fix (stashed only the source change) and pass with it.
- [x] Ran `TestPresetCatalogMultiCatalog` in full — 33 passed, no regressions.


mnriem:
Thank you!

PR #4321, opened 2026-08-25
fix(workflows): reject falsy non-mapping step.yml in step add

Noor-ul-ain001:
## Summary
- `workflow_step_add` (`src/specify_cli/workflows/_commands.py`) parses a fetched `step.yml` with `_yaml.safe_load(step_yml_content.decode("utf-8")) or {}`.
- That `or {}` coerces a **falsy non-mapping** top-level document (`[]`, `false`, `0`, `''`) to `{}` *before* the `if not isinstance(meta, dict): ...` shape check runs. The command then proceeds with `meta = {}`, derives `step_meta = meta.get("step", {})` → `{}` and `type_key = ""`, and reports the unrelated `"step.yml missing 'step.type_key' field"` instead of the real problem: `"step.yml must be a YAML mapping"`.
- A **truthy** non-mapping document (e.g. a bare string) already hits the correct `isinstance` check and reports the right error — this was an inconsistency between falsy and truthy malformed inputs.
- Same falsy-or-coerce shape as the catalog-config bugs already fixed this cycle in `workflows/catalog.py` (`WorkflowCatalog`/`StepCatalog`), `presets/__init__.py` (`PresetCatalog._load_catalog_config`), and `integrations` (#4187).

## Test plan
- [x] Added `test_add_rejects_falsy_non_mapping_step_yml` to `TestWorkflowStepAddCLI`, parametrized over `[]`, `false`, `0`, `''`, mocking the HTTP fetch so `step.yml`'s body is the falsy document.
- [x] Verified all 4 new parametrized cases fail without the fix (stashed only the source change, confirmed the misleading `"missing 'step.type_key'"` message) and pass with it.
- [x] Ran `TestWorkflowStepAddCLI` in full — 24 passed; the 1 remaining failure (`test_add_rejects_symlinked_steps_base_dir`) is a pre-existing Windows symlink-elevation failure unrelated to this change (reproduces on an unmodified checkout).

### DIFF
--- a/src/specify_cli/workflows/_commands.py
+++ b/src/specify_cli/workflows/_commands.py
@@ -3364,12 +3364,18 @@ def workflow_step_add(
         try:
             import yaml as _yaml
 
-            meta = _yaml.safe_load(step_yml_content.decode("utf-8")) or {}
+            meta = _yaml.safe_load(step_yml_content.decode("utf-8"))
         except Exception as exc:
             console.print(f"[red]Error:[/red] Invalid step.yml: {exc}")
             raise typer.Exit(1)
 
-        if not isinstance(meta, dict):
+        # Do NOT coerce with ``or {}`` here: that also turns a FALSY non-mapping
+        # (top-level ``[]``, ``false``, ``0``, ``''``) into ``{}`` and silently
+        # bypasses this shape check, surfacing the unrelated "missing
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):

mnriem (approved):

mnriem:
Thank you!

PR #4326, opened 2026-08-25
fix(events): stop `event run` crashing on every piped stdin payload

Noor-ul-ain001:
## Summary
- `event_run` (`src/specify_cli/commands/event.py`) capped its stdin read at 1 MiB to prevent a DoS (#3857), but the truncation check reads `sys.stdin.eof` — **that attribute does not exist** on any Python file-like object, including `sys.stdin` (`hasattr(sys.stdin, "eof")` is `False`).
- Every piped-stdin invocation raised `AttributeError: '...' object has no attribute 'eof'` instead of running. Piped stdin is this command's documented primary use case ("Resolve and run an event-driven command script with **stdin payload**" — a native hook feeds it a JSON payload this way), and `isatty()` is `False` whenever stdin isn't an interactive terminal, so this fired on essentially every real invocation, not only oversized ones — the #3857 DoS fix left the feature entirely broken.
- The intended oversized-payload branch was also broken a second, independent way: `raise typer.Exit(code=1, message="...")` — `typer.Exit.__init__` accepts only `code` (verified via `inspect.signature`), not `message` — so that path raised `TypeError` instead of the documented clean error.
- Fix: detect truncation the standard way — after reading the 1 MiB cap, read one more byte; a non-empty result means more data was waiting beyond it. Report the oversized-payload error via `typer.echo(..., err=True)` before `raise typer.Exit(code=1)`.

## Test plan
- [x] Added `tests/test_event_command.py` (no prior test coverage existed for this command): a normal piped payload reaches the handler intact, a TTY/no-stdin invocation falls back to `"{}"`, and an oversized piped payload exits 1 with the limit message instead of crashing.
- [x] Verified all 3 tests fail without the fix — reproduced the exact `AttributeError('...' object has no attribute 'eof')` on every case, including the "no stdin" one (confirms `isatty()` is `False` under non-interactive invocation, matching real hook usage) — and pass with it.
- [x] Ran the new test module standalone — 3 passed.

### DIFF
--- a/src/specify_cli/commands/event.py
+++ b/src/specify_cli/commands/event.py
@@ -28,12 +28,18 @@ def event_run(
     MAX_STDIN_BYTES = 1 * 1024 * 1024
     if not sys.stdin.isatty():
         raw = sys.stdin.read(MAX_STDIN_BYTES)
-        if not sys.stdin.eof:
-            raise typer.Exit(
-                code=1,
-                message="stdin payload exceeds 1 MiB limit; "
+        # File-like objects have no ``.eof`` attribute in Python; that check
+        # always raised AttributeError, crashing every piped-stdin invocation
+        # (the command's primary use case) regardless of payload size. Detect
+        # truncation instead by reading one more byte once the cap is hit: a
+        # non-empty result means more data was waiting beyond the cap.
+        if len(raw) >= MAX_STDIN_BYTES and sys.stdin.read(1):
+            typer.echo(
+                "stdin payload exceeds 1 MiB limit; "
                 "truncate or pipe a smaller payload",
+                err=True,
             )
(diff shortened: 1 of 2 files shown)


KSchlobohm:

KSchlobohm:
Thanks for the quick update to add the try/catch. Our standard is that any new behavior branch carries both a positive and a negative test — the happy path and the failure path. The valid-UTF-8 path is already covered, but the new `UnicodeDecodeError` guard (the negative case) isn't, so nothing proves it exits cleanly or guards against regression.


Could you add a test that pipes invalid UTF-8 (e.g. `b"\xff\xfe"`) and asserts exit code 1, the "must be valid UTF-8" message, and that the handler isn't called?

PR #4330, opened 2026-08-26
fix(converge): do not re-append work an unchecked task already tracks (#4269)

### LINKED ISSUE #4269: [Bug]: /speckit-converge is not idempotent — re-runs append duplicate remediation tasks
## Version

spec-kit v1.0.1 (`.claude/skills/speckit-converge/SKILL.md`)

## Description

`/speckit-converge` only checks that `tasks.md` exists before assessing the codebase and appending remediation tasks. It keeps no persisted evidence that `/speckit-implement` actually ran against the *current* task list, and it does not map its findings onto existing **unchecked** tasks before appending.

Two consequences:

1. **Invoked early** (before implement has run, or after a partial implementation), it appends remediation tasks for work that is already represented by existing unchecked tasks — now under new IDs, splitting traceability for the same work across two task entries.
2. **Invoked twice** (e.g. converge → partial fix → converge again), gaps found both times are appended both times.

## Expected behavior

- Converge verifies prerequisite state (an implementation checkpoint, or at minimum warns loudly when `tasks.md` contains unchecked tasks that overlap its findings).
- Before appending, each finding is compared against existing unchecked tasks; covered work is reused/updated rather than duplicated, and only genuinely new gaps get new IDs.

## Context

Found while upgrading a template fleet from 0.5.0 to v1.0.1 and reviewing the newly vendored command. Related in spirit to the duplicate-detection work in #2968, but on the tasks.md side rather than GitHub issues.

ntdatt812:
Closes #4269.

Step 7 knew how to compute the next task ID and nothing about whether a finding was already represented:

```
1. Scan all existing task IDs; let `M` be the maximum. …
3. Emit one checklist item per actionable finding … assigning IDs `T{M+1:03d}, …`
```

One item **per finding**, unconditionally. So the two cases in the issue both land the same way: run converge twice and gaps found both times are appended twice; run it before `/speckit-implement` has worked the list and it appends remediation for work the plan already tracks. The part that makes it hard to undo is that the second run **cannot tell its own previous output apart from the plan's original tasks** — both are just unchecked lines by then — so the duplicates accumulate and one piece of work ends up split across two IDs.

## What changed

Findings are matched against the existing unchecked tasks before anything is appended, and an already-covered finding is dropped rather than re-issued.

Three details that matter more than the rule itself:

- **Matched on substance, not wording.** The comparison is on the work described and the file paths or `<source-ref>` named — converge writes its own descriptions, so a string match would never fire on its own previous output.
- **Scanned outside code fences**, the rule `/speckit-clarify` already applies and the one #4272 just brought `/speckit-implement` in line with. A checklist that documents the checkbox format is not tracked work.
- **Dropped findings are reported** (`F3 — already tracked by T017, not appended`). A silently discarded finding is indistinguishable from one that was never found. And when *every* finding is already tracked, the run takes the `converged` path **saying why** — so the operator reads "the work is known", not "the codebase is complete". That distinction is the whole value of the run.

## What I deliberately did not do

The issue's first bullet asks converge to verify a prerequisite implementation checkpoint. That needs a new piece of persisted state and a decision about what counts as "implement has run against *this* task list" — a design call for maintainers, not something to infer inside a dedup fix. This PR takes the second bullet only, which is self-contained and is what stops the duplicates.

## Tests

`tests/unit/test_converge_idempotency.py`, four cases over the template:

- the append step still exists — so the rest cannot pass vacuously if it is renamed away
- findings are compared against existing **unchecked** tasks, and the marker `- [ ]` is named so the rule is executable
- the scan excludes code fences
- an already-tracked finding is reported, not silently dropped

**Mutation-checked** — weakening "unchecked task" to "existing tasks" and dropping the marker fails exactly one case, the comparison one, and leaves the other three green.

```
python -m pytest tests/unit/test_converge_idempotency.py -q
4 passed

python -m pytest tests/unit/ -q
470 passed, 2 skipped, 2 failed
```

The two failures are `test_add_source_refuses_symlinked_specify_escape` and `test_save_records_refuses_symlinked_specify_escape`, which fail identically on `main` on this machine — creating a symlink on Windows needs elevation. Measured on a clean checkout, not assumed.

### DIFF
--- a/templates/commands/converge.md
+++ b/templates/commands/converge.md
@@ -201,6 +201,23 @@ Append to the **end** of `tasks.md`, per the append contract:
 
 1. Scan all existing task IDs; let `M` be the maximum. Determine the next phase number `N`
    (highest existing phase + 1).
+
+   **Drop findings that existing unchecked tasks already cover.** Compare each actionable
+   finding against every `- [ ]` task already in `tasks.md` — outside code fences, the same
+   rule `__SPECKIT_COMMAND_CLARIFY__` applies — matching on the work described and the file
+   paths or `<source-ref>` it names, not on wording. A finding that an unchecked task
+   already represents is **not** appended: it is unbuilt work that is already tracked, and
+   appending it again splits one piece of work across two IDs.
+
+   This is what makes the command idempotent. Converge run twice, or run before
+   `__SPECKIT_COMMAND_IMPLEMENT__` has worked through the list, would otherwise append the
+   same remediation tasks under fresh IDs each time — the second run cannot tell its own
+   previous output apart from the plan's original tasks.
+
+   Report the dropped ones in the summary rather than silently discarding them:
+   `F3 — already tracked by T017, not appended`. If **every** finding is already tracked,
+   there is nothing to append: take the `converged` path below and say why, so the operator
+   sees "the work is known" rather than "the codebase is complete".
 2. Write a single new section header `## Phase N: Convergence`.
 3. Emit one checklist item per actionable finding, ordered CRITICAL/HIGH first, assigning
    zero-padded IDs `T{M+1:03d}, T{M+2:03d}, …`:
--- a/tests/unit/test_converge_idempotency.py
+++ b/tests/unit/test_converge_idempotency.py
@@ -1,0 +1,57 @@
+"""`/speckit-converge` must not append work that an unchecked task already tracks.
+
+The command's only write is appending remediation tasks to `tasks.md`. It knew how to
+compute the next task ID, but not whether a finding was already represented — so running
+it twice, or running it before `/speckit-implement` had worked through the list, appended
+the same work again under fresh IDs and split one piece of work across two entries
+(#4269). The dedup rule is what makes the command idempotent, so it is pinned here rather
+than left to survive on prose alone.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


mnriem:
This PR addresses the duplication symptom, but the root issue is that `converge` does not enforce its prerequisite. If `tasks.md` contains unchecked tasks, it should stop before analysis and tell the user to run `implement`.

Only after all tracked tasks are complete should `converge` assess the implementation against every SDD artifact, including each task. Any newly discovered work is then appended, making the next convergence attempt stop until that work is implemented.

Please adjust the PR around this lifecycle rather than routing already-tracked incomplete work through the `converged` outcome, which incorrectly reports that implementation is complete.

ntdatt812:
Closing in favour of #4520, which consolidates this with #4313 and #4314 as asked on #4460.

Reworked there around the lifecycle you described, rather than the dedup rule. Step 1 now stops before any assessment while `tasks.md` has an unchecked task (outside code fences), lists the open IDs, points to `/speckit-implement` and leaves `tasks.md` untouched; that is neither `converged` nor `tasks_appended`. Once every task is checked, each task is assessed along with the other artifacts, so one marked done but not built becomes a finding traced to its ID, and anything appended makes the next run stop until it is implemented. The dedup paragraph is removed, and the handoff and `docs/reference/agentic-sdd.md` describe the gate. Copilot's note about pinning the substance-based matching no longer applies, since that rule is gone.

PR #4337, opened 2026-08-26
fix(events): cap stdin in the generated dispatcher, not just the CLI command

Noor-ul-ain001:
## Summary
- The #3857 fix capped stdin at 1 MiB in `specify event run` (`src/specify_cli/commands/event.py`), but that CLI command is not the code path native hooks actually invoke.
- Every installed integration writes a self-contained `.specify/events.py` dispatcher (the `_EVENTS_DISPATCHER_TEMPLATE` string in `src/specify_cli/events.py`), and native hook configs call that dispatcher directly. Its `main()` did:
  ```python
  payload = sys.stdin.read() if not sys.stdin.isatty() else "{}"
  ```
  with **no size cap at all** — the exact DoS #3857 was meant to close, wide open on the primary invocation path. `specify event run` is a secondary/manual entry point; the generated dispatcher is what actually runs on every `session_start`/`pre_tool_use`/etc. hook fire in real usage.
- Fix: apply the same byte-capped read (from the binary buffer, so the cap counts encoded bytes rather than decoded characters) inside the dispatcher template, so every newly-installed or refreshed dispatcher enforces the limit.

## Test plan
- [x] Added 3 tests in `tests/integrations/test_events.py::TestCommandRunner`: an oversized payload exits 1 with the limit message instead of running unbounded, a multibyte payload (~300k emoji, ~1.14 MiB UTF-8 but only 300k characters) is still rejected by the byte-based cap, and a normal under-the-cap payload still reaches the handler script unchanged.
- [x] Verified both new tests fail without the fix (test-the-test): the oversized-payload test failed because the dispatcher silently accepted the full payload and returned "not found" instead of exiting 1 with the limit message — reproducing the exact bug.
- [x] Ran the full `tests/integrations/test_events.py` suite: 124/128 pass; the remaining 4 are pre-existing Windows symlink-elevation failures unrelated to this change.

https://claude.ai/code/session_01PJHJ2dHP2RVCNncHqN8Qm9


mnriem (changes requested):

mnriem:
Good catch — #3857 capped `specify event run` but left the generated dispatcher (the path native hooks actually invoke) uncapped, so this completes that DoS fix on the real hot path. Prioritizing it. One merge-blocker from the review before it lands: on Windows (Python 3.11–3.14) the UTF-8 decode feeds `subprocess.run(..., text=True)` with no explicit `encoding=`, so locale encoding can corrupt UTF-8 payloads (`events.py:397`). Please pass an explicit `encoding="utf-8"` there and add a test for the non-ASCII payload path. Re-request review once that's in.

mnriem (approved):

mnriem:
Thank you!

Issue #4345, opened 2026-08-26
[Bug]: Bundled extensions (agent-context, git) change without extension.yml version bumps — 'extension update' reports Up to date and installed copies never get fixes


CrazyBaran:
### Summary

`specify extension update` decides whether an installed extension needs updating by comparing semver versions (`src/specify_cli/extensions/_commands.py`, `extension_update`, ~line 1660):

```python
if catalog_version > installed_version:
    ...offer update...
else:
    console.print(f"✓ {ext_id}: Up to date (v{installed_version})")
```

Both bundled extensions (`agent-context`, `git`) have carried `version: "1.0.0"` since the day they were created, while their content kept changing — including bug fixes for failures that make an extension unusable on some platforms. Any project that installed them before those fixes is told **"Up to date (v1.0.0)"** forever and never receives them. Nothing else catches it: the registry's `manifest_hash` covers the manifest (which didn't change), not the scripts.

### Quantified drift (checked 2026-08-26 against `main`)

**`extensions/agent-context`** — `version: "1.0.0"` set at extraction (#2546, 2026-05-30), never bumped since (the manifest was never touched again). Changes an installed v1.0.0 copy is missing:

- **15 commits**, aggregate **7 files changed, +1,120 / −169 lines** (the extension is 8 files total — nearly every file rewritten)
- Among them:
  - fix: PowerShell script failing on **every run** under Windows PowerShell 5.1 (multi-line `python -c` argument mangling) — fixed incidentally in #2969
  - feat: `context_files` multi-file support + path-containment validation (#2969)
  - fix: derive plan path from `feature.json` (#3069)
  - feat: full Python port (`scripts/python/update_agent_context.py`, 368 lines) (#3387)
  - fix: nested/scoped plan discovery (#3024, #3405, #3734)
  - fix: blank config markers (#3736), temp-file cleanup masking errors (#3803)

**`extensions/git`** — `version: "1.0.0"` set at creation (#1941, 2026-04-07), never bumped; the manifest was edited once more (#3293, 2026-07-06) and the version *still* stayed 1.0.0. Changes an installed v1.0.0 copy is missing:

- **23 commits**, aggregate **21 files changed, +2,191 / −567 lines** (the extension is 21 files total — every file touched)
- Among them:
  - fix: PS 5.1 compat — non-ASCII chars in shipped PowerShell scripts (#2709)
  - feat: Python port of all scripts (#3400)
  - feat: configurable Conventional Commit support (#3390)
[...]

PR #4350, opened 2026-08-27
Fix #4345: Extension version staleness

ranka23:
Bumps agent-context and git versions to 1.1.0. Manually tested. AI-assisted.

### DIFF
--- a/extensions/agent-context/extension.yml
+++ b/extensions/agent-context/extension.yml
@@ -3,7 +3,7 @@ schema_version: "1.0"
 extension:
   id: agent-context
   name: "Coding Agent Context"
-  version: "1.0.0"
+  version: "1.1.0"
   description: "Manages coding agent context/instruction files (e.g., CLAUDE.md, copilot-instructions.md) with project-specific plan references and configurable markers"
   author: spec-kit-core
   repository: https://github.com/github/spec-kit
--- a/extensions/catalog.json
+++ b/extensions/catalog.json
@@ -6,7 +6,7 @@
     "agent-context": {
       "name": "Coding Agent Context",
       "id": "agent-context",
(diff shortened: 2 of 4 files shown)


KSchlobohm:
**Maintainer note:** We’re carrying this PR forward as a narrow version-metadata correction for future installs of the bundled `agent-context` and `git` extensions. Existing installations require the known workaround after upgrading to a Spec Kit release containing these versions.

**Known workaround**

```bash
specify extension add agent-context --force
specify extension add git --force
```

mnriem:
To make this mergeable as a complete fix for #4345, the following still needs to be addressed:

1. Update the exact `agent-context` pins from `1.0.0` to `1.1.0` in the four `examples/bundles/*/bundle.yml` manifests and the fixtures in `tests/integration/test_bundler_{init_install,local_install}.py`. These are examples/tests rather than a production catalog bundle, but the bundler enforces their pins against the embedded manifest.
2. Fix the bundled-extension update path. `extension update` currently discovers `1.1.0` from the live catalog and then calls `download_extension()`, which rejects bundled entries without a `download_url`. For a bundled extension, the available version should come from the CLI's embedded `extension.yml`, and the update should reinstall that embedded directory atomically with `force=True`, preserving the existing config and rollback behavior. This also prevents an older CLI embedding `1.0.0` from offering a live-catalog `1.1.0` update it cannot provide.
3. Add regression coverage for updating an installed older bundled extension from a newer embedded copy, and for catalog/manifest version parity. The proposed CI guard requiring a manifest version bump whenever shipped extension content changes should also be included or tracked explicitly as follow-up work.

The extension version remains independent from the CLI version; this only requires the catalog, embedded manifest, and exact example/test pins to agree on the extension release being shipped.

Issue #4328, opened 2026-08-26
[Bug]:  __SPECKIT_COMMAND_<NAME>__ still can't represent hyphenated command names — #4204 stalled after review


minzzang144:
### Bug Description

`__SPECKIT_COMMAND_<NAME>__` still can't represent command names containing a hyphen (e.g. the bundled `speckit.agent-context.update`), because the token's character class doesn't allow `-`:

```python
# src/specify_cli/integrations/base.py
r"__SPECKIT_COMMAND_([A-Z][A-Z0-9_]*)__"
```

This was reported in #4198. A fix landed in #4204 by adding a second token form, `__SPECKIT_COMMAND(...)__`, carrying the command id verbatim. During review, a simpler alternative was proposed instead: keep the single existing token form and just widen its character class to allow the literal hyphen. The maintainer (@mnriem) requested this direction explicitly ("Please go for option 2", `CHANGES_REQUESTED`, 2026-08-21), but #4204 has seen no further activity since. Filing this so the agreed direction doesn't get lost.

### Suggested fix

Widen the character class in both `resolve_command_refs()` (`src/specify_cli/integrations/base.py`) and the extension-skills resolver (`src/specify_cli/extensions/__init__.py`):

```python
# before
r"__SPECKIT_COMMAND_([A-Z][A-Z0-9_]*)__"
# after
r"__SPECKIT_COMMAND_([A-Z][A-Z0-9_-]*)__"
```

No decode-logic change is needed — `.replace("_", separator)` already leaves a literal `-` untouched, so e.g. `__SPECKIT_COMMAND_AGENT-CONTEXT_UPDATE__` resolves to `/speckit.agent-context.update`. This works for any number of segments and any number of hyphens per segment. A tested reference implementation (including test coverage and a docs update) is available at:
https://github.com/github/spec-kit/compare/main...minzzang144:spec-kit:fix/4198-command-ref-token-hyphen

### Steps to Reproduce

```python
from specify_cli.integrations.base import IntegrationBase as I

# actual bundled command: speckit.agent-context.update
I.resolve_command_refs("__SPECKIT_COMMAND_AGENT_CONTEXT_UPDATE__", ".")
```

### Expected Behavior

The token resolves to the command id that is actually installed: `/speckit.agent-context.update`.

### Actual Behavior

The hyphen is interpreted as a segment separator, so a three-segment command renders as four segments, with no warning:

```
[...]

PR #4356, opened 2026-08-27
allow hyphen in command ref token names

### LINKED ISSUE #4198: [Bug]:  __SPECKIT_COMMAND_*__ tokens can't reference commands whose names contain hyphens
### Bug Description

`__SPECKIT_COMMAND_<NAME>__` tokens encode a command name in upper case with underscores, and at render time each underscore is replaced with the agent's separator (`.` or `-`):

  ```python
  # src/specify_cli/integrations/base.py
  r"__SPECKIT_COMMAND_([A-Z][A-Z0-9_]*)__"
  ... + m.group(1).lower().replace("_", separator)
  ```

  Command names themselves, however, are allowed to contain hyphens:

  ```python
  # src/specify_cli/extensions/__init__.py
  EXTENSION_COMMAND_NAME_PATTERN = re.compile(r"^speckit\.([a-z0-9-]+)\.([a-z0-9-]+)$")
  ```

  Since the token grammar `[A-Z][A-Z0-9_]*` has no character available to represent a hyphen, a command whose name contains one cannot be referenced by a token. What makes this hard to notice is that it doesn't fail — it silently renders as a **different** command name.

  It is especially easy to miss because in skills mode, where `invoke_separator` is `-`, the result happens to come out correct. The problem only surfaces once you switch agents or modes.

  The extension submission checklist asks for IDs in `lowercase-with-hyphens`, so following that convention appears to make an extension's own commands unreferenceable by token.

  The same condition applies to `speckit.agent-context.update`, from the `agent-context` extension bundled in `core_pack/extensions/`. I could not find any place in `core_pack` that references it through a token, and every token used in `core_pack` refers to a name without hyphens. I have no way of knowing whether that reflects a deliberate workaround or simply a case that has not come up yet, so I am mentioning it rather than assuming.

  ### Some options that came to mind as a user (I don't know the design intent)

  I don't have a deep understanding of speckit's internals, so please read the following as ideas from someone who ran into this rather than as proposals. If any of them were already considered and rejected, I would be genuinely interested in the reasoning.

  - A token form that carries the command id verbatim — something like `__SPECKIT_COMMAND(speckit.agent-context.update)__` — would seem to make the round trip safe regardless of which characters a name uses. Whether that can coexist with the current form, I can't tell.
  - Resolving against the list of installed commands instead of transforming the string might also work, though that depends on what information is available at resolution time, which I'm not in a position to judge.
  - Some check that reports a token which fails to resolve would help. That said, since substitution happens during `integration use`, it would only be discoverable at that point. Checking at install time seems like it would produce false positives, since a referenced extension may not be installed yet.

  If none of these are practical, even a single line in the extension/preset development guide saying that command names intended to be token-referenced should avoid hyphens would have been enough for me. Right now the naming rules point toward hyphens, so there was no way to anticipate the constraint.

### Steps to Reproduce

```python
from specify_cli.integrations.base import IntegrationBase as I

# actual bundled command: speckit.agent-context.update
I.resolve_command_refs("__SPECKIT_COMMAND_AGENT_CONTEXT_UPDATE__", ".")
I.resolve_command_refs("__SPECKIT_COMMAND_AGENT_CONTEXT_UPDATE__", "-")
```

In a real project:

1. Install an extension that provides a command with a hyphen in its name (e.g. the bundled `agent-context`, or an extension providing `speckit.<id>.my-command`).
2. Reference that command from a preset or extension command body as `__SPECKIT_COMMAND_MY_COMMAND__`.
3. Run `specify integration use <agent>` for an integration whose `invoke_separator` is the default `.`.
4. Check the rendered command file.

### Expected Behavior

The token resolves to the command id that is actually installed — or, if it cannot be resolved, something tells me about it instead of silently emitting a wrong string.

### Actual Behavior

The hyphen is interpreted as a segment separator, so a three-segment command renders as four segments. No warning is emitted.

```text
dot separator (default):
  __SPECKIT_COMMAND_AGENT_CONTEXT_UPDATE__ → /speckit.agent.context.update   ← no such command

hyphen separator (skills mode):
  __SPECKIT_COMMAND_AGENT_CONTEXT_UPDATE__ → /speckit-agent-context-update   ← correct by coincidence
```

I checked the `v0.16.4` sources and both places (the substitution logic and the name pattern) are unchanged.

### Specify CLI Version

0.16.0

### AI Agent

Claude Code

### Operating System

macOS 15.3.1

### Python Version

3.13.9 (specify CLI venv)

### Error Logs

```shell

```

### Additional Context

_No response_

HuzaifaChaudary:
## problem

the command ref token can not hold a command name that has a hyphen in it. the pattern is

    __SPECKIT_COMMAND_([A-Z][A-Z0-9_]*)__

the bundled command speckit.agent-context.update has a hyphen in the middle segment so there is no way to write a token for it. the text __SPECKIT_COMMAND_AGENT-CONTEXT_UPDATE__ matches nothing and it stays in the generated file as plain text. the agent then reads a raw token instead of a real command.

## fix

allow a literal hyphen in the character class in the two places that resolve the token

- `src/specify_cli/integrations/base.py` in `resolve_command_refs`
- `src/specify_cli/extensions/__init__.py` in the extension skills resolver

no decode change is needed. the existing replace call already leaves a hyphen untouched. so `AGENT-CONTEXT_UPDATE` becomes `agent-context.update` with the dot separator and `agent-context-update` with the hyphen separator.

this is option 2 from the review on #4204 which is the direction that was asked for there.

## tests

i added 4 tests

- hyphenated name with the dot separator
- hyphenated name with the hyphen separator
- a leading hyphen still does not match because the first character is still `[A-Z]`
- extension skill registration resolves a hyphenated token

all 4 fail before the change and pass after it.

full suite after the change is 7139 passed and 9 failed. the same 9 fail on main without my change so they are not from this. they are the python parity template tests.

`ruff 0.15.0 check src tests` passes which is what the workflow runs.

## ai disclosure

i used an ai assistant to help me read the code and write the tests. i reviewed the change myself and ran the suite locally.

Fixes #4198

### DIFF
--- a/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
+++ b/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
@@ -296,15 +296,16 @@ A command body is a *template* that Spec Kit renders once per agent. Different a
 
 Instead use the agent-neutral token `__SPECKIT_COMMAND_<NAME>__`. Spec Kit resolves it to a `/speckit<separator>...` invocation using the active integration's `invoke_separator` (and integrations may post-process that further in skills output).
 
-Encode the command name in upper case, dropping the `speckit.` prefix and turning each dotted segment separator into an underscore:
+Encode the command name in upper case, dropping the `speckit.` prefix and turning each dotted segment separator into an underscore. A hyphen inside a segment is kept as a hyphen:
 
 | Command file | Token |
 | --- | --- |
 | `speckit.plan.md` | `__SPECKIT_COMMAND_PLAN__` |
 | `speckit.bug.fix.md` | `__SPECKIT_COMMAND_BUG_FIX__` |
 | `speckit.git.commit.md` | `__SPECKIT_COMMAND_GIT_COMMIT__` |
+| `speckit.agent-context.update.md` | `__SPECKIT_COMMAND_AGENT-CONTEXT_UPDATE__` |
 
-The resolver maps each underscore back to the active agent's separator, so use tokens to reference commands whose name segments are single words. (Command names are dotted segments like `git.commit`; the token scheme rebuilds those dots and does not carry hyphens within a segment.)
+The resolver maps each underscore back to the active agent's separator. An underscore separates segments and a hyphen belongs to the segment it sits in, so `AGENT-CONTEXT_UPDATE` is the two segments `agent-context` and `update` rather than three.
 
 **Example** — a command body that points the user at the next step:
 
--- a/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
+++ b/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
@@ -315,15 +315,13 @@ Once the assessment exists, the next step is `__SPECKIT_COMMAND_BUG_FIX__ slug=<
 
 This renders as `/speckit.bug.fix slug=<slug>` for a slash-based agent, `/speckit-bug-fix slug=<slug>` for a skills-based agent, and so on — the author writes it once and it stays portable. The first-party `bug` and `git` extensions use this token exclusively; see `extensions/bug/commands/` for working examples.
 
-> **Current limitation — skills mode.** Token resolution runs in the
-> command-rendering path (`CommandRegistrar`), so it applies when an extension
-> installs *command files*. It does **not** yet run when an extension is
-> registered as *skills* for a skills-based agent: `_register_extension_skills`
-> resolves placeholders and post-processes content but never calls
-> `resolve_command_refs`, so a `__SPECKIT_COMMAND_<NAME>__` token reaches
-> agents such as Codex, ZCode, and Kimi verbatim in that mode. Until that
-> rendering step lands, prefer the token for command-file extensions and avoid
-> relying on it inside skill bodies destined for skills-based agents.
+> **Skills mode.** Token resolution runs in both paths, so the token is safe to
+> use either way. Command files go through the command-rendering path
+> (`CommandRegistrar`). Skill bodies go through `_resolve_command_ref_tokens` in
+> `_register_extension_skills`, which resolves the same token shape against the
+> active skill style: `$speckit-bug-fix` for a dollar-prefixed agent such as
+> Codex or ZCode, `/speckit-bug-fix` for a slash-prefixed one such as Kimi, and
+> the integration's own invocation otherwise.
 
 ### Script Path Rewriting
 
--- a/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
+++ b/extensions/EXTENSION-DEVELOPMENT-GUIDE.md
@@ -319,9 +319,15 @@ This renders as `/speckit.bug.fix slug=<slug>` for a slash-based agent, `/specki
 > use either way. Command files go through the command-rendering path
 > (`CommandRegistrar`). Skill bodies go through `_resolve_command_ref_tokens` in
 > `_register_extension_skills`, which resolves the same token shape against the
-> active skill style: `$speckit-bug-fix` for a dollar-prefixed agent such as
-> Codex or ZCode, `/speckit-bug-fix` for a slash-prefixed one such as Kimi, and
-> the integration's own invocation otherwise.
+> active skill style, so `__SPECKIT_COMMAND_BUG_FIX__` renders as:
+>
+> | Agent | Rendered |
+> | --- | --- |
+> | Codex, ZCode, Command Code | `$speckit-bug-fix` |
+> | Claude, Copilot, Cursor, Devin, Droid, Grok and the other slash agents | `/speckit-bug-fix` |
+> | Kimi | `/skill:speckit-bug-fix` |
+>
+> Anything else falls through to the integration's own `build_command_invocation`.
 
 ### Script Path Rewriting
 
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -1596,7 +1596,7 @@ class ExtensionManager:
                 )
 
             return re.sub(
-                r"__SPECKIT_COMMAND_([A-Z][A-Z0-9_]*)__", _replacement, body
+                r"__SPECKIT_COMMAND_([A-Z][A-Z0-9_-]*)__", _replacement, body
             )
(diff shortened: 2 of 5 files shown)


mnriem:
This is now the canonical implementation for #4198. We selected this PR because it applies the requested option 2 with the smallest backward-compatible code change. Please address the existing focused documentation feedback so the authoring guide reflects the newly supported literal-hyphen syntax; no broader refactor is requested. #4204, #4361, and duplicate issue #4328 are being closed to consolidate the work here.

HuzaifaChaudary:
documentation feedback is addressed in `a40b2bf2`, and nothing else changed.

`extensions/EXTENSION-DEVELOPMENT-GUIDE.md` had the opposite of this behaviour written into it:

> the token scheme rebuilds those dots and does not carry hyphens within a segment

that line is gone. the encoding rule now says a hyphen inside a segment is kept, and the table has a hyphenated row using the real bundled command rather than a made up one:

| Command file | Token |
| --- | --- |
| `speckit.agent-context.update.md` | `__SPECKIT_COMMAND_AGENT-CONTEXT_UPDATE__` |

i used `speckit.agent-context.update` on purpose. it ships in `extensions/agent-context/commands/` and it is the command that was unreachable before this change, so the example is the actual case rather than an illustration.

the sentence after the table now states the rule both ways, since a hyphen and an underscore doing different jobs in the same token is the part that is easy to get wrong:

> An underscore separates segments and a hyphen belongs to the segment it sits in, so `AGENT-CONTEXT_UPDATE` is the two segments `agent-context` and `update` rather than three.

markdownlint clean, 0 issues. no code changed in this commit, the regex is still the one character `[A-Z0-9_-]` widening from the first commit.

mnriem (changes requested):

HuzaifaChaudary:
addressed in `fd8ed36e`. the finding was right, and it is worth being precise about which half of it was.

the callout said this:

> `_register_extension_skills` resolves placeholders and post-processes content but never calls `resolve_command_refs`, so a `__SPECKIT_COMMAND_<NAME>__` token reaches agents such as Codex, ZCode, and Kimi verbatim in that mode

the first half is still literally true. `resolve_command_refs` is never called there. the conclusion drawn from it is not, because `_resolve_command_ref_tokens` in the same function does its own substitution over the same token shape and handles exactly the agents the callout named:

```python
if is_dollar_skills_agent(selected_ai, ai_skills_enabled):
    return "$" + command_name.replace("speckit.", "speckit-").replace(".", "-")
if is_slash_skills_agent(selected_ai, ai_skills_enabled):
    return "/" + command_name.replace("speckit.", "speckit-").replace(".", "-")
```

so a token becomes `$speckit-bug-fix` for codex or zcode and `/speckit-bug-fix` for kimi rather than arriving verbatim.

i ran it rather than take my own reading of it:

```
tests/test_extension_skills.py  8 passed
```

that includes `test_skill_registration_resolves_hyphenated_command_ref_tokens`, which registers an extension as skills for a skills agent and asserts the hyphenated token comes out resolved.

one thing worth flagging since it changes who owns the wording. the callout was already wrong before this branch. the skills side substitution and its `re.sub` were both there, and all this branch did to that line was widen the character class by one character to allow `-`. so this is not a limitation my change lifted, it is one the guide had been describing incorrectly for a while, and i noticed it because copilot pointed at the paragraph next to the one i was editing.

the replacement says what both paths do rather than claiming a limitation either way. markdownlint 0 issues.

HuzaifaChaudary:
the red on `Test & Lint Python` is a cancelled macos runner rather than a test failure, in case it reads as one.

run [33519638199](https://github.com/github/spec-kit/actions/runs/33519638199), 6 of 7 jobs green:

```
success    ruff
success    pytest (ubuntu-latest, 3.13)
success    pytest (ubuntu-latest, 3.14)
success    pytest (windows-latest, 3.13)
success    pytest (windows-latest, 3.14)
success    pytest (macos-latest, 3.13)
cancelled  pytest (macos-latest, 3.14)
```

the cancelled one ran 14:28:48 to 14:43:56 and reports no step conclusions at all, so it never got to a result. macos 3.13 passed and both 3.14 cells on the other two platforms passed, so it is not the python version and not the platform.

the workflow conclusion is failure because a job was cancelled, not because anything asserted false. a re-run should clear it, and i cannot trigger one from here.

`Lint`, `CodeQL` and `Security Audit` are all green on the same commit.

mnriem:

HuzaifaChaudary:
@mnriem addressed in `db5ff0d5`, and the finding was correct.

i had written that kimi renders `/speckit-bug-fix`. it does not. kimi is in `SKILL_COLON_AGENTS` only, so `_resolve_command_ref_tokens` misses both the dollar and the slash branch and falls through to the integration:

```python
DOLLAR_SKILLS_AGENTS   -> kimi present: False
ALWAYS_SLASH_AGENTS    -> kimi present: False
CONDITIONAL_SLASH_AGENTS -> kimi present: False
SKILL_COLON_AGENTS     -> kimi present: True
```

and `KimiIntegration.build_command_invocation` returns `"/skill:speckit-" + stem.replace(".", "-")`. so the rendered form is `/skill:speckit-bug-fix`.

the callout is now a table rather than a sentence, because naming one example per style is what got it wrong in the first place:

| Agent | Rendered |
| --- | --- |
| Codex, ZCode, Command Code | `$speckit-bug-fix` |
| Claude, Copilot, Cursor, Devin, Droid, Grok and the other slash agents | `/speckit-bug-fix` |
| Kimi | `/skill:speckit-bug-fix` |

markdownlint 0 issues. no code changed in this commit, the only production change on the branch is still the one character `[A-Z0-9_-]` widening.

HuzaifaChaudary:
Disclosure, which I should have added when I opened this and did not.

Per CONTRIBUTING.md, "AI contributions in Spec Kit": this contribution was made with AI assistance. Claude Code was used for the code change, the tests, the documentation edits, and for writing the pull request description and my comments on this thread, working from my direction and with the results checked and run locally before each push.

Apologies for the late disclosure. I read the issue and the labels before starting but not the contributing guide, which is my mistake. Happy to answer anything about the change directly.

mnriem (approved):

mnriem:
Thank you!

PR #4367, opened 2026-08-28
add --require-spec to check-prerequisites

### LINKED ISSUE #4364: [Bug]: check-prerequisites.sh never validates spec.md although it resolves and exports FEATURE_SPEC
### Bug Description

`scripts/bash/check-prerequisites.sh` validates `FEATURE_DIR`, `plan.md`, and — under `--require-tasks` — `tasks.md`, each with an actionable error naming the command to run:

```sh
echo "ERROR: plan.md not found in $FEATURE_DIR" >&2
echo "Run /speckit-plan first to create the implementation plan." >&2
```

`FEATURE_SPEC` is resolved, exported in the JSON payload and printed in text mode (5 occurrences in the file), but **its existence is never checked**.

Both `/speckit-analyze` and `/speckit-converge` invoke the script as `check-prerequisites.sh --json --require-tasks --include-tasks` and then read the spec directly — `SPEC = FEATURE_DIR/spec.md`, followed by a "From spec.md" extraction section. When `spec.md` is missing, the prerequisite check passes and the failure surfaces later, without the "Run /speckit-specify first" guidance the script provides for every other artifact.

For accuracy: `/speckit-implement` does not read `spec.md`, so this affects `analyze` and `converge`.

### Steps to Reproduce

1. Create a feature, then generate `plan.md` and `tasks.md`.
2. Delete or rename `specs/<feature>/spec.md`.
3. Run `.specify/scripts/bash/check-prerequisites.sh --json --require-tasks --include-tasks`.

### Expected Behavior

The same treatment as `plan.md` and `tasks.md`: an explicit error pointing at `/speckit-specify`. If `spec.md` is legitimately optional for some callers, gating it behind a `--require-spec` flag would keep the check opt-in while still making it available to `analyze` and `converge`.

### Actual Behavior

Exit 0, with `FEATURE_SPEC` reported as a path that does not exist. `analyze` and `converge` then fail further downstream with a less actionable error, which is precisely what this script exists to prevent.

### Specify CLI Version

1.0.1 — and still present on `main` at the time of writing.

### AI Agent

Claude (not agent-specific: the script is shared by all integrations).

### Operating System

macOS — not OS-specific.

HuzaifaChaudary:
Closes #4364

## problem

`check-prerequisites` validates `FEATURE_DIR`, `plan.md`, and `tasks.md` under `--require-tasks`, each with an error naming the command to run. it resolves `FEATURE_SPEC`, exports it in the json payload and prints it in text mode, but never checks that the file exists.

`analyze` and `converge` both call it and then read `spec.md` directly. with the spec missing the prerequisite check exits 0, and the failure surfaces later without the "run specify first" guidance the script gives for everything else, which is the thing this script exists to prevent.

## fix

a `--require-spec` flag, shaped exactly like `--require-tasks`:

```
ERROR: spec.md not found in specs/001-my-feature
Run /speckit.specify first to create the feature specification.
```

it is opt in, as suggested in the issue, so nothing changes for callers that do not read the spec. `analyze` and `converge` pass it because they do.

## all three variants

the flag is in bash, powershell and python. that is not gold plating, the parity tests compare help text, json output and error shapes across the three, so adding it to one would have broken them.

| file | change |
|---|---|
| `scripts/bash/check-prerequisites.sh` | `--require-spec` |
| `scripts/powershell/check-prerequisites.ps1` | `-RequireSpec` |
| `scripts/python/check_prerequisites.py` | `--require-spec` |
| `templates/commands/analyze.md` | passes the flag in sh ps and py |
| `templates/commands/converge.md` | passes the flag in sh ps and py |

## tests

`test_python_require_spec_matches_bash` covers three states across bash and python:

- spec missing and no flag, both exit 0, so the opt in stays opt in
- spec missing with the flag, both exit 1 with byte identical stderr
- spec present with the flag, both exit 0 with identical json

it fails on `main` with `ERROR: Unknown option '--require-spec'`.

## verification

| check | result |
|---|---|
| new test on `main` | fails |
| new test with the change | passes |
| full suite before | 10 failed, 7135 passed |
| full suite after | 10 failed, 7137 passed |
| new failures introduced | none, i diffed the two failure lists |
| `ruff check scripts tests` | all checks passed |

the 10 pre existing failures are the composed template parity tests and are unrelated to this.

---

disclaimer: this contribution was prepared with the assistance of an ai agent. i reproduced the gap first, checked which callers actually read the spec, and ran the parity suite locally before opening this.

### DIFF
--- a/scripts/bash/check-prerequisites.sh
+++ b/scripts/bash/check-prerequisites.sh
@@ -9,6 +9,7 @@
 #
 # OPTIONS:
 #   --json              Output in JSON format
+#   --require-spec      Require spec.md to exist (for analysis phase)
 #   --require-tasks     Require tasks.md to exist (for implementation phase)
 #   --include-tasks     Include tasks.md in AVAILABLE_DOCS list
 #   --paths-only        Only output path variables (no validation)
@@ -24,6 +25,7 @@ set -e
 
 # Parse command line arguments
 JSON_MODE=false
+REQUIRE_SPEC=false
 REQUIRE_TASKS=false
 INCLUDE_TASKS=false
 PATHS_ONLY=false
@@ -33,6 +35,9 @@ while [[ $# -gt 0 ]]; do
     case "$1" in
         --json)
             JSON_MODE=true
+            ;;
+        --require-spec)
+            REQUIRE_SPEC=true
             ;;
         --require-tasks)
             REQUIRE_TASKS=true
@@ -59,6 +64,7 @@ Consolidated prerequisite checking for Spec-Driven Development workflow.
 
 OPTIONS:
   --json              Output in JSON format
+  --require-spec      Require spec.md to exist (for analysis phase)
   --require-tasks     Require tasks.md to exist (for implementation phase)
   --include-tasks     Include tasks.md in AVAILABLE_DOCS list
   --paths-only        Only output path variables (no prerequisite validation)
@@ -139,6 +145,13 @@ fi
 if [[ ! -f "$IMPL_PLAN" ]]; then
     echo "ERROR: plan.md not found in $FEATURE_DIR" >&2
     echo "Run $(format_speckit_command plan "$REPO_ROOT") first to create the implementation plan." >&2
+    exit 1
+fi
+
+# Check for spec.md if required
+if $REQUIRE_SPEC && [[ ! -f "$FEATURE_SPEC" ]]; then
+    echo "ERROR: spec.md not found in $FEATURE_DIR" >&2
+    echo "Run $(format_speckit_command specify "$REPO_ROOT") first to create the feature specification." >&2
     exit 1
 fi
 
--- a/scripts/powershell/check-prerequisites.ps1
+++ b/scripts/powershell/check-prerequisites.ps1
@@ -10,8 +10,7 @@
 # OPTIONS:
 #   -Json               Output in JSON format
 #   -RequireSpec        Require spec.md to exist (for analysis phase)
-#   -RequireSpec        Require spec.md to exist (for analysis phase)
-  -RequireTasks       Require tasks.md to exist (for implementation phase)
+#   -RequireTasks       Require tasks.md to exist (for implementation phase)
 #   -IncludeTasks       Include tasks.md in AVAILABLE_DOCS list
 #   -PathsOnly          Only output path variables (no validation)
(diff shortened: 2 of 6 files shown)


HuzaifaChaudary:
both findings were right and both are fixed in `3c608ed3`.

**the parse break.** exactly as described. the `-RequireSpec` line got duplicated over the `-RequireTasks` line and the second copy lost its leading hash, so line 14 was bare powershell sitting above `[CmdletBinding()]`. the parser is unambiguous about it:

```
# committed 09bf3f15
PARSE ERRORS: 2
  line 20: Unexpected attribute 'CmdletBinding'.
  line 21: Unexpected token 'param' in expression or statement.

# after 3c608ed3
PARSES CLEAN
```

that is why every `test_ps_*` case went red on ubuntu 3.13, not just the ones near this change. the header block now differs from `main` by exactly one added comment line and `-RequireTasks` is back where it was.

**the missing powershell coverage.** added `test_powershell_require_spec_matches_python`, behind the existing `HAS_PWSH or _WINDOWS_POWERSHELL` guard, covering the same three states the bash and python test covers: no flag with a missing spec passes, the flag with a missing spec exits 1 and says `spec.md not found`, and the flag with the spec present exits 0 with matching json.

worth saying plainly that this test is what would have caught the break, and i did not write it the first time. the bash and python halves passed the whole way through.

local run after the fix, with pwsh 7.5.2:

```
2 failed, 52 passed, 1 skipped
```

the two are `test_all_variants_resolve_requested_template[composed]` and `test_all_variants_validate_requested_template_in_text_mode[composed]`. those are not mine. i ran them on clean `main` with no changes to check before saying so:

```
main, no changes:  2 failed, 2 passed
```

same two, same way. they were also in the ubuntu 3.13 failure list, so on ci they may well have been collateral from the parse break, but either way they fail without this branch.

mnriem:
Thank you!

HuzaifaChaudary:
Disclosure, which I should have added when I opened this and did not.

Per CONTRIBUTING.md, "AI contributions in Spec Kit": this contribution was made with AI assistance. Claude Code was used for the code change, the tests, the documentation edits, and for writing the pull request description, working from my direction and with the results checked and run locally before each push.

Apologies for the late disclosure. I read the issue and the labels before starting but not the contributing guide, which is my mistake. Happy to answer anything about the change directly.

PR #4372, opened 2026-08-31
fix(github-http): reject malformed explicit URL ports

WOLIKIMCHENG:
## Summary

- Validate explicit HTTP(S) ports before GitHub token lookup or urllib request construction.
- Preserve valid request URLs, methods, and exact-host bearer authentication behavior.
- Add regression coverage for malformed and valid explicit ports.

## Validation

- `.venv/bin/python -m pytest tests/test_github_http.py tests/test_authentication.py -q`
- `git diff --check`


mnriem:
Thank you!

PR #4390, opened 2026-09-01
feat(evaluator): add evaluator contract extension with model routing

tbitcs:
## Summary

Implements the standard evaluator result contract proposed in #4290. This is a provider-neutral protocol for extensions that evaluate artifact quality between Spec-Driven Development phases.

### Extension (`extensions/evaluator/`)
- **JSON Schema** for evaluator results (6 outcomes, 14 finding kinds, 5 evidence kinds, 6 uncertainty levels)
- **4 commands**: `speckit.evaluator.run`, `.compose`, `.report`, `.route`
- **3 scripts**: Python, Bash, PowerShell (parity across all runtimes)
- **4 lifecycle hooks**: `after_specify`, `after_plan`, `after_tasks`, `after_implement`
- **Model routing**: recommends budget/standard/premium tier per phase based on evaluator findings
- **Composition**: strict/majority/optimistic strategies with deterministic contradiction detection
- **Quick-start demo**: `python extensions/evaluator/examples/demo.py` — self-contained, zero deps

### Tests (`tests/extensions/evaluator/`)
- **70 tests**: layout, catalog, install, compose logic, benchmarks, model routing
- **0 regressions** against full test suite (4,296 passed, 114 skipped)

### Benchmarks (`benchmarks/evaluator/`)
- SDD workflow simulation with 8 evaluators across 4 phases
- Composition at scale: up to 20 evaluators × 100 findings (2,000 total)
- Report generation in all 5 formats (terminal, markdown, JSON, CI annotations, gate)
- Token-economic simulation with Monte Carlo (500 runs/scenario)
- Portfolio approach: budget for routine generation, premium for critical decisions

### Catalog
Registered as bundled extension in `extensions/catalog.json`.

## Description

This PR adds the **evaluator contract extension** — a standardized protocol for extensions that evaluate artifact quality between SDD phases. Currently, every evidence, provenance, factuality, security, or governance extension invents its own result format. This contract provides a shared schema and composition semantics so independently developed evaluators can interoperate.

The contract defines:
- A versioned JSON Schema for evaluator results
- 6 outcomes: `pass`, `warn`, `iterate`, `clarify`, `gather_evidence`, `block`
- 14 finding kinds
- 5 evidence kinds: `observed`, `inferred`, `asserted`, `contradicted`, `unsupported`
- Deterministic composition with contradiction detection
- Model routing recommendations for the portfolio approach

## Key Design Decisions

[...]

Issue #4290, opened 2026-08-23
[Feature]: Standard evaluator contract for evidence, provenance, uncertainty, and recovery


tbitcs:
### Problem Statement

Spec Kit already has a strong lifecycle and an extensible hook system, but there is no common contract for extensions that evaluate the quality or trustworthiness of artifacts between phases.

That means every evidence, provenance, factuality, security, or governance extension has to invent its own answers to questions such as:

- What did the evaluator find?
- Is the result a pass, warning, iteration request, or hard block?
- What source or artifact supports the finding?
- Is a statement observed, inferred, contradicted, or simply unsupported?
- What should the workflow do next?
- How should evaluator state survive pause/resume or context reduction?
- How do multiple independent evaluators compose?

This is closely related to #23, where the discussion moved from a single provenance integration toward a generic registerable “linter” / “evaluator” concept that could run after workflow phases and trigger iteration.

I think the missing piece is not another workflow engine. It is a small, provider-neutral **evaluator result contract** that existing hooks and extensions can use.

### Proposed Solution

Define a standard evaluator protocol that extensions can implement while continuing to use Spec Kit's existing lifecycle and hook execution model.

The smallest useful contract would let an evaluator:

1. register for one or more lifecycle events;
2. receive the relevant resolved artifacts plus stable source/provenance references;
3. return a versioned machine-readable result;
4. distinguish observed evidence from generated assertions;
5. preserve contradictory findings instead of forcing false consensus;
6. represent insufficient evidence or unresolved uncertainty explicitly;
7. request a bounded next action such as `pass`, `warn`, `iterate`, `clarify`, `gather_evidence`, or `block`;
8. persist enough compact state to survive pause/resume;
9. compose deterministically with other evaluators;
10. remain implementation-neutral: deterministic, model-backed, local, remote, private, paid, or hybrid.

The current hook system can remain the execution mechanism. I am mainly proposing a shared evaluator **result schema and lifecycle semantics**, not a second orchestration layer.

### Illustrative result

```json
{
  "schema_version": "1.0",
  "evaluator": {
    "id": "epistemic",
    "version": "0.1.0"
  },
  "phase": "after_plan",
  "outcome": "iterate",
  "summary": "Two high-impact claims are unsupported and one requirement conflicts with an accepted constraint.",
  "findings": [
    {
      "id": "EPI-001",
      "severity": "high",
      "kind": "unsupported_claim",
      "subject": "REQ-014",
      "evidence_refs": [],
[...]

PR #4393, opened 2026-09-01
feat(workflows): propagate step output model to context for evaluator-driven routing

tbitcs:
## Summary

Enables automatic model routing in the workflow engine. When a step's output includes a `model` field, the engine propagates it to `context.default_model` so subsequent steps automatically use the routed model.

This is the follow-up to #4390 (evaluator contract extension). Together they enable the full portfolio approach: evaluators recommend a model tier, and the engine applies it automatically.

## Changes

- [`engine.py`](src/specify_cli/workflows/engine.py): 7 lines — after each step, check `result.output.get("model")` and propagate string values to   `context.default_model`
- [`test_workflows.py`](tests/test_workflows.py): 5 new tests covering propagation, no-op when absent, non-string rejection, subsequent step inheritance, and explicit override priority

## Design

- **Only strings propagate.** Non-string values (None, lists, dicts) are silently ignored — they're already rejected by step validation.
- **Explicit overrides win.** A step with `model: "sonnet-4"` always uses that model, regardless of what was routed.
- **Backward compatible.** If no step outputs a model, behavior is unchanged — `context.default_model` stays at its initial value.

## Testing

```
$ uv run pytest tests/test_workflows.py::TestModelRoutingPropagation -v
5 passed in 5.73s
```

## AI Disclosure

Generated by GitHub Copilot (model: deepseek-v4-pro, autonomous).

Refs: #4290
Depends on: #4390

### DIFF
--- a/src/specify_cli/workflows/engine.py
+++ b/src/specify_cli/workflows/engine.py
@@ -1232,6 +1232,14 @@ class WorkflowEngine:
             }
             self._record_result(context, state, step_id, step_data)
 
+            # Propagate model routing: if a step output specifies a model,
+            # use it as the default for subsequent steps. This enables
+            # evaluator-driven model routing (e.g., evaluator recommends
+            # "premium" tier → next phase uses premium model).
+            routed_model = result.output.get("model")
+            if routed_model and isinstance(routed_model, str):
+                context.default_model = routed_model
+
             state.append_log(
                 {
                     "event": "step_completed",
--- a/tests/test_workflows.py
+++ b/tests/test_workflows.py
@@ -17204,3 +17204,103 @@ steps:
         result = runner.invoke(app, ["workflow", "disable", "align-wf"])
         assert result.exit_code == 0
         assert "already disabled" in result.output
+
+
+class TestModelRoutingPropagation:
+    """Test that step output model propagates to context.default_model.
+
+    When a step's output includes a ``model`` field (e.g., from an evaluator
+    recommending a tier), the engine propagates it to ``context.default_model``
+    so subsequent steps use the routed model automatically.
+    """
+
+    def test_step_output_model_propagates_to_context(self):
+        """A step that outputs a model updates context.default_model."""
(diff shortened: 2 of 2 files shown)


mnriem:
Thanks for putting this together and for exploring evaluator-driven model selection. I agree that workflows should be able to use an evaluator's recommendation, but I don't think the engine should apply that recommendation automatically.

Step outputs are already recorded in the workflow context and can be referenced by later steps. A workflow author can therefore make the routing decision explicit:

```yaml
- id: evaluate
  type: command
  command: speckit.evaluator.route

- id: implement
  type: command
  command: speckit.implement
  model: "{{ steps.evaluate.output.model_routing.recommended_tier }}"
```

Keeping this connection in the workflow definition has a few benefits:

- The dataflow is visible and intentional.
- Producing a field named `model` does not unexpectedly change subsequent steps.
- Workflow authors can validate, transform, or ignore the recommendation.
- Each downstream step can independently decide whether to use the output.
- The engine remains neutral about the meaning of extension-defined outputs.

In particular, `output.model` may describe the model that executed a step rather than a recommendation for future steps. Treating it as both metadata and a command to mutate `context.default_model` would introduce implicit behavior that may be difficult to understand from the workflow YAML.

For those reasons, I'd prefer that outputs remain data and that applying them to later inputs remain a conscious decision by the workflow author. The existing expression mechanism already supports that, so I don't think this change is needed in the core workflow engine.

tbitcs:
Thanks for the review — you're right that the existing expression mechanism already handles this cleanly. Closing in favor of the explicit workflow wiring pattern documented in the evaluator extension's route command.

Workflow authors can wire model routing explicitly:

```yaml
- id: evaluate
  type: command
  command: speckit.evaluator.route

- id: implement
  type: command
  command: speckit.implement
  model: "{{ steps.evaluate.output.model_routing.recommended_tier }}"
```

This keeps the dataflow visible and intentional without any engine changes.

PR #4396, opened 2026-09-01
fix(scripts): stop wrap composition looping on a token in core content

### LINKED ISSUE #4385: [Bug]: wrap template composition can loop forever when core content contains {CORE_TEMPLATE}
## Bug Description

The Bash template resolver's `wrap` strategy can loop indefinitely when the resolved base content contains the literal `{CORE_TEMPLATE}` token.

The current loop replaces one token in `layer_content` with `content`, then scans the modified `layer_content` again. If `content` contains `{CORE_TEMPLATE}`, every iteration reintroduces a token and the resolver never terminates.

This is present on current `main` at commit `31e6b84f701b108a848847a8bc9ea86a1e2890b7` in `scripts/bash/common.sh`.

## Steps to Reproduce

1. Initialize a project with Bash scripts.
2. Add this preset registry:

   `.specify/presets/.registry`

   ```json
   {
     "presets": {
       "a-wrap": { "enabled": true, "priority": 1 },
       "b-base": { "enabled": true, "priority": 2 }
     }
   }
   ```

3. Add a wrapping preset:

   `.specify/presets/a-wrap/preset.yml`

   ```yaml
   provides:
     templates:
       - type: template
         name: review-template
         file: wrapper.md
         strategy: wrap
   ```

   `.specify/presets/a-wrap/wrapper.md`

   ```text
   BEFORE
   {CORE_TEMPLATE}
   AFTER
   ```

4. Add a base preset whose content intentionally contains the literal token:

   `.specify/presets/b-base/preset.yml`

   ```yaml
   provides:
     templates:
       - type: template
         name: review-template
         file: base.md
         strategy: replace
   ```

   `.specify/presets/b-base/base.md`

   ```text
   BASE {CORE_TEMPLATE}
   ```

5. Run:

   ```bash
   timeout 5 .specify/scripts/bash/resolve-template.sh review-template
   echo $?
   ```

## Expected Behavior

The resolver replaces each `{CORE_TEMPLATE}` occurrence that existed in the original wrapper exactly once, preserves literal token text contained in the inserted base content, and terminates. For the example above, the output should contain:

```text
BEFORE
BASE {CORE_TEMPLATE}
AFTER
```

## Actual Behavior

The process consumes CPU until `timeout` terminates it with exit code 124. Each iteration replaces a token with base content that contains another token, so the loop condition remains true forever.

## Specify CLI Version

`1.0.3.dev0`

## AI Agent

Codex CLI

## Operating System

Linux 7.2.0 x86_64

## Python Version

Python 3.14.5 (Specify CLI runtime)

## Error Logs

```shell
$ timeout 5 .specify/scripts/bash/resolve-template.sh review-template
$ echo $?
124
```

## Additional Context

A single-pass replacement over an unmodified `rest` variable fixes the issue: consume tokens only from the original wrapper while appending `content` to a separate output buffer.

The issue was found while reviewing generated Spec Kit tooling in [mforce/cluckwork#635](https://github.com/mforce/cluckwork/pull/635). A verified local fix is in [commit c70a5ad1](https://github.com/mforce/cluckwork/commit/c70a5ad1a098e44a5bc44e191cc61c36af6a4e27).

I searched open and closed issues and PRs using `{CORE_TEMPLATE}`, wrap strategy, infinite loop, hang, and placeholder-reintroduction terms. Related issues #3991 and #4044 concern frontmatter corruption during wrap composition; they do not cover this non-terminating Bash replacement loop.

Yash-Chindam:
Fixes #4385

## Problem

`scripts/bash/common.sh` rewrote `layer_content` in place and then re-tested the string it had just modified:

```bash
while [[ "$layer_content" == *'{CORE_TEMPLATE}'* ]]; do
    local before="${layer_content%%\{CORE_TEMPLATE\}*}"
    local after="${layer_content#*\{CORE_TEMPLATE\}}"
    layer_content="${before}${content}${after}"
done
```

When the resolved core content holds a literal `{CORE_TEMPLATE}`, every pass reinserts it and the condition is true forever.

## Why this shape of fix

The PowerShell and Python ports were already correct, so the intended semantics did not need inventing — only matching:

```powershell
# scripts/powershell/common.ps1:789
$content = $layerContent.Replace('{CORE_TEMPLATE}', $content)
```

```python
# scripts/python/common.py:466
content = layer_content.replace(placeholder, content)
```

`String.Replace` and `str.replace` scan the original subject once and never revisit inserted text. The new loop consumes the wrapper left to right into an accumulator, so work is bounded by the placeholders in the original wrapper and inserted content is never re-examined. This aligns bash with the other two ports rather than introducing a third behaviour.

`tests/parity_helpers.py` already treats `.replace()` as the oracle for expected wrap output (`install_composition_stack`), so the three ports now agree with the harness by construction.

## Verification

I extracted the wrap block verbatim from `upstream/main` and from this branch and ran both under Git Bash:

| case | wrapper | core content | before | after |
|---|---|---|---|---|
| A — normal | `BEFORE\n{CORE_TEMPLATE}\nAFTER` | `BASE` | `BEFORE\nBASE\nAFTER` | identical |
| B — this bug | `BEFORE\n{CORE_TEMPLATE}\nAFTER` | `BASE {CORE_TEMPLATE}` | **hangs (rc=124)** | `BEFORE\nBASE {CORE_TEMPLATE}\nAFTER` |
| C — two tokens | `A{CORE_TEMPLATE}B{CORE_TEMPLATE}C` | `X` | `AXBXC` | identical |

A and C confirm no behaviour change on the paths that already worked, including multiple placeholders in the wrapper — the fix is not "replace the first occurrence". B matches `str.replace` byte for byte.

## Test

`test_all_variants_treat_core_token_in_core_content_as_literal` runs all three variants through the existing parity harness with core content carrying a literal token.

The regression mode here is a **hang, not a wrong value**, so a plain assertion would never fail — it would just never finish. `run()` therefore grows an optional `timeout` parameter (default `None`, so every existing call is unchanged) and the new test passes `timeout=30`. Without it, a reintroduced bug would stall the suite instead of failing it.

## Notes on local test runs

The bash parity tests skip on my machine: Windows resolves `bash` to the WSL launcher, which `tests/conftest.py::_has_working_bash` correctly rejects. I verified the bash behaviour by driving Git Bash directly, as in the table above; CI will exercise the new test through the normal harness.

`pytest tests/` shows 20 pre-existing failures locally, all `WinError 1314: A required privilege is not held by the client` from symlink creation. I confirmed the identical set fails on unmodified `upstream/main`, so they are environmental and unrelated.

---

*Disclosure: this change was developed with AI assistance (Claude). The AI helped locate the defect, compare the three ports, draft the fix and test, and write this description. All reasoning and results above were verified by running the code; I reviewed the change before submitting.*

### DIFF
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -902,12 +902,20 @@ except Exception as exc:
                     *'{CORE_TEMPLATE}'*) ;;
                     *) echo "Error: wrap strategy missing {CORE_TEMPLATE} placeholder" >&2; return 2 ;;
                 esac
-                while [[ "$layer_content" == *'{CORE_TEMPLATE}'* ]]; do
-                    local before="${layer_content%%\{CORE_TEMPLATE\}*}"
-                    local after="${layer_content#*\{CORE_TEMPLATE\}}"
-                    layer_content="${before}${content}${after}"
+                # Consume the wrapper left to right instead of rewriting it in
+                # place. Rewriting re-scanned the string just modified, so base
+                # content holding a literal {CORE_TEMPLATE} reintroduced the
+                # token every pass and the loop never terminated. Advancing over
+                # ``rest`` bounds the work by the tokens in the original wrapper
+                # and leaves inserted content untouched, matching the single-pass
+                # semantics of .Replace()/.replace() in the PowerShell and Python
+                # ports.
+                local wrapped="" rest="$layer_content"
+                while [[ "$rest" == *'{CORE_TEMPLATE}'* ]]; do
+                    wrapped="${wrapped}${rest%%\{CORE_TEMPLATE\}*}${content}"
+                    rest="${rest#*\{CORE_TEMPLATE\}}"
                 done
-                content="$layer_content"
+                content="${wrapped}${rest}"
                 ;;
             *) echo "Error: unknown strategy '$strat'" >&2; return 2 ;;
         esac
--- a/tests/parity_helpers.py
+++ b/tests/parity_helpers.py
@@ -83,8 +83,17 @@ def clean_env() -> dict[str, str]:
 
 
 def run(
-    cmd: list[str], repo: Path, env: dict[str, str] | None = None
+    cmd: list[str],
+    repo: Path,
+    env: dict[str, str] | None = None,
+    timeout: float | None = None,
 ) -> subprocess.CompletedProcess[str]:
+    """Run a script variant.
+
+    ``timeout`` guards cases whose regression mode is a hang rather than a bad
+    value; without it such a failure would stall the suite instead of failing
+    it. ``subprocess.TimeoutExpired`` propagates so the test reports the hang.
+    """
     return subprocess.run(
         cmd,
         cwd=repo,
@@ -92,6 +101,7 @@ def run(
         text=True,
         check=False,
         env=env if env is not None else clean_env(),
+        timeout=timeout,
     )
 
 
--- a/tests/test_resolve_template_python_parity.py
+++ b/tests/test_resolve_template_python_parity.py
@@ -92,6 +92,40 @@ def test_all_variants_preserve_composition_parity(
 
 
 @requires_bash
+def test_all_variants_treat_core_token_in_core_content_as_literal(
+    tmp_path: Path,
+) -> None:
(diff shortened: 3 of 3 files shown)


mnriem:
Thank you!

PR #4397, opened 2026-09-01
fix(scripts): name setup-plan's feature directory key FEATURE_DIR

### LINKED ISSUE #4017: Bug: SPECS_DIR key in setup-plan.sh JSON output contains FEATURE_DIR value
## Summary

The JSON output of `scripts/bash/setup-plan.sh` exposes a key named `SPECS_DIR` but populates it with the `$FEATURE_DIR` value (the per-feature subdirectory, e.g. `specs/001-my-feature`) rather than the actual specs root directory (`specs/`).

## Affected file

`scripts/bash/setup-plan.sh`

## Bug locations

**Line ~71** (jq path):
```bash
# Current (wrong):
--arg specs_dir "$FEATURE_DIR"

# Should be:
--arg feature_dir "$FEATURE_DIR"
```

**Line ~73** (jq JSON key):
```bash
# Current (wrong):
'{FEATURE_SPEC:$feature_spec,IMPL_PLAN:$impl_plan,SPECS_DIR:$specs_dir,BRANCH:$branch}'

# Should be:
'{FEATURE_SPEC:$feature_spec,IMPL_PLAN:$impl_plan,FEATURE_DIR:$feature_dir,BRANCH:$branch}'
```

**Line ~75** (printf fallback):
```bash
# Current (wrong):
printf '{"FEATURE_SPEC":"%s","IMPL_PLAN":"%s","SPECS_DIR":"%s","BRANCH":"%s"}\n'

# Should be:
printf '{"FEATURE_SPEC":"%s","IMPL_PLAN":"%s","FEATURE_DIR":"%s","BRANCH":"%s"}\n'
```

**Line ~81** (text-mode output):
```bash
# Current (wrong):
echo "SPECS_DIR: $FEATURE_DIR"

# Should be:
echo "FEATURE_DIR: $FEATURE_DIR"
```

## Impact

Consumers of the JSON output who treat `SPECS_DIR` as the specs root directory will receive the wrong path instead of the feature subdirectory.

## Fix

Rename the key from `SPECS_DIR` to `FEATURE_DIR` in all 4 locations listed above.

Yash-Chindam:
Fixes #4017

## Problem

`setup-plan` emitted a JSON key named `SPECS_DIR` holding `$FEATURE_DIR` — the per-feature subdirectory, not the specs root.

The name is already in use elsewhere with the other meaning. In `scripts/bash/create-new-feature.sh`:

```bash
SPECS_DIR="$REPO_ROOT/specs"           # the specs root
FEATURE_DIR="$SPECS_DIR/$BRANCH_NAME"  # one feature inside it
```

So `SPECS_DIR` meant `specs/` in one script and `specs/001-my-feature/` in another.

## Why `FEATURE_DIR` is the right name

`setup-plan` was the only script in the suite using `SPECS_DIR`. Every sibling already emits `FEATURE_DIR` for exactly this value:

```
scripts/bash/setup-tasks.sh:75           FEATURE_DIR:$feature_dir
scripts/bash/check-prerequisites.sh:122  FEATURE_DIR:$feature_dir
scripts/bash/check-prerequisites.sh:208  FEATURE_DIR:$feature_dir
```

This brings `setup-plan` in line rather than inventing a convention.

## Scope

The issue lists four bash locations. The same key is emitted by all three ports, and there is a consumer that parses it by name — fixing only bash would leave the ports disagreeing, which is worse than the current state. Full surface:

```
scripts/bash/setup-plan.sh         73, 75, 77, 83
scripts/powershell/setup-plan.ps1  79, 86
scripts/python/setup_plan.py       85, 93
templates/commands/plan.md         62   <- parses the key by name
```

`plan.md` ships alongside the scripts, so it stays in sync automatically.

## On the compatibility question

I raised rename-vs-deprecated-alias on the issue. Going with the straight rename here, since `plan.md` is the only in-repo consumer and the value was never the specs root — anything relying on `SPECS_DIR` to mean the specs root was already reading the wrong path. Happy to switch to emitting `FEATURE_DIR` while keeping `SPECS_DIR` as a deprecated alias for a release or two if you would rather not break external parsers.

## Verification

Ran all three variants against a scratch project and compared the emitted keys:

```
bash    rc=0 keys=['BRANCH', 'FEATURE_DIR', 'FEATURE_SPEC', 'IMPL_PLAN']
python  rc=0 keys=['BRANCH', 'FEATURE_DIR', 'FEATURE_SPEC', 'IMPL_PLAN']
pwsh    rc=0 keys=['BRANCH', 'FEATURE_DIR', 'FEATURE_SPEC', 'IMPL_PLAN']
```

Also confirmed no other reference to the old key remains outside `create-new-feature.sh`, where `SPECS_DIR` correctly means the specs root and is untouched.

`pytest tests/ -k "plan or command_template or preset"` — 852 passed. The 8 failures are all `WinError 1314: A required privilege is not held by the client` from symlink creation; the identical set fails on unmodified `upstream/main`, so they are environmental and unrelated.

---

*Disclosure: this change was developed with AI assistance (Claude). The AI helped map the affected call sites across the three ports, apply the rename, and draft this description. The three-port key comparison above was produced by running the scripts; I reviewed the change before submitting.*

### DIFF
--- a/scripts/bash/setup-plan.sh
+++ b/scripts/bash/setup-plan.sh
@@ -70,16 +70,16 @@ if $JSON_MODE; then
         jq -cn \
             --arg feature_spec "$FEATURE_SPEC" \
             --arg impl_plan "$IMPL_PLAN" \
-            --arg specs_dir "$FEATURE_DIR" \
+            --arg feature_dir "$FEATURE_DIR" \
             --arg branch "$CURRENT_BRANCH" \
-            '{FEATURE_SPEC:$feature_spec,IMPL_PLAN:$impl_plan,SPECS_DIR:$specs_dir,BRANCH:$branch}'
+            '{FEATURE_SPEC:$feature_spec,IMPL_PLAN:$impl_plan,FEATURE_DIR:$feature_dir,BRANCH:$branch}'
     else
-        printf '{"FEATURE_SPEC":"%s","IMPL_PLAN":"%s","SPECS_DIR":"%s","BRANCH":"%s"}\n' \
+        printf '{"FEATURE_SPEC":"%s","IMPL_PLAN":"%s","FEATURE_DIR":"%s","BRANCH":"%s"}\n' \
             "$(json_escape "$FEATURE_SPEC")" "$(json_escape "$IMPL_PLAN")" "$(json_escape "$FEATURE_DIR")" "$(json_escape "$CURRENT_BRANCH")"
     fi
 else
     echo "FEATURE_SPEC: $FEATURE_SPEC"
     echo "IMPL_PLAN: $IMPL_PLAN"
-    echo "SPECS_DIR: $FEATURE_DIR"
+    echo "FEATURE_DIR: $FEATURE_DIR"
     echo "BRANCH: $CURRENT_BRANCH"
 fi
--- a/scripts/powershell/setup-plan.ps1
+++ b/scripts/powershell/setup-plan.ps1
@@ -76,13 +76,13 @@ if ($Json) {
     $result = [PSCustomObject]@{
         FEATURE_SPEC = $paths.FEATURE_SPEC
         IMPL_PLAN = $paths.IMPL_PLAN
-        SPECS_DIR = $paths.FEATURE_DIR
+        FEATURE_DIR = $paths.FEATURE_DIR
         BRANCH = $paths.CURRENT_BRANCH
     }
     $result | ConvertTo-Json -Compress
 } else {
     Write-Output "FEATURE_SPEC: $($paths.FEATURE_SPEC)"
     Write-Output "IMPL_PLAN: $($paths.IMPL_PLAN)"
-    Write-Output "SPECS_DIR: $($paths.FEATURE_DIR)"
+    Write-Output "FEATURE_DIR: $($paths.FEATURE_DIR)"
     Write-Output "BRANCH: $($paths.CURRENT_BRANCH)"
 }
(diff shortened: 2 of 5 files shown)


mnriem (changes requested):

Yash-Chindam:
Addressed in the latest push — good catch, and the criticism was accurate: every existing `setup-plan` test compares the ports against each other, so all three could have regressed to `SPECS_DIR` together and still gone green.

`test_all_variants_emit_feature_dir_not_specs_dir` now pins the contract absolutely, parametrised over JSON and text mode and covering bash, Python and PowerShell:

- `SPECS_DIR` absent from stdout
- JSON keys exactly `['BRANCH', 'FEATURE_DIR', 'FEATURE_SPEC', 'IMPL_PLAN']`
- text mode carries a `FEATURE_DIR:` line
- the value is the feature directory, not the specs root

One wrinkle worth flagging: the value is matched by **suffix** rather than by full path. The ports legitimately disagree on path flavour — under MSYS, bash reports `/tmp/.../specs/001-my-feature` where the Python and PowerShell ports report `C:\...\specs\001-my-feature`. An equality assertion would have passed locally on Linux and failed the Windows matrix. The suffix comparison still separates `specs/001-my-feature` from a bare `specs`, which is the regression actually being guarded; I checked it rejects both `/tmp/proj/specs` and `C:\proj\specs`.

Verified all six port × mode combinations produce the expected key and value.

Also rebased onto current `main`.

*Disclosure: AI assistance (Claude) was used for this change and comment, as with the original PR.*

Yash-Chindam:
Rebased onto current `main` (was 10 behind). The Copilot feedback is addressed in the commit above — `test_all_variants_emit_feature_dir_not_specs_dir` pins the output contract in JSON and text mode across bash, Python and PowerShell, rather than only comparing the ports against each other.

Workflow runs on this branch are sitting at `action_required`, so CI needs a maintainer to approve them before it can report. Nothing needed from me until then, but flagging it since the checks look absent rather than pending.

mnriem (approved):

mnriem:
Thank you!

PR #4398, opened 2026-09-01
fix(workflows): drop the bundled speckit workflow's inert scope input

### LINKED ISSUE #4384: [Bug]: bundled speckit workflow exposes a scope input that has no effect
## Bug Description

The bundled `speckit` workflow declares a required-choice `scope` input with `full`, `backend-only`, and `frontend-only` values, but no workflow step references `{{ inputs.scope }}`. All four command steps receive only `{{ inputs.spec }}`, so every scope choice has identical behavior.

This is present on current `main` at commit `31e6b84f701b108a848847a8bc9ea86a1e2890b7` in `workflows/speckit/workflow.yml`.

## Steps to Reproduce

1. Initialize a project with Spec Kit `1.0.3.dev0` so the bundled workflow is installed.
2. Inspect `.specify/workflows/speckit/workflow.yml`.
3. Observe the declared input:

   ```yaml
   scope:
     type: string
     default: "full"
     enum: ["full", "backend-only", "frontend-only"]
   ```

4. Search its consumers:

   ```bash
   rg -n 'inputs\.scope|inputs\.spec' .specify/workflows/speckit/workflow.yml
   ```

5. Optionally run the same specification with each accepted value:

   ```bash
   specify workflow run speckit -i spec="Add an example feature" -i scope=full
   specify workflow run speckit -i spec="Add an example feature" -i scope=backend-only
   specify workflow run speckit -i spec="Add an example feature" -i scope=frontend-only
   ```

## Expected Behavior

Each exposed input affects workflow execution. Either:

- the selected scope is propagated using a documented command/workflow contract so the three choices behave differently; or
- the unsupported selector is removed from the bundled workflow.

## Actual Behavior

`scope` appears only in its declaration. The `specify`, `plan`, `tasks`, and `implement` steps all receive the same argument:

```yaml
input:
  args: "{{ inputs.spec }}"
```

Users are therefore prompted to choose among three values that do not affect the run.

## Specify CLI Version

`1.0.3.dev0`

## AI Agent

Codex CLI

## Operating System

Linux 7.2.0 x86_64

## Python Version

Python 3.14.5 (Specify CLI runtime)

## Error Logs

```shell
$ rg -n 'inputs\.scope|inputs\.spec' .specify/workflows/speckit/workflow.yml
48:      args: "{{ inputs.spec }}"
60:      args: "{{ inputs.spec }}"
72:      args: "{{ inputs.spec }}"
78:      args: "{{ inputs.spec }}"
```

There is no `inputs.scope` consumer.

## Additional Context

The issue was found while reviewing generated Spec Kit tooling in [mforce/cluckwork#635](https://github.com/mforce/cluckwork/pull/635). The local correction removes the unsupported selector in [commit c70a5ad1](https://github.com/mforce/cluckwork/commit/c70a5ad1a098e44a5bc44e191cc61c36af6a4e27) rather than inventing downstream argument semantics.

I searched open and closed issues and PRs using `scope`, `backend-only`, `frontend-only`, bundled workflow, and unused-input terms. Issue #2142 is the closed workflow-engine proposal and includes intended conditional scope routing as an example; it does not report that the shipped bundled workflow leaves the selector unused.

Yash-Chindam:
Fixes #4384

## Problem

The bundled `workflows/speckit/workflow.yml` declared a required-choice `scope` input that no step referenced. All four steps pass `{{ inputs.spec }}`, so every choice behaved identically.

## What actually happens

Two corrections to the report's framing, both checked against `main`:

**Workflow inputs are never prompted for.** The report says users "are prompted to choose among three values", but nothing in the engine reads an input's `prompt:` field to ask anything — `_resolve_inputs` only merges provided values with defaults, and the sole interactive `input()` call in `workflows/` belongs to the gate step. The `prompt:` key is documentation-only metadata today, which is also why `scope` being the one input without one went unnoticed.

**The value is validated, then discarded.** Calling `_coerce_input` with the declared definition:

```
-i scope=full           -> ACCEPTED as 'full'
-i scope=backend-only   -> ACCEPTED as 'backend-only'
-i scope=frontend-only  -> ACCEPTED as 'frontend-only'
-i scope=bogus          -> REJECTED: Input 'scope' value 'bogus' not in allowed values: [...]
```

That is worse than being ignored outright: a bogus value errors, so the acceptance of `backend-only` reads to the user as confirmation the flag does something.

## Why removal rather than propagation

The issue offered both. Propagating scope has no contract to hang it on — the four steps take a single `args` string, so a scope could only be smuggled in by appending prose to `{{ inputs.spec }}`, putting a scoping instruction into the same free-text field the user's own description occupies and leaving its interpretation to the agent. That is a feature with a design question attached, not a bug fix.

Removing the selector restores the invariant that every declared input affects the run, and leaves a real scope feature free to land later on its own terms.

## Also updated

`workflows/README.md` demonstrated "Multiple Inputs" using this exact input against the bundled workflow:

```bash
specify workflow run speckit \
  --input spec="..." \
  --input scope="backend-only"
```

That documented the inert input as if it worked, so the example now uses a declared one (`integration`). The `inputs.scope` reference further down in the If/Then/Else section is a generic authoring illustration with its own hypothetical workflow, not the bundled one, so it stays.

## Verification

```
declared  : ['integration', 'spec']
referenced: ['integration', 'spec']
referenced-but-undeclared: none
declared-but-unused      : none
```

`pytest tests/test_workflows.py` — 941 passed. The 20 failures are all `WinError 1314: A required privilege is not held by the client` from symlink creation; I confirmed the identical set fails on unmodified `upstream/main`, so they are environmental and unrelated.

## Follow-up worth considering separately

`validate_workflow` already rejects malformed `enum` shapes and out-of-enum defaults, but has no check for a declared input that no step references. A warning there would have caught this at authoring time. Left out of this PR to keep it to the fix.

---

*Disclosure: this change was developed with AI assistance (Claude). The AI helped trace the input-resolution path, establish the accept-and-discard behaviour, and draft this description. The coercion results and the declared/referenced comparison above were produced by running the code; I reviewed the change before submitting.*

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -415,10 +415,6 @@ inputs:
     type: string
     default: "copilot"
     prompt: "Integration to use (e.g. claude, copilot, gemini)"
-  scope:
-    type: string
-    default: "full"
-    enum: ["full", "backend-only", "frontend-only"]
 
 steps:
   - id: specify
--- a/workflows/README.md
+++ b/workflows/README.md
@@ -73,7 +73,7 @@ specify workflow run ./my-workflow.yml --input spec="Build a user authentication
 ```bash
 specify workflow run speckit \
   --input spec="Build a user authentication system with OAuth support" \
-  --input scope="backend-only"
+  --input integration="claude"
 ```
 
 ## Step Types
--- a/workflows/speckit/workflow.yml
+++ b/workflows/speckit/workflow.yml
@@ -35,10 +35,6 @@ inputs:
     type: string
     default: "auto"
     prompt: "Integration to use (e.g. claude, copilot, gemini; 'auto' uses the project's initialized integration)"
-  scope:
-    type: string
-    default: "full"
-    enum: ["full", "backend-only", "frontend-only"]
 
 steps:
   - id: specify


Yash-Chindam:
Fixed — `scope` is now removed from the reference copy in `docs/reference/workflows.md` too. That file reproduces the shipped workflow verbatim, so leaving it would have made the documented definition disagree with the workflow this PR changes. Good catch; my original sweep for consumers missed it.

I checked the other `scope` references and left them deliberately, since none claims to reproduce the bundled workflow:

- `workflows/README.md:407` — the input **type-coercion** section, where `scope` is the illustrative `enum` example alongside `task_count` (number) and `dry_run` (boolean)
- `workflows/PUBLISHING.md:72` — a skeleton for authoring your *own* workflow (`speckit_version: ">=0.6.1"`, `any: ["claude", "gemini"]`)
- `workflows/ARCHITECTURE.md:146` — the schema table row documenting what `enum` does

## Pre-existing drift in the same reference copy

While confirming that block, I noticed the copy has drifted from the shipped workflow in three other ways, all predating this PR:

| | `docs/reference/workflows.md` | `workflows/speckit/workflow.yml` |
|---|---|---|
| `speckit_version` | `">=0.7.2"` | `">=0.8.5"` |
| `integrations.any` | `["copilot", "claude", "gemini"]` | also `alquimia`, `opencode` |
| `integration` default | `"copilot"` | `"auto"` |

The last one is the most user-visible — the docs state the default integration is `copilot` when it is actually `auto`, resolved from the project's initialized integration.

I have left these out to keep the PR to the reported bug. Happy to sync them here if you would prefer the reference copy fully accurate in one go, or to file it separately — your call.

Also rebased onto current `main`.

*Disclosure: AI assistance (Claude) was used for this change and comment, as with the original PR.*

Yash-Chindam:
Closing — #4401 landed the same fix and resolved #4384. Thanks @gyanu2507; that PR also covered `workflows/catalog.json` and added bundled-workflow test coverage, which this one did not.

One thing from my review here that #4401 did not pick up, in case it is worth a separate issue: the reference copy in `docs/reference/workflows.md` still differs from the shipped workflow in three ways that predate both PRs.

| | `docs/reference/workflows.md` | `workflows/speckit/workflow.yml` |
|---|---|---|
| `speckit_version` | `">=0.7.2"` | `">=0.8.5"` |
| `integrations.any` | `["copilot", "claude", "gemini"]` | also `alquimia`, `opencode` |
| `integration` default | `"copilot"` | `"auto"` |

The last is the most user-visible — the docs state the default integration is `copilot`, when it is actually `auto`, resolved from the project's initialized integration. Happy to open an issue or a PR for it if that would be useful.

mnriem:
@Yash-Chindam Go for the mentioned PR :)

Yash-Chindam:
Opened as #4424 — syncs all four drifted fields and adds a guard (comparing parsed YAML, so the guide keeps its own formatting) that fails against the pre-sync copy and passes after. It extends the test file added in #4401. Thanks!

PR #4405, opened 2026-09-02
fix(integrations): atomically record generated skill files

### LINKED ISSUE #4273: [Bug]: specify init writes speckit.manifest.json without the speckit-converge skill it just installed
## Version

spec-kit v1.0.1, `specify init --here --integration claude --force`

## Description

After an in-place upgrade, `.specify/integrations/speckit.manifest.json` lists 9 `speckit-*` SKILL.md entries under `files` — but init also installed `.claude/skills/speckit-converge/SKILL.md`, which is absent from the manifest.

Any tooling that uses the manifest to classify vendored files (pristine / modified / missing / untracked — e.g. deciding whether the next `specify init` will clobber local work) sees converge as **untracked**: it cannot be classified at all, which is the worst of the four states for a file the installer itself owns.

Observed identically on two independent upgrades (a template repo and a workflows repo, both 0.5.0 → v1.0.1 in-place).

## Expected behavior

The generated manifest includes every skill file the init run installs, converge included.

Gambit-Checkmate:
## Summary
- route generated integration files through IntegrationManifest.record_file so writing and manifest tracking happen together
- add Claude integration coverage ensuring every generated SKILL.md, including speckit-converge, persists in claude.manifest.json

Fixes #4273

## Validation
- python -m compileall -q completed for both modified Python files

The full repository test suite is not available in the local environment, so CI should run the integration test.

### DIFF
--- a/src/specify_cli/integrations/base.py
+++ b/src/specify_cli/integrations/base.py
@@ -563,17 +563,14 @@ class IntegrationBase(ABC):
     ) -> Path:
         """Write *content* to *dest*, hash it, and record in *manifest*.
 
-        Creates parent directories as needed.  Writes bytes directly to
-        avoid platform newline translation (CRLF on Windows).  Any
-        ``\r\n`` sequences in *content* are normalised to ``\n`` before
-        writing.  Returns *dest*.
+        Uses the manifest as the single write-and-record entry point, so a
+        generated file cannot be written without its hash being recorded.
+        Any ``\r\n`` sequences in *content* are normalised to ``\n`` before
+        writing. Returns the path written by the manifest.
         """
-        dest.parent.mkdir(parents=True, exist_ok=True)
         normalized = content.replace("\r\n", "\n")
-        dest.write_bytes(normalized.encode("utf-8"))
         rel = dest.resolve().relative_to(project_root.resolve())
-        manifest.record_existing(rel)
-        return dest
+        return manifest.record_file(rel, normalized)
 
     def integration_scripts_dir(self) -> Path | None:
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
The two modified files need to be normalized back to LF before this can be reviewed. The repository enforces `* text=auto eol=lf`, but this commit introduces mixed CRLF/LF endings across almost every line: the displayed diff is 2,778 additions and 2,772 deletions, while `git diff -w` reduces it to roughly 22 changed lines. This makes the substantive change effectively unreadable and creates unnecessary merge risk.

Please also demonstrate that the regression test fails on `main` before the implementation change. As written, it appears to assert behavior the existing `write_file_and_record()` path already provides, so it does not yet establish that switching from write-then-`record_existing()` to `record_file()` fixes the reported upgrade scenario.

Gambit-Checkmate:
Thanks for the review. I normalized the local investigation and re-ran the reported upgrade path in an isolated environment: initialize with v0.5.0, then run v1.0.1 init --here --integration claude --force in the same project. The upgraded project contains 10 .claude/skills/speckit-*/SKILL.md files, and claude.manifest.json records all 10, including speckit-converge/SKILL.md. A fresh v1.0.1 install also records all 10.\n\nBecause I cannot make the reported behavior fail before the proposed implementation change, this PR does not demonstrate a valid fix. I am closing it rather than retain a noisy or unproven change. If you can provide the minimal project state or environment condition that reproduces the missing entry, I would be glad to revisit it with a failing regression test first.

PR #4416, opened 2026-09-03
fix(workflows): resolve negative list indices in expressions

NgoQuocViet2001:
## Problem

`_resolve_dot_path` matches the index bracket with `^([\w-]+)\[(\d+)\]$`, which accepts digits only. A negative index therefore never enters the indexing branch — it falls through to the dict lookup and asks for the literal key `"task_list[-1]"`, which is absent, so the whole path resolves to `None`:

```python
ctx = StepContext(steps={"tasks": {"output": {"task_list": [{"file": "a.md"}, {"file": "b.md"}]}}})

evaluate_expression("{{ steps.tasks.output.task_list[0].file }}", ctx)   # 'a.md'
evaluate_expression("{{ steps.tasks.output.task_list[-1].file }}", ctx)  # None
```

`list[-1]` is valid in both Python and the Jinja2 subset the module documents itself as providing, and "the last item a step produced" is a natural thing for a workflow template to want. There is no error: the template renders empty, and `evaluate_condition` on the same path reads false, so a step can be skipped for a reason that never surfaces.

## Fix

Accept the negative form in the pattern and bound the index from both ends. Out-of-range in either direction keeps returning `None` rather than raising, matching the existing behaviour for `[9]` on a short list.

## Scope

Two lines in `src/specify_cli/workflows/expressions.py` plus a docstring note, and one test next to the existing `test_list_indexing`. Positive indices and non-index path segments are untouched.

## Test plan

- Ran: `pytest tests/test_workflows.py -k "indexing or literal"` → 11 passed.
- Ran: `pytest tests/test_workflows.py` → 942 passed. The 20 failures are the `TestWorkflowCliAlignment` symlink cases, which fail identically on an unmodified checkout here (Windows, no symlink privilege).
- Checked: reverting only `expressions.py` fails the new test with `assert None == 'b.md'`.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -1048,8 +1048,9 @@ def _has_incomplete_operand(text: str) -> bool:
 # None, so a correction built on one turns a truthy condition false.
 _NAMESPACE_ROOTS = ("inputs", "steps", "item", "fan_in", "context")
 
-# Exactly what _resolve_dot_path accepts: a name, optionally one numeric index.
-_PATH_SEGMENT = re.compile(r"^[\w-]+(\[\d+\])?$")
+# Exactly what _resolve_dot_path accepts: a name, optionally one index, which
+# may be negative.
+_PATH_SEGMENT = re.compile(r"^[\w-]+(\[-?\d+\])?$")
 
 
 class _ProbeNamespace(dict):
@@ -1214,7 +1215,7 @@ def _unresolvable_term(text: str) -> str | None:
     # branch returns None for those however it is written -- so the index is
     # stripped for `item` alone rather than for roots in general.
     root = segments[0].strip()
-    indexed_root = re.fullmatch(r"([\w-]+)\[\d+\]", root)
+    indexed_root = re.fullmatch(r"([\w-]+)\[-?\d+\]", root)
     if indexed_root is not None and indexed_root.group(1) == "item":
         root = indexed_root.group(1)
     if root not in _NAMESPACE_ROOTS:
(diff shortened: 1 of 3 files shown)


mnriem (changes requested):

NgoQuocViet2001:
Addressed — the Copilot finding was correct, thanks.

`_PATH_SEGMENT` and the indexed-root fullmatch were both still on `\d+`, so a condition the resolver handles was classified unresolvable and the correction withheld:

```
item[0] == 'x'   -> Wrap the expression: "{{ item[0] == 'x' }}".
item[-1] == 'x'  -> No correction is offered because 'item[-1]' is not a name
                    the evaluator can resolve
```

Both now allow `-?\d+`, and the two negative cases are in the existing parametrize beside `test_an_indexed_item_root_keeps_the_correction`. The refusal side is unchanged — `inputs[0]` and `steps[-1]` are still rejected as non-roots.

`pytest tests/unit/test_condition_expression_block.py` → 337 passed. Reverting only `expressions.py` fails the two new cases.

mnriem:
Thanks for the quick turnaround — the fix to `_PATH_SEGMENT` and the indexed-root fullmatch looks right, and I appreciate you catching that the correction path had the same `\d+` gap. Two things before merge: (1) please add the **AI-disclosure** per CONTRIBUTING (the PR body has none); (2) the Copilot 🟡 review predates your fix — I'll re-request it to confirm it's resolved. Assuming it comes back clean and disclosure's added, this is good to go.

mnriem:
The re-run review is green now — the negative-index handling and its regression coverage look solid, thanks for turning the earlier findings around.


One thing before I can merge: please add the **AI-disclosure** per [[CONTRIBUTING](https://github.com/github/spec-kit/blob/main/CONTRIBUTING.md#ai-contributions-in-spec-kit)](https://github.com/github/spec-kit/blob/main/CONTRIBUTING.md#ai-contributions-in-spec-kit) — either describe the AI assistance and its extent, or check the "no AI assistance" box. That's the only item still outstanding on your side.


Heads-up on sequencing: this and #4417 both touch `expressions.py`, and there's a larger refactor in flight (#4460) that reworks how the correction gate resolves leaves — which overlaps the grammar area you're patching here. I'm sorting out the merge order across those PRs; depending on which lands first, I may ask you to rebase. Nothing to do now — just flagging so a rebase request later isn't a surprise. Your fix is good either way.

PR #4424, opened 2026-09-03
docs(workflows): sync the reference copy with the shipped workflow

Yash-Chindam:
Follow-up to #4384 / #4398 — @mnriem asked me to open this in [#4398](https://github.com/github/spec-kit/pull/4398#issuecomment-5526283124).

## Problem

`docs/reference/workflows.md` introduces its YAML block with:

> Here is the built-in **Full SDD Cycle** workflow that ships with Spec Kit:

A reader is entitled to treat that as the real definition. Diffing the block against `workflows/speckit/workflow.yml` shows it had drifted on four points:

| | reference guide | shipped workflow |
|---|---|---|
| `workflow.version` | `"1.0.0"` | `"1.0.1"` |
| `requires.speckit_version` | `">=0.7.2"` | `">=0.8.5"` |
| `requires.integrations.any` | `copilot`, `claude`, `gemini` | also `alquimia`, `opencode` |
| `inputs.integration.default` | `"copilot"` | `"auto"` |

The last is the most user-visible. The guide stated the default integration is `copilot`, when it is actually `auto`, resolved from the project's initialized integration — so someone reading the guide to learn what they get by default was told the wrong thing. The `prompt:` text is updated to match for the same reason.

The `steps:` section was already accurate and is untouched.

## Guard

`test_reference_doc_matches_the_shipped_workflow` extends the file added in #4401 and pins the two together, so this cannot drift again silently.

It compares **parsed YAML rather than text**, so the guide stays free to format lists however reads best — only the content has to agree.

Verified in both directions:

- against the pre-sync copy it **fails**, and names all four differences:
  ```
  {'requires': {'speckit_version': '>=0.7.2', 'integrations': {'any': ['copilot', 'claude', 'gemini']}}}
    != {'requires': {'speckit_version': '>=0.8.5', 'integrations': {'any': ['alquimia', 'claude', 'copilot', 'gemini', 'opencode']}}}
  {'workflow': {... 'version': '1.0.0' ...}} != {'workflow': {... 'version': '1.0.1' ...}}
  ```
- after the sync it **passes**

## Notes

Only the reference guide's copy is touched. The other `scope`/`enum` examples elsewhere are generic authoring illustrations rather than reproductions of the bundled workflow, so they are correctly left alone:

- `workflows/README.md` — the input type-coercion section
- `workflows/PUBLISHING.md` — a skeleton for authoring your own workflow
- `workflows/ARCHITECTURE.md` — the schema table row documenting `enum`

`pytest tests/workflows/` — 200 passed. The 10 failures are all `WinError 1314: A required privilege is not held by the client` from symlink creation on my Windows machine; zero non-symlink failures, and the identical set fails on unmodified `main`.

---

*Disclosure: this change was developed with AI assistance (Claude). The AI helped diff the two definitions, write the guard, and draft this description. The four differences and both directions of the guard were verified by running the tests; I reviewed the change before submitting.*

### DIFF
--- a/docs/reference/workflows.md
+++ b/docs/reference/workflows.md
@@ -431,14 +431,19 @@ schema_version: "1.0"
 workflow:
   id: "speckit"
   name: "Full SDD Cycle"
-  version: "1.0.0"
+  version: "1.0.1"
   author: "GitHub"
   description: "Runs specify → plan → tasks → implement with review gates"
 
 requires:
-  speckit_version: ">=0.7.2"
+  speckit_version: ">=0.8.5"
   integrations:
-    any: ["copilot", "claude", "gemini"]
+    any:
+      - "alquimia"
+      - "claude"
+      - "copilot"
+      - "gemini"
+      - "opencode"
 
 inputs:
   spec:
@@ -447,8 +452,8 @@ inputs:
     prompt: "Describe what you want to build"
   integration:
     type: string
-    default: "copilot"
-    prompt: "Integration to use (e.g. claude, copilot, gemini)"
+    default: "auto"
+    prompt: "Integration to use (e.g. claude, copilot, gemini; 'auto' uses the project's initialized integration)"
 
 steps:
   - id: specify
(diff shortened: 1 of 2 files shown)


mnriem:
Thank you!

PR #4425, opened 2026-09-03
feat(workflows): add per-step integration configuration

### LINKED ISSUE #4275: [Feature]: Per step integration configuration in workflows
### Problem Statement

Workflow steps currently allow specifying an integration, but do not provide a way to configure that integration independently for a particular step.

This is limiting for integrations whose CLI requires additional runtime arguments or options. For example, [Docker Agent](https://docs.docker.com/ai/docker-agent/getting-started/introduction/#how-it-works) may be launched like:

```bash
docker-agent run ./agent.yaml --exec "<prompt>"
```

The `./agent.yaml` path must be passed before the prompt as a positional argument. Also, an integration may require additional key-value options such as:

```text
--agent root
--model provider/model
--safety balanced
```

The current `SPECKIT_INTEGRATION_<KEY>_EXTRA_ARGS` environment variable applies to the **entire process** rather than to an individual workflow step. This prevents a workflow from using different integration configurations for different steps and makes the effective configuration invisible in the workflow definition.

### Proposed Solution

Add per-step integration configuration to the workflow step definition.

The existing `integration` field remains a string containing the integration key:

```yaml
- id: specify
  command: speckit.specify
  integration: docker-agent
  input:
    args: "{{ inputs.spec }}"
```

Add two optional fields:

```yaml
- id: specify
  command: speckit.specify
  integration: docker-agent
  integration_args:
    - ./agent.yaml
  integration_options:
    agent: root
    safety: balanced
  input:
    args: "{{ inputs.spec }}"
```

Where

**integration_args**

`integration_args` is an ordered list of positional arguments passed to the selected integration before the generated prompt.

For Docker Agent:

```yaml
integration_args:
  - ./agent.yaml
```

should produce a command equivalent to:

```bash
docker-agent run ./agent.yaml --exec --json "<prompt>"
```

The order of values in `integration_args` must be preserved.

**integration_options**

`integration_options` is a mapping of key-value options passed to the selected integration.

```yaml
integration_options:
  agent: root
  safety: balanced
```

The integration is responsible for translating these options into its native CLI representation:

```bash
--agent root --safety balanced
```

Integration-specific options should be validated by the selected integration. Unknown options must result in an actionable validation error.

**Runtime API**

The workflow runner should resolve and pass these values only to the current step:

```python
integration.dispatch_command(
    command,
    args=command_args,
    project_root=project_root,
    model=model,
    integration_args=resolved_integration_args,
    integration_options=resolved_integration_options,
)
```

The integration API should expose the values to command construction:

```python
def build_exec_args(
    self,
    prompt: str,
    *,
    model: str | None = None,
    output_json: bool = True,
    integration_args: Sequence[str] | None = None,
    integration_options: Mapping[str, Any] | None = None,
) -> list[str] | None:
    ...
```

The existing integrations must remain compatible. Integrations that do not use per-step configuration should ignore empty `integration_args` and `integration_options`.

### Alternatives Considered

**Global environment variables**

```bash
export SPECKIT_INTEGRATION_DOCKER_AGENT_EXTRA_ARGS=./agent.yaml
specify workflow run speckit -i spec="..."
```

This is already available for some use cases, but it applies to the whole workflow process. It cannot express different configurations for individual steps and hides important workflow configuration outside the workflow definition.

**Step-level environment variables**

```yaml
- id: specify
  type: command
  env:
    SPECKIT_INTEGRATION_DOCKER_AGENT_EXTRA_ARGS: ./agent.yaml
```

This would require adding per-step environment handling and carefully isolating child-process environments. It also retains the ambiguity of treating a path as an untyped arbitrary environment string.

**A dedicated `agent_config` field**

```yaml
integration:
  name: docker-agent
  agent_config: ./agent.yaml
```

This is clear for Docker Agent, but introduces an integration-specific field into the common workflow schema. It does not generalize well to integrations that need multiple positional arguments or different key-value options.

**Changing `integration` from a string to a mapping**

```yaml
integration:
  name: docker-agent
  options:
    ...
```

This could express the desired configuration, but would change the existing type and break compatibility with current workflow definitions that use:

```yaml
integration: docker-agent
```

Keeping `integration` as a string and adding `integration_args` and `integration_options` avoids that breaking change.

### Component

Agent integrations (command files, workflows)

### AI Agent (if applicable)

None

### Use Cases

_No response_

### Acceptance Criteria

- [ ] Workflow steps accept an optional `integration_args` list.
- [ ] Workflow steps accept an optional `integration_options` mapping.
- [ ] `integration_args` values are resolved using the existing workflow expression mechanism.
- [ ] `integration_options` values are resolved using the existing workflow expression mechanism.
- [ ] Positional arguments preserve their declared order.
- [ ] Integration arguments are passed only to the step where they are declared.
- [ ] Integration options are passed only to the step where they are declared.
- [ ] Integrations that do not use these fields continue to work unchanged.
- [ ] Unknown or malformed integration options produce actionable validation errors.
- [ ] Relative file paths are resolved relative to the workflow project root where appropriate.
- [ ] Per-step integration configuration is persisted in workflow run state so resume behavior is deterministic.
- [ ] Configuration values are not interpolated into shell commands as unescaped strings.
- [ ] Existing `SPECKIT_INTEGRATION_<KEY>_EXTRA_ARGS` behavior remains backward compatible.
- [ ] Tests cover:
  - [ ] no per-step configuration;
  - [ ] positional arguments;
  - [ ] key-value options;
  - [ ] expression-based values;
  - [ ] separate configurations across multiple steps;
  - [ ] workflow resume.

### Additional Context

_No response_

philo-x:
Assisted-by: OpenAI Codex (model: GPT-5, autonomous)

## Summary

- add optional per-step `integration_args` and `integration_options` to workflow command steps
- resolve, validate, isolate, and persist runtime integration configuration for deterministic resume behavior
- add Docker Agent support for per-step agent references and validated `agent` and `safety` options, using the command-step `model` field for model selection while retaining the legacy extra-args path for steps without per-step `integration_args`
- document the workflow syntax and audit all built-in execution signatures for compatibility

Closes #4275

## Testing

- [x] `.venv/bin/python -m pytest tests/integrations/test_base.py tests/integrations/test_integration_docker_agent.py tests/test_workflows.py -q` — 1121 passed on `b4bcd5f7` after syncing main
- [x] Ruff 0.15.0 targeted checks (`E9,F63,F7,F82,S602,S604,S605`)
- [x] `git diff --check`
- Earlier full suite result (not rerun for this precedence follow-up): 7523 passed, 195 skipped, with 3 failures in untouched areas (PowerShell launcher availability and two bundler reference tests)

## AI Disclosure

Implementation, tests, documentation, and this PR description were authored by OpenAI Codex (model: GPT-5, autonomous) on behalf of @philo-x.

Precedence follow-up, regression tests, documentation updates, and main synchronization authored by OpenAI Codex (model: GPT-6, autonomous) on behalf of @philo-x.


mnriem (changes requested):

philo-x:
Posted on behalf of @philo-x by OpenAI Codex (model: GPT-5).

Addressed the review feedback in `83b6b9a2f743bf505f0de59ea0618ebb6bbfa9e4`:

- replaced the legacy Docker Agent config reference when per-step `integration_args` is present while preserving legacy flags
- made the command-step `model` field the single Docker Agent model source
- re-resolved the complete dispatch configuration consistently on resume
- enforced runtime configuration validation for direct execution-argument builders
- added regression coverage and updated the workflow reference documentation

Validation: 1,111 targeted tests pass; Ruff 0.15.0 and `git diff --check` pass. The broader integration suite reports 2,892 passed and 5 skipped, with only the pre-existing PowerShell launcher availability failure remaining.

mnriem:
Thanks — and good that this rides on the agreed #4275 design, with disclosure. One merge-blocker to resolve from the review: when the legacy Docker-Agent env var has its documented shape (e.g. `./default.yaml --hide-tool-results`) *and* a step supplies `integration_args`, the current code appends **both** agent references, emitting two conflicting ones. Since preserving the legacy path is an explicit goal of this PR, please define the precedence (I'd expect per-step `integration_args` to win and the legacy value to be ignored when both are present) and add a test for that collision. Re-request review once that's in.

philo-x:
Posted on behalf of @philo-x by OpenAI Codex (model: GPT-6, autonomous).

Thanks @mnriem. The stricter precedence is implemented in `88d6abd`: when per-step `integration_args` supplies an agent reference, the entire `SPECKIT_INTEGRATION_DOCKER_AGENT_EXTRA_ARGS` value is ignored for that step, including malformed quoting. The legacy environment behavior remains unchanged when no per-step agent reference is supplied.

For completeness, the previous head (`83b6b9a2`) already replaced the legacy agent reference with the per-step one, but still inherited any trailing legacy flags. This follow-up makes the precedence stricter by ignoring the legacy value entirely whenever a per-step agent reference is present.

Regression coverage now includes legacy agent-plus-flags, flags-only, and malformed quoting. Both workflow guides document the precedence, and the PR description identifies the command-step `model` field as the model source.

Synced with main at `0c8e31ff`; final head is `b4bcd5f`. Validation on that head: 1,121 targeted tests passed, Ruff 0.15.0 targeted checks passed, and `git diff --check` passed. CI workflows are currently awaiting maintainer approval.

@mnriem, this is ready for another review. Thanks again.

mnriem:
Thanks — that's exactly the precedence I wanted: when a per-step `integration_args` agent reference is present, ignore the legacy `SPECKIT_INTEGRATION_DOCKER_AGENT_EXTRA_ARGS` entirely (including malformed quoting), and the regression coverage for legacy-agent-plus-flags / flags-only / malformed quoting nails the cases down. Re-requesting review/CI on `88d6abd`; once it's green I'll merge (closes #4275).

mnriem:
Thank you!

PR #4432, opened 2026-09-03
docs(templates): clarify /constitution's Sync Impact Report is temporary, review-only material (#4431)

### LINKED ISSUE #4431: [Bug]:  /constitution prepends a new Sync Impact Report HTML   comment on every run, growing agent context unboundedly
### Bug Description

The /constitution command is instructed to produce a Sync Impact Report and prepend it as an HTML comment at the top of .specify/memory/constitution.md (Step 4 of the Outline section in templates/commands/constitution.md, commit 7b2f3684):

`"Produce a Sync Impact Report (prepend as an HTML comment at top of the constitution file after update)`

HTML comments are invisible in rendered Markdown but are fully present in the raw file. Because the instruction says "prepend" and says nothing about removing a previous report, each invocation of /constitution stacks another  block above the last. Every agent or command that loads .specify/memory/constitution.md at runtime, including /specify, /review, and any extension hooks registered under before_constitution or after_constitution, reads the raw file and pays a growing token cost for accumulated historical changelog data that has no governance value at inference time.

### Steps to Reproduce

1. Initialise a project and run /constitution to create an initial .specify/memory/constitution.md.
2. Make a minor amendment (e.g. reword one principle) and run /constitution again.
3. Repeat step 2 a third time.
4. Open .specify/memory/constitution.md and inspect the raw file (not the rendered preview).

### Expected Behavior

The file contains at most one Sync Impact Report comment, representing the most recent update. On each run the previous comment is overwritten, not accumulated. Alternatively, the report is not written to the constitution file at all, because the information (version change, modified/added/removed sections, deferred TODOs) is already surfaced in the Step 7 summary output to the user, which is the natural place to derive a commit message from.

### Actual Behavior

Each run prepends a new  block abovw the previous one. After n runs the raw file contains n stacked report blocks. Any agent loading the constitution for downstream work reads all of them, spending tokens on historical diffs that are irrelevant to the current task.

### Specify CLI Version

v1.0.4

### AI Agent

Claude Code

### Operating System

macOS

### Python Version

Python 3.14.7

### Error Logs

```shell

```

### Additional Context

_No response_

chelsealong:
Fixes #4431

## Problem

`templates/commands/constitution.md` (Step 4 of the Outline section) instructed
agents to "prepend" the Sync Impact Report as an HTML comment at the top of
`.specify/memory/constitution.md`, with no instruction to remove a previous
report first. Because HTML comments are invisible in rendered Markdown but
fully present in the raw file, every `/constitution` run stacked another
`<!-- ... -->` report block on top of the last one. Any agent or extension
hook that loads the raw constitution file at runtime (`/specify`, `/review`,
`before_constitution`/`after_constitution` hooks, etc.) pays a growing,
unbounded token cost reading historical changelog data that has no
governance value at inference time.

## Fix

Step 4 now explicitly instructs the agent to remove any existing HTML
comment at the top of the file before adding the new Sync Impact Report,
so the file never carries more than one report — it is replaced, not
stacked.

This is a one-line-of-behavior instruction change to the command template
(`templates/commands/constitution.md`); no other copies of this text exist
in the repo.

## Test plan

Added `tests/test_constitution_template_sync_report.py`, which asserts the
Step 4 instructions require removing any prior Sync Impact Report comment
rather than only prepending.

- Confirmed the test fails against the pre-fix template
  (`git checkout HEAD~1 -- templates/commands/constitution.md`):
  ```
  tests/test_constitution_template_sync_report.py::test_sync_impact_report_step_instructs_removing_prior_report FAILED
  AssertionError: assert ('never stack' in '...' or 'not stack' in '...')
  ```
- Restored the fix and confirmed the test passes:
  ```
  tests/test_constitution_template_sync_report.py::test_sync_impact_report_step_instructs_removing_prior_report PASSED
  1 passed in 0.18s
  ```
- Ran the full suite: `uv run --extra test pytest -q`
  ```
  7705 passed, 12 skipped, 48 warnings in 499.34s (0:08:19)
  ```

## AI disclosure

This change, including the code, tests, and this PR description, was
generated autonomously by an AI coding agent (Claude, model: claude-sonnet-5)
acting on behalf of the repository owner, with no line-by-line human review
prior to commit.

Assisted-by: Claude (model: claude-sonnet-5, autonomous)

### DIFF
--- a/templates/commands/constitution.md
+++ b/templates/commands/constitution.md
@@ -109,10 +109,12 @@ Follow this execution flow:
    - Ensure each Principle section: succinct name line, paragraph (or bullet list) capturing non‑negotiable rules, explicit rationale if not obvious.
    - Ensure Governance section lists amendment procedure, versioning policy, and compliance review expectations.
 
-4. Produce a Sync Impact Report as an HTML comment at the top of the constitution file after update:
-   - If the file already starts with an HTML comment (e.g. a Sync Impact Report from a prior run),
-     remove it entirely before adding the new one. The file must never carry more than one Sync
-     Impact Report; replace, never stack.
+4. Produce a Sync Impact Report as an HTML comment at the top of the constitution file after update.
+   This report is temporary scratch material for human review of the amendment, not governance
+   content; it is expected to be removed before the amended constitution file is committed.
+   - If the file already starts with an HTML comment (e.g. a Sync Impact Report left over because
+     it was not removed before a prior commit), remove it entirely before adding the new one. The
+     file must never carry more than one Sync Impact Report; replace, never stack.
    - Version change: old → new
    - List of modified principles (old title → new title if renamed)
    - Added sections
--- a/templates/commands/constitution.md
+++ b/templates/commands/constitution.md
@@ -109,7 +109,10 @@ Follow this execution flow:
    - Ensure each Principle section: succinct name line, paragraph (or bullet list) capturing non‑negotiable rules, explicit rationale if not obvious.
    - Ensure Governance section lists amendment procedure, versioning policy, and compliance review expectations.
 
-4. Produce a Sync Impact Report (prepend as an HTML comment at top of the constitution file after update):
+4. Produce a Sync Impact Report as an HTML comment at the top of the constitution file after update:
+   - If the file already starts with an HTML comment (e.g. a Sync Impact Report from a prior run),
+     remove it entirely before adding the new one. The file must never carry more than one Sync
+     Impact Report; replace, never stack.
    - Version change: old → new
    - List of modified principles (old title → new title if renamed)
    - Added sections
--- a/templates/commands/constitution.md
+++ b/templates/commands/constitution.md
@@ -112,9 +112,6 @@ Follow this execution flow:
 4. Produce a Sync Impact Report as an HTML comment at the top of the constitution file after update.
    This report is temporary scratch material for human review of the amendment, not governance
    content; it is expected to be removed before the amended constitution file is committed.
-   - If the file already starts with an HTML comment (e.g. a Sync Impact Report left over because
-     it was not removed before a prior commit), remove it entirely before adding the new one. The
-     file must never carry more than one Sync Impact Report; replace, never stack.
    - Version change: old → new
    - List of modified principles (old title → new title if renamed)
    - Added sections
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
As indicated by the comment on the issue please change this to a documentation issue

chelsealong:
Agreed, retitled as a docs clarification. Pushed 609dc58: Step 4 now states explicitly that the Sync Impact Report is temporary, review-only material expected to be removed before the constitution file is committed, alongside the existing replace-not-stack instruction for runs where a report was left in place (`templates/commands/constitution.md`). Test suite still green (`tests/test_constitution_template_sync_report.py`).

mnriem:
Thanks for adding the lifecycle wording — that part is exactly right. But I'd like this to be **documentation-only**, and right now it still carries the behavioral change I was pushing back on.

Two requests:

1. **Drop the de-dupe instruction.** Please remove the bullet that tells the agent to strip a pre-existing HTML comment:

*"If the file already starts with an HTML comment … remove it entirely before adding the new one. The file must never carry more than one Sync Impact Report; replace, never stack."*

In the intended workflow the Sync Impact Report is temporary scratch material that's removed before the amended constitution is committed, so `/constitution` shouldn't be auto-stripping prior reports — that's a behavior change to work around a state that shouldn't occur. Keep just the lifecycle sentence you added.

2. **Retarget the test.** `test_constitution_template_sync_report.py` currently asserts the *removal/never-stack behavior* (`"remove"`, `"never stack"`). Please change it to assert the **documented lifecycle** instead — e.g. that Step 4 states the report is temporary/review-only and is expected to be removed before commit. The test should pin the documentation, not the de-dupe logic we're removing.

Net: Step 4 gains only the clarification that the report is temporary and removed before commit; no new agent behavior. Re-request review once that's in.

chelsealong:
Pushed e3870d6: removed the de-dupe bullet from Step 4 (no more auto-stripping a leading HTML comment), keeping only the temporary/removed-before-commit lifecycle wording. Retargeted `tests/test_constitution_template_sync_report.py` to assert that documented lifecycle instead of the removal/never-stack behavior. Full suite green (7705 passed, 12 skipped).

mnriem:
Approving. To be explicit for the record: Copilot's 🟡 wants the de-dup / at-most-one-report behavior enforced, but that's the command-behavior change I deliberately scoped **out** of this PR. In the intended workflow the Sync Impact Report is temporary and removed before commit, so accumulation shouldn't occur — documenting that lifecycle is the correct fix, not code-enforcing a de-dup. [[@chelsealong](https://github.com/chelsealong)](https://github.com/chelsealong) implemented exactly what I asked. Overriding the automated review and merging.

mnriem:
Thank you!

PR #4445, opened 2026-09-04
fix(scripts): honor SPECKIT_PYTHON override for preset manifest parsing

### LINKED ISSUE #4443: [Bug] Preset composition fails under isolated installs: scripts run on system python3 but PyYAML lives in the CLI venv
## Summary

Preset template composition fails with `PyYAML is required to resolve preset template composition` whenever the CLI is installed the recommended way (`uv tool install` or `pipx`), because the automation scripts run under the system `python3` while `pyyaml` lives in the CLI's isolated venv.

The better the installation isolation, the more reliably the bug shows up.

## The chain

1. Command skills instruct the agent to run the **script**, not the CLI. From `speckit-plan`:
   ```
   Run `python3 .specify/scripts/python/setup_plan.py --json` from repo root
   ```
2. That script reaches `common.py`, which does `import yaml` (line 391 in v1.0.4) and raises `TemplateResolutionError("PyYAML is required to resolve preset template composition")` when a `preset.yml` manifest exists.
3. `pyyaml>=6.0` is declared in `pyproject.toml` and is correctly installed — but inside the tool venv created by `uv tool install` / `pipx`.
4. The script is invoked as bare `python3`, which resolves to the system interpreter that cannot see that venv.

## Reproduction

```bash
uv tool install specify-cli
specify init myproj --integration claude --script py
cd myproj
specify preset add <any preset with a preset.yml>

# CLI path — works (runs inside the tool venv):
specify preset resolve plan-template

# Script path — fails (runs under system python3):
python3 .specify/scripts/python/resolve_template.py plan-template --json
# ERROR: PyYAML is required to resolve preset template composition
```

Same failure with the bash variant (`.specify/scripts/bash/resolve-template.sh`), since `_python3_command()` in `common.sh` only probes `python3` on PATH.

## Environment

- spec-kit `1.0.4`
- macOS 15 (arm64), Homebrew Python 3.14.6
- Reproduced with both `uv tool install specify-cli` and `pipx install specify-cli`

## Why the usual workarounds don't apply

- `--script py` doesn't help: the Python variant is invoked by the same bare `python3`.
- There is no environment variable to point at another interpreter; `_python3_command()` only checks PATH.
- Installing `pyyaml` into the system interpreter is refused on Homebrew/Debian Python under [PEP 668](https://peps.python.org/pep-0668/) (`externally-managed-environment`). `uv pip install --python "$(which python3)" pyyaml` is refused for the same reason and suggests a venv — but a venv doesn't help, because the skills invoke bare `python3` with no activation.

The only way out today is bypassing PEP 668 (`pip install --user --break-system-packages pyyaml`), which asks users to disable a safety mechanism to run a documented feature.

## Impact

Any project using a preset with a manifest — i.e. any preset that needs `append`, `prepend` or `wrap`, since the manifest-less path is hardcoded to `replace`:

```python
except ImportError as exc:
    if manifest_path.is_file():
        raise TemplateResolutionError("PyYAML is required ...")
    return (conventional, "replace") if conventional is not None else None
```

So composition strategies are unavailable precisely for users who installed the CLI in an isolated environment, which is what the docs recommend.

## Suggested fixes

Any one of these would close it:

1. Have the scripts resolve the interpreter that ships with spec-kit (record its path at `init` time, similar to how some tools pin the interpreter into generated hooks) instead of assuming bare `python3`.
2. Honor an override such as `SPECIFY_PYTHON` in `_python3_command()`.
3. Vendor a minimal YAML parser for manifest reading, or parse the small, well-known `preset.yml` schema without the dependency.
4. If none of the above, document the requirement explicitly in the installation guide, since today it surfaces only as a runtime error.

Happy to send a PR for option 2 if that's the direction you'd prefer.

chelsealong:
## Summary

Fixes #4443.

Preset composition raises `PyYAML is required to resolve preset template
composition` whenever the resolve/setup scripts are invoked under a bare
`python3` that lacks PyYAML, while a `uv tool install` / `pipx`-isolated
CLI venv (which *does* have PyYAML, per `pyproject.toml`) is invisible to
that bare interpreter. This happens both when the bash/PowerShell twins
shell out to Python to parse a `preset.yml` manifest, and when the Python
twin (`resolve_template.py` / `setup_plan.py` / etc., via `common.py`)
tries to `import yaml` directly in its own process.

`SPECKIT_PYTHON` is already the established override for exactly this class
of problem elsewhere in the codebase (see
`extensions/agent-context/scripts/bash/update-agent-context.sh` and its
tests). This PR extends that same override to the three
template-resolution twins:

- `scripts/bash/common.sh`: `_python3_command()` now prefers `SPECKIT_PYTHON`
  (verified to be a working Python 3 interpreter *that also has PyYAML*)
  before falling back to `python3` / `python` / `py -3` on PATH.
- `scripts/powershell/common.ps1`: `Get-Python3Command` does the same.
- `scripts/python/common.py`: since this twin runs *in-process* rather than
  shelling out, `_preset_template_layer` now tries `import yaml` first and,
  if that fails, delegates manifest parsing to `SPECKIT_PYTHON` via a small
  subprocess-based `safe_load`/`YAMLError` shim (`_import_yaml` /
  `_DelegatedYAML`) so the rest of the function's logic is unchanged.

No dependencies, lockfiles, or CI config were touched — `pyyaml` was already
a declared dependency; this only fixes which interpreter picks it up.

**Update:** an earlier revision of this PR let the bash/PowerShell twins
adopt `SPECKIT_PYTHON` as soon as it resolved to *any* Python 3 interpreter,
without checking it actually had PyYAML — unlike the precedent this PR
cites, which validates both. That meant a `SPECKIT_PYTHON` set to a valid
but PyYAML-less interpreter (e.g. pinned for an unrelated reason) would
shadow a working `python3` on PATH and turn previously-succeeding preset
composition into a `PyYAML is required` failure. Both `_python3_command()`
and `Get-Python3Command` now also probe `import yaml` before accepting
`SPECKIT_PYTHON`, matching `update-agent-context.sh`'s candidate-validation
loop; if the probe fails they fall through to the existing `python3`/
`python`/`py -3` chain exactly as if `SPECKIT_PYTHON` were unset.

## Test plan

Added `test_all_variants_honor_speckit_python_override_when_yaml_missing` to
`tests/test_resolve_template_python_parity.py`. It creates a real
PyYAML-less interpreter (`python -m venv --without-pip`) and:
1. runs the bash and Python twins with that interpreter as the default
   `python3` and no `SPECKIT_PYTHON` set — asserts both fail (reproduces
   #4443), then
2. sets `SPECKIT_PYTHON` to the interpreter that has PyYAML — asserts both
   succeed and produce the correct composed template content.
(PowerShell is included in the same assertions when `pwsh` is available.)

Also added `test_all_variants_fall_back_when_speckit_python_lacks_pyyaml`,
covering the regression an earlier revision introduced: it sets
`SPECKIT_PYTHON` to a real Python-3 interpreter *without* PyYAML while
leaving PATH's own `python3` (which has PyYAML) untouched, and asserts
composition still succeeds via that PATH fallback across the bash and
PowerShell twins.

Confirmed both tests fail without their respective fixes:
- `test_all_variants_honor_speckit_python_override_when_yaml_missing`: with
  `git checkout HEAD~2 -- scripts/bash/common.sh
  scripts/powershell/common.ps1 scripts/python/common.py`, fails with
  `assert False` on the override-succeeds assertion.
- `test_all_variants_fall_back_when_speckit_python_lacks_pyyaml`: with only
  the latest commit's two files reverted (`git checkout HEAD~1 --
  scripts/bash/common.sh scripts/powershell/common.ps1`, i.e. the state the
  independent reviewer blocked), fails the same way — reproducing the
  reviewer's exact regression scenario.
Restored the fix in both cases and reran; both pass.

```
$ .venv/bin/python -m pytest tests/test_resolve_template_python_parity.py -q
============================= 42 passed in 39.14s ==============================

$ .venv/bin/python -m pytest tests/extensions/test_extension_agent_context.py -q
======================== 39 passed, 1 skipped in 10.62s ========================

$ .venv/bin/python -m pytest tests/ -q
========== 7706 passed, 12 skipped, 48 warnings in 416.00s (0:06:56) ===========

$ shellcheck --severity=error scripts/bash/common.sh
(no output, exit 0)

$ uvx ruff@0.15.0 check tests/test_resolve_template_python_parity.py src
All checks passed!
```

## AI disclosure

This PR was written by Claude Code, Anthropic's agentic coding CLI, based on
the issue's description and existing `SPECKIT_PYTHON` precedent in the
codebase. An independent reviewer agent blocked the first revision for
letting the bash/PowerShell twins honor `SPECKIT_PYTHON` without checking
it actually had PyYAML (a regression relative to the cited precedent); that
revision and subsequent review-response revisions fixed that gap and the
issues raised in later rounds, each adding a regression test reproducing
the specific issue and rerunning the full test suite, shellcheck, and ruff
before pushing.

The most recent revision (preserving non-recursive shared YAML aliases in
`_stringify_keys` while still catching true cycles) was generated by Claude
Code running Claude Sonnet 5 (model ID `claude-sonnet-5`). Earlier revisions
in this thread were also produced with Claude Code; the specific underlying
Claude model used for each of those earlier revisions was not recorded at
the time.

### DIFF
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -400,7 +400,8 @@ check_dir() { [[ -d "$1" && -n $(ls -A "$1" 2>/dev/null) ]] && echo "  ✓ $2" |
 
 _python3_command() {
     if [[ -n "${SPECKIT_PYTHON:-}" ]] && command -v "$SPECKIT_PYTHON" >/dev/null 2>&1 &&
-        "$SPECKIT_PYTHON" -c 'import sys; raise SystemExit(sys.version_info.major != 3)' >/dev/null 2>&1; then
+        "$SPECKIT_PYTHON" -c 'import sys; raise SystemExit(sys.version_info.major != 3)' >/dev/null 2>&1 &&
+        "$SPECKIT_PYTHON" -c 'import yaml' >/dev/null 2>&1; then
         printf '%s\n' "$SPECKIT_PYTHON"
     elif command -v python3 >/dev/null 2>&1 &&
         python3 -c 'import sys; raise SystemExit(sys.version_info.major != 3)' >/dev/null 2>&1; then
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -399,7 +399,10 @@ check_file() { [[ -f "$1" ]] && echo "  ✓ $2" || echo "  ✗ $2"; }
 check_dir() { [[ -d "$1" && -n $(ls -A "$1" 2>/dev/null) ]] && echo "  ✓ $2" || echo "  ✗ $2"; }
 
 _python3_command() {
-    if command -v python3 >/dev/null 2>&1 &&
+    if [[ -n "${SPECKIT_PYTHON:-}" ]] && command -v "$SPECKIT_PYTHON" >/dev/null 2>&1 &&
+        "$SPECKIT_PYTHON" -c 'import sys; raise SystemExit(sys.version_info.major != 3)' >/dev/null 2>&1; then
+        printf '%s\n' "$SPECKIT_PYTHON"
+    elif command -v python3 >/dev/null 2>&1 &&
         python3 -c 'import sys; raise SystemExit(sys.version_info.major != 3)' >/dev/null 2>&1; then
         printf '%s\n' "python3"
     elif command -v python >/dev/null 2>&1 &&
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -411,7 +411,7 @@ _python3_command() {
         printf '%s\n' "python"
     elif command -v py >/dev/null 2>&1 &&
         py -3 -c 'import sys' >/dev/null 2>&1; then
-        printf '%s\n' "py -3"
+        printf '%s\n' "py" "-3"
     else
         return 1
     fi
@@ -419,10 +419,9 @@ _python3_command() {
 
 _sorted_extension_ids() {
     local ext_dir="$1"
-    local python_spec
-    if python_spec=$(_python3_command); then
-        local -a python_cmd
-        read -r -a python_cmd <<< "$python_spec"
+    local -a python_cmd=()
+    mapfile -t python_cmd < <(_python3_command)
+    if [ "${#python_cmd[@]}" -gt 0 ]; then
         local py_stderr sorted_ids
         py_stderr=$(mktemp)
         if sorted_ids=$(SPECKIT_EXTENSIONS="$ext_dir" "${python_cmd[@]}" -c "
@@ -517,11 +516,8 @@ resolve_template() {
     local presets_dir="$repo_root/.specify/presets"
     if [ -d "$presets_dir" ]; then
         local registry_file="$presets_dir/.registry"
-        local python_spec=""
         local -a python_cmd=()
-        if python_spec=$(_python3_command); then
-            read -r -a python_cmd <<< "$python_spec"
-        fi
+        mapfile -t python_cmd < <(_python3_command)
         if [ -f "$registry_file" ] && [ "${#python_cmd[@]}" -gt 0 ]; then
             # Read preset IDs sorted by priority (lower number = higher precedence).
             # The python3 call is wrapped in an if-condition so that set -e does not
@@ -640,11 +636,8 @@ resolve_template_content() {
         local registry_file="$presets_dir/.registry"
         local sorted_presets=""
         local registry_parsed=false
-        local python_spec=""
         local -a python_cmd=()
-        if python_spec=$(_python3_command); then
-            read -r -a python_cmd <<< "$python_spec"
-        fi
+        mapfile -t python_cmd < <(_python3_command)
         if [ -f "$registry_file" ] && [ "${#python_cmd[@]}" -gt 0 ]; then
             if sorted_presets=$(SPECKIT_REGISTRY="$registry_file" "${python_cmd[@]}" -c "
 import json, re, sys, os
--- a/scripts/bash/common.sh
+++ b/scripts/bash/common.sh
@@ -420,7 +420,10 @@ _python3_command() {
 _sorted_extension_ids() {
     local ext_dir="$1"
     local -a python_cmd=()
-    mapfile -t python_cmd < <(_python3_command)
+    local _python_cmd_line
+    while IFS= read -r _python_cmd_line; do
+        python_cmd+=("$_python_cmd_line")
+    done < <(_python3_command)
     if [ "${#python_cmd[@]}" -gt 0 ]; then
         local py_stderr sorted_ids
         py_stderr=$(mktemp)
@@ -517,7 +520,10 @@ resolve_template() {
     if [ -d "$presets_dir" ]; then
         local registry_file="$presets_dir/.registry"
         local -a python_cmd=()
-        mapfile -t python_cmd < <(_python3_command)
+        local _python_cmd_line
+        while IFS= read -r _python_cmd_line; do
+            python_cmd+=("$_python_cmd_line")
+        done < <(_python3_command)
         if [ -f "$registry_file" ] && [ "${#python_cmd[@]}" -gt 0 ]; then
             # Read preset IDs sorted by priority (lower number = higher precedence).
             # The python3 call is wrapped in an if-condition so that set -e does not
@@ -637,7 +643,10 @@ resolve_template_content() {
         local sorted_presets=""
         local registry_parsed=false
         local -a python_cmd=()
-        mapfile -t python_cmd < <(_python3_command)
+        local _python_cmd_line
+        while IFS= read -r _python_cmd_line; do
+            python_cmd+=("$_python_cmd_line")
+        done < <(_python3_command)
         if [ -f "$registry_file" ] && [ "${#python_cmd[@]}" -gt 0 ]; then
             if sorted_presets=$(SPECKIT_REGISTRY="$registry_file" "${python_cmd[@]}" -c "
 import json, re, sys, os
--- a/scripts/powershell/common.ps1
+++ b/scripts/powershell/common.ps1
@@ -320,6 +320,10 @@ function Format-SpecKitCommand {
 # Find a usable Python 3 executable (python3, python, or py -3).
 # Returns the command/arguments as an array, or $null if none found.
 function Get-Python3Command {
+    if ($env:SPECKIT_PYTHON -and (Get-Command $env:SPECKIT_PYTHON -ErrorAction SilentlyContinue)) {
+        $ver = & $env:SPECKIT_PYTHON --version 2>&1
+        if ($ver -match 'Python 3') { return @($env:SPECKIT_PYTHON) }
+    }
     if (Get-Command python3 -ErrorAction SilentlyContinue) { return @('python3') }
     if (Get-Command python -ErrorAction SilentlyContinue) {
         $ver = & python --version 2>&1
--- a/scripts/powershell/common.ps1
+++ b/scripts/powershell/common.ps1
@@ -322,7 +322,10 @@ function Format-SpecKitCommand {
 function Get-Python3Command {
     if ($env:SPECKIT_PYTHON -and (Get-Command $env:SPECKIT_PYTHON -ErrorAction SilentlyContinue)) {
         $ver = & $env:SPECKIT_PYTHON --version 2>&1
-        if ($ver -match 'Python 3') { return @($env:SPECKIT_PYTHON) }
+        if ($ver -match 'Python 3') {
+            & $env:SPECKIT_PYTHON -c 'import yaml' *> $null
+            if ($LASTEXITCODE -eq 0) { return @($env:SPECKIT_PYTHON) }
+        }
     }
     if (Get-Command python3 -ErrorAction SilentlyContinue) { return @('python3') }
     if (Get-Command python -ErrorAction SilentlyContinue) {
--- a/scripts/python/common.py
+++ b/scripts/python/common.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import os
 import re
+import subprocess
 import sys
 from dataclasses import dataclass
 from pathlib import Path
@@ -380,6 +381,76 @@ def _validate_manifest_template_entry(entry: object) -> None:
         )
 
 
+class _DelegatedYAMLError(Exception):
+    """Raised when a SPECKIT_PYTHON-delegated manifest parse fails."""
+
+
+class _DelegatedYAML:
+    """``yaml.safe_load`` proxy that shells out to SPECKIT_PYTHON.
+
+    Used when this interpreter lacks PyYAML but SPECKIT_PYTHON names one
+    that has it (e.g. a `uv tool install` / `pipx` venv invisible to the
+    bare `python3` a script is launched with). See #4443.
+    """
+
+    YAMLError = _DelegatedYAMLError
+
+    def __init__(self, python_exe: str) -> None:
+        self._python_exe = python_exe
+
+    def safe_load(self, text: str) -> object:
+        try:
+            proc = subprocess.run(
+                [
+                    self._python_exe,
+                    "-c",
+                    "import sys, json, yaml; "
+                    "json.dump(yaml.safe_load(sys.stdin.read()), sys.stdout)",
+                ],
+                input=text,
+                capture_output=True,
+                text=True,
+                timeout=10,
+            )
+        except (OSError, subprocess.TimeoutExpired) as exc:
+            raise _DelegatedYAMLError(
+                f"SPECKIT_PYTHON could not parse the manifest: {exc}"
+            ) from exc
+        if proc.returncode != 0:
+            raise _DelegatedYAMLError(
+                proc.stderr.strip() or "SPECKIT_PYTHON could not parse the manifest"
+            )
+        try:
+            return json.loads(proc.stdout)
+        except json.JSONDecodeError as exc:
+            raise _DelegatedYAMLError(
+                f"SPECKIT_PYTHON returned invalid JSON: {exc}"
+            ) from exc
+
+
+def _import_yaml() -> object | None:
+    """Import PyYAML, delegating to SPECKIT_PYTHON if this interpreter lacks it."""
+    try:
+        import yaml
+
+        return yaml
+    except ImportError:
+        pass
+
+    python_override = os.environ.get("SPECKIT_PYTHON")
+    if not python_override:
+        return None
+    try:
+        probe = subprocess.run(
+            [python_override, "-c", "import yaml"], capture_output=True, timeout=10
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        return None
+    if probe.returncode != 0:
+        return None
+    return _DelegatedYAML(python_override)
+
+
 def _preset_template_layer(
     preset_dir: Path, template_name: str
 ) -> tuple[Path, str] | None:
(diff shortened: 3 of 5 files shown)


chelsealong:
Addressed all 5 findings from the Copilot review:

- `scripts/bash/common.sh`: `_python3_command` now emits one argv element per line (`mapfile -t`), so a `SPECKIT_PYTHON` path containing spaces is no longer split across `read -r -a`.
- `scripts/python/common.py`: the delegated-YAML subprocess now dumps with `default=str`, so non-JSON-native YAML values (e.g. an unquoted date) no longer crash the parse.
- `tests/test_resolve_template_python_parity.py`: the two SPECKIT_PYTHON tests now pick venv paths via a new `venv_python3_exe()` helper (POSIX `bin/python3` vs Windows `Scripts/python.exe`), and explicitly unset `SPECKIT_PYTHON` before building the baseline env. Added two new regression tests covering the spaces-in-path and non-JSON-YAML-value fixes.

Verified both new tests fail on the pre-fix code (`git checkout HEAD~1 -- <file>`) and pass after; full suite (`pytest tests/ -q`) is green at 7708 passed.

mnriem:
The re-review and CI both flag a real regression: your `mapfile -t` change in `_python3_command` (from the argv-splitting fix) uses a **Bash 4.0+ builtin, but macOS ships Bash 3.2** — CI shows `common.sh: line 640: mapfile: command not found` on macOS, which fails preset composition (`test_all_variants_resolve_requested_template[composed]`). Please replace `mapfile` with a 3.2-safe read loop (e.g. `while IFS= read -r line; do arr+=("$line"); done < <(...)`), and address the delegated-YAML transport/validation point the review raised. Re-request once CI is green across macOS/Windows.

chelsealong:
Addressed both points from the latest review:

- `scripts/bash/common.sh`: replaced all three `mapfile -t python_cmd < <(_python3_command)` calls with a Bash 3.2-compatible `while IFS= read -r ...; do ...; done` loop (macOS system Bash is 3.2 and lacks `mapfile`).
- `scripts/python/common.py`: `_DelegatedYAML.safe_load` now passes `encoding="utf-8"` (instead of `text=True`) and sets `PYTHONIOENCODING=utf-8` for the child process, so non-ASCII manifest metadata survives an ASCII-locale environment. It also replaced the blanket `default=str` JSON dump with a marker-based `default`/`object_hook` pair, so a validated field (e.g. `file`) that PyYAML parses into a non-JSON-native type (an unquoted date) still fails the `isinstance(str)` check under delegation, matching native `yaml.safe_load` behavior instead of silently passing as a string.

Added three regression tests to `tests/test_resolve_template_python_parity.py`, each confirmed to fail on the pre-fix code and pass after: `test_bash_resolves_composed_template_without_bash4_mapfile_builtin` (simulates Bash 3.2 by disabling the `mapfile` builtin), `test_python_variant_delegates_manifest_with_non_ascii_metadata_under_ascii_locale`, and `test_python_variant_rejects_delegated_manifest_with_non_string_validated_field`. Full suite: `pytest tests/ -q` → 7711 passed, 12 skipped. `shellcheck --severity=error scripts/bash/*.sh` and `ruff@0.15.0 check src tests` are clean.

chelsealong:
Addressed both points from the latest Copilot review:

- `scripts/python/common.py`: the delegated-YAML child now stringifies non-JSON-native mapping *keys* before `json.dump` (its `default=` hook only applies to values, not keys, so a manifest like `metadata: {2026-09-08: value}` raised `TypeError` under delegation even though the in-process parser accepts it). The child also catches `yaml.YAMLError` and prints a concise message instead of letting a raw Python traceback leak into `TemplateResolutionError` on a malformed manifest.
- `tests/test_resolve_template_python_parity.py`: reworked `test_bash_honors_speckit_python_path_containing_spaces` to symlink `sys.executable` into the spaced path (guaranteed PyYAML, unlike relying on `--system-site-packages` inheritance) and put a PyYAML-less venv on `PATH`, so the test can only pass if `_python3_command` actually selects the spaced override. Added regression tests for the mapping-key and traceback-leak fixes.

Confirmed both new regression tests fail on the pre-fix code and pass after. Full suite: `pytest tests/ -q` → 7713 passed, 12 skipped. `shellcheck --severity=error scripts/bash/*.sh` and `ruff@0.15.0 check tests/test_resolve_template_python_parity.py src` are clean.

mnriem:
Thanks @chelsealong — confirmed both are handled: the child now emits the concise `yaml.YAMLError` message instead of a raw traceback, and `_stringify_keys` correctly covers non-JSON-native mapping keys that `default=` misses. This is on me next: I'll trigger CI on the new head and review.

chelsealong:
Addressed both points from the latest Copilot review, plus the failing CI:

- `scripts/python/common.py`: `_stringify_keys` now tracks visited container ids, so a self-referential YAML alias (`metadata: &m {self: *m}`, which `yaml.safe_load` accepts natively) no longer causes a `RecursionError` under delegation. `_import_yaml`'s probe now also checks `sys.version_info.major == 3` before accepting `SPECKIT_PYTHON`, matching the bash/PowerShell twins, so a PyYAML-equipped Python 2 is rejected instead of failing later in the child process.
- Also fixed the CI failure on `test_bash_honors_speckit_python_path_containing_spaces` (all 6 matrix jobs): reproduced locally under `uv run pytest` — symlinking `sys.executable` from outside its own venv directory breaks Python's `pyvenv.cfg` discovery, so the symlinked interpreter silently loses access to the venv's PyYAML. Replaced the symlink with a shim script that `exec`s `sys.executable` directly (keeping `argv[0]` at its real, discoverable path), which also drops the need for `os.symlink` (addressing the earlier Windows-privilege concern on that same test).

Added `test_python_variant_delegates_manifest_with_recursive_yaml_alias` and `test_python_variant_rejects_speckit_python_override_without_python_3`, each confirmed to fail on the pre-fix code and pass after. Full suite: `uv run pytest tests/` → 7715 passed, 12 skipped. `ruff@0.15.0 check` is clean.

mnriem:
Thanks @chelsealong — the latest changes address the recursive-metadata example and remove the symlink dependency from the spaced-path test.


One parsing-parity issue remains in `_stringify_keys`: the global `seen` set treats ordinary shared aliases as cycles. I reproduced this with a template list anchored under metadata and referenced under `provides.templates`. Native parsing preserves the list; delegated parsing replaces the reference with `<non-native YAML value>`, which the resolver then rejects.


Please preserve non-recursive shared aliases while preventing recursive traversal, and add that case to the parity coverage.


Please also complete the disclosure with the specific agent/tool and model(s) used. Autonomous operation and extent are already stated, but “Claude/Anthropic” doesn’t identify the model.


chelsealong:
Addressed the shared-alias regression from the latest review:

- `scripts/python/common.py`: `_stringify_keys` now tracks an active recursion stack (adds a container's id before recursing into it, discards it in a `finally` once that subtree finishes) instead of a set that persists for the rest of the document. A second, unrelated reference to the same anchor (e.g. a templates list anchored under `metadata` and reused under `provides.templates`) is no longer mistaken for a cycle, while true self-references still get caught and replaced with the non-native marker.
- Added `test_python_variant_delegates_manifest_with_shared_non_recursive_alias`, confirmed to fail on the pre-fix code (`manifest templates must be a list`) and pass after. Full suite: `pytest tests/ -q` → 7716 passed, 12 skipped. `ruff check tests/test_resolve_template_python_parity.py src` is clean.

Also completed the AI disclosure in the PR body with the specific tool/model for this revision.

chelsealong:
Addressed the PYTHONPATH isolation issue from the latest Copilot review:

- `tests/parity_helpers.py`: `clean_env()` now also strips `PYTHONPATH` (a `--without-pip` venv still honors an inherited one), and a new `make_yaml_less_venv()` helper creates the venv and asserts the interpreter actually can't `import yaml`, replacing the ad-hoc creation blocks in `tests/test_resolve_template_python_parity.py`.
- Added `test_clean_env_strips_pythonpath`, confirmed to fail on the pre-fix `clean_env()` (`AssertionError: assert 'PYTHONPATH' not in {...}`) and pass after.

Full suite: `pytest tests/ -q` → 7717 passed, 12 skipped. `ruff check tests/test_resolve_template_python_parity.py tests/parity_helpers.py` is clean.

chelsealong:
Addressed the latest review: `_stringify_keys` in `scripts/python/common.py` now recurses into tuples too (PyYAML represents `!!omap`/`!!pairs` entries as tuples), so a nested non-JSON-native mapping key inside one is stringified instead of reaching `json.dump` raw and raising `TypeError`.

Added `test_python_variant_delegates_manifest_with_omap_metadata`, confirmed to fail on the pre-fix code (`TypeError: keys must be str, int, float, bool or None, not date`) and pass after. Full suite: `pytest tests/ -q` → 7718 passed, 12 skipped. `ruff check` (0.15.0) on the changed files is clean.

chelsealong:
Addressed both points from the latest Copilot review:

- `scripts/python/common.py`: `_preset_template_layer` now checks for `preset.yml` before importing/probing PyYAML, so a manifest-less preset no longer pays for a `SPECKIT_PYTHON` probe (or child-process spawn) it doesn't need.
- `tests/parity_helpers.py`: `clean_env()` now also strips an ambient `SPECKIT_PYTHON`, so tests like `test_all_variants_fail_when_yaml_parser_is_unavailable` aren't environment-dependent; tests exercising the override already set it explicitly.

Added `test_python_variant_skips_yaml_probe_for_manifest_less_preset` and `test_clean_env_strips_speckit_python`, both confirmed to fail on the pre-fix code and pass after. Full suite: `pytest tests/ -q` → 7720 passed, 12 skipped. `ruff check` (0.15.0) on the changed files is clean.

mnriem:
One naming adjustment before this lands: please use `SPECKIT_PYTHON_EXECUTABLE` as the canonical variable rather than adding further usage of the more ambiguous `SPECKIT_PYTHON`.

The value identifies a specific executable, and the name aligns with the existing `SPECKIT_INTEGRATION_<KEY>_EXECUTABLE` convention while leaving a clear `SPECKIT_PYTHON_*` namespace for future Python-related configuration.

Since `SPECKIT_PYTHON` is already consumed by the agent-context scripts, please retain it as a deprecated compatibility fallback, with `SPECKIT_PYTHON_EXECUTABLE` taking precedence when both are set. Apply that contract consistently across the Bash, PowerShell, and Python implementations and their tests.

Drafted for @mnriem by GitHub Copilot (model: GPT-5.6 Sol, autonomous); comment fully AI-drafted.

chelsealong:
Renamed to \`SPECKIT_PYTHON_EXECUTABLE\` as the canonical override, retaining \`SPECKIT_PYTHON\` as a deprecated fallback (still used by `update-agent-context.sh`) with the new name taking precedence when both are set. Applied across `scripts/bash/common.sh`, `scripts/powershell/common.ps1`, and `scripts/python/common.py`.

Added `test_all_variants_prefer_speckit_python_executable_when_both_set` (bash/python/PowerShell) and `test_python_variant_prefers_speckit_python_executable_over_deprecated_alias` covering the precedence, plus extended `test_clean_env_strips_speckit_python` for the new var. Confirmed both new precedence tests fail on the pre-rename code (`git checkout HEAD~1 -- scripts/bash/common.sh scripts/powershell/common.ps1 scripts/python/common.py`) and pass after. Full suite: `pytest tests/ -q` → 7722 passed, 12 skipped. `shellcheck --severity=error scripts/bash/common.sh` and `ruff check` (0.15.0) on the changed files are clean.

chelsealong:
Fixed the Windows CI failure (\`test_python_variant_delegates_manifest_with_omap_metadata\` on windows-latest, which cascaded into the macos/windows-3.14 jobs getting canceled by fail-fast).

Root cause: \`install_composition_stack()\` in \`tests/parity_helpers.py\` wrote fixture template content with \`Path.write_text(..., encoding="utf-8")\`, whose default \`newline=None\` translates \`\n\` to \`os.linesep\` on write — baking CRLF into the files on Windows. \`resolve_template_content()\` reads composed content back with \`read_bytes().decode("utf-8")\` (no newline translation, matching how production code writes resolved templates via \`write_bytes()\`), so the CRLF survived into \`TEMPLATE_CONTENT\` and failed the byte-for-byte comparison against the LF-only \`expected\` string.

Switched those two fixture writes to \`write_bytes(content.encode("utf-8"))\`, which cannot be translated by the platform regardless of OS. Verified full parity suites pass locally (Linux, where the bug can't reproduce since \`os.linesep\` is already \`\n\`); the fix itself is platform-independent by construction (\`write_bytes\` never applies newline translation, unlike \`write_text\`'s default). Pushed as 8bd0e84.

mnriem:

chelsealong:
Investigated the CI failure and current lint status on this PR's head (8bd0e84):

- `ruff` and `shellcheck` both currently pass as separate CI checks, and I confirmed locally against this PR's changed files: `ruff check tests/test_resolve_template_python_parity.py tests/parity_helpers.py src scripts/python/common.py` → all checks passed; `shellcheck --severity=error scripts/bash/common.sh` → clean. No lint errors are present.
- The only failing check is `pytest (windows-latest, 3.14)`, and it's a single test: `tests/extensions/test_extension_agent_context.py::TestBundledUpdaterPathValidation::test_powershell_script_discovers_nested_plan`. It fails with `subprocess.TimeoutExpired` spawning `powershell.exe` after 30s — not an assertion failure. That test exercises `extensions/agent-context/scripts/powershell/update-agent-context.ps1`, which this PR does not modify; this PR's diff is limited to `scripts/bash/common.sh`, `scripts/powershell/common.ps1`, `scripts/python/common.py`, `tests/parity_helpers.py`, and `tests/test_resolve_template_python_parity.py`.
- Ran the full suite locally: `pytest tests/ -q` → 7722 passed, 12 skipped (matches the passing macOS/Ubuntu/Windows-3.13 jobs' counts on this same commit), with `tests/test_resolve_template_python_parity.py` at 58/58.

This looks like a Windows-runner timeout unrelated to this PR's changes rather than something introduced here (the test only runs on Windows and is skipped elsewhere, so it can't be reproduced in this environment). Given the evidence, I haven't changed any code — let me know if a re-run reproduces it and I'll dig further.

Posted on behalf of @chelsealong by Claude Code (model: claude-sonnet-5, autonomous).

chelsealong:
Addressed both findings from the latest Copilot review:

- `scripts/python/common.py`: `_DelegatedYAML`'s child now drops every manifest field except `provides.templates` before normalizing/serializing, since that's all `_preset_template_layer` consumes. Previously an ignored `metadata` value built from shared (non-recursive) YAML aliases could expand exponentially in the JSON round trip even though the cycle-detection stack alone let it terminate.
- `tests/test_resolve_template_python_parity.py` / `tests/parity_helpers.py`: `test_all_variants_fall_back_when_speckit_python_lacks_pyyaml` now puts a deterministic `python3` shim (execs the pytest interpreter) first on PATH via a new `make_python3_path_shim()` helper, instead of relying on the host's own `python3` having PyYAML.

Added `test_python_variant_delegates_manifest_with_exponential_alias_dag`, confirmed to fail on the pre-fix code (7.9s, near the 10s child timeout, vs <1s after) and pass after. Also reproduced the PATH-dependent failure directly: with a PyYAML-less `python3` shimmed first on PATH, the fallback test failed on the pre-fix version and passes with the new shim. Full suite: `pytest tests/ -q` → 7723 passed, 12 skipped. `ruff check` (0.15.0) on the changed files is clean.

Posted on behalf of @chelsealong by Claude Code (model: claude-sonnet-5, autonomous).

chelsealong:
Addressed the latest Copilot finding: `_only_templates` pruned unused top-level manifest fields but still passed every field of each `provides.templates` entry through to recursive normalization, so an ignored per-entry field (e.g. `description`) built from a non-recursive YAML alias DAG could still expand exponentially and hit the 10s child timeout.

- `scripts/python/common.py`: the delegated child now prunes each template entry to just `type`, `name`, `file`, and `strategy` (the fields the resolver reads) before `_stringify_keys` normalizes it.
- `tests/test_resolve_template_python_parity.py`: added `test_python_variant_delegates_manifest_with_exponential_alias_dag_in_entry`, confirmed to fail on the pre-fix code (times out after 10s) and pass after (~0.3s). Full suite: `pytest tests/ -q` → 7724 passed, 12 skipped. `ruff check` (0.15.0) on the changed files and `shellcheck --severity=error scripts/bash/common.sh` are clean.

Posted on behalf of @chelsealong by Claude Code (model: claude-sonnet-5, autonomous).

Issue #4227, opened 2026-08-20
[Preset]: Add Inventory Alignment (inventory-alignment)


Yash-Chindam:
### Preset ID

inventory-alignment

### Preset Name

Inventory Alignment

### Version

0.1.0

### Description

Classifies each requirement against a read-only inventory of live IDs before writing, so reworded requirements are updated instead of duplicated.

### Author

Yash Chindam

### Repository URL

https://github.com/Yash-Chindam/spec-kit-inventory-alignment

### Download URL

https://github.com/Yash-Chindam/spec-kit-inventory-alignment/releases/download/v0.1.0/inventory-alignment.zip

### Documentation URL

https://github.com/Yash-Chindam/spec-kit-inventory-alignment/blob/main/inventory-alignment/README.md

### License

MIT

### Required Spec Kit Version

>=0.9.0

### Required Extensions (optional)

`speckit-inventory` >=0.1.0 — submitted separately alongside this one.

Note that `preset.yml`'s `requires:` can only declare `speckit_version`, so this dependency cannot currently be declared formally in the manifest. The wrappers degrade safely: without the extension installed, the core workflow runs unchanged and the inventory checks are skipped. @mnriem flagged this same gap in https://github.com/github/spec-kit/issues/4164#issuecomment-5348616852 as a small, separable core enhancement worth tracking on its own issue.

### Templates Provided

1 template — `tasks-template` (`append` strategy), documenting the optional `covers:` field that links a task to the requirement IDs it implements.

### Commands Provided

2 commands, both using the `wrap` strategy with the `{CORE_TEMPLATE}` placeholder so they fall through to the original command body:

- `speckit.specify` — prepends a pre-pass that loads the inventory and classifies each proposed requirement as already-true / edit-existing / conflict / genuinely-new before writing.
- `speckit.analyze` — prepends a pre-pass that loads the complete live ID set as an input to cross-artifact coverage and conflict review.

### Number of Scripts (optional)

0

### Tags

inventory, alignment, requirements, traceability, workflow

### Key Features

Filed at @mnriem's request in #4164 — this is the preset half of the extension + preset pair proposed in https://github.com/github/spec-kit/issues/4164#issuecomment-5348616852. The companion extension `speckit-inventory` is submitted separately.

[...]

Issue #4226, opened 2026-08-20
[Extension]: Add Spec Inventory (speckit-inventory)


Yash-Chindam:
### Extension ID

speckit-inventory

### Extension Name

Spec Inventory

### Version

0.1.0

### Description

Read-only inventory of live requirement and task IDs, with focused per-task context packs instead of whole-file dumps.

### Author

Yash Chindam

### Repository URL

https://github.com/Yash-Chindam/spec-kit-inventory-alignment

### Download URL

https://github.com/Yash-Chindam/spec-kit-inventory-alignment/releases/download/v0.1.0/speckit-inventory.zip

### License

MIT

### Homepage (optional)

https://github.com/Yash-Chindam/spec-kit-inventory-alignment

### Documentation URL (optional)

https://github.com/Yash-Chindam/spec-kit-inventory-alignment/blob/main/speckit-inventory/README.md

### Changelog URL (optional)

https://github.com/Yash-Chindam/spec-kit-inventory-alignment/blob/main/speckit-inventory/CHANGELOG.md

### Required Spec Kit Version

>=0.9.0

### Required Tools (optional)

None. Python 3.11+ only, no third-party dependencies.

### Number of Commands

2

### Number of Hooks (optional)

2

### Tags

inventory, requirements, context, traceability, alignment

### Key Features

Filed at @mnriem's request in #4164 — this is the read-only primitive half of the extension + preset pair proposed in https://github.com/github/spec-kit/issues/4164#issuecomment-5348616852. The companion preset `inventory-alignment` is submitted separately.

- **Regenerates every live ID on each run.** Extracts all `FR-`, `NFR-`, `SC-`, `AC-`, and `T-` IDs from the existing `spec.md` and `tasks.md` and prints JSON. Nothing is persisted, so there is no sidecar and no second source of truth that can drift out of sync with the Markdown.
- **Read-only.** The script contains no write calls. It never edits `spec.md`, `tasks.md`, or creates files.
- **Focused context packs.** `speckit.speckit-inventory.context` returns one task plus only the requirements that task covers, instead of loading whole artifacts into context.
- **Dangling-reference detection.** Each record reports its `source` artifact. `spec.md` is scanned first and wins as the definition site, so any ID whose `source` is `tasks.md` is a reference to a requirement that was never defined.
- **Lenient task-ID matching.** `T014`, `T14`, `t14`, and `TASK-14` all resolve to the same record.
[...]

PR #4449, opened 2026-09-05
fix(catalog): repoint speckit-inventory and inventory-alignment at live release artifacts

Yash-Chindam:
## Problem

Both community entries I submitted in #4226 and #4227 point at release assets under a repository that no longer exists:

```json
"download_url": "https://github.com/Yash-Chindam/spec-kit-inventory-alignment/releases/download/v0.1.0/speckit-inventory.zip"
```

`repository`, `homepage`, `documentation`, `changelog`, and `download_url` all 404. Installation fails at the download, before `verify_archive_sha256` is ever reached:

```
specify extension add speckit-inventory
specify preset add inventory-alignment
```

This is my fault, and I'm sorry it reached the catalog in that state.

## Fix

I've restored the source in a standalone repository and cut [v0.1.1](https://github.com/Yash-Chindam/spec-kit-inventory-alignment/releases/tag/v0.1.1). The v0.1.0 archives are **unrecoverable** — nothing reproduces their pinned digests — so v0.1.1 supersedes them and the pinned SHA-256 values change with it.

| Entry | Was | Now |
|---|---|---|
| `speckit-inventory` | `v0.1.0`, `9ebf004e…` | `v0.1.1`, `df735fe7…` |
| `inventory-alignment` | `v0.1.0`, `8ea62813…` | `v0.1.1`, `227db99a…` |

Release artifacts are built with sorted members, a fixed timestamp, and two canonical permission modes — mirroring `bundler/services/packager.py` — and `.gitattributes` pins `eol=lf` so a Windows checkout cannot yield a different digest. A clean clone reproduces both digests byte for byte.

## One correctness fix beyond the URLs

The preset entry declares `requires.extensions`, but claimed `speckit_version: ">=0.9.0"`. That floor is wrong: `requires.extensions` first shipped in **v1.0.4** (#4250) — `v1.0.3` has no `_validate_requires_extensions`, `v1.0.4` does. A v0.9 CLI would install the preset and ignore the dependency it declares, which is the exact silent-no-op failure #4250 set out to prevent. Raised to `>=1.0.4`.

## Verification

- Both `download_url`s return **HTTP 200**, and the downloaded bytes hash to the pinned `sha256` (checked against the live URLs, not local files).
- `CatalogEntry.from_dict()` parses both entries.
- `tests/contract/test_catalog_schema.py` — **45 passed**.
- Both bundle manifests validate against `PresetManifest` / `ExtensionManifest`, and the extension's own suite passes (9 tests).

Diff is 9 lines across the two catalogs; no other entry is touched.

### DIFF
--- a/extensions/catalog.community.json
+++ b/extensions/catalog.community.json
@@ -4790,9 +4790,9 @@
       "id": "speckit-inventory",
       "description": "Read-only inventory of live requirement and task IDs, with focused per-task context packs instead of whole-file dumps.",
       "author": "Yash Chindam",
-      "version": "0.1.0",
-      "download_url": "https://github.com/Yash-Chindam/spec-kit-inventory-alignment/releases/download/v0.1.0/speckit-inventory.zip",
-      "sha256": "9ebf004ef6494323f6dccfab2554a04898c9e92bc7f25e0638b5aee916566e7f",
+      "version": "0.1.1",
+      "download_url": "https://github.com/Yash-Chindam/spec-kit-inventory-alignment/releases/download/v0.1.1/speckit-inventory.zip",
+      "sha256": "df735fe7c7ca7afe1ac279ffe5768061a29710a9d61956f90ee2cf2d66da04be",
       "repository": "https://github.com/Yash-Chindam/spec-kit-inventory-alignment",
       "homepage": "https://github.com/Yash-Chindam/spec-kit-inventory-alignment",
       "documentation": "https://github.com/Yash-Chindam/spec-kit-inventory-alignment/blob/main/speckit-inventory/README.md",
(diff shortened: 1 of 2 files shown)


mnriem:
Thanks for jumping on this, and no worries — the fix is right, but like other catalog changes it should go through the **submission-issue flow** rather than a direct edit of `catalog.community.json`. Since v0.1.1 changes the pinned digests, it's effectively an update and needs the same validation. Please re-submit v0.1.1 as two submissions — one `[Extension]` for `speckit-inventory`, one `[Preset]` for `inventory-alignment` — and the workflow will verify the new archives/digests and regenerate the catalog entries. Closing in favor of those.

PR #4460, opened 2026-09-07
refactor(workflows): let the evaluator report its own leaves (#4274)

ntdatt812:
Implements #4274, which was the write-up you asked for at the end of #4230.

## The problem

`_unresolvable_term` answers one question — *does every operand in this condition resolve to something?* — by walking the expression itself: filters, then `or`/`and`/`not`, then comparisons, then list literals, down to the leaves.

That walk is a second implementation of the parsing in `_evaluate_simple_expression`, kept in step with it by hand. Two helpers exist purely to restate rules the evaluator already has, and both say so:

- `_looks_numeric` — *"Mirror the evaluator's numeric literal test exactly."* A bare `float()` accepts `1e3`; the evaluator does not.
- `_is_literal` — *"Mirror the evaluator's literal tests exactly."* `startswith`/`endswith` accepts `'a' 'b'`; the evaluator does not.

When the two drift, nothing breaks loudly. The gate keeps answering, just wrongly, and the wrong answer is a paste-ready correction that silently inverts a condition. Seven of the nine findings in #4230 were that same defect in different clothes — the gate disagreeing with the evaluator about *where the operands are*. Each round fixed one shape; nothing stopped a tenth.

## The change

`_evaluate_simple_expression` has exactly one place where a substring stops being grammar and becomes a name to resolve — its final line, `_resolve_dot_path`. Literals return before it; operands, filter arguments and list elements all arrive there by construction. So let the evaluator report what it reaches:

```python
    # Variable reference (dot-path)
    sink = _leaf_sink.get()
    if sink is not None:
        sink.append(expr)
    return _resolve_dot_path(namespace, expr)
```

`_collect_leaves` runs the probe `_evaluator_rejects` already uses, with the sink armed. `_unresolvable_leaf` keeps only the namespace rules — root membership, path-segment shape, and the `item` index narrowing from the last round of #4230. The gate now contains no grammar at all.

A `ContextVar` rather than a module global, so concurrent probes cannot append into each other's list; it is `None` outside a probe, so a normal evaluation costs one `.get()`.

## Two properties this rests on

Both are asserted rather than assumed, because if either changed the gate would go quietly blind rather than fail:

**`or`/`and` are not short-circuited.** `_evaluate_simple_expression` evaluates both sides and only then combines them, so a leaf is recorded whatever the other side is worth. `test_both_sides_of_a_boolean_are_reported` pins it.

**A probe run can raise on its own placeholder values**, which is what `_evaluator_rejects` sorts out. The leaves seen before that point are real — the evaluator reached them — so they are kept rather than discarded. Discarding them would lose `bogus` in `inputs.tags | join(bogus)`, which is the filter-argument case an earlier round of #4230 had to add by hand. `test_leaves_seen_before_a_probe_error_are_kept` pins it.

## Verification

`expressions.py`: **109 lines removed, 84 added.**

**All 336 existing tests pass unchanged**, including the 20 cases of `test_operands_must_be_literals_or_known_paths` that took eight rounds to get right. That is the main evidence: the new gate agrees with the old one on every shape review found, without knowing about any of them.

One test changed rather than passed: `test_literal_test_mirrors_the_evaluator` tested the mirror, and the mirror is gone. It becomes `test_literal_handling_comes_from_the_evaluator` and asserts the same knowledge — `1e3` is not a number to the evaluator, `'a' 'b'` is not one literal — through `_unresolvable_term`. That is the property that actually mattered; the old test could pass while the two had drifted.

Five tests added for the mechanism itself. Checked by breaking it:

| reverted | killed |
|---|---|
| evaluator stops reporting its leaves | **38 tests**, including the whole `_unresolvable_term` suite |
| discard the leaves seen before a probe error | the filter-argument tests |
| drop the `item[0]` narrowing | `test_an_indexed_item_root_keeps_the_correction` |
| leave the sink armed after a probe | the two sink-hygiene tests |

`uvx ruff@0.15.0 check src tests` — clean. Wider run (`tests/unit`, `test_workflows.py`, `test_extensions.py`): 1988 passed, and the set of failing test names is identical before and after — 24 symlink tests that cannot run unprivileged on Windows.

---

## Update: the segment grammar was still duplicated, and #4416 / #4417 (2026-09-09)

`c6c3ef7` finishes what the first commit started. The gate had stopped restating the *operator* grammar, but it still restated the shape of a **path segment**: `_PATH_SEGMENT` and an inline `re.fullmatch` both described the index form that `_resolve_dot_path` matches with its own regex — three copies of one rule, kept in step by hand. That form is now named once as `_INDEXED_SEGMENT` beside `_resolve_dot_path`, and the gate asks it. The regex is copied verbatim, so behaviour is unchanged; what changes is that widening indexing now reaches the gate for free.

That was not cosmetic. I measured it before writing it:

| on this branch, before `c6c3ef7` | evaluator | gate |
|---|---|---|
| `task_list[-1].file` | `None` | rejects — `'task_list[-1]' is not a valid path segment` |
| `(inputs.a or inputs.b) and inputs.c` | `False` | rejects |

**Answering your question directly.** I took the *evaluator-side hunks only* of #4416 and #4417 — no edit to `_unresolvable_term`, `_PATH_SEGMENT` or anything else in the gate — and applied them on top of this branch:

| case | evaluator | gate |
|---|---|---|
| `task_list[-1].file` | `'b.md'` | **accepts** |
| `task_list[0].file` | `'a.md'` | accepts |
| `(inputs.a or inputs.b) and inputs.c` | `True` | **accepts** |
| `(inputs.n)` | `5` | **accepts** |
| `inputs.a or inputs.b and inputs.c` | `True` | accepts |

So yes: once this lands, both of @NgoQuocViet2001's PRs reduce to their evaluator hunks. The gate-side changes they each carry — #4416's widened `_PATH_SEGMENT` and `indexed_root`, #4417's "mirror the evaluator's group unwrapping" block in `_unresolvable_term` — become unnecessary, because the gate no longer has an opinion of its own to keep in sync. Their diffs shrink and stop touching the region this PR rewrites, which should make the rebase mechanical rather than a merge argument.

One correction to my own framing, since it matters for sequencing: **this refactor does not fix their bugs.** `task_list[-1]` still resolves to `None` on this branch and `(a or b) and c` still reads false — those are evaluator defects and their PRs are what fix them. What this branch changes is that the gate now agrees with the evaluator instead of holding a second opinion, so their one-line evaluator fixes are sufficient on their own.

`dd4b723` adds two regression tests for exactly that property, because it is easy to claim and easy to lose — a fresh copy of the grammar in the gate would keep every other test green. One patches `_INDEXED_SEGMENT` and asserts the gate follows; the other makes the evaluator stop treating a group as a leaf and asserts the gate stops checking it. I verified both by reintroducing the drift (giving the gate its own segment regex again), which fails the first with the real message rather than an import error. I deliberately did *not* mirror their test bodies: asserting `task_list[-1] == 'b.md'` here would be a test for a fix this PR does not contain, and would pass or fail based on whether their PR is present.

`tests/unit/test_condition_expression_block.py`: 343 passed. `tests/test_workflows.py` + that file: the set of failing test names is byte-identical before and after these two commits — 20 symlink tests that cannot run unprivileged on Windows.

## AI disclosure

Per CONTRIBUTING: this pull request was developed with AI assistance — I used Claude Code as a coding agent for the code, the tests and the measurements above, reviewing and directing it throughout, and the experiment applying #4416/#4417's hunks was run and checked by me. This comment and the PR body were also written with that assistance. Apologies for the omission on the original submission; it was an oversight, not an attempt to hide it.

### DIFF
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -560,8 +560,27 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
                 f"operand in its own expression instead."
             )
         value = _evaluate_simple_expression(head, namespace)
+        sink = _leaf_sink.get()
         for segment in segments[1:]:
-            value = _apply_filter(value, segment.strip(), namespace)
+            if sink is None:
+                value = _apply_filter(value, segment.strip(), namespace)
+                continue
+            # Probing. A filter handed a placeholder can raise on it -- from_json
+            # on a mapping is the common one -- and letting that end the walk
+            # hides every leaf further along the chain, which is the one thing
+            # this collection exists to report: `inputs.blob | from_json |
+            # contains(bogus)` recorded `inputs.blob` and stopped, so `bogus`
+            # was never offered to _unresolvable_leaf. _apply_filter evaluates a
+            # filter's argument before it can raise on the value, so the leaves
+            # of the failing segment are already recorded when we get here.
+            # Carry a fresh placeholder so the next filter sees the same kind of
+            # unknown the namespace hands out. Real evaluation is untouched: the
+            # sink is armed only by _collect_leaves, and _evaluator_rejects runs
+            # its own probe without it, so a mis-wired filter is still reported.
+            try:
+                value = _apply_filter(value, segment.strip(), namespace)
+            except Exception:  # noqa: BLE001 - probe values, not the author's text
+                value = _ProbeNamespace()
         return value
 
     # Boolean operators — parse 'or' first (lower precedence) so that
--- a/src/specify_cli/workflows/expressions.py
+++ b/src/specify_cli/workflows/expressions.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import json
 import re
+from contextvars import ContextVar
 from typing import Any
 
 
@@ -479,6 +480,11 @@ def _apply_filter(value: Any, filter_expr: str, namespace: dict[str, Any]) -> An
 # evaluator will actually split on.
 _COMPARISON_OPERATORS = ("!=", "==", ">=", "<=", ">", "<", " not in ", " in ")
 
+# Set only while `_collect_leaves` probes an expression; None everywhere else, so
+# a normal evaluation costs one `.get()`. A ContextVar rather than a module global
+# so concurrent probes cannot append into each other's list.
+_leaf_sink: ContextVar[list[str] | None] = ContextVar("_leaf_sink", default=None)
+
 
 def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
     """Evaluate a simple expression against the namespace.
@@ -627,7 +633,12 @@ def _evaluate_simple_expression(expr: str, namespace: dict[str, Any]) -> Any:
         ]
         return items
 
-    # Variable reference (dot-path)
+    # Variable reference (dot-path). This is the one place a substring stops being
+    # grammar and becomes a name to resolve, so it is where a probe can learn what
+    # the evaluator will actually look up. Literals have all returned above.
+    sink = _leaf_sink.get()
+    if sink is not None:
+        sink.append(expr)
     return _resolve_dot_path(namespace, expr)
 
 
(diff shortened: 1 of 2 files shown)


mnriem:
Thanks — this is the right root-cause fix: having the evaluator report its own leaves instead of `_unresolvable_term` re-implementing the grammar is exactly what removes the drift. Two things: (1) please add the **AI-disclosure** per CONTRIBUTING — the body has none. (2) Coordination: #4416 and #4417 (by [[@NgoQuocViet2001](https://github.com/NgoQuocViet2001)](https://github.com/NgoQuocViet2001)) are point-fixes to the *same* `_unresolvable_term` grammar duplication this PR deletes, touching the same file. I'd like to land this refactor first and then have those two rebase/verify on top (their specific cases should be covered once the gate stops re-implementing the grammar) — could you confirm your refactor handles the negative-index and parenthesised-group cases they fixed, ideally with tests mirroring theirs? I'll sequence the merges accordingly.

ntdatt812:
Thanks — both points addressed, and the second one turned up something I had missed in my own PR.

**(1) AI disclosure** — added to the PR body, and it applies to this comment too. I used Claude Code as a coding agent for the code, the tests and the measurements below, reviewing and directing throughout. Sorry for the omission on the original submission; an oversight rather than an attempt to hide it.

**(2) The coordination question.** I did not want to answer this from reading, so I applied the *evaluator-side hunks only* of #4416 and #4417 on top of this branch — no edit to `_unresolvable_term`, `_PATH_SEGMENT`, or anything else in the gate — and measured:

| case | evaluator | gate |
|---|---|---|
| `task_list[-1].file` | `'b.md'` | **accepts** |
| `(inputs.a or inputs.b) and inputs.c` | `True` | **accepts** |
| `(inputs.n)` | `5` | **accepts** |

So the answer is yes, but only after a fix I had to add. **My refactor as you first saw it did not cover the negative-index case.** The gate had stopped restating the operator grammar, but it still restated the shape of a *path segment* — `_PATH_SEGMENT` plus an inline `fullmatch`, both describing the index form `_resolve_dot_path` matches with its own regex. Three copies of one rule. So `task_list[-1]` was still rejected by the gate with `'task_list[-1]' is not a valid path segment` even with the evaluator fixed. `c6c3ef7` names that form once as `_INDEXED_SEGMENT` beside `_resolve_dot_path` and has the gate ask it; the regex is copied verbatim, so behaviour is unchanged.

The parenthesised case needed nothing — it already worked, because once the evaluator stops treating a group as a leaf, the leaf sink stops reporting it.

**One correction to my own framing**, since it bears on your sequencing: this refactor does *not* fix their bugs. `task_list[-1]` still resolves to `None` on this branch and `(a or b) and c` still reads false. Those are evaluator defects and their PRs are what fix them. What lands here is that the gate stops holding a second opinion, so their evaluator hunks become sufficient on their own — #4416's widened `_PATH_SEGMENT`/`indexed_root` and #4417's "mirror the evaluator's group unwrapping" block can both be dropped, which also takes their diffs out of the region this PR rewrites.

On tests: `dd4b723` adds two, but they pin the *property* rather than mirror their cases. One patches `_INDEXED_SEGMENT` and asserts the gate follows; the other makes the evaluator stop treating a group as a leaf and asserts the gate stops checking it. I verified both by reintroducing the drift — giving the gate its own segment regex again fails the first with the real message, not an import error. I deliberately did not copy their assertions: `task_list[-1] == 'b.md'` here would be a test for a fix this PR does not contain, and would pass or fail depending on whether their branch is present. Their own tests are the right home for those, and they should keep passing unchanged on top of this.

Happy to rebase whenever suits the order you pick.

mnriem:
Thanks — the disclosure's added, and I really appreciate you actually *running* the #4416/#4417-hunks experiment rather than just asserting compatibility. That settles the sequencing: I'll land this refactor first, then #4416/#4417 rebase down to just their evaluator hunks. No worries on the original disclosure omission. Approving the CI run; once it's green I'll review for merge.

mnriem:
The re-review found a real correctness bug in the refactor: catching the first probe-value exception stops leaf collection, so later valid operands vanish — e.g. `inputs.blob | from_json | contains(bogus)` records only `inputs.blob`. Since the whole point here is the evaluator reporting *all* its leaves, this needs fixing before merge (probe failures should be recorded and traversal continue, not halt). Heads-up on impact: this PR is the keystone for the `expressions.py` cluster — I'm planning to land it first and then have #4416/#4417 rebase to their evaluator-only hunks, so getting the leaf collection right here unblocks all of them. Re-request once addressed.

ntdatt812:
Fixed in `d829b7e`. You were right, and it was worse than a gap in the new code: it was a regression against `main`.

*AI disclosure, per CONTRIBUTING: I used Claude Code as a coding agent for this fix, its tests and the measurements below, reviewing and directing throughout.*

## What it was

`_collect_leaves` ran the probe inside a single `try/except`, so the first exception a probe value raised ended the whole walk. `from_json` receives the probe's placeholder mapping and raises, and `contains(bogus)` was never reached.

The hand-written walk this PR deletes read `bogus` straight out of its own grammar rules, so the refactor *lost* coverage rather than merely failing to add it:

| `inputs.blob \| from_json \| contains(bogus)` | `_unresolvable_term` |
| --- | --- |
| `origin/main` | reports `'bogus'` |
| this PR, as you reviewed it | **missed** |
| this PR + `d829b7e` | reports `'bogus'` |

## The fix

Carry on past a failing filter, but only while the sink is armed:

```python
value = _evaluate_simple_expression(head, namespace)
sink = _leaf_sink.get()
for segment in segments[1:]:
    if sink is None:
        value = _apply_filter(value, segment.strip(), namespace)
        continue
    try:
        value = _apply_filter(value, segment.strip(), namespace)
    except Exception:  # noqa: BLE001 - probe values, not the author's text
        value = _ProbeNamespace()
```

Two things make this safe rather than a broad swallow:

- `_apply_filter` evaluates a filter's **argument** before it can raise on the value, so the failing segment's own leaves are already recorded when the handler runs. Nothing is lost by continuing.
- The sink is armed only by `_collect_leaves`. `_evaluator_rejects` runs its own probe **without** it, so a mis-wired filter is still reported as a rejection, and a real evaluation still raises rather than quietly returning the unfiltered value — which is exactly what `_apply_filter` refuses to do.

A fresh `_ProbeNamespace()` goes into the next filter, so it sees the same kind of unknown the namespace hands out.

## Tests

Three added to `tests/unit/test_condition_expression_block.py`, next to the existing sink tests:

- the chain keeps walking past a probe error (your exact expression, asserting the full leaf list, not just membership);
- **two** failing links do not hide a third's leaf — `... | from_json | map(bogus) | join(alsobogus)`;
- the carry-on is probe-only: a real `evaluate_expression` still raises on both `from_json` against non-JSON and an unknown filter, and `_evaluator_rejects` still reports the latter.

Mutation, three ways:

| mutation | result |
| --- | --- |
| re-raise on a probe error (the bug you found) | **2 fail** |
| never arm the carry-on | **2 fail** |
| arm it for *every* evaluation, not just the probe | **16 fail**, including the scoping test |

The third one matters most: without it the fix could have been a blanket swallow that passed the first two.

`tests/unit/` + `tests/test_workflows.py`: 1475 passed. The 22 failures there are the `TestWorkflowCliAlignment` symlink tests, which need a privilege this Windows box does not have — identical counts (8 failed / 177 passed in that class) with `origin/main`'s `expressions.py` swapped in, so they are unrelated to this branch.

Ready for re-review whenever suits you. Happy to keep this as a separate commit or squash it into the refactor, whichever you prefer for the `expressions.py` sequencing with #4416/#4417.

mnriem:
Thanks for your contributions here, @ntdatt812 — a quick note on review prioritization. You currently have **5 open pull requests**, above the three-open-PR guideline in [CONTRIBUTING](https://github.com/github/spec-kit/blob/main/CONTRIBUTING.md). This isn't a freeze, and it doesn't hold up this PR: #4460 is in good shape — green CI and an approving review — and the `author-over-cap` label here is just a tracking marker for your overall open count, not a block on merging. Beyond three open PRs, additional submissions may be placed behind other contributors' work, so the most effective path is to **consolidate where you can, or tell us the few you'd most like prioritized**. Merging #4460 will bring you back toward the guideline.

mnriem:
Thank you!

Issue #4439, opened 2026-09-04
[Extension]: Update OKF Knowledge Bundle Generator to v0.5.0


alexcpn:
### Extension ID

okf

### Extension Name

OKF Knowledge Bundle Generator

### Version

0.5.0

### Description

Generates and maintains an Open Knowledge Format (OKF v0.1) knowledge bundle from a source-code repository, mining git history for rationale and parking what it cannot verify as open questions.

### Author

Alex Punnen

### Repository URL

https://github.com/alexcpn/speckit_okf

### Download URL

https://github.com/alexcpn/speckit_okf/archive/refs/tags/v0.5.0.zip

### License

MIT

### Homepage (optional)

https://github.com/alexcpn/speckit_okf

### Documentation URL (optional)

https://github.com/alexcpn/speckit_okf/blob/main/README.md

### Changelog URL (optional)

https://github.com/alexcpn/speckit_okf/blob/main/CHANGELOG.md

### Required Spec Kit Version

>=0.12.0

### Required Tools (optional)

- bash - required
- python3 (>=3.9) - required
- git - optional (used for churn ranking, history mining and incremental update; everything else works without it)

### Number of Commands

4

### Number of Hooks (optional)

0

### Tags

knowledge, okf, documentation, metadata, catalog

### Key Features

- Generates a conformant OKF v0.1 bundle: cross-linked markdown concepts with YAML frontmatter for services, modules, APIs, data models and operations
- Mines git history for design rationale, following reverts back to the commit they reverted, so invariants recorded nowhere else are captured with commit citations
- Parks what it cannot establish as `open_questions` instead of guessing, and `/speckit.okf.clarify` folds human answers back in behind `<!-- clarified -->` sentinels that later updates will not overwrite
- Incremental `/speckit.okf.update` diffs since the last logged commit and refreshes only affected concepts, deprecating rather than deleting
- `/speckit.okf.validate` enforces the OKF §9 conformance rules (4 error classes, 9 warning classes)

### Testing Checklist

- [x] Extension installs successfully via download URL
- [x] All commands execute without errors
- [x] Documentation is complete and accurate
- [x] No security vulnerabilities identified
- [x] Tested on at least one real project

### Submission Requirements

- [x] Valid `extension.yml` manifest included
- [x] README.md with installation and usage instructions
- [x] LICENSE file included
- [x] GitHub release created with version tag
- [x] All command files exist and are properly formatted
[...]

PR #4462, opened 2026-09-07
[extension] Update OKF Knowledge Bundle Generator to v0.6.0

alexcpn:
Updates the `okf` entry in `extensions/catalog.community.json`. Submitted as an issue in #4439; opening a PR as well, since that seems to be the path that actually lands.

### Two things this fixes

**The version is three releases behind.** The catalog has 0.3.0; current is [v0.6.0](https://github.com/alexcpn/speckit_okf/releases/tag/v0.6.0).

**Every URL in the entry is wrong.** They point at `speckit_ofk` — a transposition of `speckit_okf`. They currently resolve only because GitHub is still 301-redirecting from the repository's old name, which is not something a catalog should depend on.

### Diff

Ten lines, all inside the single `okf` object. Nothing else in the file is touched.

| Field | Was | Now |
| --- | --- | --- |
| `version` | `0.3.0` | `0.6.0` |
| `download_url` | `…/speckit_ofk/…/v0.3.0.zip` | `…/speckit_okf/…/v0.6.0.zip` |
| `repository`, `homepage`, `documentation`, `changelog` | `speckit_ofk` | `speckit_okf` |
| `provides.commands` | `4` | `5` |
| `updated_at` | `2026-07-21` | `2026-09-07` |
| `description` | — | rewritten, below |

> Generates and maintains an Open Knowledge Format (OKF v0.1) knowledge bundle from a source-code repository. Mines git history for the rationale behind the code, parks what it cannot establish as open questions rather than guessing, and verifies that the bundle's claims are supported by the repository.

### What changed in the extension

The fifth command is `/speckit.okf.verify`. The other four are unchanged.

`validate` only ever asked whether a bundle is well-formed. `verify` asks whether it is *true*: every cited commit must exist and belong to the concept citing it and must not be a commit that changed only test files; every symbol in an `# Interfaces` table must appear in non-test source; every `source_files` path must be tracked by git; and no invariant may be asserted with no evidence.

The motivation was concrete. Generating a bundle for `kubernetes/kubernetes` twice, independently, produced the same five false claims both times — three of them commits that had only touched `_test.go` files, written up as production invariants. The instruction not to trust a bare commit subject already existed and was ignored both times. `verify` catches all five unaided.

Also in 0.6.0: history mining now lists the files each flagged commit touched and marks test-only ones; a new co-change script mines coupling that no import graph can see; untracked subtrees are refused rather than documented as the project's own; cross-links are relative, because a leading `/` means bundle root to OKF and repository root to GitHub; and bundles carry a `README.md`, since forges render that and ignore `index.md`.

[Full changelog](https://github.com/alexcpn/speckit_okf/blob/main/CHANGELOG.md)

### Checks

All URLs in the new entry return 200. The file parses as valid JSON. No new required tools, no new permissions, `effect` unchanged at `read-write`. Repository CI (shellcheck, skill validation, asset-sync) passes on the v0.6.0 commit.

### DIFF
--- a/extensions/catalog.community.json
+++ b/extensions/catalog.community.json
@@ -3412,14 +3412,14 @@
     "okf": {
       "name": "OKF Knowledge Bundle Generator",
       "id": "okf",
-      "description": "Generates and maintains an Open Knowledge Format (OKF v0.1) knowledge bundle from a source-code repository, mining git history for significance and rationale, and resolving open questions with the user.",
+      "description": "Generates and maintains an Open Knowledge Format (OKF v0.1) knowledge bundle from a source-code repository. Mines git history for the rationale behind the code, parks what it cannot establish as open questions rather than guessing, and verifies that the bundle's claims are supported by the repository.",
       "author": "Alex Punnen",
-      "version": "0.3.0",
-      "download_url": "https://github.com/alexcpn/speckit_ofk/archive/refs/tags/v0.3.0.zip",
-      "repository": "https://github.com/alexcpn/speckit_ofk",
-      "homepage": "https://github.com/alexcpn/speckit_ofk",
-      "documentation": "https://github.com/alexcpn/speckit_ofk/blob/main/README.md",
-      "changelog": "https://github.com/alexcpn/speckit_ofk/blob/main/CHANGELOG.md",
+      "version": "0.6.0",
+      "download_url": "https://github.com/alexcpn/speckit_okf/archive/refs/tags/v0.6.0.zip",
+      "repository": "https://github.com/alexcpn/speckit_okf",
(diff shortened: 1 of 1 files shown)


KSchlobohm:
Closed by #4475 

We use [GitHub Agentic Workflows](https://github.github.com/gh-aw/) to help merge Community Extensions into `main` and that this PR was closed when that flow was completed.

Thank you for your contribution to the community.

alexcpn:
@KSchlobohm the website still show the old version https://speckit-community.github.io/extensions/okf

KSchlobohm:
@alexcpn it appears to be updated and showing version 0.6.0 now.

Also, please note that this site is a community-maintained website. https://speckit-community.github.io/extensions/about

PR #4469, opened 2026-09-08
fix: roll back bundle installs when record save fails

marcelsafin:
## Description

Include bundle provenance persistence in the install rollback boundary.

Previously, a failure while saving `bundles.json` occurred after the protected
install block. Newly installed components remained on disk without a
provenance record, so later bundle operations could not manage them. Record
creation and saving now trigger the existing bounded rollback for components
installed by the current call.

## Testing

Targeted: `tests/integration/test_bundler_install_flow.py` (22 passed).

Full suite: 7,532 passed, 195 skipped. One existing PowerShell-launcher test
was omitted because `pwsh` is unavailable locally; it fails identically on the
unchanged upstream commit.

## AI Disclosure

GitHub Copilot (GPT-5.6 Sol) autonomously reproduced the bug, wrote the
regression test and implementation, and ran verification under
@marcelsafin's direction and review.

### DIFF
--- a/src/specify_cli/bundler/services/installer.py
+++ b/src/specify_cli/bundler/services/installer.py
@@ -155,6 +155,16 @@ def install_bundle(
                 if installer.is_installed(project_root, component):
                     installer.remove(project_root, component)
                     result.uninstalled.append(component)
+
+        record = InstalledBundleRecord.create(
+            bundle_id=plan.bundle_id,
+            version=plan.version,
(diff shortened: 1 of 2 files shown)


marcelsafin:
Consolidated into #4470 at maintainer request; closing this PR as superseded. Posted on behalf of @marcelsafin by GitHub Copilot (GPT-5.6 Sol).

Issue #4446, opened 2026-09-04
[Extension]: Add pay-x402 v1.1.0 (update to existing catalog entry)


shawn93676:
### Extension ID

pay-x402

### Extension Name

AgentPay x402 — Spend Controls for Spec Kit Agents

### Version

1.1.0

### Description

Set USDC spending caps, check a counterparty's credit before paying it, and pay x402-enabled APIs during spec implementation. Settles on Base or Solana.

### Author

AgentPay Team

### Repository URL

https://github.com/shawnhvac/spec-kit-pay-x402

### Download URL

https://github.com/shawnhvac/spec-kit-pay-x402/archive/refs/tags/v1.1.0.zip

### License

MIT

### Homepage

https://x402-agent-pay.com

### Documentation URL

https://github.com/shawnhvac/spec-kit-pay-x402#readme

### Changelog URL

https://github.com/shawnhvac/spec-kit-pay-x402/blob/main/CHANGELOG.md

### Required Spec Kit Version

>=0.1.0

### Required Tools

None. The commands are agent prompts; they use plain HTTP and the agent's existing signer.

### Number of Commands

3

### Number of Hooks

1

### Tags

payments, x402, budget, usdc, api, solana

### Key Features

- **Spend cap per spec** — `/speckit.pay-x402.budget` sets the most an agent may spend on one implementation, enforced before signing rather than reconciled afterwards. Offered automatically via an optional `after_plan` hook.
- **Counterparty credit check** — `/speckit.pay-x402.trust` reads the recipient wallet's credit file from SolvScore before any money moves. Free, no key, accepts Solana or Base addresses. A wallet too new to price returns an explicit refusal with a reason rather than a misleading zero.
- **Payment execution over x402** — `/speckit.pay-x402.pay` reads the endpoint's live 402 terms, signs a payment for the chosen rail, retries with the `X-Payment` header, and confirms a settlement transaction hash came back.
- **Two verified rails** — Base (`eip155:8453`, EIP-3009 `transferWithAuthorization`) and Solana mainnet (partially-signed `TransferChecked` with a facilitator fee payer, per the x402 exact scheme for SVM).
- **Terms always come from the endpoint** — amount, recipient and facilitator are read from each live 402, so stale documentation in the extension cannot cause a wrong payment.

### Testing Checklist

- [X] Extension installs successfully via download URL
- [X] All commands execute without errors
- [X] Documentation is complete and accurate
- [X] No security vulnerabilities identified
[...]

PR #4476, opened 2026-09-08
Update pay-x402 extension entry to v1.1.0

shawn93676:
Updates the `pay-x402` entry in `extensions/catalog.community.json` from **v1.0.0 to v1.1.0**.

**Why this matters:** the catalog currently serves v1.0.0, whose `pay` command directed agents to POST to an endpoint that is itself a paid x402 resource — an agent following the bundled instructions received a `402` instead of a payment header. v1.1.0 fixes the flow (the payer signs its own payment) and removes XRPL claims the facilitator never supported.

Changes:
- `version`: 1.0.0 → 1.1.0
- `download_url`: points at the v1.1.0 tag (verified: the zip resolves)
- `description`: updated to match v1.1.0 (spend caps, counterparty credit check via SolvScore, x402 payments during spec implementation; settles on Base or Solana)
- `updated_at`: entry-level and catalog-level timestamps

Follows the Extension Submission template (filed as issue #4446 on Sept 4); the catalog was last refreshed Sept 3, so the entry has not yet picked up the v1.1.0 fix.

Tag: `🤖🤖🤖` (prepared by an AI agent on behalf of the extension author, per repo convention).

### DIFF
--- a/extensions/catalog.community.json
+++ b/extensions/catalog.community.json
@@ -1,6 +1,6 @@
 {
   "schema_version": "1.0",
-  "updated_at": "2026-09-03T00:00:00Z",
+  "updated_at": "2026-09-08T00:00:00Z",
   "catalog_url": "https://raw.githubusercontent.com/github/spec-kit/main/extensions/catalog.community.json",
   "extensions": {
     "adrkit": {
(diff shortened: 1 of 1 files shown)


mnriem:
Thanks — but catalog changes shouldn't be hand-edited into `catalog.community.json` via a PR; they go through the `[Extension]` submission-issue flow, which validates the release/digests and generates the catalog PR automatically. Good news: you already filed this correctly as #4446, so this PR is a duplicate of that submission. Closing this in favor of #4446 — the v1.1.0 update will land through the automated pipeline there. (No need to also open the PR; the issue is enough.)

Issue #2416, opened 2026-04-29
[Bug]: Systematic build_exec_args() audit — CLI dispatch flags unverified for 11+ integrations


markuswondrak:
## Scope extends beyond opencode — systematic `build_exec_args()` audit

While investigating the opencode dispatch bug, I audited **all 28 integrations** and found that the root cause is not specific to opencode — it's a systematic architectural weakness in how `build_exec_args()` is inherited.

### Current state

Only **4 out of 15 CLI-capable integrations** have verified, correct `build_exec_args()` implementations:

| Integration | Base class | CLI invocation | Status |
|---|---|---|---|
| **Claude** | `SkillsIntegration` | `claude -p "prompt" --model X --output-format json` | ✅ Verified (`-p` is documented) |
| **Gemini** | `TomlIntegration` | `gemini -p "prompt" -m X --output-format json` | ✅ Verified (`-p` and `-m` are documented) |
| **Codex** | `SkillsIntegration` | `codex exec "prompt" --model X --json` | ✅ Custom override with `exec` subcommand |
| **Copilot** | `IntegrationBase` | `copilot -p "prompt" --agent X --yolo --model X --output-format json` | ✅ Custom override with dynamic `--yolo` |

The remaining **11 integrations with `requires_cli: True`** all inherit `build_exec_args()` from their base class without verification that the CLI actually supports the generated flags:

| Integration | Inherited invocation | Risk |
|---|---|---|
| **OpenCode** | `opencode -p "prompt" --model X --output-format json` | ❌ **Confirmed broken** — should be `opencode run "prompt" -m X` |
| **Amp** | `amp -p "prompt" --model X --output-format json` | ⚠️ Unverified |
| **Auggie** | `auggie -p "prompt" --model X --output-format json` | ⚠️ Unverified |
| **CodeBuddy** | `codebuddy -p "prompt" --model X --output-format json` | ⚠️ Unverified |
| **Forge** | `forge -p "prompt" --model X --output-format json` | ⚠️ Unverified |
| **iFlow** | `iflow -p "prompt" --model X --output-format json` | ⚠️ Unverified |
| **Junie** | `junie -p "prompt" --model X --output-format json` | ⚠️ Unverified |
| **Kiro CLI** | `kiro-cli -p "prompt" --model X --output-format json` | ⚠️ Unverified |
| **Pi** | `pi -p "prompt" --model X --output-format json` | ⚠️ Unverified |
[...]

PR #4481, opened 2026-09-08
fix(agy): support --model, --add-dir workspace, and pre-prompt extra args

### LINKED ISSUE #4480: bug(agy): --model ignored, EXTRA_ARGS appended after --print, and --add-dir workspace not passed
### Summary

In `src/specify_cli/integrations/agy/__init__.py`, `AgyIntegration.build_exec_args()` currently hardcodes ignoring the `model` parameter:

```python
def build_exec_args(
    self,
    prompt: str,
    *,
    model: str | None = None,
    output_json: bool = True,
) -> list[str] | None:
    # agy does not support --model or JSON output; both params are ignored
    args = [self._resolve_executable(), "--print", prompt]
    self._apply_extra_args_env_var(args)
    return args
```

### Issues Observed

1. **`--model` is supported by `agy` CLI**:
   Modern Antigravity CLI (`agy`) supports `--model <model_name>` (e.g. `agy --model claude-sonnet-4-6 -p "..."` or `agy --model gemini-3.1-pro-high -p "..."`). Because `model` is omitted from `build_exec_args`, stage-level model tiering defined in workflows (such as `.specify/workflows/...`) fails to route to the desired model, silently falling back to the user's default local model.

2. **Working directory / workspace flag**:
   When `agy` is launched in headless `--print` mode, it does not automatically bind the current directory unless `--add-dir <path>` is passed, causing it to report:
   ```text
   You do not currently have an active workspace open.
   By default, scratch files and projects are located at: ~/.gemini/antigravity-cli/scratch
   ```

3. **Flag positioning before `--print`**:
   Any extra args appended *after* `--print <prompt>` are treated by `agy`'s CLI parser as part of the positional prompt text rather than CLI flags. Flags (such as `--model`, `--add-dir`, `--dangerously-skip-permissions`, `--print-timeout`) should precede `--print`.

### Proposed Fix

In `AgyIntegration.build_exec_args()`:

```python
def build_exec_args(
    self,
    prompt: str,
    *,
    model: str | None = None,
    output_json: bool = True,
) -> list[str] | None:
    args = [self._resolve_executable()]
    if model:
        args.extend(["--model", model])
    self._apply_extra_args_env_var(args)
    args.extend(["--print", prompt])
    return args
```

And in `AgyIntegration.dispatch_command()`, pass `--add-dir` with `project_root` when invoking the subprocess so `agy` recognizes the project repository workspace.

Happy to submit a PR with tests if this looks good to the maintainers!

timoyan:
## Summary

Fixes three bugs in `AgyIntegration` that together prevented `specify workflow run` from working correctly with `agy`. Closes #4480.

### 1. `--model` was silently ignored

`build_exec_args()` had a comment `# agy does not support --model ... both params are ignored` and hardcoded no `--model` flag. `agy >=1.20` fully supports `agy --model <name> --print <prompt>`. Workflow YAML `model:` pins had no effect.

**Fix:** Pass `--model <model>` before `--print` when `model` is set.

### 2. `EXTRA_ARGS` were appended after `--print` (silently absorbed as prompt text)

`agy` treats every token **after** `--print` as part of the prompt string, not as CLI flags. Operator flags like `--dangerously-skip-permissions` or `--print-timeout 30m` set via `SPECKIT_INTEGRATION_AGY_EXTRA_ARGS` were appended after `--print` and silently became part of the prompt.

**Fix:** `_apply_extra_args_env_var()` is now called **before** `args.extend(['--print', prompt])`.

### 3. No workspace passed to `agy` -- 'no active workspace'

The base `dispatch_command()` sets `cwd=project_root` for the subprocess, but `agy` does not read `cwd` as its workspace root -- it requires an explicit `--add-dir <path>` flag. Without it, `agy` falls back to its own scratch directory, cannot locate `.agents/skills/`, and reports 'no active workspace' -- making all installed Spec Kit skills invisible.

**Fix:** `build_exec_args()` now accepts `project_root: Path | None` and `agy` injects `--add-dir <project_root>` before `--print` when `project_root` is given. The implementation now uses the shared base integration to thread `project_root` through to `build_exec_args()` instead of an agy-specific `dispatch_command()` override.

## Flag Order Guarantee

All flags appear before `--print` so `agy` parses them correctly:

`agy [--model <m>] [--add-dir <d>] [<EXTRA_ARGS>] --print <prompt>`

## Tests

- Updated ignores_model -> honors_model (asserts --model is now included)
- Updated extra_args_honors -> extra_args_before_print (asserts extra args come BEFORE --print)
- Added: test_build_exec_args_add_dir_for_workspace
- Added: test_build_exec_args_no_add_dir_when_project_root_is_none
- Added: test_build_exec_args_no_model_flag_when_model_is_none
- Added: test_build_exec_args_combined_flag_order

All 47 tests pass.

## AI Assistance Disclosure

This PR was drafted with substantial AI assistance (Google Antigravity / agy). The bugs were discovered while personally running `specify workflow run` against a real project and observing the failures firsthand. The root cause analysis, fix design, and test verification were reviewed and validated by me. AI was used to write the code, tests, commit message, and PR body.

### DIFF
--- a/scripts/powershell/common.ps1
+++ b/scripts/powershell/common.ps1
@@ -1,6 +1,9 @@
 #!/usr/bin/env pwsh
 # Common PowerShell functions analogous to common.sh
 
+[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
+$OutputEncoding = [System.Text.Encoding]::UTF8
+
 # Find repository root by searching upward for .specify directory
 # This is the primary marker for spec-kit projects
 function Find-SpecifyRoot {
--- a/scripts/powershell/common.ps1
+++ b/scripts/powershell/common.ps1
@@ -1,8 +1,6 @@
 #!/usr/bin/env pwsh
 # Common PowerShell functions analogous to common.sh
 
-[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
-$OutputEncoding = [System.Text.Encoding]::UTF8
 
 # Find repository root by searching upward for .specify directory
 # This is the primary marker for spec-kit projects
--- a/src/specify_cli/integrations/agy/__init__.py
+++ b/src/specify_cli/integrations/agy/__init__.py
@@ -90,13 +90,26 @@ class AgyIntegration(SkillsIntegration):
         output_json: bool = True,
         integration_args: Sequence[str] | None = None,
         integration_options: Mapping[str, Any] | None = None,
+        project_root: Path | None = None,
     ) -> list[str] | None:
-        # agy does not support --model or JSON output; both params are ignored
         self.validate_runtime_config(integration_args, integration_options)
-        args = [self._resolve_executable(), "--print", prompt]
-        # Honor SPECKIT_INTEGRATION_AGY_EXTRA_ARGS (operator-supplied flags),
-        # appended after the positional prompt like the devin integration.
+        # agy does not support JSON output; output_json is ignored.
+        args = [self._resolve_executable()]
+        # Pass --model before --print so agy can parse it as a flag.
+        # agy >=1.20 supports: agy --model <name> --print <prompt>
+        if model:
+            args.extend(["--model", model])
+        # Inject --add-dir so agy discovers the project workspace when invoked
+        # from an arbitrary working directory (e.g. the workflow engine's cwd).
+        # Without this agy falls back to its own scratch directory and cannot
+        # locate .agents/skills/, reporting "no active workspace".
+        if project_root is not None:
+            args.extend(["--add-dir", str(project_root)])
+        # Honor SPECKIT_INTEGRATION_AGY_EXTRA_ARGS (operator-supplied flags).
+        # These MUST be inserted before --print because agy treats every token
+        # that follows --print as part of the prompt, not as CLI flags.
         self._apply_extra_args_env_var(args)
+        args.extend(["--print", prompt])
         return args
 
     def setup(
--- a/src/specify_cli/integrations/agy/__init__.py
+++ b/src/specify_cli/integrations/agy/__init__.py
@@ -104,7 +104,7 @@ class AgyIntegration(SkillsIntegration):
         # Without this agy falls back to its own scratch directory and cannot
         # locate .agents/skills/, reporting "no active workspace".
         if project_root is not None:
-            args.extend(["--add-dir", str(project_root)])
+            args.extend(["--add-dir", str(project_root.resolve())])
         # Honor SPECKIT_INTEGRATION_AGY_EXTRA_ARGS (operator-supplied flags).
         # These MUST be inserted before --print because agy treats every token
         # that follows --print as part of the prompt, not as CLI flags.
--- a/src/specify_cli/integrations/base.py
+++ b/src/specify_cli/integrations/base.py
@@ -248,6 +248,7 @@ class IntegrationBase(ABC):
         output_json: bool = True,
         integration_args: Sequence[str] | None = None,
         integration_options: Mapping[str, Any] | None = None,
+        project_root: Path | None = None,
     ) -> list[str] | None:
         """Build CLI arguments for non-interactive execution.
 
@@ -410,6 +411,7 @@ class IntegrationBase(ABC):
             output_json=not stream,
             integration_args=integration_args,
             integration_options=integration_options,
+            project_root=project_root,
         )
 
         if exec_args is None:
@@ -1067,6 +1069,7 @@ class MarkdownIntegration(IntegrationBase):
         output_json: bool = True,
         integration_args: Sequence[str] | None = None,
         integration_options: Mapping[str, Any] | None = None,
+        project_root: Path | None = None,
     ) -> list[str] | None:
         self.validate_runtime_config(integration_args, integration_options)
         if not self.config or not self.config.get("requires_cli"):
@@ -1161,6 +1164,7 @@ class TomlIntegration(IntegrationBase):
         output_json: bool = True,
         integration_args: Sequence[str] | None = None,
         integration_options: Mapping[str, Any] | None = None,
+        project_root: Path | None = None,
     ) -> list[str] | None:
         self.validate_runtime_config(integration_args, integration_options)
         if not self.config or not self.config.get("requires_cli"):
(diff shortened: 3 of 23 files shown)


mnriem:
Thanks — and this is a model report→fix loop (#4480 + this), with genuinely exemplary disclosure and the bugs verified on a real project. The `build_exec_args` fix is right: `--model`/`--add-dir`/extra-args before `--print`, prompt last. One design question before merge: this **overrides `dispatch_command`** to thread `project_root` through as `--add-dir`, which duplicates the base class's subprocess/`shutil.which`/streaming/timeout logic in the agy integration — that's a maintenance surface that can drift from the base. Is there a lighter path (e.g. threading `project_root` through the base `dispatch_command` so all integrations can use `--add-dir`-style hints), rather than a full override here? I'll trigger the automated review in parallel. Also relates to the systemic `build_exec_args()` audit in #2416.

timoyan:
> Thanks — and this is a model report→fix loop (#4480 + this), with genuinely exemplary disclosure and the bugs verified on a real project. The `build_exec_args` fix is right: `--model`/`--add-dir`/extra-args before `--print`, prompt last. One design question before merge: this **overrides `dispatch_command`** to thread `project_root` through as `--add-dir`, which duplicates the base class's subprocess/`shutil.which`/streaming/timeout logic in the agy integration — that's a maintenance surface that can drift from the base. Is there a lighter path (e.g. threading `project_root` through the base `dispatch_command` so all integrations can use `--add-dir`-style hints), rather than a full override here? I'll trigger the automated review in parallel. Also relates to the systemic `build_exec_args()` audit in #2416.

@mnriem  I think you're right, we could have project_root from base. Please check the latest commmit. :)

mnriem:
This is a great generalization — threading `project_root` through the base `dispatch_command` is the right call, and it likely fixes the same class for other integrations (relates to #2416, and to the bob dispatch bug in #4491/#4492). One consequence to name: this grew from an agy fix into a base-layer change touching ~15 integrations, so it needs (a) a rebase — it's currently conflicting with `main` — and (b) green CI across the full matrix, since every integration's dispatch is now affected. Given the blast radius I'll give it a thorough review rather than merging on the automated pass alone. Could you rebase and confirm the per-integration dispatch tests still pass? Then re-request review.

timoyan:
@mnriem   Rebased onto the latest main, threaded project_root across base classes and all integrations, and added a regression test for dispatching project_root as --add-dir. All integration and base test suites are green locally!

mnriem:
Thanks — the base-class threading is the right call and it's actually leaner than the per-integration override. Rebased and green-locally is great, but given this touches ~15 integrations I want a full CI run across the matrix (macOS/Windows have surprised us elsewhere today) plus a proper review of the base change before merge — I'll trigger CI now. Nothing needed from you meanwhile.

timoyan:
@mnriem sorry there is Ruff error in opencode integration. I just submit the fix.

mnriem:
Good — the base-class threading and rebase look right, and CI's now running. One real gap the re-review found before merge: workflow **`prompt`** steps still don't get `project_root`. `PromptStep._try_dispatch()` calls `build_exec_args(prompt, model=..., output_json=False)` without threading it through, so agy's "no active workspace" failure (the original #4480 bug) can still recur on prompt steps. Please thread `project_root` on the prompt path too, with a test covering it. Re-request once addressed (and let's confirm the full CI matrix goes green given this touches ~15 integrations).

timoyan:
@mnriem Changes pushed, please verify it in latest commit :)

timoyan:
@mnriem I also found unit test edge case about encoding and execution environment variables pollution. I have submitted these fixes in commit e49845b so they are clearly isolated for review. All test suites pass cleanly.

timoyan:
@mnriem  Following Copilot’s suggestions, I submitted two changes: one for encoding in testing and another for the project_root path. Sorry for the back and forth.

timoyan:
I discussed this with my AI coding assistant (Antigravity), and it correctly identified that when `project_root` is a relative path, passing it directly to `--add-dir`      
  causes `agy` to resolve it incorrectly within the subprocess's updated `cwd`. We fixed this by calling `.resolve()` to ensure it always passes an absolute path.               
                                                                                                                                                                                 
    I've pushed the fix and added a regression test for this specific relative-path edge case. I also ran a full end-to-end test on a local dummy project to verify the workflow 
  engine dispatch works as expected.                                                                                                                                             
                                                                                                                                                                                 
    *(Note: The code changes, tests, and this explanation were generated in collaboration with the Antigravity AI assistant).*                                                   
                                                                                                                                                                                 
    ## Manual test results                                                                                                                                                       
                                                                                                                                                                                 
    **Agent**: Antigravity (agy) | **OS/Shell**: Windows / PowerShell                                                                                                            
                                                                                                                                                                                 
    | Command tested | Notes |                                                                                                                                                   
    | --- | --- |                                                                                                                                                                
    | `/speckit.specify` | Verified via `specify workflow run speckit`. The workflow engine now correctly passes the absolute path to `--add-dir`. `agy` successfully located the
  `.agents/skills` directory without throwing the `no active workspace` error. |                                                                                                 
    All integration test suites and the agent config consistency checks (`test_agent_config_consistency.py`) are green locally! Ready for another CI run whenever you are.

timoyan:
I've just rebased the branch onto the latest main to keep everything up to date.

timoyan:
@mnriem any ncesssary updates need me to do ?

mnriem:
The previous workspace, prompt-dispatch, and relative-path fixes are present. One compatibility update is now needed because #4492 merged afterward: Bob has a new `build_exec_args()` override that does not accept `project_root`.


Please incorporate that override into the shared signature update and verify the registry-wide and dispatch tests against current `main`, including Bob—not only agy.


Also update the PR description: it still describes an agy-specific `dispatch_command()` override, whereas the implementation now uses the shared base path.


timoyan:
@mnriem all changes are delivered, please verify when you are avaliable. Thanks.

timoyan:
@mnriem quick question, there is an error I found when run ruff in local. Not sure it should be fixed here or in another PR.
https://github.com/github/spec-kit/pull/4481/changes/021529ddfbd918f3dca86679be162e25a8ab0ba8#diff-7d5f7affaf6d0acb77415a6e7078b932706f1daece744ed5cf78c5becb0ca2cdR287

mnriem:
Thank you!

PR #4492, opened 2026-09-09
fix(integrations): dispatch bob commands via `bob run`

### LINKED ISSUE #4491: [Bug]: Bob integration cannot dispatch — build_exec_args() unimplemented
### Bug Description

`BobIntegration` doesn't override `build_exec_args()`, so it inherits the `IntegrationBase` no-op returning `None`. Callers read `None` as "this CLI is unavailable", so every workflow `command:`/`prompt:` step targeting Bob fails, even with `bob` installed and on PATH.

Separately, `BobIntegration` extends `IntegrationBase` directly and inherits its hardcoded `.` separator, so `build_command_invocation("speckit.specify")` returns `/speckit.specify` while skills-mode projects install skills at `.bob/skills/speckit-specify/`. The class already computes the right separator in `effective_invoke_separator()`.

Related: #2416 (build_exec_args audit — Bob isn't in its table, since it doesn't currently dispatch).

### Steps to Reproduce

```
$ specify workflow run speckit -i spec="hello"

Running workflow: Full SDD Cycle (speckit)
Version: 1.0.0

  ▸ [specify] speckit.specify …

Status: failed
Run ID: a9de034e
Error: Cannot dispatch command 'speckit.specify': integration 'bob' CLI not found or not installed. Install the CLI tool or check
'specify integration list'.
```

`bob` is installed and on PATH:

```
$ bob -h
Usage: bob [options] [command]
```

### Expected Behavior

The step dispatches to Bob, as it does for claude, codex, cursor-agent and goose.

### Actual Behavior

The run fails at the first step with "CLI not found or not installed", which is misleading — the CLI is present and working.

### Specify CLI Version

1.0.5

### AI Agent

IBM Bob

### Operating System

macOS Tahoe 26.3

### Python Version

Python 3.13

### Error Logs

```shell
Status: failed
Run ID: a9de034e
Error: Cannot dispatch command 'speckit.specify': integration 'bob' CLI not found or not installed. Install the CLI tool or check 'specify integration list'.
```

### Additional Context

_No response_

temp-droid:
## Description

Fixes #4491.

`BobIntegration` never overrode `build_exec_args()`, so it inherited the
`IntegrationBase` no-op returning `None`. Every workflow `command:`/`prompt:`
step targeting Bob failed with "CLI not found or not installed" even with `bob`
on PATH. Bob Shell has a headless entry point (`bob run`), so this dispatches
through it.

Also overrides `build_command_invocation()`, which was inherited from
`IntegrationBase` with a hardcoded `.` and rendered `/speckit.specify` where
skills-mode projects install `.bob/skills/speckit-specify/`. The class already
computes the right separator in `effective_invoke_separator()`.

`requires_cli` stays `False` — dispatch is opt-in by overriding the method,
mirroring `CursorAgentIntegration`. Same fix shape as #2631.

## Testing

Added 6 tests to `tests/integrations/test_integration_bob.py` (62 pass in that
file); `tests/test_agent_config_consistency.py` passes.

Verified end-to-end against Bob Shell 2.0.1: a workspace skill at
`.bob/skills/speckit-marker/` was reached through
`dispatch_command → build_exec_args → bob run`, returning `status: success`
with one `UseSkillTool` call.

Note: the full suite shows 11 failures in the bash/Python parity tests. They are
identical on `main` without this change — 11 failed / 7702 passed on `main`,
11 failed / 7708 passed here, same test set — and pass when run in isolation.

## AI Disclosure

Investigation, patch and tests written with Claude; reviewed and verified by me.


mnriem:
Thanks — the `bob run` dispatch approach is right, and the `requires_cli=False` reasoning is sound. The re-review found two invocation-rendering bugs to fix before merge: (1) `effective_invoke_separator()` is called without `project_root`, so it always falls back to Bob's fresh-project default (`-`) and gets legacy mode wrong — thread the project context through; (2) only the *first* separator is converted, so a dotted extension command like `speckit.git.commit` renders `/speckit-git.commit` in skills mode, but the installed skill is `speckit-git-commit` — convert all segments. Re-request once addressed.

mnriem:
Thanks @temp-droid — the project-aware dispatch now preserves legacy Bob invocations, and skills-mode extension commands correctly flatten every separator. The runtime-configuration compatibility update is also included.


Both requested corrections are addressed, and CI and the follow-up review pass on the current head. No further author changes requested from this pass; this is ready for my final review and merge decision.


mnriem:
Thank you!

PR #4506, opened 2026-09-10
Copy core command templates into .specify/templates/commands on init

### LINKED ISSUE #3086: specify init does not copy templates/commands/ to .specify/templates/commands/ — wrap composition strategy fails

gyanu2507:
## Description

`specify init` copied only top-level files from `templates/`, so `.specify/templates/commands/` never existed after install. Wrap presets then had no core base layer and `registered_commands` stayed empty.

The wheel maps `templates/commands` to `core_pack/commands`, which that file loop could not see. `install_shared_infra` and `refresh_shared_templates` now copy that tree with the same overwrite and manifest policy as the other shared templates.

## Testing

- [x] `uv run pytest tests/test_shared_infra_commands.py tests/test_shared_infra_gitignore.py tests/test_command_template_py_scripts.py`

## AI Disclosure

LLM-assisted implementation. I chose the issue, wrote the tests, and ran them locally.

Fixes #3086

### DIFF
--- a/src/specify_cli/shared_infra.py
+++ b/src/specify_cli/shared_infra.py
@@ -145,6 +145,28 @@ def shared_templates_source(
     return repo_root / "templates"
 
 
+def shared_commands_source(
+    *,
+    core_pack: Path | None,
+    repo_root: Path,
+) -> Path:
+    """Return the bundled/source core command templates directory.
+
+    Wheel installs force-include ``templates/commands`` as
+    ``core_pack/commands`` (not nested under ``core_pack/templates``), so
+    ``shared_templates_source()`` cannot see that tree. Source checkouts
+    keep the files at ``repo_root/templates/commands``.
+    """
+    if core_pack:
+        wheel_commands = core_pack / "commands"
+        if wheel_commands.is_dir():
+            return wheel_commands
+        nested = core_pack / "templates" / "commands"
+        if nested.is_dir():
+            return nested
+    return repo_root / "templates" / "commands"
+
+
 def shared_scripts_source(
     *,
     core_pack: Path | None,
@@ -354,7 +376,8 @@ def refresh_shared_templates(
 ) -> None:
     """Refresh default-sensitive shared templates without touching scripts."""
     templates_src = shared_templates_source(core_pack=core_pack, repo_root=repo_root)
-    if not templates_src.is_dir():
+    commands_src = shared_commands_source(core_pack=core_pack, repo_root=repo_root)
+    if not templates_src.is_dir() and not commands_src.is_dir():
         return
 
     manifest = load_speckit_manifest(project_path, version=version, console=console)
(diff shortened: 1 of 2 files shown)


mnriem:
Thanks [@gyanu2507](https://github.com/gyanu2507) — clean, well-tested, focused change, and the disclosure is exactly right. But I think this may be fixing a problem that isn't actually present on current `main`, so I want to check the premise before we add this.


The `wrap` resolver doesn't require `.specify/templates/commands/`. In `PresetResolver` (`presets/__init__.py`, the core-resolution path), `.specify/templates/commands/` is only an optional **Priority 4** override; **Priority 5** falls back to the bundled `core_pack/commands` (wheel) or `templates/commands` (source checkout), and the comment there notes it *"must always be checked so that strategy:wrap presets can locate {CORE_TEMPLATE}."* So a `wrap` preset should resolve its `{CORE_TEMPLATE}` base from `core_pack/commands` even when `.specify/templates/commands/` is absent — which is the same tree you note the wheel maps `templates/commands` onto.


Copying the tree into `.specify/templates/commands/` makes it a tracked, overwrite-managed override that shadows that canonical `core_pack` source, which is drift surface I'd rather not add unless there's a real gap.


Could you share a concrete reproduction on current `main` — a `wrap` command override that fails to resolve its core base **without** `.specify/templates/commands/` present? If the Priority-5 fallback isn't firing in some install layout, that's the actual bug to fix (in the resolver), and I'll prioritize it. If it resolves fine, then #3086 is already addressed by that fallback and we can close both out.

gyanu2507:
You're right. I re-checked wrap on current main with an empty `.specify/templates/commands/`.

`collect_all_layers("speckit.implement", "command")` still gets a Priority 5 base: `core (bundled)` pointing at `commands/implement.md`. Same for `resolve()`. That's already pinned in `test_collect_all_layers_finds_bundled_core_without_specify_commands` and `test_resolve_command_falls_back_to_bundled_core`.

Copying the tree into `.specify/templates/commands/` would only add an overwrite-managed shadow of `core_pack`. Closing this PR. #3086 looks addressed by that fallback.

PR #4510, opened 2026-09-10
fix(workflows): harden community submission workflows

mnriem:
## Summary

A focused hardening change to the community catalog submission workflows
(bundle, extension, preset).

- Threat detection is configured to block safe outputs instead of only
  emitting a warning.
- Adds a regression test that keeps this fail-closed behavior from silently
  reverting on a future regeneration.

Generated workflow files were regenerated to match; action pins are unchanged
from the base. No catalog data changes and no change to the maintainer
PR-review step.

### Note on review feedback

An earlier revision also switched issue parsing to the sanitized event
snapshot. That was reverted: the snapshot redacts URLs on hosts outside the
workflow's allowed-domains list, which would corrupt valid submissions whose
metadata legitimately points at non-GitHub hosts. Reading the triggering issue
is retained, with the separate maintainer review remaining the control for any
post-label edits.

---
Opened on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8).


mnriem:
Thanks for the review — addressed in a6d52076 (and the two preceding commits).

**Sanitized-snapshot input (High, bundle + extension).** You're right that the regenerated snapshot redacts HTTPS hosts outside the workflow's allowed-domains list. These workflows record off-allowlist URLs verbatim (extension homepage/documentation/changelog, bundle required component catalogs, and the proposed catalog-entry JSON), so the snapshot would corrupt valid submissions. I reverted the snapshot change and kept reading the triggering issue; the separate maintainer review remains the control for any post-label edits.

**Action pin downgrade (Medium ×3).** Unintended side effect of the regeneration. The lock files now restore the base `actions/checkout` (v7.0.1) and `actions/setup-node` (v7.0.0) pins, and the pin database is unchanged from the base. The only remaining lock delta versus base is the threat-detection change.

**Missing fail-closed regression coverage (Low).** Added a test that asserts each workflow source sets `threat-detection.continue-on-error: false` and each lock compiles the fail-closed detection gate, so a later regeneration can't silently restore warning-only behavior.

Net change is now scoped to fail-closed threat detection plus that test.

Posted on behalf of @mnriem by GitHub Copilot (model: Claude Opus 4.8), acting autonomously.

Issue #1717, opened 2026-02-27
Bug: /speckit.clarify 在 spec 阶段将 spec 级别条目错误 defer 到 plan


bigsmartben:
## 问题描述

`/speckit.clarify` 在 checklist 驱动模式（`checklists/spec-quality.md`）下，将 24 个未勾选条目**全部**标记为"Plan 阶段关注"并 defer，跳过了 spec 阶段应承担的澄清职责。Agent 同时输出了自相矛盾的结论："No critical ambiguities detected"，但实际上存在大量未解决的 spec 级别模糊项。

## 复现步骤

```
/speckit.clarify checklists/spec-quality.md
```

Agent 输出大量 remaining 条目，将其全部归类为 Plan-stage concerns，未向用户提出任何 spec 级别的澄清问题。

## 被错误 defer 的条目分析

按 `clarify.md` 规范的 spec-oriented taxonomy，以下条目应在 spec 阶段澄清，而非 defer：

| 条目 | 被错误标记为 | 实际所属 spec taxonomy 分类 |
|------|------------|---------------------------|
| CHK014 并发用户量 | Plan: 容量规划 | Non-Functional Quality Attributes |
| CHK032-036 NFR 细化 | Plan: 运维设计 | Non-Functional Quality Attributes |
| CHK020-022 验收量化 | Plan: 测试策略 | Completion Signals |
| CHK025, CHK028-031 边界与异常 | Plan: 集成/容错设计 | Edge Cases & Failure Handling |
| CHK004 图片约束 | Plan: 存储方案设计 | Domain & Data Model / Constraints |
| CHK008 空状态 | Plan: 标准 UX 模式 | Interaction & UX Flow |
| CHK037-040 依赖契约 | Plan: 契约冻结 | Integration & External Dependencies |
| CHK001, CHK007 API 分页 | Plan: 接口设计 | 边界（"是否需要分页"属 spec，"实现方式"属 plan） |

## 根因分析

Agent 缺乏**阶段归属判断机制**——在决定"某条目是否 defer 到下游"时，未对照 `clarify.md` 中定义的 spec-oriented taxonomy 进行分类匹配，而是基于条目关键词做了粗粒度的直觉判断，导致"懒惰 defer"。

### 1. 在 Step 2 与 Step 3 之间插入 Stage Gate 判断步骤

```
对于每个未勾选的 checklist 条目：
1. 将条目内容与当前阶段 taxonomy 分类做逐项匹配
[...]

PR #4515, opened 2026-09-10
Add opt-in clarify-spec-gate wrap preset

gyanu2507:
## Description

Follow-up to #4507 / #1717. Core `/speckit.clarify` keeps a bounded deferral. The stage-gate procedure, the >60% defer-ratio audit, and the MUST-NOT block are methodology, so they live in an opt-in wrap of `speckit.clarify`.

Install with `specify preset add clarify-spec-gate`. The wrapper uses `{CORE_TEMPLATE}`, so core command updates still compose.

## Testing

- [x] `uv run pytest tests/test_presets.py::TestClarifySpecGatePreset tests/contract/test_wheel_bundled_presets.py tests/test_presets.py::TestConstitutionSyncPreset`

### DIFF
--- a/presets/catalog.json
+++ b/presets/catalog.json
@@ -1,6 +1,6 @@
 {
   "schema_version": "1.0",
-  "updated_at": "2026-08-04T00:00:00Z",
+  "updated_at": "2026-09-10T00:00:00Z",
   "catalog_url": "https://raw.githubusercontent.com/github/spec-kit/main/presets/catalog.json",
   "presets": {
     "lean": {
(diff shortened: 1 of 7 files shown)


mnriem:
@gyanu2507— genuinely, thank you for this. You took the #1717 problem, accepted the rescope on #4507 without friction, and turned the methodology into a clean, well-built, well-tested preset. This is exactly the kind of thoughtful, responsive contribution that makes the project better, and I want to be clear the work is landing — just in the right place.


Now that I can see the preset itself, the call is about *hosting*, not quality: an opinionated clarify methodology belongs in the **community catalog** (externally hosted), not as a first-party preset bundled inside spec-kit. The bundled set (`presets/catalog.json` — `lean`/`constitution-sync`) is a small, foundational, github-maintained group, and keeping methodology opinions in the community catalog is consistent with #4507 removing this from core so it isn't imposed on everyone. Per `presets/PUBLISHING.md`, community presets live in your own public repo and come in via the **`[Preset]` submission issue** → validated → added to `catalog.community.json`.


So I'm closing this against the repo itself, but everything you built carries straight over: publish `clarify-spec-gate` from your own repo (the README, `preset.yml`, and wrapped `speckit.clarify.md` are already done), tag a release, and open a `[Preset]` submission issue — I'll validate it and get it into the community catalog. The core bounded-deferral change from #4507 stays in-repo separately, so the two compose exactly as you designed. Thanks again, and I'm looking forward to the catalog submission.

PR #4534, opened 2026-09-11
fix(workflows): cap gate `show_file` line width, not just its line count

jawwad-ali:
## Problem

`GateStep._read_show_file` bounds the number of lines:

```python
if len(lines) >= GateStep.MAX_SHOW_FILE_LINES:
    truncated = True
    break
```

but nothing bounds a single line's **length**. A file with no newlines — minified JSON, a lockfile, a base64 blob, all plausible review material to put in front of an approver — is one line of arbitrary size, so the cap never triggers and the whole file floods the prompt. That is exactly the outcome `MAX_SHOW_FILE_LINES` exists to prevent.

```
many short lines : returned 201 lines, total chars=1323    (cap=200)
one long line    : returned   1 lines, total chars=400008  <-- unbounded
```

The 5,000-line file is correctly clipped to 1,323 characters. The single-line 400 KB file passes through whole.

The content is also rendered inside the gate box one character at a time by `_prompt`, so the operator's terminal is buried and the choice prompt is pushed far off screen.

## Fix

Add a per-line cap alongside the line cap, with a notice in the same style as the existing one:

```python
MAX_SHOW_FILE_LINE_CHARS = 500
...
if len(clean) > GateStep.MAX_SHOW_FILE_LINE_CHARS:
    clean = clean[:GateStep.MAX_SHOW_FILE_LINE_CHARS] + "… (line truncated at 500 characters)"
```

After the fix:

```
many short lines : 201 lines, 1323 chars (unchanged)
one long line    :   1 lines,  536 chars
short file       : ['hello', 'world']    (untouched)
```

## Verification

- **Fail-before / pass-after:** 1 new-vs-baseline failure with the source reverted to `upstream/main` → **49 passed** with the fix.
- A companion test asserts short lines are still rendered verbatim with no notice, so the cap cannot start mangling ordinary files.
- The existing `test_read_show_file_truncates_large_file`, `test_read_show_file_strips_control_chars` and `test_read_show_file_empty` all use short lines and are unaffected — they pass unchanged.

**Behaviour change, disclosed:** a `show_file` line longer than 500 characters is now rendered truncated with a notice instead of in full. That is the intent; no existing test exercises such a line.

**Note on overlap:** this touches `steps/gate/__init__.py`, as does my #4529, but a different function (`_read_show_file` vs `_prompt`). They should merge independently — happy to rebase whichever lands second.

---

### DIFF
--- a/src/specify_cli/workflows/steps/gate/__init__.py
+++ b/src/specify_cli/workflows/steps/gate/__init__.py
@@ -35,6 +35,10 @@ class GateStep(StepBase):
     #: Maximum number of ``show_file`` lines rendered at the prompt, so a
     #: large file cannot flood the terminal before the choice.
     MAX_SHOW_FILE_LINES = 200
+    # A line cap alone does not bound the output: a file with no newlines --
+    # minified JSON, a lockfile, a base64 blob -- is a single line of arbitrary
+    # length and floods the prompt exactly as the line cap exists to prevent.
+    MAX_SHOW_FILE_LINE_CHARS = 500
 
     def execute(self, config: dict[str, Any], context: StepContext) -> StepResult:
         message = config.get("message", "Review required.")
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
Please resolve conflicts

PR #4535, opened 2026-09-11
fix(bundler): validate a catalog source before persisting it

jawwad-ali:
## Problem

`add_source` writes the new entry and only *then* constructs it:

```python
catalogs.append(entry)
_write(project_root, catalogs)
return CatalogSource.from_dict(entry, Scope.PROJECT)   # <-- validation, too late
```

`CatalogSource.from_dict` already rejects an empty id — but by the time it runs, the entry is on disk.

```
before     : []
add_source -> raised: A catalog source is missing its 'id'.
AFTER      : [{"id": "", "url": "https://example.test/c.json", "priority": 5, "install_policy": "install-allowed"}]
```

The command reports a failure, so the user reasonably assumes nothing happened.

### It wedges the project

Every command that loads the catalog stack now fails on that stray entry:

```
load_source_stack -> BROKEN: BundlerError A catalog source is missing its 'id'.
```

So a single mistyped `catalog add --id "   "` leaves the project's catalog config unusable until `bundle-catalogs.yml` is hand-edited — by hand, for a file the `bundle catalog` commands exist to manage.

## Fix

Construct before writing, reusing the validation that already exists rather than adding a second check that could drift:

```python
source = CatalogSource.from_dict(entry, Scope.PROJECT)
catalogs.append(entry)
_write(project_root, catalogs)
return source
```

After the fix:

```
add with blank id -> refused: A catalog source is missing its 'id'.
config after      : []                      <-- nothing persisted
stack still loads : ['default', 'community']
normal add works  : ok https://example.test/ok.json
```

## Verification

- **Fail-before / pass-after:** 1 new-vs-baseline failure with the source reverted to `upstream/main` → passing with the fix.
- The test asserts both halves: the config is left empty **and** `load_source_stack` still succeeds — the latter is what the stray entry actually broke.
- Scoped regression over `tests/unit`: **536 passed** vs a clean-`main` baseline of **535 passed**, same 2 pre-existing failures, none new.

**No behaviour change for any input that previously succeeded.** The reordered call is the same construction, just earlier, and it raises for exactly the input that already raised — the difference is that the failure no longer leaves a file behind.

**Note on overlap:** this touches `commands_impl/catalog_config.py`, as does my #4533, but a different function (`add_source` vs `remove_source`). Happy to rebase whichever lands second.

---

### DIFF
--- a/src/specify_cli/bundler/commands_impl/catalog_config.py
+++ b/src/specify_cli/bundler/commands_impl/catalog_config.py
@@ -198,9 +198,17 @@ def add_source(
         "priority": int(priority),
         "install_policy": install_policy.value,
     }
+    # Construct BEFORE writing. ``CatalogSource.from_dict`` already rejects an
+    # empty id, but it used to run after ``_write`` had persisted the entry, so
+    # a whitespace-only ``--id`` (or a url from which no id can be derived)
+    # reported "A catalog source is missing its 'id'" while leaving
+    # ``{"id": "", ...}`` behind in bundle-catalogs.yml. Every later command
+    # that loads the stack then failed with that same error, wedging the
+    # project's catalog config until the file was hand-edited.
+    source = CatalogSource.from_dict(entry, Scope.PROJECT)
     catalogs.append(entry)
     _write(project_root, catalogs)
-    return CatalogSource.from_dict(entry, Scope.PROJECT)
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
Please resolve conflicts

PR #4538, opened 2026-09-11
fix(bundler): report a newline-containing version constraint instead of crashing

jawwad-ali:
## Problem

`_normalize_constraint` matches each comma-separated clause and uses the result without checking it:

```python
match = _SPECIFIER_CLAUSE.match(raw)
operator, version = match.groups()
```

`_SPECIFIER_CLAUSE` is anchored with `^`/`$`, and `.` does not cross newlines — so a clause containing an **embedded** newline does not match at all, and `match.groups()` raises a raw `AttributeError`. That escapes `parse_constraint`'s contract, which is to surface bad input as a `BundlerError`:

```python
except InvalidSpecifier as exc:
    raise BundlerError(f"Invalid version constraint '{value}': {exc}") from exc
```

A YAML **block literal** — an ordinary way to write a multi-clause constraint if you don't know the comma form — reaches it directly:

```yaml
requires:
  speckit_version: |
    >=1.0.0
    <2.0.0
```

loads as `'>=1.0.0\n<2.0.0\n'`, and:

```
manifest value: '>=1.0.0\n<2.0.0\n'
  parse_constraint  -> AttributeError: 'NoneType' object has no attribute 'groups'   <-- raw crash
  satisfies         -> AttributeError: 'NoneType' object has no attribute 'groups'   <-- raw crash
```

Worth noting why this survived: **leading and trailing** newlines are fine, because `\s*` absorbs them. Only an *embedded* one fails:

```
'>=1.0.0\n'   -> OK >=1.0.0
'\n>=1.0.0'   -> OK >=1.0.0
'>=1.0\n.0'   -> AttributeError
```

## Fix

Raise `InvalidSpecifier` for an unmatched clause, which the existing handler in `parse_constraint` already converts to the `BundlerError` callers expect — so the fix reuses the established error path rather than introducing a second one.

After the fix:

```
block literal  -> BundlerError: Invalid version constraint ...
embedded nl    -> BundlerError: Invalid version constraint ...
garbage nl     -> BundlerError: Invalid version constraint ...
--- unchanged ---
'>=1.0.0' / '>=1.0.0,<2' / '~=1.2' / '>=v1.0.0' / '>=1.0.0\n' / '\n>=1.0.0'  all OK
satisfies('1.5.0', '>=1.0.0,<2.0.0') = True
```

## Verification

- **Fail-before / pass-after:** 3 new-vs-baseline failures with the source reverted to `upstream/main` → **33 passed** with the fix.
- A companion parametrization pins that every valid constraint — including the leading/trailing-newline forms that already worked — is unaffected, so the guard cannot start rejecting good input.
- Scoped regression over `tests/unit`: **543 passed** vs a clean-`main` baseline of **535 passed**, same 2 pre-existing failures, none new.

**No behaviour change for any input that previously succeeded** — the only inputs affected are those that previously crashed with an `AttributeError`, which now report the documented error instead.

---

### DIFF
--- a/src/specify_cli/bundler/lib/versioning.py
+++ b/src/specify_cli/bundler/lib/versioning.py
@@ -54,6 +54,19 @@ def _normalize_constraint(value: str) -> str:
         if not raw.strip():
             continue
         match = _SPECIFIER_CLAUSE.match(raw)
+        if match is None:
+            # ``_SPECIFIER_CLAUSE`` is anchored with ``^``/``$`` and ``.`` does
+            # not cross newlines, so a clause containing an EMBEDDED newline
+            # does not match at all and ``match.groups()`` raised a raw
+            # AttributeError -- escaping ``parse_constraint``'s contract to
+            # report bad input as a BundlerError. A YAML block literal is an
+            # ordinary way to reach this:
+            #     requires:
+            #       speckit_version: |
+            #         >=1.0.0
+            #         <2.0.0
+            # which loads as ">=1.0.0\n<2.0.0\n".
+            raise InvalidSpecifier(f"Invalid specifier: {raw!r}")
         operator, version = match.groups()
         clauses.append(f"{operator or ''}{_normalize_semver(version)}")
     return ",".join(clauses)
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
Please resolve conflicts

PR #4542, opened 2026-09-11
fix(workflows): state the equal-priority tie-break in `workflow resolve` output

jawwad-ali:
## Problem

`workflow resolve` prints the layer list under a bare header:

```python
console.print("Layers (highest precedence first):")
```

`collect_all_layers` sorts by `(priority, source)` **ascending**, which puts the winning layer first only while priorities *differ*. On a tie the sort is alphabetical by source, while the merge gives the conflict to the **last** id — exactly as documented:

> Equal-priority overlays are applied alphabetically by ID, with the last ID winning conflicts. — `docs/reference/workflows.md:126`

So on a tie the header states the opposite of the outcome printed directly beneath it, in the one command whose documented purpose is explaining which overlay contributed or overrode a step.

```
DIFFERENT priorities (alpha=5, beta=10)
   listed first : project:alpha
   actual winner: echo FROM-ALPHA
   header correct? YES

EQUAL priorities (both 10)
   listed first : project:alpha
   actual winner: echo FROM-BETA
   header correct? NO  <-- inverted
```

The full output contradicted itself two lines apart:

```
Layers (highest precedence first):
  • [project-overlay] project:alpha (priority=10)
  • [project-overlay] project:beta (priority=10)
Step attribution:
  • build: project:beta          <-- beta won, alpha was labelled highest
```

## Fix — the label, not the order

The ordering is **deliberate**: `test_workflow_resolve_equal_priority_layers_sort_by_source` pins it, and its own comment acknowledges that the alphabetically-last layer is the one that wins. Reordering the list would fight a decision the maintainers made on purpose, so this spells the tie-break out instead:

```
Layers (highest precedence first; on equal priority the last ID wins):
  • [project-overlay] project:alpha (priority=10)
  • [project-overlay] project:beta (priority=10)
Step attribution:
  • build: project:beta
```

Now self-consistent, and the behaviour is untouched.

## Verification

- **Fail-before / pass-after:** 1 new-vs-baseline failure with the source reverted to `upstream/main` → **26 passed** with the fix.
- The new test asserts the header **and** the attribution together, so it pins the label against the real outcome rather than against a hard-coded string alone.
- No existing test asserts the old header text (checked: `git grep "highest precedence first" -- tests/` is empty), and `test_workflow_resolve_equal_priority_layers_sort_by_source` passes unchanged.

**No behaviour change** — output text only.

**Note on overlap:** this touches `overlays/_commands.py`, as does my #4141, but a different function (`workflow_resolve` vs `workflow_overlay_add`). Happy to rebase whichever lands second.

---

### DIFF
--- a/src/specify_cli/workflows/overlays/_commands.py
+++ b/src/specify_cli/workflows/overlays/_commands.py
@@ -424,7 +424,17 @@ def workflow_resolve(project_root: Path, workflow_id: str) -> dict[str, Any] | N
         return None
 
     console.print(f"Resolved workflow '{workflow_id}':")
-    console.print("Layers (highest precedence first):")
+    # The list is sorted by (priority, source) ascending, which puts the
+    # winning layer first only while priorities DIFFER. On a tie the sort is
+    # alphabetical by source while the merge gives the conflict to the LAST id
+    # (docs/reference/workflows.md:126), so the plain "highest precedence
+    # first" label stated the opposite of the outcome printed directly beneath
+    # it in the same output -- in the one command whose job is explaining which
+    # overlay won. Spell the tie-break out rather than reorder the list, which
+    # `test_workflow_resolve_equal_priority_layers_sort_by_source` pins.
+    console.print(
+        "Layers (highest precedence first; on equal priority the last ID wins):"
+    )
     for layer in layers:
         priority = (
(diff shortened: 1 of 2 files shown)


mnriem (changes requested):
Please resolve conflicts

PR #4552, opened 2026-09-12
fix: use structured YAML parsing for frontmatter and scripts in process_template

darion-yaphet:
## Description

    In `IntegrationBase.process_template`, extracting and stripping the `scripts:` block from command templates previously relied on a hand-rolled line-by-line
  scanning loop and regex matching (`script_pattern = re.compile(r"^\s*([A-Za-z0-9_-]+):\s*(.+)$")`).

    This approach had multiple issues:
    - It failed on valid YAML syntax in frontmatter, such as comments between script entries, quoted script paths, or folded/literal multi-line block scalars (`>` or
  `|`).
    - It risked accidental matches if `scripts:` was referenced inside Markdown body text or code examples.
    - Stripping `scripts:` relied on line indentation checks rather than structured object manipulation.

    This PR replaces the hand-rolled text traversal with structured YAML handling using `yaml.safe_load` and `yaml.dump`:
    1. Safely parses frontmatter into a Python dictionary to extract `scripts`.
    2. Replaces `{SCRIPT}` in the template body.
    3. Removes `scripts` (`frontmatter_dict.pop("scripts", None)`) and dumps the cleaned frontmatter back via `yaml.dump(..., sort_keys=False, allow_unicode=True,
  width=float("inf"))`.
    4. Adds unit tests in `tests/integrations/test_base.py` covering YAML comments, quoted script commands, and folded scalar descriptions.

## Testing

      - `pytest tests/integrations/test_base.py tests/test_command_template_py_scripts.py` (118 tests passed)
      - `ruff check src/specify_cli/integrations/base.py tests/integrations/test_base.py` (All checks passed)

    ## AI Disclosure

    Assisted by Google Antigravity / Gemini for refactoring `process_template` to use `yaml.safe_load` and `yaml.dump`, adding test coverage in
  `tests/integrations/test_base.py`, and drafting commit/PR descriptions.

### DIFF
--- a/src/specify_cli/integrations/base.py
+++ b/src/specify_cli/integrations/base.py
@@ -826,29 +826,37 @@ class IntegrationBase(ABC):
         6. Rewrite paths: ``scripts/`` → ``.specify/scripts/`` etc.
         7. Replace ``__SPECKIT_COMMAND_<NAME>__`` with invocation strings
         """
-        # 1. Extract script command from frontmatter
+        # 1. Extract script command from frontmatter using structured YAML
         script_commands: dict[str, str] = {}
-        script_pattern = re.compile(r"^\s*([A-Za-z0-9_-]+):\s*(.+)$")
-        # Find the scripts: block
-        in_frontmatter = False
-        in_scripts = False
-        for line in content.splitlines():
-            if line == "---":
-                if in_frontmatter:
-                    break
-                in_frontmatter = True
-                continue
-            if not in_frontmatter:
-                continue
-            if line == "scripts:":
-                in_scripts = True
-                continue
-            if in_scripts and line and not line[0].isspace():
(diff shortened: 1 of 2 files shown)


mnriem:
Supporting quoted and folded YAML script commands is worthwhile, but this implementation needs corrections before merge. Please require an exact opening frontmatter delimiter: `----` currently causes ordinary Markdown content to be consumed and discarded. Also report YAML parsing failures explicitly rather than emitting empty frontmatter with unresolved `{SCRIPT}`.

Both added tests currently pass against the pre-change implementation. Please assert the actual rendered invocation for quoted and folded **script commands**, with tests that fail before the fix and pass afterward, and cover the content-preservation/error cases above. The current CI runs also await maintainer approval.

Issue #4433, opened 2026-09-03
[Bug] Windows Codex PATH can crash HTTPS updates with OPENSSL_Applink


joey-orbyss:
### Spec Kit version

`specify 1.0.4`, installed as a uv tool on Windows.

### Environment

- Codex Desktop local task on Windows
- The task PATH prepends Codex-owned native dependency directories, including Poppler and Git directories that contain their own `libssl-3-x64.dll` and `libcrypto-3-x64.dll`.
- `specify.exe` is the uv-installed launcher under `%USERPROFILE%\.local\bin`.

### Reproduction

From an initialized consumer with registered HTTPS catalogs, run either:

```powershell
specify workflow update program-kit-bootstrap
specify bundle update program-kit --integration codex
```

Both commands reproducibly exit 1 before completing the remote update with:

```text
OPENSSL_Uplink(...): no OPENSSL_Applink
```

The failure occurs specifically in the Codex Desktop process environment. A local/offline component installation can proceed because it does not enter the failing HTTPS path.

### Expected behavior

Spec Kit HTTPS/catalog operations should use the OpenSSL runtime packaged with its Python environment and must not load an ABI-incompatible DLL merely because another application prepended native tools to PATH. If the launcher cannot isolate its DLL search path, it should fail with an actionable dependency diagnostic rather than the OpenSSL applink abort.

### Impact

Documented `workflow update` and `bundle update` commands cannot run from Codex Desktop on affected Windows hosts. This blocks unattended governed upgrades and encourages consumers to use incomplete local recovery sequences.

mnriem:
Thanks — this is an unusually clear report. To pinpoint whether DLL-search-path isolation on our side would fix it, could you help narrow it: (1) does the same command succeed from a plain PowerShell (i.e. **not** launched by Codex Desktop), or with the Poppler/Git directories removed from `PATH`? (2) which OpenSSL DLL resolves first — e.g. `where.exe libssl-3-x64.dll` and `where.exe libcrypto-3-x64.dll` in the failing environment? That'll confirm the ABI-incompatible-DLL-on-PATH root cause and tell us whether pinning the DLL search dir to Spec Kit's Python env resolves it. In the meantime we'll look at making the HTTPS path fail with an actionable dependency diagnostic instead of the raw applink abort.

Yi-111-a:
[...]

PR #4556, opened 2026-09-12
feat: report the OpenSSL runtime in `specify version`

Yi-111-a:
## Description

`specify version` reported CLI version, Python, Platform, Architecture and OS Version, but nothing about the OpenSSL runtime the interpreter actually loaded.

That is the first thing HTTPS triage needs, and on Windows it is genuinely not inferable from outside the process: several unrelated toolchains ship their own `libssl-3-x64.dll` (Git for Windows alone has two distinct builds — `mingw64\bin\libssl-3-x64.dll` and `usr\bin\msys-ssl-3.dll`), and only the one actually loaded matters.

This came out of #4433, where the report is an OpenSSL-level abort (`OPENSSL_Uplink(...): no OPENSSL_Applink`) and the working hypothesis is a PATH-preceded OpenSSL DLL. Today, answering "which OpenSSL is in use?" requires the reporter to run a separate `python -c` snippet. After this change, `specify version` answers it:

```
┌────────────────────────── Specify CLI Information ──────────────────────────┐
│                                                                             │
│     CLI Version    1.0.7.dev0                                               │
│                                                                             │
│          Python    3.14.6                                                   │
│        Platform    Windows                                                  │
│    Architecture    AMD64                                                    │
│      OS Version    10.0.26200                                               │
│         OpenSSL    OpenSSL 3.5.7 9 Jun 2026                                 │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘
```

The row is skipped when `ssl.OPENSSL_VERSION` is unavailable — including on interpreters built without the `ssl` extension, where the command still succeeds rather than failing on `import ssl`.

Related to #4433 — this does **not** fix the abort, it only makes the runtime visible to whoever triages it.

## Evidence

New test fails on `main`, passes with the change:

```
# baseline (unmodified main)
$ .venv/Scripts/python -m pytest tests/test_cli_version.py -q
tests\test_cli_version.py ......F                                        [100%]
FAILED tests/test_cli_version.py::TestVersionCommand::test_version_reports_openssl_runtime
1 failed, 6 passed in 1.44s

# with this change (bfeea287)
$ python -m pytest tests/test_cli_version.py -q
tests/test_cli_version.py .........                                      [100%]
9 passed in 0.21s
```

The +2 tests cover the no-`ssl` paths: `version` skips the OpenSSL row, and `version --features --json` never touches `ssl` at all.

Lint matches what CI runs:

```
All checks passed!
```

## Testing

- [ ] Tested with a sample project (if applicable) — not applicable, `specify version` does not read a project

On the second box: `tests/test_cli_version.py` passes (9/9), and the `specify version` CLI surface was exercised directly. I could not complete the full suite on this machine — `tests/conftest.py` probes for a working bash and invokes `wsl.exe`, which is blocked by my host's security policy, and long runs get cut short. CI will cover the full matrix; I'd rather flag the gap than tick a box I didn't verify.

## AI Disclosure

- **Agent/tool:** Devin CLI — Cognition's interactive command-line agent
- **Model:** SWE-2 High
- **Mode/settings:** normal (autonomous) terminal session on a Linux VM — the agent ran shell commands, edited files, installed the package into a venv, ran pytest/ruff, and used `gh` under my GitHub account. No Devin Cloud web sessions.

The agent drafted the change and its tests and ran the local verification shown above; I reviewed the diff and the reported outputs.

## Note: why the loaded DLL path is not in this PR

The obvious companion field is the resolved path of the loaded `libssl-3-x64.dll`. I left it out on purpose.

It would require process-module enumeration, and it only means something once we know a PATH-preceded DLL *can* be loaded at all — which is exactly what is still unresolved in #4433. On my Windows host, `libssl-3-x64.dll` resolves to the interpreter's own `DLLs\` directory regardless of what is on `PATH` (CPython calls `SetDefaultDllDirectories` at startup, bpo-36085, so `PATH` is no longer part of extension-module resolution). Adding the path now would encode an unverified assumption into the UI.

If it turns out to be useful once #4433 is understood, I'm happy to follow up with it as a separate focused change.

### DIFF
--- a/src/specify_cli/__init__.py
+++ b/src/specify_cli/__init__.py
@@ -455,6 +455,7 @@ def version(
 ):
     """Display version and system information."""
     import platform
+    import ssl
 
     cli_version = get_speckit_version()
 
@@ -489,6 +490,12 @@ def version(
     info_table.add_row("Platform", platform.system())
     info_table.add_row("Architecture", platform.machine())
     info_table.add_row("OS Version", platform.version())
+    # The OpenSSL runtime the interpreter actually loaded. HTTPS failure
+    # reports (#4433) hinge on which OpenSSL is in play, and on Windows it is
+    # not obvious from the outside, so surface it here.
+    openssl_version = getattr(ssl, "OPENSSL_VERSION", "")
+    if openssl_version:
+        info_table.add_row("OpenSSL", openssl_version)
 
     panel = Panel(
         info_table,
--- a/tests/test_cli_version.py
+++ b/tests/test_cli_version.py
@@ -1,6 +1,7 @@
 """Tests for CLI version reporting."""
 
 import json
+import ssl
 from unittest.mock import patch
 
 from typer.testing import CliRunner
@@ -77,3 +78,20 @@ class TestVersionCommand:
 
         assert result.exit_code != 0
         assert "--json requires --features" in result.output
+
+    def test_version_reports_openssl_runtime(self):
+        """specify version reports the OpenSSL runtime the interpreter loaded.
+
+        Regression test for the triage gap in #4433: HTTPS failures on Windows
+        are commonly blamed on a PATH-preceded OpenSSL DLL, but ``specify
+        version`` reported no OpenSSL information at all, so a report had no way
+        to show which runtime was actually in use.
+        """
+        with patch("specify_cli.get_speckit_version", return_value="1.2.3"):
+            result = runner.invoke(app, ["version"])
+
+        assert result.exit_code == 0
+
+        expected = ssl.OPENSSL_VERSION
+        assert expected, "test host reports no ssl.OPENSSL_VERSION to assert against"
+        assert expected in result.output


mnriem:
This is a useful, focused diagnostic addition; no need to expand it into DLL enumeration or a fix for the separate crash report.

Please complete the AI disclosure with the agent/tool, model(s), and mode/settings used. The extent of assistance is already clear. A PR-description edit is sufficient—no code commit is needed for that.

The current CI runs require maintainer approval before final review.

Yi-111-a:
Done — AI disclosure completed in the PR description (agent/tool: Devin CLI; model: SWE-2 High; mode: autonomous terminal session with shell/editor/gh access).

I also took care of the Copilot-flagged import ordering while I was here (bfeea287): `import ssl` no longer runs before the `--features`/`--json` early returns — it's imported lazily at the point of use inside `try/except ImportError`, so an interpreter built without `_ssl` skips the OpenSSL row instead of failing `specify version`. Two regression tests added (`sys.modules["ssl"] = None` simulates the missing extension); `tests/test_cli_version.py` is 9/9 and `ruff@0.15.0` is clean.

mnriem:

Yi-111-a:
Copilot feedback addressed in ca85643:

1. **Module-level `import ssl` removed from `tests/test_cli_version.py`.** On an interpreter built without `_ssl`, pytest failed at collection before the no-SSL tests could run. The positive-path test now uses `pytest.importorskip("ssl")`, so the file collects cleanly either way.
2. **New regression test for an importable ssl without `OPENSSL_VERSION`.** `test_version_skips_openssl_row_when_version_attr_missing` substitutes `sys.modules["ssl"]` with a bare namespace object, pinning the `getattr(ssl, "OPENSSL_VERSION", "")` fallback — the row is omitted instead of raising AttributeError.

All 10 tests in `tests/test_cli_version.py` pass locally.

mnriem:
Thank you!

PR #4558, opened 2026-09-12
fix(extensions): reject aliases that shadow core commands

### LINKED ISSUE #4555: An extension alias can shadow a core command; `_validate_install_conflicts` only checks installed extensions
### Problem Statement

`_validate_install_conflicts` is documented as rejecting installs that "would shadow core or installed extension commands", but it only checks installed **extensions**. Core command names are never in the map it consults, so an extension can claim a core command name as an **alias** and install successfully.

Split out of #4421 at a maintainer's request, so that stage 1 (#4488) stays additive and no core-validation change is smuggled into it.

### Reproduction

`main` at `d848fb4`. A manifest whose command carries `speckit.taskstoissues` — a live core command — as an alias:

```yaml
schema_version: "1.0"
extension:
  id: probe
  name: "Probe"
  version: "1.0.0"
  description: "probe"
requires:
  speckit_version: ">=0.2.0"
provides:
  commands:
    - name: speckit.probe.taskstoissues
      file: commands/cmd.md
      aliases: ["speckit.taskstoissues"]     # core command name
```

Result:

```
1) manifest validation:
   ACCEPTED — aliases: [['speckit.taskstoissues']]
2) install into a project (core taskstoissues exists):
   ACCEPTED — installed: probe
```

Both stages accept it, with the core command present in the same project.

### Cause

Two gaps that line up:

1. **Aliases are not pattern-checked.** `EXTENSION_COMMAND_NAME_PATTERN` and the core-namespace guard in `src/specify_cli/extensions/__init__.py` apply to *primary* command names only. A primary `speckit.taskstoissues` is correctly rejected (it fails the two-segment pattern, and an extension `id` equal to a core command name is rejected separately) — but an alias goes through neither check.

2. **The conflict check does not know about core.** `_get_installed_command_name_map` walks `self.registry` — installed extensions — so core command names never appear in the map that `_validate_install_conflicts` compares against. Confirmed by inspection: the function references `self.registry` and nothing that loads core command names (`_load_core_command_names` / `_FALLBACK_CORE_COMMAND_NAMES`).

So the docstring describes a guarantee ("shadow core or installed extension commands") that is only half implemented.

### Impact

A community extension can claim a core command's name as an alias and shadow it. This is most likely to bite during the #4421 migration, where the whole point of stage 1 is that `speckit.taskstoissues` and `speckit.github.taskstoissues` coexist without either shadowing the other — but it applies to any core command, not just this one.

The bundled `github` extension in #4488 deliberately registers no alias, and has a test pinning that. Worth being explicit that this is currently **author discipline, not an enforced invariant** — which is the reason to file this separately rather than rely on the convention holding.

### Suggested Fix

Either:

- extend the core-namespace guard to aliases as well as primary names, and include core command names in the map `_validate_install_conflicts` consults; or
- if shadowing core via an alias is intended to be allowed (an override mechanism), narrow the docstring to describe what is actually checked.

Whichever way it goes, a test should pin the decision — an install that claims `speckit.taskstoissues` as an alias currently passes silently either way.

### Component

---

*Disclosure: I used an AI assistant (Claude Code, model Claude Opus 5) to investigate this and draft this report. The results above came from running the probe manifest through `ExtensionManifest` and `ExtensionManager.install_from_directory` against `main`; I reviewed the findings before filing.*

chelsealong:
## Summary

Fixes #4555.

`_validate_install_conflicts` is documented as rejecting installs that "would shadow core or installed extension commands", but `_get_installed_command_name_map` only walks `self.registry` — installed **extensions**. Core command names were never in the map it consults, so an extension could declare a core command's fully-qualified name (e.g. `speckit.taskstoissues`) as an **alias** and install successfully, shadowing the core command.

Primary command names are already namespace-checked against `CORE_COMMAND_NAMES` in `_collect_manifest_command_names` (a primary name like `speckit.taskstoissues` is correctly rejected — it fails the required two-segment `speckit.{extension}.{command}` pattern). Aliases are intentionally free-form (existing community extensions use short aliases like `speckit.verify`), so they can't be pattern-checked the same way — the only safe fix is to compare declared alias names directly against the fully-qualified core command names in the install-conflict check.

## Fix

In `_validate_install_conflicts` (`src/specify_cli/extensions/__init__.py`), build `core_command_names` from `CORE_COMMAND_NAMES` (qualified as `speckit.<name>`) and reject any declared command/alias name that matches, in addition to the existing installed-extension check.

## Test plan

- Added `test_install_rejects_alias_shadowing_core_command` to `tests/test_extensions.py`, reproducing the exact scenario from the issue: a manifest declaring `speckit.probe.taskstoissues` with alias `speckit.taskstoissues` (a live core command).
- **Fail-before / pass-after:** reverted just the source change (`git checkout HEAD~1 -- src/specify_cli/extensions/__init__.py`), confirmed the new test fails (`Failed: DID NOT RAISE ValidationError`), restored the fix, confirmed it passes.
- Full `tests/test_extensions.py`: **536 passed**.
- Full suite (`python -m pytest tests -q`): **8054 passed, 12 skipped**, no failures.

## AI Disclosure

This fix was implemented with Claude Code (Claude Sonnet 5), based on the reproduction and root-cause analysis in the linked issue. I reviewed the code paths involved (`_collect_manifest_command_names`, `_get_installed_command_name_map`, `_validate_install_conflicts`), confirmed the fail-before/pass-after test behavior described above, and ran the full test suite myself before opening this PR.

### DIFF
--- a/src/specify_cli/extensions/__init__.py
+++ b/src/specify_cli/extensions/__init__.py
@@ -1185,17 +1185,31 @@ class ExtensionManager:
         return installed_names
 
     def _validate_install_conflicts(self, manifest: ExtensionManifest) -> None:
-        """Reject installs that would shadow core or installed extension commands."""
+        """Reject installs that would shadow core or installed extension commands.
+
+        Primary command names are already namespace-checked against
+        ``CORE_COMMAND_NAMES`` in ``_collect_manifest_command_names``, but
+        aliases are intentionally free-form (see the comment there) and so
+        can only be caught here, by comparing declared names directly
+        against the fully-qualified core command names (``speckit.<name>``)
+        rather than relying on ``_get_installed_command_name_map``, which
+        only knows about installed extensions.
+        """
         declared_names = self._collect_manifest_command_names(manifest)
         installed_names = self._get_installed_command_name_map(
             exclude_extension_id=manifest.id
         )
+        core_command_names = {f"speckit.{name}" for name in CORE_COMMAND_NAMES}
 
-        collisions = [
-            f"{name} (already provided by extension '{installed_names[name]}')"
-            for name in sorted(declared_names)
-            if name in installed_names
-        ]
+        collisions = []
+        for name in sorted(declared_names):
+            if name in installed_names:
+                collisions.append(
+                    f"{name} (already provided by extension '{installed_names[name]}')"
+                )
+            elif name in core_command_names:
+                collisions.append(f"{name} (conflicts with core command)")
+
         if collisions:
             raise ValidationError(
                 "Extension commands conflict with installed extensions:\n- "
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -3126,6 +3126,48 @@ class TestExtensionManager:
         with pytest.raises(ValidationError, match="already provided by extension 'ext-one'"):
             manager.install_from_directory(second_dir, "0.1.0", register_commands=False)
 
+    def test_install_rejects_alias_shadowing_core_command(self, temp_dir, project_dir):
+        """An alias equal to a core command's qualified name must not install.
+
+        Regression test for #4555: a primary name is namespace-checked
+        against CORE_COMMAND_NAMES, but aliases are intentionally free-form
+        and previously went unchecked against core commands entirely, so an
+        extension could claim e.g. 'speckit.taskstoissues' as an alias and
+        shadow the core command of the same name.
+        """
+        import yaml
+
+        ext_dir = temp_dir / "probe-ext"
+        ext_dir.mkdir()
+        (ext_dir / "commands").mkdir()
+
+        manifest_data = {
+            "schema_version": "1.0",
+            "extension": {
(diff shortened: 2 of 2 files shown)


chelsealong:
Addressed both review comments in 4a6ac69:

- Added `_normalize_shadow_name` and used it to compare declared alias names against core commands using their on-disk output form (dots → hyphens, `speckit-` prefix), matching the normalization `CommandRegistrar._compute_output_name` and the Cline/Forge/Junie formatters apply. This catches `taskstoissues` and `speckit-taskstoissues`, not just the exact `speckit.taskstoissues` spelling. Added a parametrized regression test for both forms, confirmed it fails without the fix (`DID NOT RAISE ValidationError`) and passes with it.
- Fixed the error heading to say "conflict with core or installed extension commands" instead of implying only installed extensions can collide.

Full suite: 8056 passed, 12 skipped. `ruff check src tests`: clean.

mnriem:
Thank you!

PR #4559, opened 2026-09-12
fix: strip `scripts:` from extension command frontmatter in command mode

### LINKED ISSUE #4554: Extension command rendering does not strip `scripts:` from the agent-facing frontmatter
### Problem Statement

Command rendering strips the `scripts:` key from the agent-facing frontmatter on the **core** path, but not on the **extension** path. An extension command that declares `scripts:` therefore ships a rendered artifact carrying a key that core's equivalent does not, leaking an internal build detail into the agent-visible file.

Split out of #4421 at a maintainer's request, so that stage 1 (#4488) stays additive and this shared-render fix is not smuggled into it.

### Reproduction

`main` at `d848fb4`, with the bundled `github` extension from #4488 installed into a Copilot commands-mode project:

```bash
specify init proj --integration copilot --integration-options="--commands" --script sh
cd proj && specify extension add github
```

**Extension render** — `.github/agents/speckit.github.taskstoissues.agent.md`:

```yaml
---
description: Convert existing tasks into actionable, dependency-ordered GitHub issues ...
tools:
- github/github-mcp-server/list_issues
- github/github-mcp-server/issue_write
scripts:
  sh: .specify/extensions/github/scripts/bash/resolve-tasks.sh --json
  ps: .specify/extensions/github/scripts/powershell/resolve-tasks.ps1 -Json
  py: .specify/extensions/github/scripts/python/resolve_tasks.py --json
---
```

**Core render** — `.github/agents/speckit.taskstoissues.agent.md`, same project, from a source template that declares the same key:

```yaml
---
description: Convert existing tasks into actionable, dependency-ordered GitHub issues ...
tools: ['github/github-mcp-server/list_issues', 'github/github-mcp-server/issue_write']
---
```

No `scripts:` key. The path substitution happened in both — `{SCRIPT}` is resolved in each body — only the cleanup step differs.

### Cause

The core path strips the key as step 3 of `process_command_template` in `src/specify_cli/integrations/base.py`, after it has been consumed for `{SCRIPT}` substitution. The extension path (`CommandRegistrar.register_commands` / `_register_extension_skills`, `src/specify_cli/extensions/__init__.py`) calls `_adjust_script_paths` and substitutes `{SCRIPT}`, but never removes the key before serialising the frontmatter.

### Why it has gone unnoticed

No bundled extension declared `scripts:` before the `github` extension in #4488. `agent-context` and `git` both spell their script paths out in prose instead, so this branch of the render had no coverage.

### Impact

Cosmetic today rather than a functional break — the rendered paths are correct and the commands run. The concerns are that the two render paths disagree where they should agree, and that a build-time key reaches agent-visible output.

One caveat worth checking during the fix: `tools:` is emitted inline (flow style) by the core render and block style by the extension render, as visible above. Semantically identical YAML, but it suggests the two paths serialise frontmatter through different code, which may be the right place to look.

### Suggested Fix

Strip `scripts:` in the extension render once it has been consumed, matching `process_command_template`, and add a regression test asserting the key is absent from rendered extension commands in both command mode and skills mode.

### Component

---

*Disclosure: I used an AI assistant (Claude Code, model Claude Opus 5) to investigate this and draft this report. The reproduction above was produced by installing the extension into a scratch project against `main` and reading the generated files back; I reviewed the findings before filing.*

chelsealong:
## Description

Fixes #4554.

`CommandRegistrar.register_commands()` renders extension commands for
non-SKILL.md agents by consuming the source `scripts:` frontmatter
block for `{SCRIPT}` substitution via `resolve_skill_placeholders()`.
In the `markdown` branch (e.g. Copilot's `.github/agents/*.agent.md`
commands-mode layout), the now-consumed key was never removed from the
frontmatter dict before `render_markdown_command()` serialized the
full dict as the final file's frontmatter. `IntegrationBase.process_template()`
— the core template render path — already strips `scripts:` as an
explicit step, so the two render paths disagreed and a build-time key
leaked into agent-visible output, exactly as described in the issue.

The fix pops `scripts` from `frontmatter` right after the
`resolve_skill_placeholders()` call in the `markdown` branch of
`register_commands()`, once it's no longer needed. The pop is scoped
to `extension_id is not None` (i.e. extension-owned commands only)
because `register_commands()` is also shared by the preset system,
where an existing test (`test_register_commands_inherits_scripts_from_core`)
intentionally asserts that a `strategy: wrap` preset command retains
the merged `scripts:` key in this same render step. That's a separate,
pre-existing behavior outside the scope of this issue, so it's left
untouched.

The `toml` and `yaml` branches are not affected by this leak and need
no pop: `render_toml_command()` only ever reads the `description` key
off `frontmatter`, and `render_yaml_command()` only ever reads
`title`/`description` — neither serializes the full frontmatter dict,
so there is nothing for a `scripts:` key to leak into. (An earlier
version of this PR added the same pop to those two branches as
a defensive copy-paste; removed after confirming with a standalone
repro that it was dead code — neither branch's output ever contains a
`scripts:` key, fix or no fix.)

Skills-mode extension rendering (`_register_extension_skills` /
`register_skill_command`) was already unaffected — those paths build a
fresh `SKILL.md` frontmatter dict from scratch and never copy `scripts:`
into it — so no change was needed there.

## Testing

Added `test_command_mode_registration_strips_scripts_key` in
`tests/test_extensions.py`, which installs an extension command
declaring `scripts:` and registers it for the `copilot` agent (command
mode, `.agent.md` output) — mirroring the issue's reproduction.

Confirmed the test fails without the fix:

```
$ git checkout HEAD~1 -- src/specify_cli/agents.py  # (before the fix)
$ uv run pytest tests/test_extensions.py -k test_command_mode_registration_strips_scripts_key -q
FAILED tests/test_extensions.py::TestCommandRegistrar::test_command_mode_registration_strips_scripts_key
  assert 'scripts:' not in '---\ndescri..."$ARGUMENTS"'
  'scripts:' is contained here:
    ---
    description: Scripted command
    scripts:
      sh: .specify/scripts/bash/setup-plan.sh --json "{ARGS}"
      ps: .specify/scripts/powershell/setup-plan.ps1 -Json
    ---
```

And passes with the fix applied:

```
$ uv run pytest tests/test_extensions.py -k test_command_mode_registration_strips_scripts_key -q
1 passed in 0.13s
```

Full `test_extensions.py` + `test_presets.py` suite (unmodified elsewhere):

```
$ uv run pytest tests/test_extensions.py tests/test_presets.py -q
1214 passed in 14.82s
```

## AI Disclosure

This PR was written primarily by an autonomous AI coding agent (Claude
Code, model Claude Sonnet 5), including the investigation, the code
change, and the regression test. I reviewed the diff and the test
output before opening this PR.

### DIFF
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -828,8 +828,6 @@ class CommandRegistrar:
                 body = self.resolve_skill_placeholders(
                     agent_name, frontmatter, body, project_root, extension_id=extension_id
                 )
-                if extension_id:
-                    frontmatter.pop("scripts", None)
                 body = self._convert_argument_placeholder(
                     body, "$ARGUMENTS", agent_config["args"]
                 )
@@ -838,8 +836,6 @@ class CommandRegistrar:
                 body = self.resolve_skill_placeholders(
                     agent_name, frontmatter, body, project_root
                 )
-                if extension_id:
-                    frontmatter.pop("scripts", None)
                 body = self._convert_argument_placeholder(
                     body, "$ARGUMENTS", agent_config["args"]
                 )
--- a/src/specify_cli/agents.py
+++ b/src/specify_cli/agents.py
@@ -816,6 +816,8 @@ class CommandRegistrar:
                 body = self.resolve_skill_placeholders(
                     agent_name, frontmatter, body, project_root, extension_id=extension_id
                 )
+                if extension_id:
+                    frontmatter.pop("scripts", None)
                 body = self._convert_argument_placeholder(
                     body, "$ARGUMENTS", agent_config["args"]
                 )
@@ -826,6 +828,8 @@ class CommandRegistrar:
                 body = self.resolve_skill_placeholders(
                     agent_name, frontmatter, body, project_root, extension_id=extension_id
                 )
+                if extension_id:
+                    frontmatter.pop("scripts", None)
                 body = self._convert_argument_placeholder(
                     body, "$ARGUMENTS", agent_config["args"]
                 )
@@ -834,6 +838,8 @@ class CommandRegistrar:
                 body = self.resolve_skill_placeholders(
                     agent_name, frontmatter, body, project_root
                 )
+                if extension_id:
+                    frontmatter.pop("scripts", None)
                 body = self._convert_argument_placeholder(
                     body, "$ARGUMENTS", agent_config["args"]
                 )
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -3805,6 +3805,68 @@ Agent __AGENT__
         assert "{ARGS}" not in content
         assert '.specify/scripts/bash/setup-plan.sh --json "$ARGUMENTS"' in content
 
+    def test_command_mode_registration_strips_scripts_key(self, project_dir, temp_dir):
+        """Extension commands rendered in command mode (non-SKILL.md agents)
+        must not leak the build-time ``scripts:`` key into agent-facing
+        frontmatter, matching the core template render (#4554)."""
+        import yaml
+
+        ext_dir = temp_dir / "ext-scripted-commands"
+        ext_dir.mkdir()
+        (ext_dir / "commands").mkdir()
+
+        manifest_data = {
+            "schema_version": "1.0",
+            "extension": {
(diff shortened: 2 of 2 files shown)


mnriem:
Thank you!

PR #4564, opened 2026-09-13
fix: skip codex config.toml rewrite when there's nothing to merge

### LINKED ISSUE #4563: [Bug]: Codex integration rewrites existing config.toml outside manifest/status tracking
## Bug Description

On Windows, installing the Codex integration rewrites an existing tracked `.codex/config.toml` even when there is no semantic Codex event configuration to add. The file is not included in the Codex or Spec Kit managed manifests, so `specify integration status --json` reports a healthy/clean state while Git reports the repository as dirty.

The important issue is not CRLF by itself: an existing tracked file is mutated outside the managed/observed surface exposed by manifests and status.

## Steps to Reproduce

On Windows PowerShell, in an existing Git repository:

```powershell
[IO.File]::WriteAllText(
  '.codex/config.toml',
  'project_doc_max_bytes = 200000',
  [Text.UTF8Encoding]::new($false)
)
git add .codex/config.toml
git commit -m 'Add existing Codex config'

specify init --here --force --non-interactive --integration claude
specify integration install codex

git diff -- .codex/config.toml
specify integration status --json
```

The test used the standard CLI installed from PyPI in an isolated Python virtual environment. No generated files were patched.

## Expected Behavior

One of the following would make ownership observable and coherent:

1. A no-op Codex integration install preserves the existing config byte-for-byte when it has no fragment to add; or
2. If `.codex/config.toml` is intentionally rewritten/managed, the touched file is represented in the relevant manifest/status surface, and status reports modification consistently.

## Actual Behavior

The tracked file changes from no final newline/LF state to CRLF with a final newline:

```diff
-project_doc_max_bytes = 200000
+project_doc_max_bytes = 200000\r
```

`git diff --numstat` reports `1  1  .codex/config.toml`, while `specify integration status --json` reports:

```json
{
  "status": "ok",
  "installed_integrations": ["claude", "codex"],
  "multi_install_safe": true,
  "missing_managed_files": 0,
  "modified_managed_files": 0,
  "invalid_manifest_paths": 0,
  "findings": []
}
```

Inspection of v1.0.6 shows the Codex integration declaring `.codex/config.toml` as its event config file, and the TOML merge path writing `existing.rstrip() + "\n\n" + fragment + "\n"`. On Windows, the text write translates the newline sequence. The generated `codex.manifest.json` and `speckit.manifest.json` do not list the pre-existing `.codex/config.toml`.

## Specify CLI Version

`1.0.6`

## AI Agent

Codex CLI (multi-install test also included Claude Code)

## Operating System

Windows 11, AMD64, PowerShell; Git worktree

## Python Version

Python 3.12.14

## Error Logs

There is no CLI error. The unexpected evidence is the Git diff combined with a clean `integration status --json` result shown above.

## Additional Context

- Installing Claude first did not modify `.codex/config.toml`.
- Installing Codex second without `--force` caused the config rewrite.
- No `.gitattributes`, post-processing wrapper, or compatibility patch was added.

AI assistance disclosure: this report was drafted and filed by OpenAI Codex on behalf of the reporter. The reproduction, file diff, generated manifests, and CLI status were collected from a real isolated canary and checked before filing.

chelsealong:
Fixes #4563

## Problem

When Codex has no Specify-managed event hooks configured (the exact
scenario in #4563: a fresh `specify integration install codex` with no
extensions/overrides declaring any handlers), `resolve_events()` returns
`{}`, and `install_integration_events()` routes through its
empty-resolved-map branch into `_remove_native_event_hooks()` ->
`_remove_toml_entries()` — not through `_merge_toml_fragment()`.

`_remove_toml_entries()` always rewrote the destination file, even when
stripping Specify-owned hook blocks was a no-op (no such blocks were
present) — i.e. a true no-op install/teardown against a config.toml with
no Specify content at all.

That unconditional rewrite went through Python's text-mode `read_text()`/
`write_text()`, which perform newline translation. On Windows this
silently turned an LF-terminated pre-existing `.codex/config.toml` into
CRLF, producing a git-visible diff with no semantic content change.
Because the file is outside the Codex/Spec Kit managed manifests in this
scenario, `specify integration status --json` reports a clean/healthy
state while `git diff` shows the file as modified — exactly as described
in #4563.

An earlier version of this PR patched `_merge_toml_fragment()` instead.
That function does have the identical unconditional-rewrite shape, but it
is only reached when there is at least one supported, non-empty event to
merge — the empty-map/no-op case never calls it, so that fix had no effect
on the actual bug. This revision fixes `_remove_toml_entries()`, the
function that is actually executed on the issue's repro path (confirmed
by tracing every caller of `install_integration_events`/`resolve_events`
and by instrumented runs of the CLI's `specify init` / `specify
integration install codex` against a real `.codex/config.toml`).

## Fix

`_remove_toml_entries()` now skips the write (`return False`, leaving the
file completely untouched — byte-for-byte, including its original line
endings) when stripping Specify-owned blocks left the content unchanged,
mirroring the existing "unreadable file" skip path already in this
function and the analogous guard already present in
`_merge_toml_fragment()`.

The previous, harmless-but-insufficient `_merge_toml_fragment()` guard is
left in place (it doesn't hurt). To correct an earlier claim in this
description: that guard does not cover "identical nonempty fragment"
merges — it only short-circuits when there is no fragment to add at all
(`not fragment`) and stripping left the file unchanged. A merge with a
non-empty fragment always writes, even if the file already contains that
exact fragment.

A follow-up review (2026-09-16) found that this fix's own no-op guard in
`_remove_toml_entries()` ran too late: for a pre-existing blank or
comment-only `config.toml` with no Specify-owned blocks, the
"nothing-but-whitespace/comments" branch unlinked the file before the
`cleaned == existing` check was ever reached, deleting user-owned content
on a true no-op path. The unchanged-content check now runs first, so the
file is left untouched whenever stripping made no change; the delete-if-
empty branch below it now only ever fires after stripping has actually
removed a Specify-owned block. Added regression tests for a comments-only
config, a blank config, and confirmed the existing owned-block-deletion
behavior on teardown is unchanged.

## Test

`TestTomlNoOpMerge` in `tests/integrations/test_events.py` now has four
cases:

- `test_no_events_leaves_existing_config_untouched` — the original
  regression test: real production shape
  (`install_integration_events(integration, tmp_path, manifest, {})`)
  against a pre-existing `.codex/config.toml` with ordinary, non-comment,
  non-Specify content. Asserts bytes and mtime are unchanged.
- `test_comments_only_config_untouched_on_teardown` (new, this revision)
  — the case the 2026-09-16 review flagged: a pre-existing
  comments-only `config.toml`. Asserts the file is not deleted and is
  byte/mtime-identical.
- `test_blank_config_untouched_on_teardown` (new, this revision) — same
  for a whitespace-only pre-existing file.
- `test_owned_only_config_still_deleted_on_teardown` (new, this revision)
  — guards against the fix over-correcting: when the file contains only a
  Specify-owned block that teardown actually strips, the file must still
  be deleted, not preserved.

Confirmed the new tests fail without this revision's fix (`git checkout
HEAD~1 -- src/specify_cli/events.py`, i.e. the previous, already-merged
ordering):

```
FAILED tests/integrations/test_events.py::TestTomlNoOpMerge::test_no_events_leaves_existing_config_untouched
AssertionError: assert 1789628920564118373 == 1789628920563545456
FAILED tests/integrations/test_events.py::TestTomlNoOpMerge::test_comments_only_config_untouched_on_teardown
AssertionError: comments-only user config was deleted
FAILED tests/integrations/test_events.py::TestTomlNoOpMerge::test_blank_config_untouched_on_teardown
AssertionError: blank user config was deleted
3 failed, 1 passed, 128 deselected
```

With the fix applied (`git checkout HEAD -- src/specify_cli/events.py`):

```
tests/integrations/test_events.py::TestTomlNoOpMerge .... [100%]
4 passed, 128 deselected in 0.10s
```

Full suite (`python3 -m pytest tests -q`): `8057 passed, 12 skipped in 492.05s`.

`ruff check src/specify_cli/events.py tests/integrations/test_events.py`
shows the same pre-existing findings as before this change; none are on
the added/modified lines (verified by diffing the finding line numbers
against the changed line ranges).

## AI assistance disclosure

This PR was authored by an autonomous AI coding agent, Claude Code. This
revision (the `_remove_toml_entries` ordering fix and its tests) was
produced by Claude Code running Claude Sonnet 5 (model ID
`claude-sonnet-5`). The prior revision that first patched
`_remove_toml_entries()` was also produced by Claude Code; its exact
underlying model version was not recorded at the time and cannot be
reconstructed after the fact — this note names what is actually known
rather than guessing.

An independent, AI-assisted review found that the original version of
this PR patched the wrong function; that revision traced the issue's
actual repro path, fixed the function that is really executed
(`_remove_toml_entries`), and rewrote the regression test to exercise
that real path instead of a synthetic input no production caller
produces. A second round of review (2026-09-16) found the fix's own
no-op guard still ran too late for comments-only/blank files; this
revision corrects that ordering (see "Fix" above).

### DIFF
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -2129,7 +2129,10 @@ def _merge_toml_fragment(dst: Path, fragment: str) -> bool:
     An unreadable or undecodable pre-existing file aborts the merge instead
     of discarding the user's bytes, mirroring ``_load_user_json`` (#22).
     Returns False when skipped so callers avoid tracking the untouched file
-    (S5).
+    (S5) — including when there is no fragment to add and no owned blocks to
+    remove, so a no-op install doesn't rewrite (and, via text-mode newline
+    translation, mangle the line endings of) an untouched pre-existing file
+    (#4563).
     """
     _ensure_safe_destination(dst)
     existing = ""
@@ -2144,14 +2147,16 @@ def _merge_toml_fragment(dst: Path, fragment: str) -> bool:
             )
             logger.debug("Read error detail: %s", exc)
             return False
-    existing = re.sub(
+    stripped = re.sub(
         r'\[\[hooks\.\w+\]\]\n(?:(?!\[\[hooks\.\w+\]\]).)*?speckit_marker = true\n*',
         "",
         existing,
         flags=re.DOTALL,
     )
+    if not fragment and stripped == existing:
+        return False
     dst.parent.mkdir(parents=True, exist_ok=True)
-    dst.write_text(existing.rstrip() + "\n\n" + fragment + "\n", encoding="utf-8")
+    dst.write_text(stripped.rstrip() + "\n\n" + fragment + "\n", encoding="utf-8")
     return True
 
 
--- a/src/specify_cli/events.py
+++ b/src/specify_cli/events.py
@@ -2200,6 +2200,11 @@ def _remove_toml_entries(dst: Path) -> bool:
     """Remove Specify-marked TOML entries; delete the file if now empty (#14).
 
     Returns True if the file was deleted (no user content remained).
+
+    Leaves the file untouched (no write) when there are no Specify-owned
+    blocks to strip, so a no-op teardown/install doesn't rewrite (and, via
+    text-mode newline translation, mangle the line endings of) an untouched
+    pre-existing file (#4563).
     """
     if not dst.exists():
         return False
@@ -2235,6 +2240,8 @@ def _remove_toml_entries(dst: Path) -> bool:
     if not stripped:
         dst.unlink(missing_ok=True)
         return True
+    if cleaned == existing:
+        return False
     dst.write_text(cleaned, encoding="utf-8")
     return False
 
--- a/tests/integrations/test_events.py
+++ b/tests/integrations/test_events.py
@@ -938,18 +938,25 @@ class TestTomlUnreadableConfig:
 
 
 class TestTomlNoOpMerge:
-    """#4563: a merge with no fragment to add and no owned blocks to remove
-    must leave a pre-existing config.toml byte-for-byte untouched.
+    """#4563: installing with no resolved events must leave a pre-existing,
+    Specify-unowned config.toml byte-for-byte untouched.
 
-    Previously the merge unconditionally rewrote the file via
-    ``existing.rstrip() + "\\n\\n" + fragment + "\\n"`` even when ``fragment``
-    was empty, which both appended stray blank lines and (through Python's
-    text-mode newline translation on read/write) silently changed the file's
-    line-ending convention — turning a clean install into a spurious git diff
-    with no semantic change.
+    This is the real ``specify integration install codex`` repro: a project
+    with no Codex-specific event hooks configured resolves to ``events={}``
+    (see ``resolve_events``), which routes through
+    ``install_integration_events``'s empty-map branch into
+    ``_remove_native_event_hooks`` -> ``_remove_toml_entries`` — not through
+    ``_merge_toml_fragment``, which only runs when there is at least one
+    supported, non-empty event to merge. ``_remove_toml_entries`` computed
+    ``cleaned`` via a regex strip and then unconditionally called
+    ``dst.write_text(cleaned, ...)`` even when ``cleaned == existing`` (no
+    Specify-marked blocks present), which (through Python's text-mode
+    newline translation on read/write) silently changed the file's
+    line-ending convention on Windows — turning a clean install into a
+    spurious git diff with no semantic change.
     """
 
-    def test_no_handlers_leaves_existing_config_untouched(self, tmp_path):
+    def test_no_events_leaves_existing_config_untouched(self, tmp_path):
         from specify_cli.integrations.codex import CodexIntegration
 
         integration = CodexIntegration()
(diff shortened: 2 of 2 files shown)


mnriem:
The revised fix reaches the correct cleanup path, but the no-op guard runs too late. A comments-only `.codex/config.toml` with no Specify-owned blocks is still deleted before `cleaned == existing` is checked.


Please check for unchanged content before the empty/comments-only deletion branch. Cover preservation of comments-only and empty unowned files, while retaining cleanup when Specify-owned blocks were actually removed.


Also correct the description’s claim about identical nonempty fragments—the merge guard only handles an empty fragment—and add the model(s) used to the existing AI disclosure.


****

chelsealong:
Fixed: the unchanged-content check in `_remove_toml_entries()` now runs before the empty/comments-only deletion branch, so a pre-existing comments-only or blank `config.toml` with no Specify-owned blocks is left untouched instead of being unlinked. The delete-if-empty branch still fires when a Specify-owned block was actually stripped. Added regression tests for comments-only and blank configs (confirmed failing on the prior commit, passing now) plus one guarding that owned-block deletion on teardown still works. Also corrected the PR description's claim about the `_merge_toml_fragment` guard (it only covers an empty fragment, not identical nonempty fragments) and added the model name (Claude Sonnet 5) to the AI disclosure for this revision.

mnriem:
Thank you!

PR #4570, opened 2026-09-13
fix(integrations): stop frontmatter injection gluing onto a missing trailing newline

Noor-ul-ain001:
## Summary

`_inject_frontmatter_flag()` in `ClaudeIntegration`/`VibeIntegration`/`AlquimiaAIIntegration` detects the closing `---` line's existing EOL and reuses it when injecting a new frontmatter key. When that `---` is the file's **last line with no trailing newline**, the injected text is appended with no newline separator at all, producing `user-invocable: true---` instead of a clean, separately-lined `---`.

This corrupts the frontmatter (the closing delimiter is no longer alone on its own line), and since `post_process_skill_content()` chains multiple calls (`user-invocable`, then `disable-model-invocation`, then any fork-context keys), the **second** call's pre-scan can no longer find a second `---` line at all — so every subsequent key injection is silently dropped, not just corrupted.

`post_process_skill_content()` runs on content from "external skill generators (presets, extensions)" per its own docstring, so a trailing newline after the closing delimiter isn't guaranteed.

```python
>>> ClaudeIntegration._inject_frontmatter_flag("---\nname: x\n---", "user-invocable")
'---\nname: x\nuser-invocable: true---'          # corrupted delimiter
>>> # a second call on that output:
>>> ClaudeIntegration._inject_frontmatter_flag(_, "disable-model-invocation", "false")
'---\nname: x\nuser-invocable: true---'           # unchanged -- silently dropped
```

`DroidIntegration`'s own copy of this exact helper already emits an unconditional `\n` instead of detecting/reusing the existing EOL, avoiding this bug entirely. Ported that fix to the other three implementations that share the same code.

## Changes

- `src/specify_cli/integrations/claude/__init__.py`, `vibe/__init__.py`, `alquimia/__init__.py`: `_inject_frontmatter_flag` now always emits `f"{key}: {value}\n"`, matching `DroidIntegration`.
- Added a regression test to each affected integration's test file, covering both the single-call corruption and the chained-calls silent-drop.

## Test plan

- [x] `ruff check .` clean
- [x] New tests fail against the pre-fix code (verified via test-the-test) and pass with the fix
- [x] Full `tests/integrations/test_integration_{claude,vibe,alquimia,droid}.py` suite: 185 passed, 1 pre-existing/unrelated skip


mnriem:
The header-only reproduction demonstrates the helper defect. Could you show a supported preset or extension rendering path that produces this input? A skill containing a body does not exhibit this failure merely because its final newline is missing, so we’re keeping this as `triage-can-wait` pending that evidence.


Please also preserve existing line endings: the new unconditional LF introduces mixed endings into CRLF content. Retain the detected line ending when present and supply a separator when it is absent.


Please complete the AI disclosure with the tool, mode/settings, and extent of assistance; the Claude Sonnet 5 attribution is already present in the commit.


mnriem:
Thank you!

PR #4572, opened 2026-09-13
fix(workflows): stop list-literal-then-index expressions silently corrupting

Noor-ul-ain001:
## Summary

`_evaluate_simple_expression`'s list-literal detection (`workflows/expressions.py`) is `expr.startswith("[") and expr.endswith("]")`. This is true not only for a genuine literal like `[1, 2, 3]` but also for a list literal immediately followed by an index suffix, e.g. `[1,2,3][1]` — which reads as "index 1 of `[1,2,3]`" (i.e. `2`).

Naively stripping the outer brackets from `"[1,2,3][1]"` produces `"1,2,3][1"`, which the comma-splitter then breaks into `["1", "2", "3][1"]`. The malformed last segment resolves to `None` via the existing dot-path fallback, so the whole expression silently evaluates to `[1, 2, None]` — no error, no warning, just wrong data.

```python
>>> evaluate_expression("{{ [1,2,3][1] }}", ctx)
[1, 2, None]          # should read as index 1 of [1,2,3], i.e. 2
```

This is the exact "grabs the wrong span" failure mode the adjacent string-literal check already guards against a few lines above (verifying the matching quote is the *last* character, not just present anywhere) — the list-literal branch never got the same rigor.

**Scope note:** the evaluator has never supported indexing directly into a literal list — an author writes `steps.x.output.list[0]`, which the dot-path resolver already handles. `[1,2,3][1]` is not a documented or intentionally-supported form; this is a parser-classification/hardening fix for a shape the grammar accidentally mishandles, not a report of a real workflow breaking in the wild.

## Fix

Added `_is_single_list_literal`: a quote/bracket-depth scan (mirroring `_split_top_level_commas`'s existing tracking style in the same file) that confirms the opening `[` closes exactly at the final character before treating the expression as one literal. A misclassified expression now falls through to the existing dot-path resolution and evaluates to `None`.

## Concrete before/after (added per review request)

Same input, three real consumer contexts, values captured by running both revisions of `expressions.py` directly:

```python
tmpl = "{{ [1,2,3][1] }}"
text = "Index 1 is {{ [1,2,3][1] }}."
```

| Consumer | Before this PR | After this PR |
|---|---|---|
| Typed value (`evaluate_expression(tmpl, ctx)`) | `[1, 2, None]` | `None` |
| `condition: "{{ [1,2,3][1] }}"` (`evaluate_condition`) | `True` (a non-empty list is truthy) | `False` |
| Interpolated text (`evaluate_expression(text, ctx)`) | `"Index 1 is [1, 2, None]."` | `"Index 1 is ."` |

Narrowing the earlier "not a new failure mode" claim: that's true only in the sense that `None`/`False`/`""` for an unresolvable expression is the same output this module already produces for e.g. a typo'd dot-path (`{{ steps.typo.output }}`). It is **not** a claim that `None` is harmless everywhere it propagates — as the table shows, in a `condition:` this flips a step from unconditionally running (previous behavior, only true because *any* non-empty list is truthy, independent of its content) to unconditionally skipped. Neither before nor after is "correct" for this malformed expression (the feature it looks like it's using was never implemented), but I don't have a real workflow that hits this: `[literal][index]` isn't a pattern any shipped template, extension, or preset in this repo uses (checked via grep), and it's not a form the docs describe. This PR only stops the *specific* corruption of producing a plausible-looking wrong list; it intentionally does not change how `None` is treated by any downstream consumer, which is out of scope here per the request not to redesign the parser.

## Test plan

- [x] `ruff check .` clean
- [x] New test fails against the pre-fix code (`assert [1, 2, None] is None` → confirmed via test-the-test) and passes with the fix
- [x] Full `test_workflows.py` suite (985 tests) passes

## AI Disclosure

Written with [Claude Code](https://claude.com/claude-code) (Anthropic CLI), model Claude Sonnet 5, in an interactive agentic session with full tool access (repo search, file edits, running `pytest`/`ruff`). Extent: the AI proposed the diagnosis, wrote the `_is_single_list_literal` implementation, the regression test, and this description's before/after table (values were generated by actually running both revisions of the module, not asserted from memory). I (the human) directed the investigation, reviewed every diff line before pushing, and am responsible for the change.


mnriem:
The list-literal misclassification is reproducible, but the change replaces a malformed list result with an unresolved value rather than supporting the expression or reporting an error.


Please provide a concrete affected workflow and its before/after outcome, and narrow the claim that returning `None` is safe. Depending on the consumer, that value can become empty text or a false condition. We’re keeping this as `triage-can-wait` pending that context; this is not a request to add literal-list indexing or redesign the parser.


Please also complete the AI disclosure with the tool, mode/settings, and extent of assistance; the Claude Sonnet 5 attribution is already present.


Noor-ul-ain001:
Thanks for the push-back — both points are fair. Updated the PR description with:

**Concrete before/after**, captured by actually running both revisions of `expressions.py` against the same input (not asserted from memory):

| Consumer | Before | After |
|---|---|---|
| Typed value | `[1, 2, None]` | `None` |
| `condition: "{{ [1,2,3][1] }}"` | `True` (any non-empty list is truthy) | `False` |
| Interpolated text | `"Index 1 is [1, 2, None]."` | `"Index 1 is ."` |

So you're right to push on "safe" — I was overclaiming. In a `condition:`, this flips a step from unconditionally running to unconditionally skipped; in interpolated text it becomes empty rather than showing a garbage list. Neither outcome is "correct" (indexing a list literal was never actually implemented — `[1,2,3][1]` isn't a supported form, an author would write `steps.x.output.list[0]`), so what I can honestly claim is narrower than "safe": this only replaces the *specific* corruption (a plausible-looking wrong list) with the same `None`/`False`/`""` this module already returns for any other unresolvable expression (e.g. a typo'd `{{ steps.typo.output }}`). It doesn't newly introduce silent-`None` propagation — that convention predates this PR — it just stops one bracket-classification bug from producing a *worse*, data-shaped-wrong answer instead of the module's existing "unresolvable" answer.

I don't have a real affected workflow to point to. I checked: no template, extension, or preset shipped in this repo uses a `[literal][index]` expression (`grep -rn "{{.*\]\[" --include=*.yml --include=*.yaml .` — zero hits), and it isn't a documented form. I found this by auditing the bracket-matching logic itself (same class of bug as the adjacent string-literal check), not from a bug report. Given that and your steer that this isn't a request to redesign the parser, I'd read this as low-priority hardening rather than something with a known blast radius — happy to leave it at `triage-can-wait` and let a maintainer decide if it's worth merging as-is.

Also completed the **AI Disclosure** section in the description: Claude Code (Anthropic CLI), model Claude Sonnet 5, interactive agentic session with full tool access, human-directed and human-reviewed.

mnriem:
Thank you!

PR #4577, opened 2026-09-14
feat: post init config command

therightstuff:
# Add post-initialization project configuration

## Description

Adds `specify config` so users can inspect and change supported project settings
after initialization:

- `list` and `list --json` show persisted settings and installed extensions.
- `get <key>` reads a setting; `set feature-numbering sequential|timestamp`
  updates numbering while preserving other settings.
- Integration-owned settings direct users to `specify integration use` or
  `upgrade`, which regenerate the affected agent files. Read-only metadata and
  unknown keys receive distinct errors.
- `config extension` exposes the existing extension lifecycle commands.

Configuration writes reject malformed or unreadable initialization files.
Legacy projects without an options file must select or install an integration
first, preventing a settings update from disabling legacy command registration.

Includes configuration reference documentation and a developer verifier that
exercises the workflow in a disposable project, including when
`SPECIFY_INIT_DIR` points to an existing project.

## Testing

179 focused tests passed:

```bash
source .venv/bin/activate
python -m pytest tests/test_config_cli.py \
  tests/test_verify_post_initialization_configuration_script.py \
  tests/integrations/test_integration_subcommand.py -q
```

Coverage includes configuration validation and persistence, legacy extension
registration, disposable-project verification, script-upgrade regeneration,
and preservation of customized commands and saved settings on refused upgrades.

`uvx --offline ruff@0.15.0 check src tests` and `git diff --check` passed.
The full repository test suite was not run.

## AI Disclosure

- [x] I **did** use AI assistance.

Codex (GPT-6) assisted with implementation, review, tests, and documentation.


mnriem:
Please rescope this around a concrete post-initialization use case that the documented configuration/update workflow does not adequately support, taking the installed extensions and presets into account.

Before introducing another CLI command surface, let’s establish the gap, which component owns it, and the smallest change needed. Please propose that scope here before further implementation. Any resulting PR should address one coherent problem; leave unrelated tooling and contributor-document cleanup out.

Issue #2409, opened 2026-04-29
[Bug]: opencode integration dispatch fails — uses unsupported `-p` flag instead of `run` subcommand


markuswondrak:
### Bug Description

When dispatching workflow commands via the opencode integration, `OpencodeIntegration` inherits `build_exec_args()` from `MarkdownIntegration`, which generates `["opencode", "-p", "/speckit.specify ..."]`. The **opencode CLI does not support the `-p` short flag** — it uses `--prompt` (long form only) for prompt input, and the intended headless dispatch mode is `opencode run [message..]`.

When opencode receives the unrecognized `-p` flag, it prints its full help text and exits — the workflow step sees exit code 0 but no actual output, causing the step to fail silently.

### Steps to Reproduce

1. Initialize a project with opencode: `specify init --ai opencode`
2. Run a workflow: `specify workflow run speckit --input spec="build a login page" --input integration=opencode`
3. Observe: opencode prints its CLI help text instead of executing the `/speckit.specify` command
4. The workflow step fails with empty stdout/stderr and `exit_code: 0` (help exits successfully)

### Expected Behavior

The workflow should dispatch `opencode run "/speckit.specify build a login page"` which:
- Uses the `run` subcommand (headless/batch mode, not interactive TUI)
- Processes the slash-command from `.opencode/command/speckit.specify.md`
- Runs to completion and exits with the appropriate exit code

### Actual Behavior

The inherited `build_exec_args()` generates `["opencode", "-p", "/speckit.specify build a login page"]`. Since `-p` is not a valid opencode flag, opencode prints its help text and exits — no command is executed.

### Environment

- **Specify CLI Version:** 0.8.3.dev0
- **AI Agent:** opencode
- **Operating System:** Linux x86_64
- **Python:** 3.12.3

### Error Output

```
$ specify workflow run speckit \
    --input spec="$(cat ./01_chat_mvp.md)" \
    --input integration=opencode

Running workflow: Full SDD Cycle (speckit)
Version: 1.0.0

  ▸  speckit.specify …

[opencode prints full CLI help text — same output as `opencode --help`]
```

### Root Cause

**Code location:** `src/specify_cli/integrations/opencode/__init__.py` (21 lines)

`OpencodeIntegration` does NOT override `build_exec_args()`. It inherits from `MarkdownIntegration` which hardcodes `-p`:

```python
# src/specify_cli/integrations/base.py, lines 833-847
class MarkdownIntegration(IntegrationBase):
[...]

PR #4581, opened 2026-09-14
fix(integrations): dispatch Amp via execute mode

### LINKED ISSUE #4580: [Bug]: Amp dispatch sends `-p`/`--model`/`--output-format` — none of these exist in the Amp CLI
### Bug Description

`AmpIntegration` never overrides `build_exec_args()`, so it inherits the generic implementation from `MarkdownIntegration` (`src/specify_cli/integrations/base.py:1062`), which emits:

```
<exe> -p <prompt> --model <model> --output-format json
```

None of those three flags exist in the Amp CLI. Every workflow `command:`/`prompt:` step dispatching to Amp fails immediately at argument parsing, even with `amp` installed and authenticated.

This is the `amp` row of the audit in #2416, which lists it first among the integrations still inheriting the unverified `-p / --model / --output-format json` pattern.

### Steps to Reproduce

1. Generate the argv Spec Kit dispatches with, from a source checkout at `main` (`d848fb4`):

```console
$ uv run python -c "from specify_cli.integrations.amp import AmpIntegration; print(AmpIntegration().build_exec_args('hello', model='gpt-5', output_json=True))"
['amp', '-p', 'hello', '--model', 'gpt-5', '--output-format', 'json']
```

2. Run exactly that against the real Amp CLI:

```console
$ amp -p "hello" --model gpt-5 --output-format json
Error: error: unknown option '-p'
$ echo $?
1
```

3. Confirm the flags Amp actually accepts:

```console
$ amp --help | grep -E '^\s+-x|^\s+--stream-json$'
  -x, --execute [message]
  --stream-json
```

### Expected Behavior

Spec Kit dispatches to Amp using Amp's documented execute-mode interface, so the command runs:

```
amp --execute <prompt> [--stream-json]
```

### Actual Behavior

The dispatch fails at argument parsing with `error: unknown option '-p'` (exit 1). The agent is never invoked.

Checked against `amp --help` (v0.0.1789372854-ge954c8):

| Flag Spec Kit sends | Present in Amp CLI? | Amp's actual flag |
|---|---|---|
| `-p <prompt>` | No | `-x, --execute [message]` |
| `--model <model>` | No | none — `-m, --mode` sets an *agent mode* (`low`/`medium`/`high`/`ultra`), not a model name |
| `--output-format json` | No | `--stream-json` (used with `--execute`) |

### Specify CLI Version

1.0.7.dev0 (source checkout, git `d848fb4`)

### AI Agent

Amp

### Operating System

Windows 11

### Python Version

3.14.7

### Error Logs

```shell
$ amp -p "hello" --model gpt-5 --output-format json
Error: error: unknown option '-p'
```

For contrast, the proposed argv parses and proceeds (it only stops at the login prompt, because this machine has no API key):

```shell
$ amp --execute "hello" --stream-json
No API key found. Starting login flow...
To log in, visit:
https://auth.ampcode.com/device?user_code=...
```

### Additional Context

Amp CLI version tested: `0.0.1789372854-ge954c8`, installed via `npx @sourcegraph/amp`.

Documentation: [execute mode](https://ampcode.com/docs/cli/execute-mode) documents `-x, --execute`; [streaming JSON](https://ampcode.com/docs/cli/streaming-json) states "Use `--stream-json` with `--execute` to output streaming JSON instead of plain text."

On the `model` parameter: Amp exposes no model-selection flag. `-m, --mode` takes an agent mode (`low`, `medium`, `high`, `ultra`, or a plugin mode) which "controls the model, system prompt, and tool selection" — it is not a model identifier, so passing Spec Kit's `model` value through to `-m` would be incorrect. A fix should drop the model flag rather than remap it.

Same fix shape as the existing one-off overrides for opencode (#2409) and goose (#3781). Happy to open a PR.

**AI Assistance Disclosure:** AI assistance was used for the source inspection, reproduction, and drafting of this report. All CLI output above was executed and captured on the reporting machine.

hmohammadi:
## Description

Fixes #4580. Part of the audit in #2416.

`AmpIntegration` never overrode `build_exec_args()`, so it inherited the generic implementation from `MarkdownIntegration` (`base.py:1062`), which emits `-p <prompt> --model <model> --output-format json`. None of those three flags exist in the Amp CLI, so every workflow `command:`/`prompt:` step targeting Amp aborted at argument parsing before the agent ran:

```console
$ amp -p "hello" --model gpt-5 --output-format json
Error: error: unknown option '-p'
$ echo $?
1
```

Amp's non-interactive entry point is `-x/--execute`, and its structured-output flag is `--stream-json` (valid only alongside `--execute`), so this dispatches through those instead:

```console
$ uv run python -c "from specify_cli.integrations.amp import AmpIntegration; print(AmpIntegration().build_exec_args('hello', model='gpt-5', output_json=True))"
['amp', '--execute', 'hello', '--stream-json']
```

Verified against `amp --help` (v0.0.1789372854-ge954c8):

| Flag previously sent | In the Amp CLI? | Amp's actual flag |
|---|---|---|
| `-p <prompt>` | No | `-x, --execute [message]` |
| `--model <model>` | No | none (see below) |
| `--output-format json` | No | `--stream-json` |

**On `model`:** it is deliberately dropped rather than remapped. Amp exposes no model-selection flag. `-m/--mode` takes an agent mode (`low`/`medium`/`high`/`ultra`, or a plugin mode) which "controls the model, system prompt, and tool selection" — it is not a model identifier, so forwarding the caller's model onto it would silently select the wrong thing. This is the one judgement call in the diff; happy to change it if maintainers prefer a different mapping.

**On flag order:** `SPECKIT_INTEGRATION_AMP_EXTRA_ARGS` is applied *before* `--execute`, because `--execute [message]` takes the prompt as an optional inline value — appending between the two would consume the prompt as the operator's flag value.

Same fix shape as the existing one-off overrides for opencode (#2409) and goose (#3781); follows the `OpencodeIntegration.build_exec_args()` pattern.

## Testing

Added 4 regression tests to `tests/integrations/test_integration_amp.py` (previously a stub with no `build_exec_args` coverage). All 4 fail on `main` with `'-p' != '--execute'` and pass with this change; the file's 22 tests pass in full.

`tests/test_agent_config_consistency.py`: 28 passed.

`tests/integrations/test_extra_args.py`, `test_base.py`, `test_integration_base_markdown.py` — the shared machinery this override calls into (`_apply_extra_args_env_var`, `_resolve_executable`, the `MarkdownIntegration` base): 119 passed, 1 skipped.

Full-suite context, stated plainly: on this machine (Windows 11, Python 3.14.7) `main` already reports **182 failed / 7694 passed / 189 skipped** *before* this change. Every failure is environmental — 105 bash/parity and 63 symlink tests, on a box with no symlink privilege — and none are in `tests/integrations/test_integration_amp.py`. Of the 18 baseline failures under `tests/integrations`, all 18 are symlink tests. A full `tests/integrations` run against this branch was still in progress at the time of writing; I will post the before/after comparison as a comment when it completes.

End-to-end against the real Amp CLI (`npx @sourcegraph/amp`, v0.0.1789372854-ge954c8) on Windows 11: the previously generated argv exits 1 at parsing with `unknown option '-p'`, while the argv this PR generates is accepted and proceeds to Amp's device-login prompt (this machine has no `AMP_API_KEY`, so execution stops there rather than at flag parsing).

## AI Disclosure

**Tooling:** Claude Code, interactive CLI session. Model: **Claude Opus 5**, 1M-token context window.

**Extent:** AI-assisted for the source inspection of `build_exec_args()` inheritance, the reproduction, the patch to `src/specify_cli/integrations/amp/__init__.py`, the four regression tests, and the drafting of this description, the linked issue #4580, and my comments on this PR.

**Human review:** I reviewed the diff and the tests line by line and ran every command quoted here on my own machine. The Amp flag surface was verified against `amp --help` from the installed binary (`@sourcegraph/amp` v0.0.1789372854-ge954c8), not from model recall.


hmohammadi:
Following up on the test note in the description — the full `tests/integrations` run against this branch has finished, so here is the before/after comparison.

| | `main` (`d848fb4`) | this branch |
|---|---|---|
| failed | 18 | 18 |
| passed | 2873 | 2873 |
| skipped | 22 | 22 |

The failing set is **identical** — same 18 node IDs, verified by diffing the sorted `FAILED` lines from both runs. No test changes state in either direction.

All 18 are symlink tests, failing because this machine (Windows 11, no symlink privilege) cannot create the symlinks they set up — e.g. `test_shared_template_refresh_preflights_before_writing` calls `os.symlink()` and expects `Refusing to overwrite symlinked`. They are unrelated to this change, which touches only `AmpIntegration.build_exec_args()`.

Targeted results on this branch:

- `tests/integrations/test_integration_amp.py` — 22 passed (the 4 new regression tests plus the 18 inherited `MarkdownIntegrationTests` cases)
- `tests/integrations/test_extra_args.py`, `test_base.py`, `test_integration_base_markdown.py` — 119 passed, 1 skipped (the shared machinery this override calls into)
- `tests/test_agent_config_consistency.py` — 28 passed

For reference, the whole suite on `main` on this machine before any change: 182 failed / 7694 passed / 189 skipped, of which 105 are bash/parity and 63 symlink. Environment-specific, and CI is the authority here rather than my box.

mnriem:
Thanks for posting the follow-up. The four focused argv regressions check out.

Could you clarify whether the before/after table compares only the existing test set? Both columns have identical totals, although this PR adds four tests. Please state the comparison scope or correct the totals; this is not a request to fix unrelated Windows symlink failures.

Please also consolidate the AI model/context and operating-mode details into the PR disclosure alongside the existing Claude Code attribution. The commit already names Claude Opus 5 with a 1M context.

hmohammadi:
Scope: both columns were full runs of `tests/integrations`, not an existing-tests-only comparison. The `main` column was wrong — it carried this branch's `passed` count. `main` cannot collect 2913 there, since the four new tests exist only on this branch.

Corrected, with `main` re-run from a clean worktree at `d848fb4` on the same machine and interpreter:

|           | `main` (`d848fb4`) | this branch |
|-----------|--------------------|-------------|
| failed    | 18                 | 18          |
| passed    | **2869**           | 2873        |
| skipped   | 22                 | 22          |
| collected | 2909               | 2913        |

Only the `main` passed count changes from my earlier comment (2873 → 2869). The failing node IDs are identical on both sides — diffed, zero difference — and the entire delta between the columns is the four new regression tests. No pre-existing test changes state.

I have also consolidated the AI model, context window and operating mode into the AI Disclosure section of the PR description.

CI remains the authority on pass/fail, and it is green on all 15 checks.

*This comment was drafted with Claude Code (Claude Opus 5); I verified every number in it on my own machine.*

mnriem:
Thank you!

Issue #1049, opened 2025-10-26
Add Automated Testing for Bash and PowerShell Scripts


amondnet:
## Summary

This proposal adds comprehensive automated testing for the shell scripts in `scripts/bash/` and `scripts/powershell/` using Bats and Pester testing frameworks, integrated with GitHub Actions CI.

## Motivation

Currently, the Spec Kit project has no automated tests for its critical shell scripts (common.sh, setup-plan.sh, check-prerequisites.sh, create-new-feature.sh, update-agent-context.sh). This creates risks:

- No regression detection when scripts are modified
- Manual testing burden for contributors
- No verification that Bash and PowerShell scripts produce identical outputs
- Difficult to validate cross-platform compatibility

The CONTRIBUTING.md states "Write tests for new functionality" [1], but the scripts predating this guideline remain untested.

### Related Issues

Several existing issues demonstrate the need for automated testing:

**Script Bugs and Edge Cases**:
- [#1038](https://github.com/github/spec-kit/issues/1038) - Unable to source common.sh when CDPATH is set
- [#1029](https://github.com/github/spec-kit/issues/1029) - Dates generated by scripts are incorrect
- [#975](https://github.com/github/spec-kit/issues/975) - create-new-feature.ps1 ignores existing feature branches when determining next feature number
- [#965](https://github.com/github/spec-kit/issues/965) - Scripts REPO_ROOT variable issue in nested folders
- [#957](https://github.com/github/spec-kit/issues/957) - PowerShell script issue with apostrophes in arguments

**Script Usability Problems**:
- [#842](https://github.com/github/spec-kit/issues/842) - Copilot/Spec-kit seem unable to write PowerShell scripts that work

These issues would have been caught earlier with automated tests, preventing user friction and reducing maintenance burden.

## Proposed Solution

Implement automated testing using industry-standard frameworks:

**For Bash scripts**: Bats (Bash Automated Testing System)
- TAP-compliant testing framework [2]
- Used by major projects (Docker, Kubernetes helpers)
- CI integration with GitHub Actions [3]

**For PowerShell scripts**: Pester
- Official Microsoft-endorsed testing framework [4]
- Pre-installed on GitHub-hosted runners [5]
- Native GitHub Actions support

### Directory Structure
```
tests/
├── bash/
│   ├── test_helper/        # bats-support, bats-assert, bats-file
│   ├── unit/               # Function-level tests
│   └── integration/        # Workflow tests
├── powershell/
[...]

PR #4605, opened 2026-09-16
Remove a dead case pattern in common.sh (SC2221/SC2222)

jzayas03:
## Description

`resolve_template_content()` in `scripts/bash/common.sh` rejects unsafe manifest paths with:

```bash
/*|*../*|../*) manifest_file="" ;;
```

The third alternative can never match: `*../*` already matches every string `../*` does. shellcheck reports **SC2221** ("this pattern always overrides a later one") and **SC2222** ("this pattern never matches") on that line. This PR removes the dead alternative — a one-line change with no behavioural effect.

Verified behaviour-identical by classifying these inputs with the old and new pattern: `/etc/passwd`, `../x`, `a/../b`, `../`, `x/../../y`, `..`, `a/..`, `..a`, `a..b`, `./x`, `templates/x.md`. Every one produces the same reject/keep result before and after. After the change, shellcheck reports zero SC2221/SC2222 findings on the file and adds nothing new.

These are *warnings*, so `lint.yml`'s `--severity=error` does not currently see them. This is one of the findings standing between the current threshold and the tightening described in `CONTRIBUTING.md`. Related: #1049.

## Testing

- [x] Ran existing tests with `uv sync && uv run pytest` — **7883 passed, 195 skipped, 1 failed.** The failure is `tests/integrations/test_events.py::TestCommandRunner::test_ps_variant_prefixed_with_powershell_launcher`, and it fails identically with this change reverted: it asserts a `pwsh`/`powershell -File` prefix and `pwsh` is not installed on the machine I ran this on (macOS). It is pre-existing and environmental, not caused by this PR; your `macos-latest` runner has `pwsh` and should be green.
- [ ] Tested with a sample project — **not done, deliberately.** The change is proven behaviour-identical (table above), so running commands through an agent could not observe any difference. Mapping per `CONTRIBUTING.md` for completeness:

### Test selection reasoning

| Changed file | Affects | Test | Why |
|---|---|---|---|
| `scripts/bash/common.sh` | every command invoking `create-new-feature.sh`, `check-prerequisites.sh`, `setup-plan.sh`, `setup-tasks.sh`, `resolve-template.sh` (all source `common.sh`) | none run | the diff is a no-op on behaviour; `resolve_template_content()` accepts and rejects exactly the same inputs |

If maintainers would still like the manual agent runs, say so and I will do them.

## AI Disclosure

This PR was produced by Claude (Anthropic's AI coding agent) operating under my direction: it found the finding while running shellcheck over a project that vendors these scripts, verified the fix, wrote the commit and this description. I reviewed and approved it before it was opened. Any follow-up comments from this account on this PR may also be AI-drafted, and I will keep that disclosed.


jancarlosz039-ui (approved):

mnriem:
Thank you!

